PR #4, opened 2019-01-11
build(ci): add ci integration 

scottdickerson:
**Summary**

- Every push to development, we will automatically run testcases
- When a PR is promoted to master, it will also automatically publish a new release to git and npm

- build(.travis.yml): basic drop of CI build to trigger testcases and deployment through semantic-release
- build(package.json): add semantic release plugins for git and npm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,10 @@
 # [major]  Make the version tag to 2.0.0
 sudo: true
 language: node_js
+node_js:
+  - node
 
 before_install:
-  - nvm install 9.10.0
   - node --version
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -0,0 +1,35 @@
+# [minor]  Build and push 1.1.0 image
+# [major]  Make the version tag to 2.0.0
+sudo: true
```
cclauss: [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).

Issue #10, opened 2019-01-16
[Table] Lightweight design


bryancboyd:
Add a "lightweight" styling to the table (prop under `options`) that matches the following designs.

This table still maintains all capabilities of the full table (pagination, filter, column selection, batch actions).

![image](https://user-images.githubusercontent.com/2175647/51275050-76c63200-1996-11e9-97bd-c007db306efd.png)

![image](https://user-images.githubusercontent.com/2175647/51275060-7af24f80-1996-11e9-9e96-bf529ec0d870.png)

![image](https://user-images.githubusercontent.com/2175647/51275067-7fb70380-1996-11e9-87b6-a58ec43bcc12.png)

![image](https://user-images.githubusercontent.com/2175647/51275070-83e32100-1996-11e9-8151-07ae7fd05c49.png)

scottdickerson:
this was closed by the integrated PR

Issue #7, opened 2019-01-15
[Table] Filters


bryancboyd:
Implement filter enhancement to table component

##### Model

* Filter option and format (i.e. `canFilter`, `filterType = 'enum'`, `filterOptions = ['Connected', 'Disconnected']`) should be associated with `columns` prop in table.
* Active filter(s) should be under `view` prop (example: `view.filter.activeFilters = [{ id: 'deviceId', value: 'A123*' }, { id: 'status', value: 'Connected' }])

##### State: No filters added

* User can apply filters in the table at column level.
* When filter button is clicked, the filter icon turns blue and the filters expand

![image](https://user-images.githubusercontent.com/2175647/51208926-9b59d580-18d4-11e9-9935-a321d703b4fd.png)

##### State: With filters added

* Once a user starts typing in a filter, a cancel button would pop up similar to how the search bar functions
* User can press enter or click off the input to add their filter
* A "clear all filters" button would appear once any filters have been added

![image](https://user-images.githubusercontent.com/2175647/51208949-a57bd400-18d4-11e9-8bff-cd8984e54d0e.png)

tay1orjones:
I believe the only remaining items for this are

- [ ] clear all filters button
- [ ] 'cancel' button on text input filters (ComboBox already has it)
- [ ] minor styling fixes - toolbar toggle icons active state

PR #13, opened 2019-01-29
Initial Table Drop

tay1orjones:
**Summary**

- initial lightweight table, pagination, toolbar from @bryancboyd
- my first pass at table filtering

- initial table port from internal platform component repo
- add filter header row
- replace deprecated immutability addon with `immutability-helper`
- increase code coverage across all table components

**Related Issues**

- #10 
- #7 

- updates are visible in storybook


tay1orjones:
There is definitely more work here to do, but I figured we could get this baseline drop in as it stands.

bryancboyd:
I've added some improvements to the branch, but did a commit without verification... there were a few snapshot errors I wasn't sure how to resolve.  Here's the how it looks:

image

tay1orjones:
@scottdickerson this is ready for review.

- [x] Fix pagination functionality
- [x] Update/fix tests for filtering and pagination
- [x] Add filter toggle button in toolbar

package.json
```diff
@@ -30,6 +30,9 @@
     }
   },
   "dependencies": {
+    "carbon-components": "9.66.3",
```
scottdickerson: we want these only in dev-dependencies

src/utils/componentUtilityFunctions.js
```diff
@@ -12,3 +12,14 @@ export function scrollErrorIntoView(focus = true) {
...
 }
+
+export const handleEnterKeyDown = (evt, callback) => {
+  if (evt.key === 'Enter') {
+    callback(evt);
+  }
+};
+
+export const defaultFunction = name => {
+  console.log(this);
```
scottdickerson: this line is causing us some compilation warnings, so going to remove for now.

package.json
tay1orjones: oop, thanks for this. was likely a leftover from the yarn issues we were having during our pairing session

PR #17, opened 2019-02-07
fix(table): Clear filter(s) functionality

tay1orjones:
**Summary**

- More table filter work

- fix clear single filter
- fix clear all filters
- calm eslint down a bit on destructuring
- wrap each table story with a decorator instead of explicit `<div>`s

**Related Issues**

- #7 

- Filter clearing fully works now in the storybook

### DIFF
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import isNil from 'lodash/isNil';
 import { Button, PaginationV2, DataTable, Checkbox } from 'carbon-components-react';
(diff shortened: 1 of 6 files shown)


src/components/Table/Table.js
```diff
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { PureComponent } from 'react';
```
scottdickerson: I suspect we can't make this a PureComponent as the nested data object might change, but we would want the table to still rerender.

tay1orjones: Ah yeah - it just does the shallow compare. I changed it because eslint was [barking about needing to be a stateless function](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md) because I pulled out the uneeded constructor.

I'll refactor into a stateless function and update the PR

PR #18, opened 2019-02-08
feat(table): simplify pagination prop model

tay1orjones:
**Summary**

- We're fully encapsulating the `<Pagination/>`, so we always know the `totalItems` from the existing `data.length`.
- Same thing for the `pageSize`, we can infer if there isn't a declared `pageSize`, we can use `pageSizes[0]`

- remove `totalItems` from the `<Table/>` prop model
- add updates to `pageSize(s)`
- add better default props/usages
- update story

- Pagination renders/functions properly in storybook without errors.

### DIFF
--- a/src/components/Table/Table-story.js
+++ b/src/components/Table/Table-story.js
@@ -164,7 +164,6 @@ class TablePagination extends Component {
       },
       view: {
         pagination: {
-          totalItems: tableData.length,
           pageSize: 10,
           pageSizes: [10, 20, 30],
           page: 1,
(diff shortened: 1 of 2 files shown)


bryancboyd:
@tay1orjones if we want to support async loading (fetching new resources when the view changes), this design may not work... the `data` prop can't be assumed to contain all of the table data if we consider cases where a user is viewing 10k+ items (a use case for IoT Platform).  That's the main use case for passing in `totalItems`

tay1orjones:
Great point. I'll close this PR and forget about the `totalItems` change.

I'll incorporate the `defaultProps` improvements in my upcoming PR containing the merge of `defaultProps` and `props`

PR #19, opened 2019-02-08
feat(table): added column sort

bryancboyd:
**Summary**

- Added column sort (basic lexicographic compare, no custom sorts yet)

image

- (see commits in PR)

**Related Issues**

- (none)

- View the "Table --> sort" story in storybook, verify that the 1st, 3rd, and 4th columns are sortable.

### DIFF
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -85,9 +85,9 @@ const propTypes = {
       ),
     }),
     table: PropTypes.shape({
-      isSelectAllSelected: PropTypes.bool.isRequired,
+      isSelectAllSelected: PropTypes.bool,
       isSelectIndeterminate: PropTypes.bool,
-      selectedIds: PropTypes.arrayOf(PropTypes.string).isRequired,
+      selectedIds: PropTypes.arrayOf(PropTypes.string),
       sort: PropTypes.shape({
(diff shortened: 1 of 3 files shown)


src/components/Table/Table.js
```diff
@@ -172,13 +184,24 @@ const Table = ({ columns, data, view, actions, options }) => {
...
+                    style={filterBarActive === true ? filterBarActiveStyle : {}}
+                    isSortable={column.isSortable}
+                    isSortHeader={hasSort}
+                    onClick={() => {
+                      if (column.isSortable && actions.table.onChangeSort) {
+                        actions.table.onChangeSort(column.id);
+                      }
+                    }}
+                    sortDirection={hasSort ? view.table.sort.direction : 'NONE'}>
+                    <span className="bx--table-header-label">{column.name}</span>
```
tay1orjones: We can just use `{column.name}` without the wrapper `span` here, [`<TableHeader/>` renders children in a span](https://github.com/IBM/carbon-components-react/blob/master/src/components/DataTable/TableHeader.js#L79)

bryancboyd: my bad, I missed that when merging.  I've pushed a change onto the PR.

PR #37, opened 2019-02-27
feat(table): add empty state

tay1orjones:
**Summary**

- Adds an empty state to the table, configurable via a custom node if desired.

- Include `@carbons/icons-react` for access to new `Bee` icon
  -  This should be our new source of truth for icons. It can live alongside existing `carbon-icons` until we fully update to Carbon v10
- Add Empty state to table
- Update stories
- Update snapshots

**Related Issues**

- Closes IBM/carbon-addons-iot-react#8

- Storybook Stateful example should show proper empty state when all items are deleted
- Storybook `with no data` story should also properly render empty state with default action

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "styled-components": "^4.1.3"
   },
   "peerDependencies": {
+    "@carbon/icons-react": "^0.0.1-alpha.12",
     "carbon-components": "^9.x",
     "carbon-components-react": "^6.x",
     "carbon-icons": "^7.x",
--- a/src/components/Table/Table-story.js
+++ b/src/components/Table/Table-story.js
@@ -105,6 +105,7 @@ const actions = {
   table: {
     onRowSelected: action('onRowSelected'),
     onSelectAll: action('onSelectAll'),
+    onDefaultEmptyCallToAction: action('onDefaultEmptyCallToAction'),
   },
 };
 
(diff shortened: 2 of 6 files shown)


bryancboyd:
I'm going to push some changes; I think it works better if we move the empty state content over to `view.table` and define it like this:

```
emptyState: PropTypes.oneOfType([
        PropTypes.shape({
          message: PropTypes.string,
          ctaButtonLabel: PropTypes.string,
        }),
        /* If a React element is provided, it will be rendered in place of the default */
        PropTypes.element,
      ]),
```

bryancboyd:
Question: should we hide pagination if the table is empty?

image

bryancboyd:
Adding the ability to pass a different message and CTA if there are filters present (vs. a truly "empty state")

image

bryancboyd:
I've added static stories:

1) No data
2) No data (w/ filters)
3) No data (w/ custom empty state)

And also updated the stateful story logic to handle the empty state actions and behave like a "real table".  From the stateful story base state, try doing this to test:

1) Select all (with the starting filters still applied) and delete
2) Validate that the "no data w/ filters" state is shown, click the Clear Filters CTA button in the empty state
3) Validate that the "no data" state is shown, click the CTA button to add a new row
4) Make sure the table looks good with the row of data

PR #45, opened 2019-02-28
feat(table): column selection (no reordering yet)

bryancboyd:
**Summary**

Added column selection.  Reordering is not supported yet, and the design will change (see related issue)... but the props interface won't change.

image

(see PR)

**Related Issues**

https://github.com/IBM/carbon-addons-iot-react/issues/28

- Test column toggling on/off in the stateful storybook example
- View the new static story


CynZhang:
@bryancboyd the size of the buttons should be using the small version. I'm not sure if the use of gray is appropriate, it's the same style as carbon's tertiary button, and the color is closer to the Disabled state.

PR #54, opened 2019-03-03
refactor(tablehead): split table into tablehead component

scottdickerson:
**Summary**

- first just splitting table into separate tablehead component that can be directly instantiated

- items_here

**Related Issues**

- tests here

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -329,7 +329,11 @@ const StyledEmptyTableRow = styled(TableRow)`
 `;
 
 const Table = props => {
-  const { id, columns, data, view, actions, options } = merge({}, defaultProps(props), props);
+  const { id, columns, data, view, actions, options, className } = merge(
+    {},
+    defaultProps(props),
+    props
+  );
 
   const minItemInView =
     options.hasPagination && view.pagination
@@ -350,7 +354,7 @@ const Table = props => {
     (options.hasRowExpansion ? 1 : 0) +
     (options.hasRowActions ? 1 : 0);
   return (
-    <div id={id}>
+    <div id={id} className={className}>
       <TableContainer>
         <TableToolbar>
           <TableToolbarContent>
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -350,7 +350,7 @@ const Table = props => {
     (options.hasRowExpansion ? 1 : 0) +
     (options.hasRowActions ? 1 : 0);
   return (
-    <div>
+    <div id={id}>
       <TableContainer>
         <TableToolbar>
           <TableToolbarContent>
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import merge from 'lodash/merge';
+import pick from 'lodash/pick';
 import { Button, PaginationV2, DataTable, Checkbox } from 'carbon-components-react';
 import { iconGrid, iconFilter } from 'carbon-icons';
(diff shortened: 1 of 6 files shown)


src/components/Table/TableHead/TableHead.jsx
```diff
@@ -0,0 +1,164 @@
...
+        {hasRowSelection ? (
+          <TableHeader
+            style={Object.assign(
+              { paddingBottom: '0.5rem' },
+              filterBarActive ? filterBarActiveStyle : {}
+            )}>
+            {/* TODO: Replace checkbox with TableSelectAll component when onChange bug is fixed
+                    https://github.com/IBM/carbon-components-react/issues/1088 */}
+            <Checkbox
+              id="select-all"
```
bryancboyd: we should pass through the Table id to the child components (like this one) so that all DOM ids are uniquely scoped to the table instance

src/components/Table/TableHead/TableHead.jsx
```diff
@@ -0,0 +1,164 @@
...
+            id: PropTypes.string.isRequired,
+            text: PropTypes.string.isRequired,
+          })
+        ),
+      }),
+    })
+  ).isRequired,
+
+  /** Current state of the table */
+  state: PropTypes.shape({
```
bryancboyd: can we call this something different, to avoid overloading the "state" word?  `tableState` would suffice IMO

bryancboyd: or `tableView`, since that is the prop we are deconstructing to pass through here

src/components/Table/Table.jsx
```diff
@@ -399,75 +391,21 @@ const Table = props => {
...
-                }))}
-                ordering={view.table.ordering}
-                tableOptions={options}
-                onChangeOrdering={actions.table.onChangeOrdering}
-              />
-            )}
-          </TableHead>
-
+          <TableHead
+            options={options}
```
bryancboyd: My preference is to never pass props down that aren't specified in the child component's propTypes, to make code more traceable/readable... since we only are utilizing `hasRowExpansion` and `hasRowSelection` can we pass these explicitly?

```
options={{
  hasRowSelection: options.hasRowSelection,
  hasRowExpansion: options.hasRowExpansion,
}}
```

bryancboyd: My bad -- I see that the `options` prop is passed directly to FilterHeaderRow / ColumnSelectRow.  in that case, how about we externalize the PropTypes shape for options into an interface file?

```
TableOptionProps = {
  hasRowSelection: PropTypes.bool,
  ...
}
```

src/components/Table/Table.jsx
```diff
@@ -399,75 +391,21 @@ const Table = props => {
...
+            columns={columns}
+            filters={view.filters}
+            actions={{ ...actions.toolbar, ...actions.table }}
```
bryancboyd: See above comment... if we are only using a subset of the actions passed to the component by merging these two deconstructed objects, let's pass them explicitly

src/components/Table/TableHead/TableHead.jsx
scottdickerson: good idea

src/components/Table/TableHead/TableHead.jsx
scottdickerson: hoping to use css class selectors off the parent table id instead to limit the amount of ids we need to track and set (and reduce the amount of inbound props)

bryancboyd: alright, that works for me -- as long as everything rendered can be clearly referenced 👍

bryancboyd: just for reference can you provide the selector I could use to grab the `input` for a column filter in a table?

PR #75, opened 2019-03-08
feat(table): variable sizes for row and column

scottdickerson:
**Summary**

- allow users to specify their own row height and column width for their tables

- test(storybooksnaps): unfortunately storybook doesn't like styled components and throws a console error which I can ignore.  I don't want to fail the testcase because of that.
- feat(table): pass all props that are passed to our Table to the underlying CarbonTable, this allows users to easily use all the CarbonTable features and set a className for the table that they can use to perform their own styling
- test(table): add 3 stories to show how to apply each custom style to the table

**Related Issues**

- Closes IBM/carbon-addons-iot-react#36
- Closes IBM/carbon-addons-iot-react#46

- check the 3 new stories, zebra, column width, and row height.

### DIFF
--- a/src/components/StorybookSnapshots.test.js
+++ b/src/components/StorybookSnapshots.test.js
@@ -20,7 +20,13 @@ describe(`Storybook Snapshot tests and console checks`, () => {
     ReactDOM.findDOMNode = jest.fn(); // needed for this issue: https://github.com/facebook/react/issues/7371
     // TODO: remove once carbon PR is merged
     spy.console = jest.spyOn(console, 'error').mockImplementation(e => {
-      if (!e.includes('Warning: Function components cannot be given refs.')) {
+      if (
+        !e.includes('Warning: Function components cannot be given refs.') &&
+        !e.includes(
+          // workaround storybook console error with styled components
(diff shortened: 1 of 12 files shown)


bryancboyd:
A few things:

1) The column width isn't honored when filters are enabled, even if the column doesn't have a filter defined (as is the case with the `secretField` column)

```
<TableColumnWidth
  columns={tableColumns}
  data={tableData}
  actions={actions}
  options={{ hasFilter: true }}
  view={{ toolbar: { activeBar: 'filter' }}}
/>
```
image

2) In keeping with our desire to provide a "common design pattern" with easy escape hatch, could we add a prop on the column definition that allows the user to avoid passing in this for every column with a max width:

```
max-width: 20px; 
white-space: nowrap;
text-overflow: ellipsis; 
overflow-x: hidden;
```

Proposal:

```
{
  id: 'columnA',
  name: 'Column A',
  filter: { placeholderText: 'enter a value for this column' },
  width: 20,
}
```

This would automatically add the whitespace/overflow rules and max-width sizing.

scottdickerson:
thanks for catching this @bryancboyd I've submitted an update that allows the same css based approach to work correctly if the filter row is enabled, rather than adding a new prop.

PR #85, opened 2019-03-15
fix(table): improve null checks for stateful table reducers

tay1orjones:
**Summary**

- Fix console errors for consumers who don't define optional props sort, filters, toolbar


tay1orjones:
Those are both params in `filterSearchAndSort()`, which contains a null check for filters

PR #105, opened 2019-03-26
feat(table): add custom cell renderers

scottdickerson:
**Summary**

- Add a custom cell renderer render prop to each column that can be optionally passed

- refactor: renamed a propType to TableRowPropTypes
- feat: added renderDataFunction prop to TableBodyRow and Table to allow passing custom renderers
- test: added testcases to cover

**Related Issues**

- Closes IBM/carbon-addons-iot-react#88
- Closes IBM/carbon-addons-iot-react#50
- Closes IBM/carbon-addons-iot-react#76

- Verify the status field of the StatefulTable story
- Verify the new custom cell renderer story

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -272,6 +272,7 @@ const Table = props => {
               selectedIds={view.table.selectedIds}
               totalColumns={totalColumns}
               {...pick(options, 'hasRowSelection', 'hasRowExpansion', 'shouldExpandOnRowClick')}
+              ordering={view.table.ordering}
               actions={pick(
                 actions.table,
                 'onRowSelected',
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -26,6 +26,8 @@ const propTypes = {
   id: PropTypes.string,
   /** render zebra stripes or not */
   zebra: PropTypes.bool,
+  /**  lighter styling where regular table too visually heavy */
+  lightweight: PropTypes.bool,
   /** Specify the properties of each column in the table */
   columns: TableColumnsPropTypes.isRequired,
   /** Row value data for the body of the table */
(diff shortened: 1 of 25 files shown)


src/components/Table/Table.jsx
```diff
@@ -28,8 +28,8 @@ const propTypes = {
   zebra: PropTypes.bool,
   /** Specify the properties of each column in the table */
   columns: TableColumnsPropTypes.isRequired,
-  /** Data for the body of the table */
-  data: TableDataPropTypes.isRequired,
```
stuckless: One thing that concerns me about the table, is that there is no clear separation between model data and view data.  This table is built to force everyone to transform their model data into a counter-intuitive view data structure for this table.  These kinds of "use it my way or no way at all" makes it harder for teams to consume the component.  eg, 99% of the time, data coming from a remote site will be a json array of objects where each object is the row.  100% of the times, people are going to have to transform their data into the table's `data` structure and separate out their `id` and `values` into 2 structures and pass that to this table.   I would expect that I should be able to set a json array of objects on a table it would render it.

I understand why this is done, in that it's easier for you to `build` a table.  But that simply puts the onus on the developer to waste cycles on transforming data.  As it stands today, no-one can consume this table simply by setting some columns and data from a server.  They need to transform it into the table's view model (instead of the table doing that transparently behind the scenes, if needed).

What I'd like to see is that the data is json array of objects (as one will get almost always from a server) and then in the view configuration you tell the table what is my `id` field, because, again, almost all data will have their own id field, sometimes called, `id`, `_id`, `rowstamp`, `uniqueid`, etc.

I'd like to approach all components with the notion... `as a consumer of a component what the minimum amount of effort required for me to get data to render` and not `as a developer of the component, what's the minimum amount of effort for me to build the component`

scottdickerson: I don't think that's a bad idea @stuckless , can you open a separate issue on that?  Thanks

PR #110, opened 2019-03-28
fix(navigation): small fix to page styles

scottdickerson:
**Summary**

- the navigation area wasn't extending the tab content all the way down to the bottom of the screen, leading to a color mismatch

- Test the navigation bar fixtures and notice the tab content color should now extend down to the bottom of the screen and not map to the content


bryancboyd:
:shipit:

PR #112, opened 2019-03-28
fix(table): add translatable label props for all components and features of the table

scottdickerson:
**Summary**

- Need to expose translatable labels for all components and features of the table

- added props through the entire chain to translate labels

**Related Issues**

- Closes IBM/carbon-addons-iot-react#71

- Try out the new story on stateful table with I18N strings and notice that they're all changed.

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import merge from 'lodash/merge';
 import pick from 'lodash/pick';
 import { PaginationV2, DataTable } from 'carbon-components-react';
+import get from 'lodash/get';
 
 import { defaultFunction } from '../../utils/componentUtilityFunctions';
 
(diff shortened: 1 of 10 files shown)


bryancboyd:
How about we consolidate all of the i18n strings in the component into a single, top-level prop?  I think it's important to keep a logical difference between the "view" (the changing 'state' of the table) and the static props (data, columns, options, etc.) that define the component.  This, to me, fits in with the more static props.

(Also, most users would likely prefer providing all of the strings in a single prop and passing that in wholesale)

Example:

```
<Table
  data={data}
  columns={columns}
  view={view}
  i18n={
    pageBack
    pageForward
    pageItemRange
    ...
    clickToExpand
    ...
    clearAllFilters
  }
/>
```

@scottdickerson if you don't mind, I'm happy to do this refactor this afternoon

scottdickerson:
@bryancboyd that's a nicer idea I think thanks, don't forget to get the search placeholder text

bryancboyd:
@scottdickerson just pushed commits refactoring the props, can you review?

PR #118, opened 2019-04-01
fix(table): gvt prop redesign to make it easier to consume

scottdickerson:
**Summary**

- Small change to make the search placeholder easier to use
- Fix to the story to remove an unused prop for overflowActions

- Test the I18N story


bryancboyd:
done, and added to proptypes (renaming to `searchPlaceholder` from `searchPlaceHolder` after a quick interweb search for typical casing of the word) 👍

PR #145, opened 2019-04-03
fix(index): expose the header component so it can be reused

scottdickerson:
**Summary**

- expose the header component in our main index so downstream consumers can pick it up

- Closes IBM/carbon-addons-iot-react#142

- Pull the header to reuse


scottdickerson:
@filipecorrea here you go...

PR #152, opened 2019-04-05
Header fix

davidicus:
**Summary**

This is to bring in the Carbon components and remove a margin error.

- main changes in Header.jsx.


src/components/Header/Header.jsx
```diff
@@ -12,13 +13,12 @@ import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-import HeaderMenu from '../HeaderMenu';
+// import HeaderMenu from '../HeaderMenu';
```
scottdickerson: can we remove this commented code?

PR #153, opened 2019-04-05
SideNav and Header fixes

davidicus:
**Summary**

- Adding fixes for styling specificity issues that these two components have. Also adding 

- styles in Header.jsx and SideNav.jsx
- added test cases to increase overall code coverage in SideNav.test.jsx
- added Sidenav components to rollup config

- check storybook. colors should match AS and Platform product


src/components/SideNav/SideNav.jsx
```diff
@@ -13,22 +13,22 @@ import { COLORS } from '../../styles/styles';
 import CarbonSideNav from './CarbonSideNav';
 
 const StyledSideNav = styled(CarbonSideNav)`
-   {
+  &&& {
     background-color: ${COLORS.darkGray};
```
stuckless: @davidicus I see that IoT components are using styles from the styles jsx source file.  Is there no way we can leverage the Carbon Themes and SCSS for styling?  As it stands if the styles are inlined in code, there is no chance that consumers of these components can restyle them without forking, unless I'm missing something.

davidicus: @stuckless The styles we make here are overrides of the carbon css. We are utilizing carbon for almost all styling but certain things, like colors, or Carbon defects, we are overwriting here to line up with our IoT specific guidelines. Consumers will still be able to override both carbon and addon styles in their local applications' stylesheets. 

We try to offload any styles  we can back to carbon core through PRs on their projects.

PR #160, opened 2019-04-08
feat(wizardinline): add optional validation that can be executed for each step

scottdickerson:
**Summary**

- It's up to the consuming component to either set the new WizardInline error prop to indicate the form error to the user, or perhaps to indicate the error next to the failed form element itself.

- Check the testcases which validate the step doesn't change if validate fails
- View the new example error story for WizardInline


scottdickerson:
also 
fix(WizardHeader) remove unused backLabel and nextLabel props
fix(storybook): rtl should be in devdependencies not dependencies.

PR #184, opened 2019-04-17
Sidemav changes

davidicus:
**Summary**

- Adding changes to SideNav component to meet AS use cases. 

- made changes to Header.jsx to remove console log and to fix header along with sidenav
- changed SideNav to allow for passing along extra props and dictate the elements being rendered.

**Related Issues**

- Closes IBM/carbon-addons-iot-react#183


src/components/SideNav/CarbonSideNav.jsx
```diff
@@ -33,7 +31,9 @@ export default class SideNav extends React.Component {
     /**
      * Required props for accessibility label on the underlying menu
      */
-    ...AriaLabelPropType,
+    /* eslint-disable */
```
scottdickerson: can you only enable next to the line, otherwise it'll turn off for the whole file.  Or if you do turn it off, only turn off specific rules?

src/components/SideNav/SideNav.story.jsx
```diff
@@ -30,51 +30,70 @@ const StyledIcon = styled(Icon)`
   }
 `;
 
+// const RouterComponent = ({ children, ...rest }) => (
```
scottdickerson: did you mean to check this one in?

src/components/SideNav/CarbonSideNav.jsx
davidicus: done

src/components/SideNav/SideNav.story.jsx
davidicus: uncommented and added to story

PR #203, opened 2019-04-25
feat(table): add single selection and associated story

### LINKED ISSUE #202: [Table]: Sticky Selection
AS has a requirement need for sticky selection on the table. 

This is a new feature request.

For a table with sticky selection enabled, when clicking on a row, the row would become highlighted. This highlighting indication would stay present on the row until another row is selected. Only one may be selected at a time.

Screen Shot 2019-04-25 at 2 53 23 PM

tay1orjones:
**Summary**

- Add sticky selection to the table

- Add single selection to table
- Extend reducer to StatefulTable, enabling single selection
- Add new single selection story
- Update Snapshots, Tests

**Related Issues**

- Closes #202 

- check out the `with single selection` story, make sure the proper actions fire
- ensure there's no regressions in `with multi select and batch actions`

### DIFF
--- a/src/components/Table/AsyncTable/AsyncTable.jsx
+++ b/src/components/Table/AsyncTable/AsyncTable.jsx
@@ -209,7 +209,7 @@ const AsyncTable = ({ fetchData }) => {
         hasFilter: true,
         hasSearch: false,
         hasPagination: true,
-        hasRowSelection: false,
+        hasRowSelection: '',
         hasRowExpansion: false,
         hasRowActions: false,
         hasColumnSelection: true,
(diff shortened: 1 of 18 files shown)


tay1orjones:
In theory sticky and checkbox row select could be used together. I isolated the feature from the existing selection functionality to keep it somewhat composable and allow usage of both simultaneously.

If you'd prefer a sticky-_or_-standard approach for row select, I can refactor. Either way the sticky selection logic is slightly different in StatefulTable due to only one row can be sticky selected at a time.

tay1orjones:
@scottdickerson I refactored to use the existing functionality, adding a single selection option. Single select should work as expected, and existing multi select should be unchanged functionally.

PR #209, opened 2019-04-26
fix(list): fixed width on empty state

cgirani:
**Summary**

- When `isFixedWidth` true, empty state was not aligned with the table header.

- Empty content inside table
- New story for empty table with columns fixed
- Fixed react issue on `isfixedwidth` prop 

**Related Issues**

- Closes IBM/carbon-addons-iot-react#204

- Verify the empty state is aligned with the table header.

### DIFF
--- a/src/components/StructuredList/StructuredList.jsx
+++ b/src/components/StructuredList/StructuredList.jsx
@@ -12,9 +12,11 @@ import Bee32 from '@carbon/icons-react/lib/bee/32';
 
 const StructuredListWrapperStyled = styled(StructuredListWrapper)`
    {
+    && {
+      width: ${props => (props.isfixedwidth === 'true' ? 'inherit;' : '')};
+    }
     background-color: #ffffff;
     margin-bottom: 0;
-    ${props => (props.isFixedWidth ? 'width: inherit;' : '')}
 
     .bx--structured-list-th {
       padding-left: 16px;
@@ -27,7 +29,7 @@ const StructuredListWrapperStyled = styled(StructuredListWrapper)`
   }
 `;
 
-const EmptyDiv = styled.div`
+const EmptyContent = styled.div`
    {
     background-color: #ffffff;
     text-align: center;
@@ -36,6 +38,8 @@ const EmptyDiv = styled.div`
     padding-top: 90px;
     padding-bottom: 115px;
     font-weight: regular;
+    caption-side: bottom;
+    display: table-caption;
   }
 `;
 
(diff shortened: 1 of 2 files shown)


src/components/StructuredList/StructuredList.jsx
```diff
@@ -92,15 +96,15 @@ const StructuredList = ({ columns, data, design, isFixedWidth, onRowClick, loadi
           </StructuredListRow>
         ))}
       </StructuredListBody>
+      {!data.length ? (
+        <EmptyContent>
+          <Bee32 width={100} height={100} fill="#5a6872" />
```
tay1orjones: The `fill` hex color should come from the `COLORS` object in `styles.js`.

src/components/StructuredList/StructuredList.jsx
```diff
@@ -67,7 +71,7 @@ const StyledStructuredListCell = styled(StructuredListCell)`
  */
 const StructuredList = ({ columns, data, design, isFixedWidth, onRowClick, loadingDataLabel }) => (
   <Fragment>
-    <StructuredListWrapperStyled selection isFixedWidth={isFixedWidth}>
+    <StructuredListWrapperStyled selection isfixedwidth={isFixedWidth.toString()}>
```
tay1orjones: What was the react error you mention in the PR? The previous solution looks like it should work just fine.

cgirani: image

That's why I included `toString()` and the at the StructuredListWrapperStyled the validation is `=== "true"`

Is there a better way?

src/components/StructuredList/StructuredList.jsx
cgirani: Fixed on new commit

src/components/StructuredList/StructuredList.jsx
bryancboyd: @cgirani as a way to omit custom props from being rendered down on the DOM, you could do something like this:

```
const FilteredHeader = ({myProp, ...otherProps}) => <Header {...otherProps} />;

const StyledHeader = styled(FilteredHeader)`
  background-color: ${ props => props.myProp ? 'green' : 'red' };
`;
```

bryancboyd: So you could do this:

```
const StructuredListWrapperStyled = styled(
  ({ isFixedWidth, ...otherProps }) => <StructuredListWrapper {...otherProps} />
)`
   {
    && {
      width: ${props => (props.isfixedwidth === 'true' ? 'inherit;' : '')};
    }
    background-color: #ffffff;
    margin-bottom: 0;
    .bx--structured-list-th {
      padding-left: 16px;
    }
    .bx--structured-list-td {
      padding-left: 16px;
      line-height: 8px;
    }
  }
`;
```

PR #210, opened 2019-04-26
feat(progressindicator): vertical mode

cgirani:
**Summary**

- ProgressIndicator vertical design 

- New prop for vertical design 

**Related Issues**

- ProgressIndicator can be changed to vertical mode


scottdickerson:
you may not need to do this as I hear that this is being done in Carbon X

bryancboyd:
@scottdickerson we'll need to support a v9 style for our first use of this component, but when carbon-addons adds v10 support we can use the explicit vertical option

PR #211, opened 2019-04-28
fix(table): row actions style bugs need to show

scottdickerson:
**Summary**

- Three style fixes made to the table actions, see the bug list

- Add padding between them, make sure the buttons show when focused on, even if the row isn't hovered over, and fix the hover state of the overflow menu

**Related Issues**

- Closes IBM/carbon-addons-iot-react#199
- Closes IBM/carbon-addons-iot-react#207
- Closes IBM/carbon-addons-iot-react#208

- Verify the row actions cell in the table

### DIFF
--- a/src/components/Table/Table.story.jsx
+++ b/src/components/Table/Table.story.jsx
@@ -249,6 +249,12 @@ export const initialState = {
         labelText: 'Add',
         isOverflow: true,
       },
+      {
+        id: 'Delete',
+        icon: 'icon--delete',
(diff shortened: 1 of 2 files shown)


scottdickerson:
thanks for catching these @tay1orjones I can fix the overflow popping the wrong way.  The second tabbing issue seems to be broken in carbon and only fixed in carbon 10s version, so I'll wait for the upgrade to v7 to fix: https://codesandbox.io/s/5284kn765x

PR #213, opened 2019-04-28
fix(table): column filter value overlaps the clear and dropdown button when in RTL

scottdickerson:
**Summary**

- The column filter value contents overlap the filter controls when the table is in RTL mode

- small change to reserve enough padding on the right to display the Clear and dropdown button

**Related Issues**

- Closes IBM/carbon-addons-iot-react#167

- Verify that when you switch to RTL that the Table with filters story doesn't overlap the text


scottdickerson:
wow, thanks again resubmitted with the real fix @tay1orjones .  The rtl bug on the pagination is a carbon problem so needs to be fixed over there

PR #222, opened 2019-04-30
feat(table): add small table support with smaller pagination component

scottdickerson:
**Summary**

- summary_here

- chore(carbon-icons): upgrade to latest carbon-icons and update EmptyState and TileCatalog components to use non-built versions
- feat(Table): detect the size of the table and hide aspects of the pagination component.  Remove min-widths so our table can get narrower if it needs to
- fix(Table): passthrough the style prop to the parent component so the whole thing can be resized

**Related Issues**

- Closes IBM/carbon-addons-iot-react#221
- Closes IBM/carbon-addons-iot-react#220

- Verify the new small width table

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -5,6 +5,8 @@ import pick from 'lodash/pick';
 import { PaginationV2, DataTable } from 'carbon-components-react';
 import get from 'lodash/get';
 import isNil from 'lodash/isNil';
+import styled from 'styled-components';
+import sizeMe from 'react-sizeme';
 
 import { defaultFunction } from '../../utils/componentUtilityFunctions';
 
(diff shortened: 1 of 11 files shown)


src/components/Table/Table.jsx
```diff
@@ -24,6 +26,24 @@ import TableBody from './TableBody/TableBody';
...
+const StyledTableDiv = styled.div`
+  &&& {
+    .bx--data-table-v2-container {
+      min-width: unset;
+    }
+  }
+`;
+
+const StyledPagination = sizeMe({ noPlaceholder: true })(styled(PaginationV2)`
+  &&& {
```
filipecorrea: I think this  is not needed.

src/components/Table/Table.jsx
```diff
@@ -24,6 +26,24 @@ import TableBody from './TableBody/TableBody';
 
 const { Table: CarbonTable, TableContainer } = DataTable;
 
+const StyledTableDiv = styled.div`
+  &&& {
```
filipecorrea: I think this  is not needed.

PR #246, opened 2019-05-16
Update issue templates

davidicus:


tay1orjones:
I like this! We should update the existing issue template to follow the new folder syntax and include the `title`, `about` etc

PR #260, opened 2019-05-23
fix: StyledRowSingleSelection: added cursor as pointer

billpereira:
**Summary**

Added on the StyledRowSingleSelection

bug

**Related Issues**

259

- tests here

### DIFF
--- a/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
+++ b/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
@@ -124,6 +124,10 @@ const StyledTableRow = styled(TableRow)`
         }
       }
     }
+    ${props => (props['data-single-select'] === 'single' ? `cursor:pointer` : null)}
+    
+    }
+
   }
 `;
 
@@ -452,6 +456,7 @@ const TableBodyRow = ({
   ) : (
     <StyledTableRow
       key={id}
+      data-single-select={hasRowSelection}
       onClick={() => {
         if (hasRowSelection === 'single') {
           onRowSelected(id, true);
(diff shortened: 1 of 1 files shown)


scottdickerson:
Thank you for submitting this.   Can you add the cursor: pointer to all rows (not just the selected row?) . You should be able to conditionally branch on the prop in the styled component..

billpereira:
Sure, every clickable row should have the cursor as pointer?

filipecorrea:
Scott's right. I think you made a mistake here and added it just to the selected row.

billpereira:
i have added to the styled component if it hasRowSelection=single to change cursor to pointer

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -450,6 +456,7 @@ const TableBodyRow = ({
   ) : (
     <StyledTableRow
       key={id}
+      data-single-select={hasRowSelection}
```
filipecorrea: Wouldn't be simpler if we just propagate `hasRowSelection`?

```suggestion
      hasRowSelection={hasRowSelection}
```

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -124,6 +124,10 @@ const StyledTableRow = styled(TableRow)`
         }
       }
     }
+    ${props => (props['data-single-select'] === 'single' ? `cursor:pointer` : null)}
```
filipecorrea: If the other comment is correct, then:

```suggestion
    ${props => (props.hasRowSelection === 'single' ? `cursor:pointer` : null)}
```

PR #262, opened 2019-05-25
Fix/simple stateful example squased batch action header

billpereira:
**Summary**

Increased in 1 rem the padding of StyledCarbonableToolBar as it was smashed

- bugs

**Related Issues**

240

- Closes IBM/carbon-addons-iot-react#240

- tests here

### DIFF
--- a/src/components/Table/TableToolbar/TableToolbar.jsx
+++ b/src/components/Table/TableToolbar/TableToolbar.jsx
@@ -44,7 +44,7 @@ const StyledCarbonTableToolbar = styled(CarbonTableToolbar)`
    {
     &&& {
       width: 100%;
-      padding-top: 0.125rem;
+      padding-top: 1.125rem;
     }
   }
 `;
--- a/src/components/Table/TableToolbar/TableToolbar.jsx
+++ b/src/components/Table/TableToolbar/TableToolbar.jsx
@@ -44,7 +44,7 @@ const StyledCarbonTableToolbar = styled(CarbonTableToolbar)`
    {
     &&& {
       width: 100%;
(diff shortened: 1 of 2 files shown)


scottdickerson:
@billpereira could you make the height of the toolbar 50 pixels to correspond with the other table examples? Thanks!

billpereira:
i can ;)

billpereira:
image
Is that what you were looking for? so now it has same size of other rows...

billpereira:
I was looking here into the commit...
but if i didn't missed anything all the changes on src/components/snapshots/StorybookSnapshots.test.js.snap
are related to this fix? I have commited again to update it since it was not passing the snapshot test

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -55,6 +55,7 @@ const StyledTableToolbarContent = styled(TableToolbarContent)`
     padding-right: 1px;
     align-items: center;
     font-size: 0.875rem;
+    height: 2.25rem;
```
filipecorrea: I think that the right value here would be `3.125rem` to be equivalent to `50px`.

filipecorrea:
```suggestion
    height: 3.125rem;
```

filipecorrea: @scottdickerson Can you please confirm if this is what you requested before?

scottdickerson: I think it looks ok as is. I can merge it

PR #263, opened 2019-05-27
feat(data-vis): add experimental data viz components for dashboarding

scottdickerson:
**Summary**

- Add experimental data-viz components for dashboarding cases.  Dashboard Layout Engine and Value Cards

### DIFF
--- a/src/components/Dashboard/Dashboard.story.jsx
+++ b/src/components/Dashboard/Dashboard.story.jsx
@@ -1,0 +1,161 @@
+import React, { useState } from 'react';
+import uuidv1 from 'uuid/v1';
+import { text, boolean, object } from '@storybook/addon-knobs';
+import { storiesOf } from '@storybook/react';
+import { Button } from 'carbon-components-react';
+
+import {
+  DASHBOARD_BREAKPOINTS,
+  CARD_DIMENSIONS,
+  ROW_HEIGHT,
+  CARD_SIZES,
+  CARD_TYPES,
+} from '../../constants/LayoutConstants';
+
+import Dashboard from './Dashboard';
+
+const originalCards = [
+  {
+    title: 'SMALL',
+    id: 'facilitycard',
+    size: CARD_SIZES.SMALL,
+    type: CARD_TYPES.VALUE,
+    content: [
+      { title: 'Comfort Level', value: 89, unit: '%' },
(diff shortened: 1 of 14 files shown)


src/components/Dashboard/Dashboard.story.jsx
```diff
@@ -0,0 +1,161 @@
...
+];
+
+const StatefulDashboard = ({ ...props }) => {
+  const [cards, setCards] = useState(originalCards);
+
+  const handleAdd = () => {
+    setCards([
+      ...cards,
+      {
+        title: 'SMALL', // faker.company.companyName(),
```
filipecorrea: I think we should be using `faker` in all users inputs.

src/constants/LayoutConstants.js
```diff
@@ -0,0 +1,114 @@
...
+export const DASHBOARD_SIZES = {
+  XLARGE: 'xl',
+  LARGE: 'lg',
+  MEDIUM: 'md',
+  SMALL: 'sm',
+  XSMALL: 'xs',
+};
+
+export const DASHBOARD_COLUMNS = {
+  xl: 12,
```
stuckless: Is this "grid" supposed to emulate the Carbon 10 grid design?  If so, they are using a 16 column grid.

https://www.carbondesignsystem.com/guidelines/layout

src/constants/LayoutConstants.js
```diff
@@ -0,0 +1,114 @@
...
+  lg: 12,
+  md: 12,
+  sm: 6,
+  xs: 4,
+};
+
+export const DASHBOARD_BREAKPOINTS = {
+  xl: 1800,
```
stuckless: Should these breakpoints match the Carbon 10 design?

https://www.carbondesignsystem.com/guidelines/layout#breakpoints

src/components/Dashboard/Dashboard.story.jsx
scottdickerson: I think on this one we want a specific title to match the card size.  Can we leave the comments for now in case we need to add debugging and remove once we move out of experimental?

src/constants/LayoutConstants.js
scottdickerson: yeah I noticed that too @stuckless,  @bryancboyd I'm surprised we ended up with 12 columns instead of 16 total?

PR #264, opened 2019-05-27
Fix/single selection cursor should be pointer not arrow

billpereira:
**Summary**

- single selection cursor should be pointer not arrow

- Fixed suggestions from Filipe

**Related Issues**

- Closes IBM/carbon-addons-iot-react#259

- tests here


scottdickerson:
I think this one is just a dupe of #260

PR #284, opened 2019-06-05
Add isEmpty and isLoading states to Card

bryancboyd:
**Summary**

- Added basic loading/empty state to Card.  We can improve on the visuals in the future.
- Included new stories for Card for each state, and a "size gallery" to render all sizes of a card hooked up to the same prop knobs, to see how loading/empty scale up and down as size changes

image

image

image


bryancboyd:
image

Just pushed a commit that propagates the dashboard `isLoading` down to the cards

Issue #154, opened 2019-04-07
[v2]: Upgrade Components to Carbon v10


scottdickerson:
Containing issue for the migration of our components to Carbon v10

High level steps/pieces:
- [x] upgrade carbon deps
- [x] upgrade storybook to v5 to support new prop-type definitions
- [x] Update/Fix Components
    - [x] `ButtonEnhanced` refactored to `Button` wrapper
    - [ ] ~`Table` renamed to `DataTable`, ensure sub components exported in the same fashion as Carbon to support custom composition of components. Avoid any breaking change from Carbon's documentation/usage for `DataTable`~
- [x] review visually for defects, fix as necessary
- [ ] ... 

Maximo needs it and we need to support downstream teams

tay1orjones:
_Originally posted by @lukefirth in https://github.com/IBM/carbon-addons-iot-react/pull/300#issuecomment-501221567_:

Ok, you asked for the firehose and my under the microscope review. Feel free to break these up and handle them however.  Here goes:

------


# V10 Component Upgrade review

**1. Add card**
- I propose a slight styling update to match duo examples better here. (@amymillett please feel free to override me here)
![image](https://user-images.githubusercontent.com/30671949/59343751-768f5000-8d04-11e9-90b0-35a5b15e0afa.png)


**2. Base modal - Danger**
- I think the red bar at the top shouldn’t be there:
![image](https://user-images.githubusercontent.com/30671949/59343897-bd7d4580-8d04-11e9-8974-efd9e9da9ee4.png)


**3. Base Modal - Error states**
- Should be padding to the left and right, I believe the same as what you see for the title and X, rather than 100% of the modal width
![image](https://user-images.githubusercontent.com/30671949/59344343-c9b5d280-8d05-11e9-81a1-dfce4543dacb.png)


**4. Base modal - add description**
- Not seeing the description

**5. Header action buttons with drop downs**
- Not for this, but capturing for posterity - we should look at swapping the drop down types to right panel components.

**6. Wizardinline - Stateful example**
- Gaps between the progress indicator steps. I think in Duo they have them no longer spanning the container but all touching and left aligned.
- Title of step cut off, ‘Device inform’ in this example. Meant to wrap I believe.
- Close icon seems much heavier than the normal icon.
[...]

PR #300, opened 2019-06-10
Beta

davidicus:
**Summary**

Upgrade to Carbon v10 packages and fix defects that stem from upgrade.

- upgrade package.json to pull in new packages. 
- The rest of the files are regression fixes

**Related Issues**

- Closes IBM/carbon-addons-iot-react#154

- tests here


davidicus:
@KarenGski if you have the bandwidth to look into our v10 compliance. I would appreciate it. 
https://deploy-preview-300--carbon-addons-iot-react.netlify.com

tay1orjones:
In #304 we've identified a few more items to include in the v2.0.0 release in addition to the Carbon v10 upgrade.

The `beta` branch is now deployed to https://beta.carbon-addons-iot-react.com and will be continually updated moving forward.

Thoughts on closing this without merging in favor of #304? Once we've satisfied all those items, a new v2 PR can be opened up.

lukefirth:
Ok, you asked for the firehose and my under the microscope review. Feel free to break these up and handle them however.  Here goes:

------


# V10 Component Upgrade review

**1. Add card**
- I propose a slight styling update to match duo examples better here. (@amymillett please feel free to override me here)
image


**2. Base modal - Danger**
- I think the red bar at the top shouldn’t be there:
image


**3. Base Modal - Error states**
- Should be padding to the left and right, I believe the same as what you see for the title and X, rather than 100% of the modal width
image


**4. Base modal - add description**
- Not seeing the description

**5. Header action buttons with drop downs**
- Not for this, but capturing for posterity - we should look at swapping the drop down types to right panel components.

**6. Wizardinline - Stateful example**
- Gaps between the progress indicator steps. I think in Duo they have them no longer spanning the container but all touching and left aligned.
- Title of step cut off, ‘Device inform’ in this example. Meant to wrap I believe.
- Close icon seems much heavier than the normal icon.
- Title and progress indicator look to be a long way from the top of the modal, a ‘line’ down from the ‘line’ with the close icon. I would expect these to be on the same line, meaning a similar padding between the top of the inline and the title as seen in modals. eg:
image
@amymillett again, please feel free to override me here, I've not seen if there is a finalised V10 design for this we should be aiming for, these are just my comments on how I'd expect to respond to the current.

**7. Navigation bar - Normal**
-  Again not something for now, but for posterity. For the nav pattern stuff, we will be moving the tabs from the top of the navigation bar to the bottom of it, just above the page content.

**8. Navigation bar - Example with work area.**
- Text at the top says to click the new entity type button to interact with a work area, but nothing actually happens when I click the button.

**9. Page hero - normal with action.**
- This example seems to be a page hero with a content switcher, so the ‘with action’ name might be a touch misleading.
- The spacing on the content switcher seems off, its rendering super wide, I think there might need to be a max width set per ‘segment’. Although I can’t eyeball what a reasonable one would be.

**10. Page hero - with section**
- This is a slight design change (so feel free to take out of here and we can track elsewhere), but this may be a good time to replace the ‘Explore/‘ with a traditional breadcrumb above the page title.

**11. Progress indicator - Stateful**
- Same comment as in the modal examples, it seems like we’re spacing each step out to fill the width of the container, but carbon recommendation is that we ‘connect’ them by left aligning.
image


**12. Progress indicator - Vertical**
- Not sure if this is us or not, but the vertical spacing between steps is quite high. In the usage & style guidance carbon doesn’t talk about this for vertical so we’re on our own. I’d suggest we drop it to ~75% of what the example shows.

**13. Resource list - with action**
- On wider windows, the actions seem to be a little way ‘in’ to the list rather than anchored to the right hand side with a fixed padding. I believe it would be better if we set a fixed distance from the right hand edge of the list itself (including the row dividers) and the right hand side of the action… I guess just right aligning the action in the component.

**14. Side nav - Sidenav component.**
- Only thing here would be to reflect the hamburger saga. I personally would say go with Carbon (hamburger turns to X when open), but you heard the resistance to that. So to follow the party line, we should retain the hamburger icon in both open and closed states
image


**15. Simplepagination**
- Looks like it just needs a style update to reflect V10 https://www.carbondesignsystem.com/components/pagination/code

**16. Table - Simple stateful example**
- Checkboxes are slight off in vertical alignment
image


**17. Table - Stateful example with expansion** 
- Spacing between the filter inputs and the column titles they are filtering on could be reduced. Right now they look a little disconnected. Suggest 0 padding between the top of the filter input and the bottom of the table title ‘block’ (the bit that changes coloraturas on hover). image


- Row height has gone quite large (not a problem in itself), but the content in the row should stay middle aligned vertically. Now the action and check look to be middle aligned, but the cell content is higher up. I noticed this isn’t happening on the simple example of expansion, but does happen again when I use the ‘row expansion with actions’ so I wonder if its the actions that are causing the weirdness to be introduced. 
image


- Column picker - I think we just need to provide you with a V10 styling of this @amymillet do you have this handy you can supply?


**18. Table with simple search**
- Search focus state is strangely tiny
image


**19. Table body row**
- I can see now from these examples, the row height changes depending on if there are actions and if they are running, error or just there, which I guess is what introduced the weirdness in the earlier example in (17)

**20. Table detail wizard stateful example**
- I think the styling here needs to be aligned with the inline wizard component for the close button, title and footer (eg the button width etc). I’m not sure which one is ‘right’ or if both need to adapt a bit. I don’t have the design to hand - @Amymillet is there a design for the table inline wizard and the traditional inline wizard handy we can share?

**21. Table detail wizard static with sidebar**
- In the example, I’m not seeing a difference between the static and static with sidebar versions aside from the main content going from text to the green box example. Am I missing something?

**22. Table detail wizard with error.**
- Same comment as (3) the base modal one. Padding on either side should be the same for the error as the general content in the wizard rather than 100% I believe.

**23. Tile catalog - default**
- Mainly just the pagination style, but we’ve got that noted earlier.
- I see the notes in the example on wrapping and truncation of title and description. I suggest:
- Title can go to 2 lines wrapped, then should truncate.
- Description will fill the rest of the available space with wrapping, but when it reaches the end, it should truncate. Generally we should in the UX patterns (that we’ve not written for this yet) recommend that these are used with character limited descriptions.

**24. Tile catalog  - with search**
- I suggest we drop the title from the component itself (like we don’t have the table title in the table component) and in its place, put the same search icon/behaviour (click to go into search mode) as in the table component https://deploy-preview-300--carbon-addons-iot-react.netlify.com/?path=/story/table--with-simple-search.

**25. Tile catalog - loading**
- Can we set the height of the tiles in loading state to be the same height as them once loaded?  I would treat the height of the tiles in the same way as row heights in the table with fixed options (thus the truncation I mentioned earlier). That way we can set the loading to be nice and smooth of just content filling in without a size change.

**26. WizardModal - basic**
- Same note as before about the pagination, it seems to be correctly left aligned here, but there are still small gaps between the steps.

**27. WizardModal - customfooter**
- We should treat the custom footer in the same way as we do with the inline wizard, where here we use a primary button to fill the example. Generally I expect this custom space to be used by small status indicators and more likely - short link style text to act as a tertiary action. There’s a flame war in Carbon about allowing tertiary actions or not, but Chris Conners is on the same thought process as us with these sorts of examples: 
image

Just change the labels to ‘remove’ ‘cancel’ ‘done’ and we’ve got the main AS pattern.

davidicus:
@tay1orjones I had turned #304 into an epic referencing this issue and others. We can turn that back into an issue and close this out. I have no preference.

tay1orjones:
👍 Yeah, #304 is definitely an epic. 

Lets move the v10 component upgrade discussion to the existing issue we have for it - #154 _[v2]: Upgrade Components to Carbon v10_ 

@lukefirth I copied your above comment to there.

We'll reopen a new `beta`->`master` PR once we've completed all the v2 items.

PR #312, opened 2019-06-12
Line chart card

cgirani:
**Summary**

- summary_here

- items_here

**Related Issues**

- tests here


src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -1,124 +1,436 @@
...
-          x: timeDataSourceId,
-          xFormat: '%Y-%m-%dT%H:%M:%S.%LZ',
-          colors: {
-            [series.dataSourceId]: series.color,
+class TimeSeriesCard extends Component {
+  constructor(props) {
+    super(props);
+    // Create the ref
+    this.c3ChartComponent = React.createRef();
+    this.state = {
```
scottdickerson: is there anyway we could not store local state of colors?

src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -1,124 +1,436 @@
...
+                  <StyledTooltipValue>Value: {valueformat}</StyledTooltipValue>
+                </Fragment>
+              )}
+            </Fragment>
+          );
+        })}
+      </Fragment>
+    );
+
+    return ReactDOMServer.renderToString(content);
```
scottdickerson: This is very interesting!  I assume you had to do this because this is what C3 accepts as a tooltip, but it took me a while of thinking about it to figure it out.  Can you add a comment to explain the 'why' of it?

PR #318, opened 2019-06-13
fix(valuecard): improve style for the various layouts

scottdickerson:
**Summary**

- Style and layout fixes for all of the value card sizes and common content patterns

- Verify that all the Value Card stories layout their content in the best possible use of space


bryancboyd:
@scottdickerson I added a few commits to the PR, now the multi-value xsmallwide rendering looks good.  There are a few final edge cases we'll want to address eventually, but we can "configure around" those issues for now -- I think with this PR we can call value card closed.

image
image
image
image

PR #338, opened 2019-06-20
Time-series charts and card time ranges

bryancboyd:
**Summary**

- Initial drop of time series cards with carbon charts
- Added time range dropdown for cards


package.json
```diff
@@ -32,6 +32,7 @@
     }
   },
   "dependencies": {
+    "@carbon/charts-react": "0.12.1",
```
scottdickerson: we need to remove this dependency and only have it in a peer dep I think...

src/components/Card/Card.jsx
```diff
@@ -220,8 +238,7 @@ const Card = ({
     </StyledToolbar>
   ) : (
     <StyledToolbar>
-      {// TODO: if we keep this, expose capability under prop
-      false && timeBoxSelection}
+      {mergedAvailableActions.range && timeBoxSelection}
```
scottdickerson: I think we should also hide the range picker if the Card is in the editable state for now...

src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -1,438 +1,142 @@
...
+
+  const sameYear =
+    moment(moment.unix(values[0].timestamp / 1000)).isSame(moment(), 'year') &&
+    moment(moment.unix(values[values.length - 1].timestamp / 1000)).isSame(moment(), 'year');
+
+  const formatInterval = (timestamp, index) => {
+    const m = moment.unix(timestamp / 1000);
+
+    return sameDay && interval === 'hour' && index === 0
+      ? m.format('DD MMM YYYY')
```
scottdickerson: are we tracking the localization work somewhere?

src/components/Card/Card.jsx
bryancboyd: yep, that's covered by the ternary branch on `isEditable` -- if it's in editable state the range picker (and expansion button) won't be rendered

package.json
bryancboyd: good call thanks

src/components/TimeSeriesCard/TimeSeriesCard.jsx
bryancboyd: not at the moment

PR #379, opened 2019-07-08
Hero component

cgirani:
**Summary**

- New Hero component

1. Simple with description
image

2. Hero with breadcrumb
image

3. Hero with breadcrumb and secundary navigator
image


cgirani:
The Carbon Tabs has some default margin-right

Issue #355, opened 2019-06-25
[v2] Provide wrapper components for all Carbon react components


tay1orjones:
Wrap/re-export Carbon components from our package so consumers can import from one place. Implementation from proposal in #268 

- [x] Re-export components we don't currently have in our package
- [x] Evaluate publishing sass as well (so consumers can install component styles from our package, vs carbon)
  - export single entrypoint like Carbon, we can override any components via extending their mixins
- [x] Update storybook with the stories of newly exported components
  - Segment storybook nav into sections: `Stock`, `Extended`, `Custom`

tay1orjones:
Things remaining after the above PR is merged
- [ ] fix naming collisions in js exports, make sure we're exporting every component that carbon-components-react does, plus ours.
- [ ] import all the stories for the stock carbon components
- [ ] figure out how to provide individual `.scss` files for consumers, in addition to the compiled `.css` file

From the initial issue description, we still need to look into how to segment the storybook left panel into sections, despite not being able to 'control' the name of the Carbon stories we're importing.

tay1orjones:
We've segmented the storybook into sections and used css to change 'others' to 'carbon components'

The original scope of this issue is satisfied by the above PRs. Closing.

Issue #370, opened 2019-07-01
Work out new styling workflow.


davidicus:
We have decided to move away from styled components and will be instead going back to Sass partials for styling components to better align with the Carbon base workflow.

tay1orjones:
Once the above PR is merged, we just need to refactor components away from styled-components and into sass. That PR has an example of one component being refactored.

tay1orjones:
As part of this, we should keep an eye on Carbon's v11 work slated for October. They're considering css-in-js, scss modules, and some other approaches.

tay1orjones:
Something interesting to highlight here that surfaced in #1366: Carbon uses type checking in a few spots, which break with styled components. 

Carbon's logic looks for a type `ModalFooter`, but with styled components we provide `StyledModalFooter`, which fails the check.

sstone2423:
Down to mainly Table as most of the other existing components are deprecated and will be removed in the next version

PR #394, opened 2019-07-16
Export Carbon components, add Sass support

tay1orjones:
**Summary**

- A broad first pass at proxying carbon components by re-exporting them from our package. Allows users of this package to use it standalone, without having to manage Carbon packages as peer dependencies.
- This is just a first pass to get things operational, [I've updated #355](https://github.com/IBM/carbon-addons-iot-react/issues/355#issuecomment-512005832) with work we still need to complete after this PR.

- Move all Carbon deps to be hard deps, not peer or dev.
- Export most Carbon components from our package (some have naming collisions with our current components and we need to rectify those)
- Import Accordion storybook story for use in our storybook, as an example of what we'll do with all 'stock' Carbon components.
- Add Sass compilation
  - New .scss entrypoint alongside the .js one
  - This compiles all to one .css file placed in `./lib` for consumers
- Refactor AddCard styling from styled-components to sass as an example of how to do so for the remaining components

**Related Issues**

- #355

- View the storybook, should have a new accordion story
- Can validate the rollup build works by running `yarn`, though travis should do this as well.


tay1orjones:
This also provides the groundwork for #370, we'll just need to refactor the components away from styled-components.

tay1orjones:
@davidicus pushed a fix for those. Re added the global `width: 100%` and removed the styling overriding carbon defaults for add tile/clickable tile.

PR #416, opened 2019-07-25
fix(tablecard): improve isEditable state for timestamp columns

scottdickerson:
**Summary**

- improve the isEditable state to match the mockup for TableCard that Cynthia supplied
- fix(TableCard.story): removed story that was throwing console error until we can support this usecase without throwing console error

image

- Verify the isEditable state

### DIFF
--- a/config/jest/setup.js
+++ b/config/jest/setup.js
@@ -26,3 +26,14 @@ window.SVGPathElement = SVGPathElement;
 import sizeMe from 'react-sizeme';
 
 sizeMe.noPlaceholders = true;
+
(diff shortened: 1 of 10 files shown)


CynZhang:
Should we also disable the interactions? search, filter, sort etc..

scottdickerson:
great idea @CynZhang thanks for catching, I'll add disabling the toolbar controls if we're in this state

scottdickerson:
I can't seem to disable sort, but I will disable the other toolbar controls @CynZhang

PR #432, opened 2019-08-02
Name & prop alignment  with Carbon

davidicus:
**Summary**

This PR aligns the names and prop models between Carbon components and our extended version. The goal being to support the same props as the underlining component in addition to our extended props. This PR includes all components except `Table`

- There are a ton of file changes here, apologies.

**Related Issues**

- Part of IBM/carbon-addons-iot-react#369

- This would have to be a manual process of checking props between the two repos.


src/styles.scss
```diff
@@ -175,3 +175,5 @@ $deprecations--message: 'Deprecated code was found, this code will be removed be
 // carbon-addons-iot-react components
 //-------------------------------------
 @import './components/AddCard/add-card';
+@import './components/ProgressIndicator/progress-indicator';
+@import './components/SideNav/side-nav';
```
tay1orjones: We should add these to the rollup config to be exported/available individually

like 
https://github.com/IBM/carbon-addons-iot-react/pull/432/files#diff-ff6e5f22a9c7e66987b19c0199636480R134

davidicus:
@tay1orjones I think I have it now.

rollup.config.js
```diff
@@ -132,6 +132,11 @@ export default {
           dest: 'lib/scss/components',
         },
         { src: ['src/components/AddCard/_add-card.scss'], dest: 'lib/scss/components/AddCard' },
+        {
+          src: ['src/components/ProgressIndicator/progress-indicator.scss'],
```
tay1orjones: I think this needs an underscore

PR #517, opened 2019-08-23
Add PageTitleBar, deprecate PageHero

### LINKED ISSUE #507: [Hero] [PageHero] Deprecate/remove PageHero in favor of Hero
It appears that `PageHero` should have been deprecated and removed when `Hero` was created.

`Hero` was added during the v2 beta phase, we resolved to keep `PageHero` instead of `Hero`, which was incorrect. In following semantic versioning, removing `PageHero` would be a non-backwards-compatible change, resulting in a new major version. In light of this, I think we should keep the `PageHero` export for now, but place deprecation notice on it/remove it from the storybook.

- [x] Add `Hero` component from `1.x` branch
- [x] Add `Hero` to storybook
- [x] Add deprecation notice to `PageHero` (but still export)
- [x] Remove `PageHero` from storybook
- [ ] Align `Hero` to newly established pattern from design

tay1orjones:
**Summary**

- Add PageTitleBar
  - Ported this from `Hero` in `1.x` branch as it's quite similar. I refactored so it falls in line with the [newly established pattern from design](https://pages.github.ibm.com/Watson-IoT/IoT-Design-Site/components/ui-shell-header-name/usage)
- Deprecate PageHero in favor of PageTitleBar

- add reusable deprecation notice helper story (for consistency
- add deprecation notice to PageHero (brought in `warning` and the `dev-expression` plugin for `__DEV__` support.
- add PageTitleBar, stories, associated styles (refactored off styled-components), tests
- added a FullWidthWrapper helper component that we can use to wrap components in stories to make them full width (instead of flex/centered)
  - modified the table and tilecatalog stories to take advantage of this

**Related Issues**

- Closes #507 

- View the PageHero story in storybook, if you pull down and inverse the __DEV__ var, it'll show the warning in the console.
- View the new PageTitleBar story, make sure it follows the [newly established pattern from design](https://pages.github.ibm.com/Watson-IoT/IoT-Design-Site/components/ui-shell-header-name/usage)
- View the table and tilecatalog stories - should be unaffected by the new fullwidth helper component


tay1orjones:
@lukefirth please review the Hero design on the deploy preview storybook before we merge.

Open questions I'd like your input on:
* Editable title
  * I placed the icon in a button, which makes the icon smaller than in the spec. I think this should be wrapped in a button for a11y concerns. Thoughts?
  * What is the intended behavior for edit? Modal? In place edit? If in place, we might consider baking that into this component. For now it's just providing a hook for products to implement their own mechanism.
* Would like to confirm, there should be nothing in the tooltip except for the description? v1 previously we had links, buttons, etc. I prefer the simplicity, but curious on your thoughts.
* Double check my spacing values? I think I got it to spec, but let me know.
* Do we need the breadcrumb right content, in addition to the right content? It's a [feature in v1](https://1-x.carbon-addons-iot-react.com/?selectedKind=Hero&selectedStory=with%20breadcrumb%20with%20right%20content&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs). I left it out for now.

tay1orjones:
#### Outcomes from chatting with @lukefirth 

- New name: `PageTitleBar`
- Should support tabs as children w/ default description behavior, so consumers are put on a happy path without having to implement tabs/collaposed/tooltip logic.
  - Users are guided to the happy path, but have an escape hatch (might be helpful to document this as a design principle of this repo)
  - This Composed/with-tabs version is what most products will use
  - `Tabs` optional children prop, `node` proptypes or instance of Tabs as children?
- Editable title bar is good w/ the button, might need the icon larger like the tooltip but defer for later.
- Tooltip - _does_ need the ability for text to be a link (for help documentation, onboarding concerns, etc).

tay1orjones:
@davidicus this is ready for review

I've updated to reflect the feedback from @lukefirth.

src/components/Page/PageHero.jsx
```diff
@@ -122,23 +123,31 @@ const PageHero = ({
...
-        {rightContent ? <StyledRightContent>{rightContent}</StyledRightContent> : null}
-      </Fragment>
-    )}
-  </StyledPageHero>
-);
+}) => {
+  if (__DEV__) {
+    warning(
+      false,
+      'PageHero component has been deprecated and will be removed in the next release of `carbon-addons-iot-react`. \n Refactor to use Hero component instead.'
```
davidicus: deprecation should refer to PageTitleBar instead of Hero, correct?

davidicus:
This looks good aside from the comment above

src/components/Page/PageHero.jsx
tay1orjones: oo yes, good catch! This should be `PageTitleBar` instead of `Hero`

Issue #426, opened 2019-07-30
[v2] Styling/asset pipeline feedback and improvement requests


tay1orjones:
Potential items to improve in regards to the styling assets we provide, depending on feedback and required use cases of consumers:

- [x] Fully shadow/proxy carbon .scss files - every component/global should have a file in our folder hierarchy that imports it's carbon styles, then declares overriding styles with an appropriate namespace (so selectors are easily targetable by consumers)
  - [x] Make the necessary folder/file changes - component `.scss` files should live in the component's source folder. Globals should probably live at `./src/scss/globals`. Entrypoint should live here as well.
  - [x] Refactor rollup config - sass files should not be copied from node modules
  - [x] Utilize and extend Carbon's import de-dupe strategy. Add a sub-namespace for the styles we write to override Carbon component styles and global files/styles. [example](https://github.ibm.com/security/carbon-addons-security/blob/v2/src/globals/namespace/_index.scss)
- [ ] Provide minified css - `.min.css` and `.min.css.map`

If you have additional thoughts or feedback please leave a comment below

tay1orjones:
Over time we've accomplished these items without updating this initial issue! 😅 #1045 is the most recent iteration on this pipeline.

The only outstanding item is the minified css, and we've received no requests for it so far. I'm guessing we have no requests for it because consuming apps have their own build system that will minify the assets we provide.

PR #533, opened 2019-09-02
feat(sass): provide central entrypoint

tay1orjones:
**Summary**

- add an `index.scss` for use in consuming projects

- modify import paths in styles.scss
- add .scss files to import component styles so we're not copying as many files from `node_modules` on build
- modify rollup build to copy appropriate files to globals, components, and styles.scss
  - added a rename helper function to assist with this
- add new step to build scrips to `rm -rf` remove generated directories before a new build

**Related Issues**

- #426 

- pull down the branch, run `yarn`, view the `./lib` folder
- please double check the new component scss files for proper import paths. 
  - ex: `_accordion.scss` actually imports `@import '~carbon-components/scss/components/accordion/accordion';` and not some other component by accident.
- Are we missing anything?

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -55,7 +55,7 @@ export default {
     }),
     copy({
       targets: [
-        { src: 'styles.scss', dest: 'lib/scss/globals/scss' },
+        { src: 'src/styles.scss', dest: 'lib/scss' },
         {
           src: [
             // ---------------
@@ -95,8 +95,7 @@ export default {
             // ------------------------------
             // Carbon Components Proxy Styles
             // ------------------------------
-            // These are listed individually so that we can easily override
-            // these with our own versions of these files.
+            // These are listed individually so that we can easily disinclude individual files if we override them
             'node_modules/carbon-components/scss/components/button',
             'node_modules/carbon-components/scss/components/copy-button',
             'node_modules/carbon-components/scss/components/file-uploader',
@@ -130,7 +129,6 @@ export default {
             'node_modules/carbon-components/scss/components/tag',
             'node_modules/carbon-components/scss/components/pagination',
             'node_modules/carbon-components/scss/components/accordion',
-            'node_modules/carbon-components/scss/components/progress-indicator',
             'node_modules/carbon-components/scss/components/breadcrumb',
             'node_modules/carbon-components/scss/components/toolbar',
             'node_modules/carbon-components/scss/components/time-picker',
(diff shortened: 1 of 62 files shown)


rollup.config.js
```diff
@@ -55,102 +67,36 @@ export default {
...
-            'node_modules/carbon-components/scss/components/search',
-            'node_modules/carbon-components/scss/components/select',
-            'node_modules/carbon-components/scss/components/text-input',
-            'node_modules/carbon-components/scss/components/text-area',
-            'node_modules/carbon-components/scss/components/number-input',
-            'node_modules/carbon-components/scss/components/form',
-            'node_modules/carbon-components/scss/components/link',
             'node_modules/carbon-components/scss/components/list-box',
             'node_modules/carbon-components/scss/components/list',
             'node_modules/carbon-components/scss/components/data-table',
```
tay1orjones: @davidicus There are 5 component files we're still directly copying and importing from `node_modules`

1. `'node_modules/carbon-components/scss/components/list-box'`
1. `'node_modules/carbon-components/scss/components/list'`
1. `'node_modules/carbon-components/scss/components/data-table'`
1. `'node_modules/carbon-components/scss/components/skeleton'`
1. `'node_modules/carbon-components/scss/components/pagination-nav'`

I'm not sure what to do with these. They're either not exported directly from carbon-components-react, or we have our own version (data-table). I could create a folder and file for them like I have the others, but as we're not exporting them, this feels odd. Thoughts?

src/styles.scss
```diff
@@ -108,53 +108,52 @@ $deprecations--message: 'Deprecated code was found, this code will be removed be
...
+@import 'components/Pagination/pagination';
+@import 'components/Accordion/accordion';
+@import 'components/Breadcrumb/breadcrumb';
+@import 'components/Toolbar/toolbar';
+@import 'components/TimePicker/time-picker';
+@import 'components/Slider/slider';
+@import 'components/Tile/tile';
+@import '~carbon-components/scss/components/skeleton/skeleton'; // ???
+@import 'components/InlineLoading/inline-loading';
+@import '~carbon-components/scss/components/pagination-nav/pagination-nav'; // ???
```
tay1orjones: I placed `// ???` by them to denote which ones I'm unsure of here.

rollup.config.js
davidicus: I think its probably a good idea to create folder and file for each and import them there as its consistent and if we ever do export them as modules they will have the infrastructure in place. The data-table partial can probably be brought into our table component no? I know we export both carbon tables and our own but the styles on top of these shouldnt clash

tay1orjones: 👍 Will add folders for these so we're not copying directly from `node_modules`

tay1orjones:
@davidicus ready for review

PR #575, opened 2019-09-30
update contributing guidelines

davidicus:
update contributing guidelines to mention the need for specs to be submitted prior to acceptance

### DIFF
--- a/.github/CONTRIBUTING.MD
+++ b/.github/CONTRIBUTING.MD
@@ -15,6 +15,10 @@ To explore our components and their documentation we use storybook. To start the
 You're ready to add a new component to this library, THANK YOU! However, this is only a [PRESENTATIONAL](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) component library. Please do not add components with any dependencies on internationalization libraries, redux, or react-router.
 Feel free to add any other types of lower-level dependencies you might need for your components (i.e. lodash, etc), but please use `yarn add` to add dependencies to the project not `npm install`.
 
+## For all new components and features
+
+For a component request or feature enhancement to be considered ‘ready to implement’, a design spec and usage guidance should be provided in the issue. Associated code pull requests will not be merged without these artifacts. 
+
 ## Submitting PRs and Commits
 
 We require all commits to follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0-beta.2/) standards. You won't be able to create a commit unless you follow those rules.


.github/CONTRIBUTING.MD
```diff
@@ -15,6 +15,10 @@ To explore our components and their documentation we use storybook. To start the
 You're ready to add a new component to this library, THANK YOU! However, this is only a [PRESENTATIONAL](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) component library. Please do not add components with any dependencies on internationalization libraries, redux, or react-router.
 Feel free to add any other types of lower-level dependencies you might need for your components (i.e. lodash, etc), but please use `yarn add` to add dependencies to the project not `npm install`.
 
+## For all new components and features
+
+For a component request or feature enhancement to be considered ‘ready to implement’, a design spec and usage guidance should be provided in the issue. Associated code pull requests will not be merged without these artifacts. 
+
 ## Submitting PRs and Commits
 
 We require all commits to follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0-beta.2/) standards. You won't be able to create a commit unless you follow those rules.
```
stuckless: Would be nice if you showed a sample commit.  The only time I did a commit it took me 20 minutes to figure out the right structure of a commit.  I was used to use `feat(ISSUE_NUMBER): message` but that kept failing.  Do we include the issue number in the message?  The linked document talks about this things, but, there are additional validations that appear to be in place to validate exactly what a commit should look like... it doesn't tell you what it expects... just that it doesn't like it.  Samples would be good.

.github/CONTRIBUTING.MD
```diff
@@ -15,6 +15,10 @@ To explore our components and their documentation we use storybook. To start the
 You're ready to add a new component to this library, THANK YOU! However, this is only a [PRESENTATIONAL](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) component library. Please do not add components with any dependencies on internationalization libraries, redux, or react-router.
 Feel free to add any other types of lower-level dependencies you might need for your components (i.e. lodash, etc), but please use `yarn add` to add dependencies to the project not `npm install`.
```
stuckless: How do you go about getting IBM OS approval for these libraries?  It seems scary to tell devs to just include whatever libraries they need/want without talking about the process.  I would have expected to see something almost opposite, where, DO NOT add new libraries to the project without first providing a discussion about why it's required and has it been approved.

.github/CONTRIBUTING.MD
tay1orjones: Ah, sorry you had trouble with it! The goal is the opposite, to easily conform to a convention. We can link to the repo's commit list for some recent examples. 

Issue numbers are usually linked in the PRs over commits, but can also be included in commits if you'd like. When I've linked them in a commit, I'll avoid the message 72 character limit by putting them in the commit body:
```sh
git commit -m "fix(component): my change" -m "#123"
```

.github/CONTRIBUTING.MD
tay1orjones: We should update this to include context around how including a new entry in `dependencies` requires internal products to re-certify their open source licensing approvals, which is not always a trivial task. 

`devDependencies` are less of a concern as they're not included in the final export.

In either case, justification should accompany the PR including changes to dependencies.

.github/CONTRIBUTING.MD
stuckless: That's a good sample, it should be in the doc :)

I did do a `git log` to finally review what the actual commit message structure should look like (after a couple of failed attmpts).  VSCode is not always good, either when these git hooks fail, since it just sits there spinning away :(

PR #577, opened 2019-09-30
chore(tests): separate storyshots into individual tests

tay1orjones:
**Summary**

- Switch out `snapshotWithOptions()` for [`multiSnapshotWithOptions()`](https://www.npmjs.com/package/@storybook/addon-storyshots#multisnapshotwithoptionsoptions)
  - > _"Like snapshotWithOptions, but generate a separate snapshot file for each stories file rather than a single monolithic file (as is the convention in Jest). This makes it dramatically easier to review changes."_
- Also improves test time by ~30%! Should also speed up our Travis CI builds a bit.

- Switch out function
- Update snapshots, deletes the old monolithic file, creates new separate files.

- Does this update miss any previous components that were included in the monolithic file?


tay1orjones:
I've resolved the conflicts with `master` from the new dashboard components, and generated the new separated snapshot for the added components.

@davidicus @scottdickerson This is ready for review and merge

Issue #306, opened 2019-06-11
[Dashboard] Update all dashboard components to support Carbon v10


tay1orjones:
Dashboard/monitoring components are baked with v9 and experimental as of now.

We need to update these components to support Carbon v10 either during the v2.0.0 beta phase, or can be done after the initial cut of v2 as well.

davidicus:
We are bringing these components into master with issue #566.

tay1orjones:
As a part of a v10 design evaluation/refactor for the dashboard components, we need to refactor off of `Toolbar`, `ToolbarTitle`, `ToolbarItem`, `ToolbarOption`. I believe we only use these in the header of `Card`.

[Carbon made a mistake including them in v10](https://github.com/carbon-design-system/carbon/pull/4097), and it will be removed in v11. It would be preferred to find an alternate solution now and fix.

cc @amymillett @CynZhang @lukefirth

davidicus:
This work has been completed and any discrepancies will be dealt in individual bug issues.

PR #579, opened 2019-10-01
Dashboard upgrade

### LINKED ISSUE #489: Update to Carbon v10.6
Carbon has released [v10.5](https://github.com/carbon-design-system/carbon/releases/tag/v10.5.0) which contains some fixes that will allow us to remove some recently introduced console exceptions to `StorybookSnapshots.text.js`.

As usual we'll need to perform some manual smoke tests to validate that the release doesn't impact any existing functionality.

davidicus:
**Summary**

This PR consist of the work to bring all dashboard monitoring components into the v2/ master branch. 

- mainly all new files from the dashboard components
- upgrade to Carbon 10.6.X
- minor bug fixes from upgrade

**Related Issues**

- Closes IBM/carbon-addons-iot-react#566
- Closes #489


src/components/ImageCard/ImageControls.jsx
```diff
@@ -0,0 +1,73 @@
...
+    background: '#fff',
+    boxShadow: '0px 0px 2px 0px rgba(0,0,0,0.5)',
+  };
+  return (
+    <>
+      <div style={bottomControlsStyle}>
+        {draggable && (
+          <>
+            <button type="button" style={buttonStyle} onClick={onZoomToFit}>
+              <Icon name="icon--minimize" width="100%" height="100%" />
```
tay1orjones: We need to swap out `name` for `icon` here. I'll make a fix.

stuckless:
I'm intested in seeing how the coveralls issue gets addressed.  That's significant drop in coverage (14%).  Did the V9 not have coverage metrics, and if so, why not?  When I look at the coverage report there a files in there with coverage as low as 4%.  It makes me a little concerned if we start using this Dashboard.

src/components/Dashboard/Dashboard.story.jsx
```diff
@@ -0,0 +1,1076 @@
...
+          isLoading={boolean('isLoading', false)}
+          onBreakpointChange={action('onBreakpointChange')}
+          onLayoutChange={action('onLayoutChange')}
+          actions={[{ id: 'edit', label: 'Edit', icon: 'edit' }]}
+          onDashboardAction={action('onDashboardAction')}
+          hasLastUpdated={false}
+        />
+      </FullWidthWrapper>
+    );
+  })
```
tay1orjones: This story is broken in the preview https://deploy-preview-579--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-experimental-dashboard--custom-actions

```
TypeError: Cannot read property 'height' of undefined
    at Icon
```

src/components/Dashboard/DashboardHeader.story.jsx
```diff
@@ -0,0 +1,110 @@
...
+                onClick={action('onClickDate')}
+                onChange={action('onTextChange')}
+              />
+            </DatePicker>
+          }
+        />
+      </div>
+    );
+  })
+  .add('with custom actions', () => {
```
tay1orjones: Also broken, same issue with Icon and the undefined height

src/components/Dashboard/DashboardHeader.story.jsx
```diff
@@ -0,0 +1,110 @@
...
+          actions={[
+            { id: 'edit', labelText: 'Edit', icon: Edit },
+            { id: 'delete', labelText: 'Delete', icon: Delete },
+          ]}
+          onDashboardAction={action('onDashboardAction')}
+        />
+      </div>
+    );
+  })
+  .add('with filter and custom actions', () => {
```
tay1orjones: Also broken, same issue with Icon and the undefined height

src/components/Dashboard/DashboardHeader.story.jsx
```diff
@@ -0,0 +1,110 @@
...
+            { id: 'edit', labelText: 'Edit', icon: Edit },
+            { id: 'delete', labelText: 'Delete', icon: Delete },
+            { id: 'pin', labelText: 'Pin', icon: Pin },
+          ]}
+          onDashboardAction={action('onDashboardAction')}
+        />
+      </div>
+    );
+  })
+  .add('with custom actions component', () => {
```
tay1orjones: Also broken, same issue with Icon and the undefined height

davidicus:
@stuckless components that are marked with experimental are not covered in our general test coverage. All of these components were listed experimental so not affecting the overall test coverage threshold. 

We probably need to add experimental back to ones like pie, donut and bar charts since these are not yet using carbon-react and do not have test. 

Our first goal is to get these components into the package so we can continue with our product integration and then upgrade to Carbon v10 and write more test. Perhaps we leave all experimental tags until #306 is closed.

stuckless:
@davidicus I get that we can "hide" the technical debt here, but, should we?  I'm wondering why we are delivering compoents without coverage, and making them "experimental" doesn't really mean that they shouldn't be covered or tested.  I have a problem with the approach that we can simply hide the technical debt ang give the impression of "moving fast" when at some point, someone will have to own this debt.  I'd argue that IF the platform team is using these components in a product they should not be experimental.  And if we are delivering components that people are NOT requesting, then, we should probably not do that, and remove them.

bryancboyd:
@stuckless also one thing to note is that we do in fact have some partial test coverage on the experimental components; the coverage gap @davidicus mentioned above is just our storybook Snapshot testing, which is an issue with the `react-sizeme` lib that appears to be pretty common.  The workaround to allow snapshot testing would be to mock the `SizeMe` component, but in that case is it really a useful snapshot?  🤷‍♂  something to consider

src/components/TableCard/TableCard.jsx
```diff
@@ -0,0 +1,768 @@
...
+            pageSizes: !isExpanded ? [10] : [20],
+            isItemPerPageHidden: true,
+          },
+          toolbar: {
+            activeBar: null,
+            isDisabled: isEditable,
+            customToolbarContent: (
+              <ToolbarButton
+                kind="ghost"
+                small
```
tay1orjones: This prop has been deprecated in favor of `size="small"`

stuckless:
@bryancboyd @davidicus The code coverage metric is not meant to be used as a quality metric (ie, high coverage doesn't mean good tests).  But in the world of JavaScript high code coverage on lines and branches ensure that you are exercising the code which will help when refactoring things, etc, or simply injected syntax errors in the code.  If you can exercise the code even with trivial tests, then you can at at avoid something the blows up in the browser because something else changed, etc.  If the problem is really about SizeMe which isn't really offering much to your component, except sizing it... then I would mock it.  Since likely your component is doing something "interesting" independent of the size.  So saying that 0 coverage is ok because we have an issue sizing it... doesn't seem right.  I'm ok with ignoring coverage on parts of code that is almost impossible to test, but that should be a last resort and not the "it's experimental, so we don't have worry about it".  If you push a component to a npm registry... it's not experimental form a code and quality point of view, it's just experiemental from genual usage point of view, meaning that it might change, the quality should not be experimental.

davidicus:
@stuckless We need the components in the master branch so work can continue with the AS/Monitoring upgrade. We still have issue #306 slotted for this sprint that will address the items that would prevent these components from being production ready. I want to keep the experimental tag so people know that they are here but are not ready to consume. Once #306 is completed we can remove these from experimental.

src/components/Dashboard/Dashboard.story.jsx
tay1orjones: fixed.

src/components/Dashboard/DashboardHeader.story.jsx
tay1orjones: fixed

src/components/Dashboard/DashboardHeader.story.jsx
tay1orjones: fixed

src/components/Dashboard/DashboardHeader.story.jsx
tay1orjones: fixed

tay1orjones:
The Bar, Donut, and Pie components aren't being used from the 1.x branch. They were built using C3, which uses an SVG method that jsdom doesn't have and isn't trivial to mock.

Beside this, they are going to be rebuilt soon using the new carbon-charts [guidance](https://www.carbondesignsystem.com/data-visualization/chart-types) and soon-to-be-merged refactor. 

We've removed these components for now to reduce the churn of including them and then completely rebuilding them.

davidicus:
With the increased test coverage and removal of unwanted components we are now passing our coverage threshold. Ready to merge

Issue #595, opened 2019-10-08
[Table]: Need unique IDs on overflow menu trigger so that automation can reliably target the overflow menu


stuckless:
For automation there isn't a reliable way to target to the overflow menu per row.  IDs should be added to the overflow trigger so that automation can easily trigger the overflow to get access to the overflow actions.

tay1orjones:
Closed in the above PR

PR #599, opened 2019-10-08
feat(table): add id to overflow menu trigger

stuckless:
**Summary**

Add unique IDs to row action overflow menu trigger so that automation can easily find and trigger the overflow menu.

#595

### DIFF
--- a/src/components/Table/TableBody/RowActionsCell/RowActionsCell.jsx
+++ b/src/components/Table/TableBody/RowActionsCell/RowActionsCell.jsx
@@ -191,6 +191,7 @@ class RowActionsCell extends React.Component {
                 ))}
               {hasOverflow ? (
                 <StyledOverflowMenu
+                  id={`${id}-row-actions-cell-overflow`}
                   flipped
                   ariaLabel={overflowMenuAria}
                   onClick={event => event.stopPropagation()}
--- a/src/components/Table/TableBody/RowActionsCell/RowActionsCell.jsx
+++ b/src/components/Table/TableBody/RowActionsCell/RowActionsCell.jsx
@@ -211,11 +211,15 @@ class RowActionsCell extends React.Component {
                         itemText={
                           action.renderIcon ? (
                             <OverflowMenuContent>
-                              <StyledIcon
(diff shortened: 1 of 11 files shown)


src/components/Table/TableBody/RowActionsCell/RowActionsCell.jsx
```diff
@@ -191,6 +191,7 @@ class RowActionsCell extends React.Component {
                 ))}
               {hasOverflow ? (
                 <StyledOverflowMenu
+                  id={`${id}-row-actions-cell-overflow`}
```
tay1orjones: Should this be configurable by the user from `StatefulTable`? `Table` doesn't pass down it's `id` prop to `TableBody`.

The `id` within `RowActionsCell` here appears to be unique to only the row, not the table/page.

stuckless: It's unique per the 'key' for that was used.  The RowActionCell is not passed the table's id, so, if the table concat's it's id with the row when passing this down, thit will be uniquie, but this point, the this component is unaware of a table, just has an id.

tay1orjones: It's unaware of the table, but it's tightly coupled with it and only used within `TableBodyRow`. Passing down the table id and concatenating it with the row id will help ensure these are unique.

stuckless: TableBodyRow passes down an 'id' and that ID id just `row.id`.  The table itslef is a mess of inconsistency when it comes to IDs.  I am following the pattern you guys use of using my parent's ID to make my id, since it's unaware that it's in a table or a row..  If you want to fix all the ids in the table to be consistent and prefix with the table's ID, then that should be a separate issue.  Before I did any changes to IDs putting 2 tables on the same page with the same data resulted in duplicate IDs.  If you ever decide to retrofit IDs then nothing would have to change in this code, since, this code uses it's parent ID just like checkbox, etc.  But I agree.. the table is mess and it should be fixed, but, that is not this defect.

PR #600, opened 2019-10-08
docs(docs): add docs folder, migrate a11y guide from wiki

tay1orjones:
**Summary**

- Add folder for documentation so it can live alongside the code instead of the Github wiki. The wiki is essentially a separate repository you have to clone, which isn't ideal. 
- The structure mirrors [Carbon's docs](https://github.com/carbon-design-system/carbon/tree/master/docs).

- Add folder, migrate content from wiki, turn off wiki in repository settings.


tay1orjones:
Designer considerations images/links
image
image

PR #609, opened 2019-10-10
feat: pin all carbon dependencies

### LINKED ISSUE #606: Pin Carbon dependencies
This library is intended to be an all-in-one package - everything needed to develop a React application for IBM Watson IoT. As such, this library proxies a series of Carbon components and re-exports them.

Carbon dependencies are currently hard dependencies with caret ranges. In the event a consumer isn't using a lockfile or deletes/regenerates their lockfile, the version of Carbon used could change to be an entire minor version difference on install.

- [x] Modify Carbon dependencies to be exact versions in `package.json`. 
  - The version to pin will need to come from what's currently in the `yarn.lock`.

tay1orjones:
Closes #606 

**Summary**

- Removes the caret ranges, and updates the lockfile
- There was only one discrepancy between `package.json` and the actual version installed via `yarn.lock`.
  - `@carbon/icons-react` `v10.6.1` was installed with `package.json` set to `^10.6.0`, so I bumped the pinned version to `10.6.1`. You can see the discrepancy in the `yarn.lock` diff.

- The status checks should catch any build issues here, but this is a pretty safe change overall.
- If you want to double double confirm, pull down, run yarn, and look in `node_modules`, there you can find the version being installed. Additionally you can run `yarn why <packageName` and it'll print out the version.


davidicus:
I think I fixed the git history on this and is ready to merge once checks have passed

PR #626, opened 2019-10-22
574 table support column alignment

sunilkotiyaibm:
**Summary**

Introduced a non-mandatory align property in table columns to align table content including header, align property value could be start, center or end.

src/components/DataTable/_data-table.scss
src/components/Table/Table.story.jsx
src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
src/components/Table/TableHead/TableHead.jsx
src/components/Table/TablePropTypes.js
src/components/__snapshots__/StorybookSnapshots.test.js.snap

**Related Issues**

- Closes IBM/carbon-addons-iot-react#574

- Change the table align property (start, center or end) in Table.story.jsx for const tableColumn,  changes will reflect the respective table column in storybook-Table.
e.g. const tableColumns = [
  {
    id: 'string',
    name: 'String',
    filter: { placeholderText: 'pick a string' },
    align: 'start',
  }


src/components/Table/TablePropTypes.js
```diff
@@ -82,6 +82,7 @@ export const TableColumnsPropTypes = PropTypes.arrayOf(
     name: PropTypes.string.isRequired,
     isSortable: PropTypes.bool,
     width: PropTypes.string, // ex: 150px, or 2rem
+    align: PropTypes.string, // ex: start, center, end
```
stuckless: This should be an enum with 3 options 'start', 'center' and 'end'.

Also we should set the default props on the column to be start so that we don't need to explicitly say 'start' when defining a column.

sunilkotiyaibm: Thanks Sean, we will incorporate your suggestion.

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,62 @@
...
+    span.bx--table-header-label {
+      width: 100%;
+      text-align: end;
+    }
+  }
+}
+html[dir='rtl'] {
+  td.data-table {
+    &-start {
+      padding-right: 1.25rem;
```
tay1orjones: Spacing values like padding, margin, etc. should not be hardcoded, but instead use [sass variables provided by carbon](https://www.carbondesignsystem.com/guidelines/spacing#spacing-scale)

src/components/Table/TableHead/TableHead.jsx
```diff
@@ -178,6 +178,7 @@ const TableHead = ({
                 }
               }}
               sortDirection={hasSort ? sort.direction : 'NONE'}
+              className={`bx--table-header-label-${matchingColumnMeta.align}`}
```
tay1orjones: Class names should not use a hardcoded `bx` prefix as this namespace is configurable by the end user via `$prefix` - [Link to carbon documentation](https://www.carbondesignsystem.com/get-started/develop/vanilla#namespacing)

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,62 @@
 @import '~carbon-components/scss/components/data-table/data-table';
+
+.bx--table-header-label {
+  &-start {
+    span.bx--table-header-label {
```
davidicus: if not necessary for specificity we should use the class here without the element selector.

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -322,17 +322,17 @@ const StyledExpansionTableRow = styled(({ hasRowSelection, ...props }) => <Table
...
         min-width: ${width};
         max-width: ${width};
         white-space: nowrap;
         overflow-x: hidden;
         text-overflow: ellipsis;
-        padding-right: ${offset}px;
+        text-align: ${align};
       `
-        : `padding-right: ${props.offset}px`;
+        : `text-align:${props.align}`;
```
davidicus: should be able to use `align` instead of props.align

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -419,6 +419,8 @@ const TableBodyRow = ({
             data-offset={offset}
             offset={offset}
             width={matchingColumnMeta && matchingColumnMeta.width}
+            align={matchingColumnMeta && matchingColumnMeta.align}
+            className={col.isSortable === true ? `data-table-${matchingColumnMeta.align}` : ``}
```
davidicus: is there a reason we only do this if the row is sortable?

stuckless:
@sunilkotiyaibm any followup on the items in the code review?

src/components/DataTable/_data-table.scss
sunilkotiyaibm: Now we are able to managed using saas variable i.e.  padding-right: $spacing-09 + $spacing-02;  for 3.25rem

src/components/Table/TablePropTypes.js
sunilkotiyaibm: Hi Sean, Code is updated based on your review comments, please review.

src/components/DataTable/_data-table.scss
sunilkotiyaibm: Code is updated based on your review comments, please review.

src/components/Table/TableHead/TableHead.jsx
sunilkotiyaibm: Code is updated based on your review comments, Please review.

src/components/DataTable/_data-table.scss
sunilkotiyaibm: Code is updated based on your review comments (removed span), Please review.

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
sunilkotiyaibm: Code is updated based on your review comments, Please review.

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
sunilkotiyaibm: Alignment class added in both case 1. With sorting icon 2. Without sorting icon

sunilkotiyaibm: Code is updated based on your review comments, Please review.

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+  &-end {
+    .bx--table-header-label {
+      width: 100%;
+      text-align: end;
+    }
+  }
+}
+
+.bx--data-table td:last-of-type {
+  padding-right: $spacing-09 + $spacing-02;
```
tay1orjones: Is there a design spec to show why we're deviating from the Carbon values for padding in table cells? I suggest we use the standard values to be in line with Carbon: https://github.com/carbon-design-system/carbon/blob/c87a493f8dd86b5289fda9e59a4ee84639652028/packages/components/src/components/data-table/_data-table-core.scss#L88-L94

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+
+  &-end {
+    .bx--table-header-label {
+      width: 100%;
+      text-align: end;
+    }
+  }
+}
+
+.bx--data-table td:last-of-type {
```
tay1orjones: `bx` needs to be replaced with the prefix variable

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
 @import '~carbon-components/scss/components/data-table/data-table';
+
+.table-header-label {
+  &-start {
+    .bx--table-header-label {
```
tay1orjones: prefix variable instead of `bx`

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+      width: 100%;
+      text-align: start;
+    }
+  }
+
+  &-center {
+    .bx--table-header-label {
```
tay1orjones: prefix variable instead of `bx`

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+      width: 100%;
+      text-align: center;
+    }
+  }
+
+  &-end {
+    .bx--table-header-label {
```
tay1orjones: prefix variable instead of `bx`

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+      text-align: end;
+    }
+  }
+}
+
+.bx--data-table td:last-of-type {
+  padding-right: $spacing-09 + $spacing-02;
+}
+
+.bx--data-table td:first-of-type {
```
tay1orjones: prefix variable instead of `bx`

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+  padding-left: $spacing-04 + $spacing-02;
+}
+
+.bx--table-sort__icon-unsorted,
```
tay1orjones: prefix variable instead of `bx`

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+    padding-right: $spacing-05 + $spacing-02;
+    padding-left: $spacing-05 + $spacing-02;
+  }
+
+  &-end {
+    padding-left: $spacing-05 + $spacing-02;
+    padding-right: $spacing-09 + $spacing-02;
+  }
+
+  padding-right: $spacing-04 + $spacing-02 !important;
```
tay1orjones: please justify the use of `!important` - this shouldn't be necessary to apply the required styles atop Carbon's styling

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+      padding-right: $spacing-05 + $spacing-02;
+      padding-left: $spacing-05 + $spacing-02;
+    }
+
+    &-end {
+      padding-left: $spacing-09 + $spacing-02;
+      padding-right: $spacing-05 + $spacing-02;
+    }
+
+    padding-right: $spacing-04 + $spacing-02 !important;
```
tay1orjones: please justify the use of `!important` - this shouldn't be necessary to apply the required styles atop Carbon's styling

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+  &-center {
+    .bx--table-header-label {
+      width: 100%;
+      text-align: center;
+    }
+  }
+
+  &-end {
+    .bx--table-header-label {
+      width: 100%;
```
tay1orjones: This declaration is common among this entire block, de-dupe by placing in a selector not scoped by `&-center` or `&-end`, etc

src/components/DataTable/_data-table.scss
```diff
@@ -1 +1,67 @@
...
+}
+
+td.data-table {
+  &-start {
+    padding-right: $spacing-05 + $spacing-02;
+    padding-left: $spacing-05 + $spacing-02;
+  }
+
+  &-end {
+    padding-left: $spacing-05 + $spacing-02;
```
tay1orjones: This declaration is common among `&-start`, `&-end`, and the rtl `&-start` below. De dupe this by placing in a unified selector. It seems that all table cells with `start` should have this left padding.

src/components/DataTable/_data-table.scss
sunilkotiyaibm: We are not able to align table column header and its respective data in cells while using standard spacing. i.e for 3.25 rem for which we used $spacing-09 + $spacing-02, but if we use $spacing-09 only then it is not aligning as showing in attached screenshot shared over slack

src/components/DataTable/_data-table.scss
sunilkotiyaibm: Ok replaced with #{$prefix}

src/components/DataTable/_data-table.scss
sunilkotiyaibm: Ok replaced with #{$prefix}

src/components/DataTable/_data-table.scss
sunilkotiyaibm: Ok replaced with #{$prefix}

src/components/DataTable/_data-table.scss
sunilkotiyaibm: Ok replaced with #{$prefix}

src/components/Table/Table.story.jsx
```diff
@@ -66,30 +66,36 @@ export const tableColumns = [
     id: 'string',
     name: 'String',
     filter: { placeholderText: 'pick a string' },
+    align: 'start',
```
tay1orjones: Please create a separate story to showcase this functionality and scope the feature to only that story instead of globally for all stories. Ideally this parameter would also be a knob so users can play with the different states of `start`, `end`, and `center`

tay1orjones:
@JordanWSmith15 @KarenGski @lukefirth A question for design - based on this spec, there is a question regarding the right alignment present in this PR.
image

This shows right alignment should take into account the sorting icon and align content with the label, but this seems nonstandard to me.

When left and center aligned, the sorting icon size/spacing is not taken into account. 
image

Additionally, when center aligned the content is not perfectly aligned with the label. The label may be smaller/larger than the cell contents. It seems continuing the pattern would direct us to not take the sorting icon into account when padding the right alignment. Thoughts?

JordanWSmith15:
@tay1orjones @KarenGski @lukefirth Personally, I like aligning the text perfectly and ignoring the icon, but I'm fine with changing the design here, if it's easier to code it as Taylor is suggesting. We just need to make sure that we account for columns that are right-aligned and not sortable. In that case, we need to align the text. Here's an example: 
Screen Shot 2019-10-31 at 10 16 22 AM

src/components/DataTable/_data-table.scss
davidicus: I think we need to just use the `text-align` property here. The padding of these cells should use the standard padding we get from the Carbon component.

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -419,6 +421,8 @@ const TableBodyRow = ({
             data-offset={offset}
             offset={offset}
             width={matchingColumnMeta && matchingColumnMeta.width}
+            align={align}
+            className={col.isSortable === true ? `data-table-${align}` : `data-table`}
```
davidicus: we should use the classnames module to handle these.

src/components/DataTable/_data-table.scss
davidicus: I have refactored all the padding changes out of this PR

src/components/DataTable/_data-table.scss
davidicus: done.

src/components/Table/Table.story.jsx
davidicus: done.

davidicus: Did not incorporate a knob though

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
davidicus: Not sure why isSortable is tied with cell alignment. I have removed the tertiary and align defaults to `start`

src/components/DataTable/_data-table.scss
davidicus: done

src/components/DataTable/_data-table.scss
davidicus: This has been removed.

davidicus:
Previously we seriously deviating from the Carbon defaults for padding. I have refactored the padding to use the Carbon base styles. Only padding added now is to resolve RTL issues.

sunilkotiyaibm:
> Previously we seriously deviating from the Carbon defaults for padding. I have refactored the padding to use the Carbon base styles. Only padding added now is to resolve RTL issues.



> Previously we seriously deviating from the Carbon defaults for padding. I have refactored the padding to use the Carbon base styles. Only padding added now is to resolve RTL issues.

Thanks David for refactoring the alignment code and adding the necessary changes.
In case of RTL, I added padding (td.data-table) to align table content with header.  Please have a look.

sunilkotiyaibm:
@davidicus @tay1orjones 
Please review updated changes in _data_table.scss for RTL case.

sunilkotiyaibm:
@davidicus @tay1orjones Please review the PR (574 table support column alignment
#626), fixed few alignment issue in _data_table.scss and updated storybook for alignment with few combination of Alignment and sorting.

PR #632, opened 2019-10-24
align Header component to spec

### LINKED ISSUE #541: Header App name size not in line with pattern
According to pattern on the [IoT design site](https://pages.github.ibm.com/Watson-IoT/IoT-Design-Site/components/ui-shell-header/usage#anatomy) the App header needs to be changed from the Carbon default to a `16px/1rem` size.

image

This issue is complete when we have overwritten the current `a.bx--header__name { font-size: 0.875rem;}` to be `a.bx--header__name{ font-size: 1rem; }`

### LINKED ISSUE #542: [Header] Action button icons do not follow spec
image

The flexibility to add other content aside from an icon was built into the component. This flexibility is not in line with the navigation [pattern](https://pages.github.ibm.com/Watson-IoT/IoT-Design-Site/components/ui-shell-header/usage#header-elements).

Should this flexibility be removed from the component or should the pattern be updated to be less prescriptive about the items placed here?

davidicus:
Closes #542 and closes #541

**Summary**

Aligning header component to spec. 

- changed header styles. 
- refactored syled components to sass

check sizes of app header and action buttons


src/components/Header/Header.jsx
```diff
@@ -159,29 +112,30 @@ const Header = ({
       );
     }
     return (
-      <StyledGlobalAction
+      <HeaderGlobalAction
+        className="header-action-btn"
```
tay1orjones: Do we need to begin to bake in the `$prefix` to this classname?

src/components/Header/_header.scss
```diff
@@ -0,0 +1,50 @@
...
+.#{$prefix}--header {
+  .#{$prefix}--skip-to-content:focus {
+    min-width: 240px;
+    justify-content: center;
+    border: $spacing-01 solid $ui-01;
+  }
+
+  &__menu-toggle {
+    display: block;
+    color: red;
```
tay1orjones: This should be a color token var. Is the red intended in the spec?

davidicus: that was a testing style and should not be included. will remove.

src/components/Header/Header.jsx
davidicus: i had thought about this but was worried that it may someday be used by carbon and over ride our styles.

davidicus: Not opposed to adding it though

PR #634, opened 2019-10-25
fix(imagecard): expansion and collapse doesn't work on firefox

### LINKED ISSUE #638: [TimeSeries] When the is editable bit is set the TimeSeries card isn't showing fake data
### Describe the bug

when the isEditable bit is set, the TimeSeries card isn't showing fake data

### To Reproduce

View the timeseries isEditable story, that should show fake data in the timeseries card

scottdickerson:
Closes #638

**Summary**

- Internal Bug reported where the Image Card didn't collapse from full screen correctly on firefox

- refactor(ImageHotspots): pretty large refactor to remove the local container state from ImageHotspots and instead inherit the size from the Card content parent (which correctly handles Firefox)
- fix(TimeSeriesCard): isEditable is no longer showing fake data

- Verify in Firefox the full screen image card dashboard story.  Expand and then collapse the image and ensure it is correctly positioned

### DIFF
--- a/src/components/Card/Card.jsx
+++ b/src/components/Card/Card.jsx
@@ -62,6 +62,7 @@ export const CardHeader = styled.div`
 export const CardContent = styled.div`
   flex: 1;
   position: relative;
+  overflow: hidden;
   height: ${props => props.dimensions.y - CARD_TITLE_HEIGHT}px;
 `;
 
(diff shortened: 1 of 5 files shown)


scottdickerson:
@tay1orjones that was actually a bug just in general with our zoom in and collapse, we should probably open a separate bug to track

tay1orjones:
I've opened https://github.com/IBM/carbon-addons-iot-react/issues/648 for the issue found above.

PR #645, opened 2019-10-26
fix(card): make card header consistent height/size

bryancboyd:
**Summary**

With the larger touch area for carbon 10 buttons, the card header height increased due to extra button padding.  Moving the toolbar buttons to absolute positioning (right aligned) is one possible fix.

Before:
image

After:
image

- Verify the Card header is the same height whether or not any toolbar buttons are visible


scottdickerson:
this is going to cause problems truncating the title unless you reserve space for the absolute buttons:

image

bryancboyd:
@scottdickerson thanks -- using line-height in place of vertical padding in the header now to set a "fixed" height that keeps icons in flex flow

src/components/Card/_card.scss
```diff
@@ -4,7 +4,7 @@
 .card {
   &--title {
     font-size: rem(14px);
-    line-height: rem(18px);
+    line-height: rem(48px);
```
tay1orjones: Type styles for this can be replaced with
```
@include type-style('productive-heading-01');
```

tay1orjones:
@scottdickerson long titles look correct to me
image

Issue #689, opened 2019-11-06
[Table] In table toolbar search icon should be right align and also display table record count (left align) 


sunilkotiyaibm:
[Table] In table toolbar search icon should be right align and also display  table record count (left align) (Refer attached design spec)
 
![TableToolbar](https://user-images.githubusercontent.com/56038154/68297832-f0621480-00bd-11ea-8ccd-5fa73e7f37fb.png)

creckling:
pal-tracking issue: https://github.ibm.com/Watson-IoT/pal-tracking/issues/34

PR #692, opened 2019-11-06
feat(table): 3287-Table-SearhIcon-align-Row-Count

sunilkotiyaibm:
#689

Closes #

**Summary**

Right align the search icon and toggle open when clicked to display a search component.
Show table row count and should be appear on the left. i.e "Results: X", where X is equal to the number of rows in the table.
This field (Row count) should be configuration via boolean property and Result Label should allow to change.

src/components/DataTable/_data-table.scss
src/components/Table/Table.jsx
src/components/Table/Table.story.jsx
src/components/Table/TablePropTypes.js
src/components/Table/TableToolbar/TableToolbar.jsx
src/components/Table/snapshots/Table.story.storyshot
src/components/TableCard/snapshots/TableCard.story.storyshot
src/components/Dashboard/snapshots/Dashboard.story.storyshot

Two fields are provided in Knob to test.
Search Icon should be display as right align in table toolbar and when clicked it should display a search component. 
Show total row count in table, 
Using knob, Row count label input field should allow to change row count label value, also show row count checkbox use to display and hide Row count field.

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -269,8 +269,7 @@ export const defaultProps = baseProps => ({
     batchCancel: 'Cancel',
     itemsSelected: 'items selected',
     itemSelected: 'item selected',
-    rowCountLabel: 'Results',
-    rowCountInHeader: (totalRowCount, rowCountLabel) => `${rowCountLabel}: ${totalRowCount}`,
+    rowCountInHeader: totalRowCount => `Results: ${totalRowCount}`,
     /** empty state */
     emptyMessage: 'There is no data',
     emptyMessageWithFilters: 'No results match the current filters',
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -349,10 +349,8 @@ const Table = props => {
           filterNone: i18n.filterNone,
           filterAscending: i18n.filterAscending,
           filterDescending: i18n.filterDescending,
-          rowCountHeaderField: i18n.rowCountInHeader(
-            view.pagination.totalItems,
-            i18n.rowCountLabel
-          ),
+          rowCountLabel: i18n.rowCountLabel,
+          rowCountInHeader: i18n.rowCountInHeader,
         }}
         actions={pick(
           actions.toolbar,
@@ -374,6 +372,7 @@ const Table = props => {
         tableState={{
           totalSelected: view.table.selectedIds.length,
           totalFilters: view.filters ? view.filters.length : 0,
+          totalItemsCount: view.pagination.totalItems,
           ...pick(
             view.toolbar,
             'batchActions',
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -84,6 +84,7 @@ const propTypes = {
     hasSearch: PropTypes.bool,
     hasColumnSelection: PropTypes.bool,
     shouldLazyRender: PropTypes.bool,
+    hasRowCount: PropTypes.bool,
   }),
 
   /** Initial state of the table, should be updated via a local state wrapper component implementation or via a central store/redux see StatefulTable component for an example */
@@ -346,6 +347,7 @@ const Table = props => {
           filterNone: i18n.filterNone,
           filterAscending: i18n.filterAscending,
           filterDescending: i18n.filterDescending,
+          rowCountLabel: i18n.rowCountLabel,
         }}
         actions={pick(
           actions.toolbar,
@@ -356,10 +358,18 @@ const Table = props => {
           'onToggleFilter',
           'onApplySearch'
         )}
-        options={pick(options, 'hasColumnSelection', 'hasFilter', 'hasSearch', 'hasRowSelection')}
+        options={pick(
+          options,
+          'hasColumnSelection',
+          'hasFilter',
+          'hasSearch',
+          'hasRowSelection',
+          'hasRowCount'
+        )}
         tableState={{
           totalSelected: view.table.selectedIds.length,
           totalFilters: view.filters ? view.filters.length : 0,
+          totalItemsCount: view.pagination.totalItems,
           ...pick(
             view.toolbar,
             'batchActions',
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -84,7 +84,7 @@ const propTypes = {
     hasSearch: PropTypes.bool,
     hasColumnSelection: PropTypes.bool,
     shouldLazyRender: PropTypes.bool,
-    hasRowCount: PropTypes.bool,
+    hasRowCountInHeader: PropTypes.bool,
   }),
 
   /** Initial state of the table, should be updated via a local state wrapper component implementation or via a central store/redux see StatefulTable component for an example */
@@ -269,6 +269,8 @@ export const defaultProps = baseProps => ({
     batchCancel: 'Cancel',
     itemsSelected: 'items selected',
     itemSelected: 'item selected',
+    rowCountLabel: 'Results',
+    rowCountInHeader: (totalRowCount, rowCountLabel) => `${rowCountLabel}: ${totalRowCount}`,
     /** empty state */
     emptyMessage: 'There is no data',
     emptyMessageWithFilters: 'No results match the current filters',
@@ -347,7 +349,10 @@ const Table = props => {
           filterNone: i18n.filterNone,
           filterAscending: i18n.filterAscending,
           filterDescending: i18n.filterDescending,
-          rowCountLabel: i18n.rowCountLabel,
+          rowCountHeaderField: i18n.rowCountInHeader(
+            view.pagination.totalItems,
+            i18n.rowCountLabel
+          ),
         }}
         actions={pick(
           actions.toolbar,
@@ -364,12 +369,11 @@ const Table = props => {
           'hasFilter',
           'hasSearch',
           'hasRowSelection',
-          'hasRowCount'
+          'hasRowCountInHeader'
         )}
         tableState={{
           totalSelected: view.table.selectedIds.length,
           totalFilters: view.filters ? view.filters.length : 0,
-          totalItemsCount: view.pagination.totalItems,
           ...pick(
             view.toolbar,
             'batchActions',
--- a/src/components/Table/TablePropTypes.js
+++ b/src/components/Table/TablePropTypes.js
@@ -158,6 +158,8 @@ export const I18NPropTypes = PropTypes.shape({
   filterNone: PropTypes.string,
   filterAscending: PropTypes.string,
   filterDescending: PropTypes.string,
+  rowCountLabel: PropTypes.string,
+  rowCountInHeader: PropTypes.func,
 });
 
 export const defaultI18NPropTypes = {
@@ -199,6 +201,7 @@ export const defaultI18NPropTypes = {
   filterAscending: 'Sort rows by this header in ascending order',
   filterDescending: 'Sort rows by this header in descending order',
   rowCountLabel: 'Results',
+  rowCountInHeader: (totalRowCount, rowCountLabel) => `${rowCountLabel}: ${totalRowCount}`,
 };
 
 export const TableSearchPropTypes = PropTypes.shape({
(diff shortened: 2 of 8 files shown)


sunilkotiyaibm:
@davidicus @tay1orjones @stuckless  Please review this PR.

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -173,16 +184,21 @@ const TableToolbar = ({
...
-      <StyledToolbarSearch
-        {...search}
-        translateWithId={(...args) => tableTranslateWithId(i18n, ...args)}
-        id={`${tableId}-toolbar-search`}
-        onChange={event => onApplySearch(event.currentTarget ? event.currentTarget.value : '')}
-        disabled={isDisabled}
-      />
+    {hasRowCount ? (
+      <ToolBarResultLabel>
+        {i18n.rowCountLabel}: {totalItemsCount}
```
stuckless: We can't do this for labels it will ultimately run into BIDI issues and other potential formatting issues.  The rowCountLabel should be a label like `Count: {0}` and we transform the message + arg into a final string.  Not sure how the table handles this, so might require some research.

stuckless:
@JordanWSmith15 @lukefirth There is a property on the table options for `showRecordCount` but this is ONLY affecting the record count in the header.  So a couple of quetions...
1. should this property only affect showing a record count in the header, and, it will always show the 1-N record count in the footer.  (if yes, then we rename this property to `showRecordCountInHeader`, if no, then we need to update the code to hide the `1-n` count in the footer if this value is false.
2. From a UX point of view does it make sense to show counts in the header and footer?

@davidicus @tay1orjones @bryancboyd @scottdickerson We should look at how we can inject a data formatter into the table (and possibly other components).  Right now the UI is inconsistent when rendering numbers because, we can format some data in the table (using a cell renderer) but in other places where we are showing numbers, likse "1 of 100000 items" we can't format that.  Or, in the header where we are showing "Records: 100000" we can't format that.  It would seem to me that a component level we should allow some sort of data formatter to be injected as part of the configuration.  The "default" might simply be a `String(val)` but, at least we could possibly inject a custom formatter if needed.

JordanWSmith15:
@stuckless For your first question: Part of the requirements was to make the record count in the table header optional, so the app dev should be able to hide it. I'm guessing a bit here, but if the page does not have pagination because the page size is large, such as https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--simple-stateful-example-with-alignment, you would want the result count in the header (since there is no footer).

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -70,6 +70,14 @@ const StyledTableBatchActions = styled(TableBatchActions)`
   }
 `;
 
+const ToolBarResultLabel = styled.label`
```
davidicus: Since we are moving away from styled components these style should be added to a class that is on your label element.

src/components/Table/TableToolbar/TableToolbar.jsx
tay1orjones: Instead of a string, this can be a function that returns a string. A good example of this pattern is the [`itemRangeText` prop in the Carbon `<Pagination>`](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Pagination/Pagination.js#L134)

tay1orjones:
> Right now the UI is inconsistent when rendering numbers because, we can format some data in the table (using a cell renderer) but in other places where we are showing numbers, likse "1 of 100000 items" we can't format that. Or, in the header where we are showing "Records: 100000" we can't format that. It would seem to me that a component level we should allow some sort of data formatter to be injected as part of the configuration.

@stuckless Those examples you listed should be modified to be functions that return a string, rather than a direct string. [Example from the `itemRangeText` prop in the Carbon `<Pagination>`](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/Pagination/Pagination.js#L134)

sunilkotiyaibm:
@stuckless @davidicus @tay1orjones  Thanks for your review comments, I will implement the given suggestion.

stuckless:
@tay1orjones I know that IoT uses functions as a way to return localized text, but that problematic.  How is a translation team supposed to translate a function?  In every translation system that I've seen, the localization part is just text with paramater placeholders and then some code inject dynamic parameters.  So the part that gets sent away for localization is just text.

instead of
```
rowCountLabel: (total) => `Results: ${total}`
```
it would be (java like)
```
rowCountLabel: 'Results: {0}'
```
or (react i18n like)
```
rowCountLabel: 'Results: {{total}}'
```
and some other proces basically has to scan the text and replace variables.

In the most simplest form the code to do the replacements is not hard, and, it ensures that i18n strings are in fact strings.

```
  /**
   * Replace placeholder parameters in string with actual values.
   *
   * <pre>
   * let msg = replaceParams('Hello {0}', ['world']);
   * would output 'Hello world'
   * </pre>.
   *
   * @param {string} message - The message string with optional {0}, {1}, {n} parameters.
   * @param {Array} params - Array containing strings to replace placeholders.
   */
  replaceParams(message, params) {
    if (message && params) {
      params.forEach((param, index) =>  {
        message = message.replace('{' + index + '}', param);
      });
    }
    return message;
  }
```

And yeah a function like that should be tightened up, because it has holes, but, this could be in a reusalbe util.js file and then included and used in a component when needed.

This is more of a fyi on how we could/should be handling localization and not relying on functions (which can't be localized))

sunilkotiyaibm:
@davidicus @tay1orjones @stuckless  Please review the updated code based on review comments.

bryancboyd:
@stuckless Regarding translation, I don't see functions under i18n as anything fundamentally different from the approaches you outlined above.  In our application usage we typically wrap the IoT Component in a utility that defines strings and props for the "common" use of that component.  Here's an example for the table:

```
import { injectIntl, defineMessages } from 'react-intl';
import { StatefulTable } from 'carbon-addons-iot-react';

const messages = defineMessages({
  selectAll: 'Select all',
  ...
  clickToCollapse: 'Click to collapse',
  itemsRange: '{min}–{max} items',
  currentPage: 'page {page}',
  itemsRangeWithTotal: '{min}–{max} of {total} items',
  ...
});

const CommonTable = ({ intl: { formatMessage }, i18n, options, columns, ...props }) => (
  <StatefulTable
    {...props}
    columns={columns.map(column => ({
      ...column,
      filter:
        column.filter || column.isFilterable
          ? { placeholderText: formatMessage(messages.filterColumn) }
          : null,
    }))}
    options={
      options
        ? merge({}, options, {
            hasRowSelection: options.hasRowSelection === true ? 'multi' : options.hasRowSelection,
          })
        : undefined
    }
    i18n={merge(
      {},
      {
        /** pagination */
        pageBackwardAria: formatMessage(messages.previousPage),
        pageForwardAria: formatMessage(messages.nextPage),
        ...
        itemsRangeWithTotal: (min, max, total) =>
          formatMessage(messages.itemsRangeWithTotal, { min, max, total }),
        pageRange: (current, total) => formatMessage(messages.pageRange, { current, total }),
        /** table body */
        overflowMenuAria: formatMessage(messages.moreActions),
      },
      i18n // use whatever props come in
    )}
  />
);
export default injectIntl(CommonTable);
```

bryancboyd:
@stuckless I guess the change you are suggesting is that we standardize to where i18n objects passed to IoT components are raw strings in a "react-y" format `itemsRangeWithTotal: '{min}–{max} of {total} items'`, and then have a common util function that does the translation inside of components?  The one benefit to exposing i18n props as functions is that it provides a clear signature for use by consuming components... otherwise I'd have to know that my `i18n.itemsRangeWtihTotal` prop needs to include some specific named `{value}`s.

I'm not picky on the approach we take, but I agree we should be consistent here. @tay1orjones could you open a separate issue to discuss our approach for standardizing i18n strings with parameters across the library?

(Also, what does Carbon do?)

tay1orjones:
Yeah, the translation team wouldn't be translating a function, but instead the keys/strings you provide them via your i18n library. These are then passed to the components, the function params provide the ability to 'replace' the dynamic parts of the string yourself. 

I've opened #714 to continue the conversation

src/components/Table/TablePropTypes.js
```diff
@@ -198,6 +200,8 @@ export const defaultI18NPropTypes = {
   filterNone: 'Unsort rows by this header',
   filterAscending: 'Sort rows by this header in ascending order',
   filterDescending: 'Sort rows by this header in descending order',
+  rowCountLabel: 'Results',
```
tay1orjones: I think you can remove `rowCountLabel` here, as well as the other locations. It's not being used now that `rowCountInHeader` takes a param of `rowCountLabel`

sunilkotiyaibm: @tay1orjones rowCountLabel: 'Results' is removed from TablePropTypes.js but same is using in Table.jsx because it's default value is passing in function  rowCountHeaderField: i18n.rowCountInHeader(
            view.pagination.totalItems,
            i18n.rowCountLabel
          ),
If we are not passing label value from storybook then it will set the from default props and I tested the same. 
Please review the same.

src/components/Table/Table.jsx
```diff
@@ -268,6 +269,8 @@ export const defaultProps = baseProps => ({
     batchCancel: 'Cancel',
     itemsSelected: 'items selected',
     itemSelected: 'item selected',
+    rowCountLabel: 'Results',
```
stuckless: I would just leave the rowCountLabel as a function (which is what you have previous).  My discussion with taylor was more about how we should find a better way for all component to manage this, but, using a function for the label is what we have today, and as such that should be used.  Right now, you've just turned 1 label prop into 3 which isn't better :(

src/components/Table/Table.jsx
```diff
@@ -346,6 +349,10 @@ const Table = props => {
           filterNone: i18n.filterNone,
           filterAscending: i18n.filterAscending,
           filterDescending: i18n.filterDescending,
+          rowCountHeaderField: i18n.rowCountInHeader(
```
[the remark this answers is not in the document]
sunilkotiyaibm: rowCountInHeader func calling moved in Toolbar.jsx

src/components/Table/Table.jsx
sunilkotiyaibm: rowCountInHeader func calling moved in Toolbar.jsx and removed rowCountHeaderField. Please review again.

sunilkotiyaibm:
@stuckless @tay1orjones latest suggestion are incorporated in Table.jsx and TableToolbar.jsx, Please review.

src/components/Table/Table.jsx
stuckless: I think we are still missing something here.   Just get rid of rowCountLabel since on it's own, it is not used.  You are later passing down into another function and then telling that function to compose some items plus a base label, and you should never do that from a translation point of view.  All I'm saying is that, remove all reference rowCountLabel then then in your rowCountInHeader function label, just have it be defined like...
```
rowCountInHeader: (totalRowCount, rowCountLabel) => `Results: ${totalRowCount}`,
```

The reason why you should not pass a label into a translation function is what if I wanted to display `247 Items Returned` instead of `Results: 247`.  The way you have this setup, it would basically pass `Items Returned` to the format function and then we'd have `Items Returned: 247`.

In essent, just treat your label function as if it were a static piece of text where you'll insert some number.  No need to break it up into smaller parts and then try to comose it, since, translation will fail miserable on that kind of composition of labels.

sunilkotiyaibm: @stuckless suggested changes are added in latest commit. please review.

davidicus:
Yes, Sunil you should just have to pull master into your branch and resolve the conflict.

sunilkotiyaibm:
> Sunil this looks good to me. Unfortunately there is a merge conflict with the Table snapshot and I think the only way to resolve that is to merge with master. @davidicus is that how you have to deal with this? Also I see the coverage now fails. Based on the tests becuase hasRowCountInHeader is an option, you shold be adding tests to verify that it shows or not shows. (sort of like what you did with the label)

@stuckless  new PR (https://github.com/IBM/carbon-addons-iot-react/pull/743) raised as yarn test was failing on windows, so I created a new branch and raised new PR.
Please review.

sunilkotiyaibm:
> Yes, Sunil you should just have to pull master into your branch and resolve the conflict.

@davidicus I have some issue related to yarn test which was failing on windows. so created new branch and PR from Team member Mac machine. Please review PR 743.
https://github.com/IBM/carbon-addons-iot-react/pull/743

sunilkotiyaibm:
New PR is raised so closing this PR and deleting the branch 
New PR : https://github.com/IBM/carbon-addons-iot-react/pull/744

PR #697, opened 2019-11-07
feat(tablecard): add width to proptypes that can be passed in for columns

scottdickerson:
Supports new width string proptype and field that can be passed in to limit a column width
image

### DIFF
--- a/src/constants/PropTypes.js
+++ b/src/constants/PropTypes.js
@@ -83,8 +83,8 @@ export const TableCardPropTypes = {
     columns: PropTypes.arrayOf(
       PropTypes.shape({
         dataSourceId: PropTypes.string.isRequired,
-        /** optional width, default is no enforced max width.  for example 150px */
-        width: PropTypes.string,
+        /** optional width in pixels, default is no enforced max width */
+        width: PropTypes.number,
         label: PropTypes.string.isRequired,
         priority: PropTypes.number,
         renderer: PropTypes.func,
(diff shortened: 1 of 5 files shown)


src/constants/PropTypes.js
```diff
@@ -83,6 +83,8 @@ export const TableCardPropTypes = {
     columns: PropTypes.arrayOf(
       PropTypes.shape({
         dataSourceId: PropTypes.string.isRequired,
+        /** optional width, default is no enforced max width.  for example 150px */
+        width: PropTypes.string,
```
tay1orjones: How are we going to handle these without styled-components? Imperative control with css custom properties?

Additionally, if we're placing these directly into styles without sanitation and we require the unit suffix, it might be a good idea to put a custom validator on it

```
width: function(props, propName, componentName) {
    if (!/(\d*)px/.test(props[propName])) {
      return new Error(
        'Invalid prop `' + propName + '` supplied to' +
        ' `' + componentName + '`. Must contain digits with `px` suffix, ex `125px`'
      );
    }
  },
```

Not 100% on that regex, you might doublecheck it 😅

src/utils/schemas/tableCardContent.json
```diff
@@ -64,12 +64,13 @@
       "properties": {
         "label": { "type": "string" },
         "dataSourceId": { "type": "string" },
+        "width": { "type": "string" },
```
tay1orjones: In tandem with my other comment about a custom validator, have we thought about more narrow types to support specific formats of strings (or other types) within the json config?

PR #698, opened 2019-11-07
Custom card

leetosc:
Closes #

**Summary**

- Add `CustomCard` component for use in `Dashboard` that takes in and displays a jsx element

- CustomCard

- tests here


src/components/Dashboard/CardRenderer.jsx
```diff
@@ -71,6 +72,7 @@ const CardRenderer = React.memo(
     onFetchData, // eslint-disable-line
     onSetupCard, // eslint-disable-line
     timeGrain, // eslint-disable-line
+    onClick, //eslint-disable-line
```
tay1orjones: I think we should establish the pattern of setting a noop as a default for optional function props so we can safely call these functions without having to check if they're `null`.

So in the defaults it's just
```
onClick: () => {}
```

src/components/CustomCard/_customCard.scss
```diff
@@ -0,0 +1,9 @@
+.customCard {
+  .card--header {
+    display: none;
+  }
+  overflow: auto;
+  &:hover {
+    box-shadow: 0 0 4px blue;
```
tay1orjones: Need to use a color variable here. @CynZhang @lukefirth @matsgothe which color should the cards display on hover?

Additionally, since these are clickable we'll need to establish focus styles and ensure these are accessible.

jest.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     global: {
       statements: 80,
       branches: 73,
-      functions: 80,
+      functions: 78,
```
tay1orjones: Can we add additional tests for this component to meet the threshold instead of decreasing the threshold?

src/components/Dashboard/CardRenderer.jsx
leetosc: default noop function prop on the `CustomCard` component

src/components/CustomCard/_customCard.scss
leetosc: updated to `$interactive-01`

jest.config.js
leetosc: Scott said to just push it for now

src/components/Dashboard/CardRenderer.jsx
tay1orjones: Do we need to also declare defaults within this component to remove the eslint disable?

src/components/CustomCard/_customCard.scss
tay1orjones: Also need something for `:focus` - we could follow the pattern of [Carbon's clickable tile](http://react.carbondesignsystem.com/?path=/story/tile--clickable). It's very similar in interaction to this component. They apply a 2px border w/ the same color. Is the `box-shadow` dictated from the design spec? There's very isolated use cases of box shadow within Carbon.

davidicus: yes, box shadows always feel very out of place with the flat Carbon design.

lukefirth: I want to be clear on a few things so I'll outline them together (some may not apply here so forgive me):

- Not all dashboard cards should have a hover state or interaction at the card level. In fact most will not have this (eg line charts, tables etc) as their interaction is on components within the card. (Not even the 'disabled' style/cursor as this would imply that it could at some point be enabled, which they can't).

- Those that do have the 'whole card' clickable (eg: the homepages key tasks) should take their styling for states from the Carbon tile where they exist.

- The purpose of the custom card is to allow a user to build their own one of our cards by injecting code into a tile. As we cannot tell if the creator of a custom card would like the whole card to be interactable. The wisest decision would be to have a prop like the Carbon tile already does, that can be set by the creator and would control if, and what the whole card interactivity would be. When it is set to true, then it should take on the styling of cards we have built with that ability (which in turn are taking on the carbon tiles styling for it).

- A caveat here is that when a dashboard of any kind (home or traditional) is in 'edit' mode, all cards will need a state as they are now interactable at the whole card level (to drag them around). Again I think the carbon tiles styling tools reasonable for this (when dragging it should have the solid border from the selected state).

bryancboyd: @leetosc Luke raises a good point, let's add a prop on `CustomCard` (maybe `hasHoverState`, default false) that will show the Carbon tile hover state (what @tay1orjones said above) on the rendered `<Card>`

tay1orjones: With that in mind, how is this overall different than from using `<Card>` and passing in custom content as children? The only prop here is `content`, the rest are passed through to `Card`. `onClick`, `isExpanded` etc are already props on `Card`.

bryancboyd:
@tay1orjones @leetosc following our offline conversation, here's the new direction we should take:

- we'll use `<Card>` as the fallback renderer in order to display custom card content
- if `title` and `availableActions` are both undefined in the card config, we will hide the card header
- in lieu of providing a hover state for the "card as a link" use case, we'll include a few stories that use `<Tile clickable>` to wrap the custom card content

bryancboyd:
This is replaced by #716

Issue #666, opened 2019-11-01
[PageWizard] add prop to render buttons in fixed footer


bryancboyd:
### Summary

Add a prop to `PageWizard` that will render the action buttons in a sticky page footer instead of directly under step content

**Additional context**

![image](https://user-images.githubusercontent.com/2175647/68031563-ee492180-fc89-11e9-9b90-a8a3c63b34dc.png)

### Specific timeline issues / requests

1-2 weeks

### Available extra resources

@jessieyan

JordanWSmith15:
@bryancboyd @lukefirth @creckling Let's make sure those buttons don't get coded with the Primary on the left

jessieyan:
Add a prop hasStickyFooter to PageWizard. 

If set to true, step actions will show up inside the sticky footer at the bottom of the page. 
If false, step actions button will show inside of step content.


![image](https://user-images.githubusercontent.com/24279794/68497327-ece3a000-0219-11ea-8fe5-b14c12f73d0b.png)

![image](https://user-images.githubusercontent.com/24279794/68497369-0258ca00-021a-11ea-982d-5d7626545fa3.png)

PR #699, opened 2019-11-08
fix(pagewizard): add sticky footer

jessieyan:
Tracker #666 

Add a prop hasStickyFooter to PageWizard. 

If set to true, step actions will show up inside the sticky footer at the bottom of the page. 
If false, step actions button will show inside of step content.

image

image


bryancboyd:
image

added a few tweaks to the footer styling

davidicus:
Does this component have any design specs for it? I am wondering about the deviation in button size and placement from other components that are similar.

bryancboyd:
Good points, @stuckless. This component (and the larger Form pattern) are
still in the design phase so I’m hesitant to make too many opinionated
behavioral choices now. Let’s push this through and open an issue in the
design repo to clarify this?

stuckless:
@bryancboyd Sounds good.  I think it should be accounted for at some level.  Ideally their might a layout component that knows about sticky header, sticky footer and provides automatic scolling (or not), but if it existed, then, it could dropped into these kinds of template components and provide the behaviours automatically.

bryancboyd:
@stuckless That's a solid idea; we have a shortage of layout components in the library at the moment, I think this would be a good thing to standardize on soon leveraging some of the layout models your team has developed already.

PR #711, opened 2019-11-12
596-Table checkbox misalignment 

### LINKED ISSUE #596: [Table]: Fix checkbox misalignment, it's easy to see in the Stories
the table checkboxes are not aligned with the text inside the cells

davidicus:
Closes #596 

**Summary**

Fixed the checkbox misalignment issue. 

- removed styled components for this component - TableBodyRow.jsx
- added prefixed class to component in its place - TableBodyRow.jsx
- migrated styles from styled component to sass file and change padding that was causing misalignment - _data-table.scss
- changed npm script to use `--watch` flag over `--watchAll`. This seems like a more appropriate flag and resolves the JS MEMORY issue that happens when you change a component that is in many stories. 

Look at table stories with checkbox.


package.json
```diff
@@ -25,10 +25,10 @@
     "prepare": "yarn build",
     "start": "yarn storybook",
     "storybook": "start-storybook -p 3000 -s public/development",
-    "test": "jest --detectOpenHandles",
+    "test": "cross-env NODE_OPTIONS='--max-old-space-size=8096' jest --detectOpenHandles",
```
tay1orjones: I don't think `--detectOpenHandles` should be used outside of a manual flag when debugging, it has perf implications

package.json
```diff
@@ -25,10 +25,10 @@
...
     "test:a11y": "jest --testPathPattern='src\\/([^\\s]+)\\/([a-zA-Z0-9\\s_\\\\.\\-\\(\\):])+(\\.test\\.a11y\\.)+(js|jsx)$'",
-    "test:watch": "jest --watchAll --runInBand --verbose",
-    "test:update": "jest --updateSnapshot",
+    "test:watch": "cross-env NODE_OPTIONS='--max-old-space-size=8096' jest  --watch --verbose",
+    "test:updateSnapshot": "cross-env NODE_OPTIONS='--max-old-space-size=8096' jest  --updateSnapshot",
```
tay1orjones: could we put this back to `test:update`?

davidicus: @scottdickerson

package.json
tay1orjones: I removed the flag, and saw the following output
```
Test Suites: 49 passed, 49 total
Tests:       435 passed, 435 total
Snapshots:   241 passed, 241 total
Time:        72.575s
Ran all test suites.
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```
Although there is nothing detected.

davidicus:
@tay1orjones with the test work you have been doing are the comments here still warranted to be changed in this PR?

package.json
tay1orjones: I've opened #732 to further investigate and track this issue. No need to hold up this PR

package.json
tay1orjones: I pushed a commit to resolve this.

PR #713, opened 2019-11-13
fix(button): loading maintains same height in button

### LINKED ISSUE #709: [Button] Loading state doesn't maintain height from "size" prop
### Describe the bug

When `<Button>` is in loading state, the size of the button increases due to spinner padding.  We should ensure (and add stories to verify) that loading maintains whatever size was passed

### To Reproduce

View the `loading` story for Button
View the `not loading` story for Button
Notice the sizes are different

### Expected behavior

loading state should respect height guidelines from size prop.

### Environment/versions:

any

### Additional context

image

soon would be good

jessieyan:
Make sure loading spinner within the height inside a button

Closes #709

### DIFF
--- a/src/components/Button/Button.jsx
+++ b/src/components/Button/Button.jsx
@@ -3,12 +3,18 @@ import PropTypes from 'prop-types';
 import { Button as CarbonButton, Loading } from 'carbon-components-react';
 import { ButtonTypes } from 'carbon-components-react/lib/prop-types/types';
 import styled from 'styled-components';
+import { settings } from 'carbon-components';
+
+const { prefix } = settings;
 
 const StyledButton = styled(CarbonButton)`
   &&& {
     justify-content: flex-start;
-
-    .bx--loading__stroke {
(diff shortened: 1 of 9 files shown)


jessieyan:
This is to resolve Issue #709.

src/components/Button/Button.jsx
```diff
@@ -7,7 +7,10 @@ import styled from 'styled-components';
 const StyledButton = styled(CarbonButton)`
   &&& {
     justify-content: flex-start;
-
+    .bx--loading {
```
tay1orjones: If the consumer of this component has configured their codebase to use a different prefix than the default `bx`, these styles will not apply. Instead of the hardcoded `.bx`, we can replace it with the prefix variable:

```
import {settings} from carbon-components;

const {prefix} = settings;


... then in the style declaration

.${prefix}--loading {

}
```

jessieyan: Done. Thanks!

Issue #707, opened 2019-11-11
[Repo] Raise test coverage levels to acceptable levels


davidicus:
Currently, small changes that are 100% test covered can shift the test thresholds and hold up PRs. There are several files that need to have more test written to bring them up to passable threshold levels to prevent this from happening. 

Goals are 
Functions to 82%
Branches to 80%
Set Coveralls to fail build if one PR drops us more than 2% in coverage

The offending Files:
- [x] ImageHotSpots.jsx
- [x] HeaderMenu.jsx
- [x] PageWizard.jsx
  - [x] StatefulPageWizard.jsx
- [x] CustomCard new code
- [x] AsyncTable.jsx
  - [x] MockApiClient.js
  - [x] asyncTableReducer.js
- [x] 🚫 WizardInline.jsx
  - [x] 🚫 WizardHeader.jsx
- [x] ListCard.jsx (Denny)
- [x] TableToolbar.jsx
- [ ] 🤷‍♀️ TimeSeriesCard.jsx (Blocked see https://github.com/carbon-design-system/carbon-charts/issues/594)
- [ ] 🤷‍♀️ BarChartCard.jsx (Blocked see https://github.com/carbon-design-system/carbon-charts/issues/594)
- [ ] ☢️ 🤷‍♀️ DashboardEditor.jsx (Blocked see https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2152#issuecomment-807250836)
- [x] FileDrop.jsx
- [x] thresholdsHelperFunctions.js
- [x] ImageCard.jsx
- [x] ImageUploader.jsx
- [x] 🚫 EditPage.jsx
- [x] 🚫 PageHero.jsx
- [x] 🚫 PageTitle.jsx
- [x] ColumnHeaderSelect.jsx
- [x] valueCardUtils.js
- [x] DashboardGrid.jsx
- [x] TableHeader.jsx
- [x] ColumnResize.jsx
- [x] FilterHeaderRow.jsx
- [x] RowActionsCell.jsx
- [x] RowActionsError.jsx
- [x] StatefulTable.jsx
- [x] CatalogContent.jsx
- [x] 🚫 Dashboard.jsx
  - [x] 🚫 CardRenderer.jsx  
- [x] StatefulTableDetailWizard.jsx
- [x] ☢️ ColorDropdown.jsx

---
Current exceptions to be fixed are listed in the jest.config.js

[...]

PR #721, opened 2019-11-14
test(coverage): track coverage per-component instead of global aggregate

tay1orjones:
Towards #707

**Summary**

- Reconfigure jest to utilize globs, which will independently track the coverage of individual components, instead of a global aggregate.
- Applicable note from [the Jest documentation](https://jestjs.io/docs/en/configuration#coveragethreshold-object):
  - > If globs or paths are specified alongside global, coverage data for matching paths will be subtracted from overall coverage and thresholds will be applied independently. Thresholds for globs are applied to all files matching the glob. If the file specified by path is not found, error is returned.

- Increase global threshold to 80% across the board
- Add new glob for component files (excludes certain components that currently do not meet the minimums)
- Add individual globs for some components that don't meet the minimums. These values are where the coverage stands today and should be fixed (and removed) via #707 

- View the log, you can see the coverage being reported overall. Status checks should pass.


bryancboyd:
So the idea is, we remove exceptions as coverage improves to the 80% threshold?

tay1orjones:
Yep, exactly. Once this is merged, I'll update #707 with a permalink code snippet of the list.

PR #726, opened 2019-11-15
fix(pagewizard): add bottom padding for sticky footer

### LINKED ISSUE #723: [PageWizard] bottom of content not visible with sticky footer
image

When sticky footer is enabled and the user scrolls down to the bottom of the wizard content, the `fixed` positioning of the sticky footer means that the bottom of the content is cut off.

I'd suggest we fix by adding bottom padding to the outer page wizard div (`page-wizard` class) with a size equal to the height of the sticky footer (`5rem`) only if sticky footer is enabled.

jessieyan:
This is to resolve #723 [PageWizard] bottom of content not visible with sticky footer

Add  bottom padding to PageWizard is hasStickyFooter is set to true.


bryancboyd:
@jessieyan it looks like the `page-wizard--sticky` rule isn't getting applied, in the storybook from the deployment preview above.  Could you take a look?

bryancboyd:
Fixed it, we needed to nest the scss rule with a `&` selector in order to support both class rules being applied at the same div:

```
.page-wizard {
  display: flex;

  &.page-wizard--sticky {
    padding-bottom: 5rem;
  }
```

bryancboyd:
@jessieyan https://www.sassmeister.com/ is a great online tool to see how compiled SASS/SCSS looks, I reference it all the time (including for this issue).  Here's how the SCSS above compiles:

image

PR #748, opened 2019-11-21
741 reconfigure a11y test

### LINKED ISSUE #732: [jest] async/hanging --detectOpenHandles
### Describe the bug

Something's up with the tests. We shouldn't be using the `--detectOpenHandles` flag. When removed, Jest outputs a message saying it did not exit. It seems this is due to an async action not being cleaned up.

`--detectOpenHandles` should print out something regarding which open handle could be causing the hang, but in our case it prints nothing.

Additionally, if `--detectOpenHandles` is removed and replaced with `--runInBand`, the tests pass without the message.

Both these flags dramatically slow the test runner, we should find the root cause so we can run without these flags.

### To Reproduce

1. Modify `yarn test` to not include `--detectOpenHandles`
1. Run `yarn test`
1. See message output from jest
1. Run `yarn test --runInBand`
1. See tests finish without the hanging output message from jest

### Expected behavior

Tests pass and jest exits cleanly with no additional parameters.

### Environment/versions:

- OS: MacOS
- `carbon-addons-iot-react` version: `v2.15.2`

### Additional context

Output:
```
=============================== Coverage summary ===============================
Statements   : 88.32% ( 1876/2124 )
Branches     : 78.97% ( 1510/1912 )
Functions    : 82.57% ( 559/677 )
Lines        : 88.26% ( 1789/2027 )
================================================================================

Test Suites: 49 passed, 49 total
Tests:       449 passed, 449 total
Snapshots:   244 passed, 244 total
Time:        24.206s
Ran all test suites.
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

### LINKED ISSUE #741: [Test] - remove a11y test from general test script
The stray processes caused by the @ibma/aat package is more of an issue with increased contribution. Remove a11y testing from general `test` script. This way a11y test can still be ran manually but will not cause issues on contributor's machines.

davidicus:
Closes #741 Closes #732 

**Summary**

- Removed a11y test from `test` and `test:watch` commands. 
- Added the `--forceExit` flag to `test:a11y` to deal with the hanging processes from the a11y tooling. 
- Added `--coverage` flag to `test` and `test:watch` command to run coverage reports. By removing from settings we can run and pass subsets of test since coverage isn't collected on all test by default.

- items_here

- Running `yarn test` should not run a11y test.


package.json
```diff
@@ -25,9 +25,9 @@
     "prepare": "yarn build",
     "start": "yarn storybook",
     "storybook": "start-storybook -p 3000 -s public/development",
-    "test": "cross-env TZ=America/Chicago NODE_OPTIONS='--max-old-space-size=8096' jest --detectOpenHandles",
-    "test:a11y": "cross-env TZ=America/Chicago jest --testPathPattern='src\\/([^\\s]+)\\/([a-zA-Z0-9\\s_\\\\.\\-\\(\\):])+(\\.test\\.a11y\\.)+(js|jsx)$'",
-    "test:watch": "cross-env TZ=America/Chicago NODE_OPTIONS='--max-old-space-size=8096' jest  --watch --verbose",
+    "test": "cross-env TZ=America/Chicago NODE_OPTIONS='--max-old-space-size=8096' jest --testPathPattern='.*\\.test\\.js(x)?' --detectOpenHandles --coverage",
```
tay1orjones: `--detectOpenHandles` can be removed here. This will also enable this PR to close #732, the PR body will need to be updated to auto-close it on merge.

davidicus:
> Some feedback:
> 
> * `yarn test` should run the entire test suite. The travis config calls that singular command to report status for PRs. What if two scripts were composed together? This allows independent configs for each type of test with the one downside of invoking jest twice. I wouldn't think there would be much of a perf impact.
> 
> ```
> "test" : "yarn test:base && yarn test:a11y

While I agree that `yarn test` _should_ run the entire suite, this will still cause the run away processes that add up on developers machines. Until we can fix the underlining issue I do not think it should be ran with the normal test command. This is troublesome even for developers who are aware and can go in to shut down the processes manually. We either need to make a separate command that is ran on our post commit hook or find another a11y tooling solution. As it is, I think this needs to be removed ASAP since we have increase contributions. We can then approach another solution so that we are running the entire suite. 

I agree with the other comments and will make the changes.

tay1orjones:
I agree, lets keep the a11y tests from being ran on dev machines unless a specific command is ran. The a11y tests should still be ran in CI though, or we should delete them. Otherwise they'll never be ran, could become stale, etc.

To run the current set up in CI, the travis.yml will need to be modified to run both the regular tests and a11y, as well as report the appropriate coverage.

It seems cleaner to me to have `yarn test` run the entire suite, no changes to travis config. Then to prevent the node buildup, modify the pre-commit hook to only run non-a11y tests (like `yarn test:base` in my previous comment). This way the dev machines aren't affected, a11y tests are ran/still required to pass for PRs, and there's no shifting of the definition of `yarn test`.

I'm good with either approach, just want to avoid the tests not being ran.

davidicus:
I am ok with that approach. My only worry is that people are going to default to just running `yarn test` and will be affected by the bug. I will probably add something to our docs to notate the preferred method.

PR #753, opened 2019-11-25
fix(pagewizard): add tests for stateful pagewizard

jessieyan:
Created tests for StatefulPageWizard


davidicus:
When I pulled this branch down and ran test it was still failing the individual file threshold.

davidicus:
@jessieyan I pushed a change to get us over the threshold and to remove the exception for this file in the jest config.

PR #754, opened 2019-11-27
Sidenav bug fixes

### LINKED ISSUE #717: SideNav: isActive is not highlighting the menu item when nav bar is in collapse state.
### Describe the bug

SideNav component when collapsed does not show the highlight/focus on the menu item which is set isActive. Only when it is expanded the menu item is highlighted.

### To Reproduce
isActive is set true for Explore tab.

When sidenav is collapsed Explore tab is not dispalyed highlighted as in below screen
image

Whereas when it is expanded Explore tab is highlighted.
image

I have noticed that the behaviour works fine with lower version of carbon-addons-iot-react v2.3.2. Below codesandbox is for reference.
https://codesandbox.io/s/carbon-iot-react-header-n4q3h

Whereas with higher versions 2.8.5 which we are currently using it has the issue mentioned above.

We need this bug to be fixed for our Nov release.

Relative PAL issue: https://github.ibm.com/Watson-IoT/pal-tracking/issues/72

davidicus:
Closes #717 

**Summary**

There were some styling bugs for subnav menu items when not selected.

- check the sidenav story. one of the dropdowns `isActive` and one is not. Should see the issue fixed


src/components/SideNav/_side-nav.scss
```diff
@@ -21,30 +19,26 @@
...
   }
 
   .#{$prefix}--side-nav__link {
+    align-items: center;
+    display: flex;
+    cursor: pointer;
+    font-size: 0.875rem;
     font-weight: 600;
+    letter-spacing: 0.16px;
+    line-height: 1.125rem;
```
tay1orjones: `body-long-01` appears to accomplish most of the custom type styles here, could we use it instead?

davidicus: I hesitate only because the purpose of that tag is not really aligned with our sidenav link. If at some point styles are added to accommodate body type issue we will be getting those styles as well.

tay1orjones: Yeah, we may not want to inherit tweaks to body copy within the scope of the sidenav. Sounds good to leave as is 👍

PR #767, opened 2019-12-04
719 breadcrumb

### LINKED ISSUE #719: [Breadcrumbs] Breadcrumb on Mobile
**Description**
Extension to the breadcrumb in Carbon 10 to support mobile use cases within applications that have extensive nested navigation

**Name of Pattern or Asset**
Breadcrumb

**User outcome**
User can easily navigate up a breadcrumb trail on a phone

**Proposal**
screen shot 2019-10-29 at 2 00 12 pm

**Additional context**
Trick here is to avoid wrapping, which would consume a majority of the screen. If we want to suppress wrapping, I recommend we suppress it entirely (rather than saying 2-line max). To make this work, we will need to programmatically determine the container size at runtime and trunkate the first and last breadcrumb so that the overflow (center menu) is directly in the middle of the container (which will need to be full-width on mobile but may be smaller on desktop).

changshunwang:
Closes #719 

**Summary**

- Dynamicly adjust breadcrumb items to have overflow menu or not,  based on container width. 

- [feat] Get container width and calculate breadcumb items total width.
- [feat] Compare container width and items total width to show overflow menu dynamicly
- [feat] Response container width change
- [bug] Haven't refactor all style from styled-components to sass
- [bug] No test
- [bug] No keyboard action

- View  the `with overfllow menu` story


stuckless:
I have some questions/suggestions around the implementation.
1. How the component to know when to show the overflow menu?  is it intelligent (ie, can't render 3 items, so it puts the middle one into the overflow?
2. Could we add some knobs to the story to change the threshold or size or whatever triggers it to go into the overflow mode.

changshunwang:
@stuckless 
1. the overflow menu is controlled by a parameter called threshold(default is 4) in `Breadcrumb`, which I followed `Carbon-angular`, because Carbon-react hasn't implemented this function. I am not sure is it ok? 
2. Good suggestion! I will add these knobs.

tay1orjones:
Please fill out the pull request template in the PR body above to link the appropriate issue, describe your changes, and detail any specific acceptance criteria.

stuckless:
@changshunwang Could you link to the design as well... i thought the design was around trying to render as much as it could, and then, if it could not "fit" then it would intelligently collapse the breadcrumb.  ie, if I just resized the browser it should collapse the breadcrumb.  That would be my expectation.  @JordanWSmith15 is that correct?

tay1orjones:
@stuckless yes, there were some gaps in the handoff of implementation details - we're working to define those specifics alongside the original design in [the internal issue](https://github.ibm.com/Watson-IoT/pal-tracking/issues/18).

tay1orjones:
I've opened the above issue on the Carbon repo to discuss the implementation with the Carbon team in aim to potentially contribute this directly to `carbon-components-react`

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+const propTypes = {
+  /** Pass in the BreadcrumbItem's for your Breadcrumb */
+  children: PropTypes.node,
+  /** Specify an optional className to be applied to the container node */
+  className: PropTypes.string,
+  /** Optional prop to omit the trailing slash for the breadcrumb */
+  noTrailingSlash: PropTypes.bool,
+  /** Define to show how many items */
+  threshold: PropTypes.number,
```
tay1orjones: `threshold` is the only prop change from the stock Breadcrumb component. All the others can be removed from proptypes and defaultprops

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * @param {number} t, threshold must be a valid number
+ */
+const setThreshold = t => (t < MINIMUM_OVERFLOW_THRESHOLD ? MINIMUM_OVERFLOW_THRESHOLD : t);
+
+const Breadcrumb = props => {
+  const { className, children, noTrailingSlash, threshold, ...other } = props;
+
+  const t = setThreshold(threshold);
```
tay1orjones: please use a descriptive variable name

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,89 @@
...
+  const showOverflow = children ? children.length > t : false;
+
+  return (
+    <div className="bx--col-lg-16 bx--col-md-8 bx--col-sm-4">
+      {showOverflow ? (
+        <StyledBreadcrumb {...other} className={className} noTrailingSlash={noTrailingSlash}>
+          {children[0]}
+          <StyledOverflowMenu renderIcon={StyledOverflowIcon20}>
```
tay1orjones: This should not be an icon, but an ellipsis text character, per [guidance here](https://github.com/carbon-design-system/carbon/issues/1105#issuecomment-422465785)

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,89 @@
...
+              if (i !== 0 && i !== children.length - 1) {
+                return <BreadcrumbOverflowItem>{child}</BreadcrumbOverflowItem>;
+              }
+              return null;
+            })}
+          </StyledOverflowMenu>
+          {children[children.length - 1]}
+        </StyledBreadcrumb>
+      ) : (
+        <StyledBreadcrumb {...other} className={className} noTrailingSlash={noTrailingSlash}>
```
tay1orjones: `className` and `noTrailingSlash ` can be removed here as they'll be included in `{...other}`

src/components/Breadcrumb/BreadcrumbItem.jsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { BreadcrumbItem as CarbonBreadcrumbItem } from 'carbon-components-react';
+
+const BreadcrumbItem = () => {
+  return <CarbonBreadcrumbItem />;
+};
+
+export default BreadcrumbItem;
```
tay1orjones: I don't think there's a need to redefine and export the component with no changes. 

I think it can be a one liner in this file
```js
export default Breadcrumb from `carbon-components-react`;
// or it might need to be
export default from 'carbon-components-react/lib/components/BreadcrumbItem';
```

I'm not exactly sure on the syntax.

src/components/Breadcrumb/BreadcrumbOverflowItem.jsx
```diff
@@ -0,0 +1,126 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from 'styled-components';
```
tay1orjones: Same thing here with refactoring to utilize sass. Please use carbon token variables where possible. This includes layout/sizing items, motion bezier curves, colors, and typography styles.

src/components/Breadcrumb/BreadcrumbOverflowItem.jsx
```diff
@@ -0,0 +1,126 @@
...
+    children: null,
+    closeMenu: null,
+    handleOverflowMenuItemFocus: null,
+    onClick: () => {},
+    onKeyDown: () => {},
+    index: 0,
+  };
+
+  // TODO: set variable replace number
+  setTabFocus = evt => {
```
tay1orjones: Keyboard focus currently becomes trapped on the overflow menu toggle, and can not move to or from it once it's been focused via keyboard.

src/components/Breadcrumb/BreadcrumbOverflowItem.jsx
```diff
@@ -0,0 +1,126 @@
...
+      onKeyDown,
+      handleOverflowMenuItemFocus,
+      index,
+      children,
+      ...other
+    } = this.props;
+
+    return (
+      <StyledDiv
+        className="bx--overflow-menu-options__option"
```
tay1orjones: please utilize the `${prefix}` in the classname, `bx` can not be hardcoded as this is a user configurable option

src/components/Breadcrumb/index.js
```diff
@@ -0,0 +1,8 @@
+import Breadcrumb from './Breadcrumb';
+import BreadcrumbItem from './BreadcrumbItem';
+// import BreadcrumbSkeleton from './Breadcrumb.Skeleton';
```
tay1orjones: Lets export this here alongside the others.

src/components/Breadcrumb/Breadcrumb.story.jsx
```diff
@@ -1 +1,30 @@
...
+
+storiesOf('Watson IoT|Breadcrumb', module)
+  .add('default', () => (
+    <Breadcrumb {...BreadcrumbProps}>
+      <CarbonBreadcrumbItem href="/">Breadcrumb Item1</CarbonBreadcrumbItem>
+      <CarbonBreadcrumbItem href="/">Breadcrumb Item2</CarbonBreadcrumbItem>
+      <CarbonBreadcrumbItem href="/">Breadcrumb Item3</CarbonBreadcrumbItem>
+    </Breadcrumb>
+  ))
+  .add('overview', () => (
```
tay1orjones: This story should be named with something more descriptive `with truncation`

src/components/Breadcrumb/Breadcrumb.story.jsx
```diff
@@ -1 +1,30 @@
...
+import { action } from '@storybook/addon-actions';
+import { BreadcrumbItem as CarbonBreadcrumbItem } from 'carbon-components-react';
+
+import Breadcrumb from './Breadcrumb';
+
+const BreadcrumbProps = {
+  onClick: action('click'),
+};
+
+storiesOf('Watson IoT|Breadcrumb', module)
```
tay1orjones: The stories here need to include all the same [stories provided by the Carbon storybook](https://angular.carbondesignsystem.com/?path=/story/components-breadcrumb--basic), in addition to the ones showing extra functionality provided by this PR

changshunwang:
@tay1orjones ,Hi Talyor, cause we don't use threshold parameter as angular temporarily, based on your suggestions and guidence, I refactory my code. so would you please review it again? 
here are 2 questions now:
1. The scss style is still rough, 2 reasons: 1. I'm not sure if I write as this correctly now? 2. It seems Carbon style is a little diffirent with what we need, but I only have a picture and some description, is there something I misunderstand with you?
2. when use the component, I should add a container div outside, just like what I do in storybook `with overflow menu`. 
image
but if I don't add, there maybe an error, like in storybook `default`.  (I think default should show all 3 items, without overflow menu.)
image

so is there something I did wrong?

In addition, I found a strange thing, the overflow menu function can work in my local storybook, but doesn't work in [sandbox here](https://codesandbox.io/s/blazing-http-9furg)

Thanks again!

src/components/Breadcrumb/Breadcrumb.jsx
tay1orjones: I have come to find that this guidance is outdated, we should instead be using the `overflowMenuHorizontal` icon, which is centered vertically.

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,120 @@
...
+  });
+  const [breadcrumbChildren, setBreadcrumbChildren] = useState(children);
+  const [overflowIndex, setOverflowIndex] = useState(0);
+
+  /** intial to get children total width */
+  useEffect(() => {
+    // measure child total width
+    const itemsWidth = [];
+    const breadcrumb = breadcrumbRef.current;
+    const breadcrumItems = breadcrumb.querySelectorAll('li.bx--breadcrumb-item');
```
tay1orjones: Instead of querying the DOM, breadcrumb items should be gathered for measurement via refs. You'll need to wrap each Carbon BreadcrumbItem in a div, and set the ref on the div since BreadcrumbItem does not utilize [forwardRef](https://reactjs.org/docs/forwarding-refs.html).

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,120 @@
...
+          break;
+        }
+      }
+      if (containerWidth < curTotalWidth) {
+        setOverflowIndex(-1);
+      }
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [containerWidth]);
```
tay1orjones: It's important that all dependent values are included here to prevent bugs. Was this intended? `breadcrumbItemsWidth` and `children` are both dependent values utilized in the function and should be included here.

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const showOverflowMenu = containerWidth > totalWidth ? 'no' : 'yes';
+
+  if (typeof containerWidth === 'undefined') {
+    return (
+      <div ref={breadcrumbRef} style={{ width: `100%`, visibility: `hidden` }}>
```
tay1orjones: If the ResizeObserver polyfill isn't available or fails for some reason, this is what will render. It shouldn't have `visibility: hidden` by default.

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,120 @@
...
+      }
+      if (containerWidth < curTotalWidth) {
+        setOverflowIndex(-1);
+      }
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [containerWidth]);
+
+  const showOverflowMenu = containerWidth > totalWidth ? 'no' : 'yes';
```
tay1orjones: This can be just be a boolean, with the appropriate blocks modified in the return jsx.
```diff
- const showOverflowMenu = containerWidth > totalWidth ? 'no' : 'yes';
+ const showOverflowMenu = containerWidth > totalWidth;
```

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,120 @@
...
+  children: PropTypes.node,
+};
+
+const defaultProps = {
+  children: null,
+};
+
+const Breadcrumb = props => {
+  const { children, ...other } = props;
+  const [breadcrumbItemsWidth, setBreadcrumbItemsWidth] = useState([0]);
```
tay1orjones: I think the initial value should be an empty array `[]` instead of `[0]`?

src/components/Breadcrumb/BreadcrumbItem.jsx
```diff
@@ -0,0 +1,40 @@
...
+      // getWidth(node.getBoundingClientRect().width)
+      // const getWidth1 = () => {
+      //   return node.getBoundingClientRect().width;
+      // }
+      setWidth(node.getBoundingClientRect().width);
+    }
+  }, []);
+
+  return (
+    <div ref={measuredRef} width={width}>
```
tay1orjones: Do we need to set a width here? This ref should probably be forwarded so that the containing Breadcrumb can utilize it to gather the width, instead of doing it here.

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,137 @@
...
+
+    // console.log("parent :", breadcrumbRef.current);
+    // console.log("children : ", childrenWithRef);
+    for (let i = 0; i < childrenWithRef.length; i += 1) {
+      // console.log("child with ref: ", childrenWithRef[i].ref.current);
+      // console.log("children height x width: ", childrenWithRef[i].ref.current.getBoundingClientRect());
+      itemsWidth.push(childrenWithRef[i].ref.current.getBoundingClientRect().width);
+      setBreadcrumbItemsWidth(itemsWidth);
+      const curTotalWidth = itemsWidth.reduce((acc, cur) => acc + cur);
+      setTotalWidth(curTotalWidth);
```
tay1orjones: I don't think we should set state on every iteration. It seems your goal here is to add up the total width of the children combined. That should be calculated and then `setTotalWidth` should be called _once_, after the children have been iterated through.

tay1orjones:
@changshunwang I've pushed a couple updates to this branch/PR. The core functionality is working, with a few bugs that still need addressed. The resize behavior is now opt-in and behind a prop `useResizeObserver`. I've updated the stories to reflect the change.

Please look into these remaining items:

1. when the container is very small, the text is not being ellipsed. 
  1. Due to this, an infinite loop is caused when the overflowIndex is set to `-1`, which then causes the wrapping of text, causing the overflowIndex to be set higher
1. I reset the styling for the overflow menu button, but it still needs improvement.
1. The slashes after each breadcrumb are no longer centered between the breadcrumbs, this is due to some styles from carbon that target the last-child, we need to override them so the slashes `/` are centered between the breadcrumbs
1. Stories for fixed container widths are not working, breadcrumbs are wrapping when they should be truncated into the OverflowMenu
1. Keyboard navigation/a11y
1. Test coverage

changshunwang:
@tay1orjones  Thanks! I didn't think of this method: set useObeserver as a prop. BUT, I have one question, why we also need fixed container in this situation. At first, I set a container in storybook in order to simulate window change. If we set useObeserver as a prop, then the componenet can self-responce. so what is the target for story fixed container? And based on your suggestion, I fixed some of above:
1. I use a new state `truncate` instead of overflowindex=-1. It seems that Storybook has a min-width, so can't see truncated text in story 'with useResizeObserver',  you can try in 'with useResizeObserver, 320px width'. It works.
2. when click overflow button, the menu will have a strange border, I didn't find why. :(
3. It works.
4. Like my question above. But when I reset the container width, sometimes it didn't re-render in time.
5. Keyboard navigation/a11y. Sorry I am not familiar with a11y, although have read related code in our project, but I am not very sure:

- 1. for both breadcrumb item and overflow menu item or just overflow menu item?
- 2. which actions or keys on the keyboard should we need? I mean, enter, command, arrows, etc... 

6. Test coverage need I write some code?

tay1orjones:
It looks like the functionality has regressed and is no longer appropriately placing items in the overflow menu. The stories for fixed widths (320px etc) aren't working either on the deploy preview.

The text is still not being ellipsed at very small sizes.

On the storybook containers: 
Every story is snapshotted and tracked in jest. If changes are made and an unrelated snapshot is modified, it can signal something is wrong. I think it's a good idea to have some stories with fixed containers (320px, etc) so we have some snapshots of the behavior across multiple container sizes. 

You should be able to view the storybook in browser widths less than 500px, it does take storybook a moment to redraw though.

For #4, it's likely related to the general bug of items not being placed in the overflow menu appropriately. We may want to look into throttling the useResizeObserver, this may help with the sputtery rendering issues. There's an example in the documentation for the hook package of how to accomplish that. 

a11y - a user should be able to tab focus through all breadcrumbs. Additionally the user should be able to open the overflow menu by tabbing focus to the overflow menu, and then pressing enter/space, then using tab/arrows to select an item in the overflow menu.

Testing does look to be sufficient with the storyshots, though we might want to write some unit tests once we've eliminated the bugs to ensure they don't pop up again.

davidicus:
closing in favor of this [PR](https://github.com/IBM/carbon-addons-iot-react/pull/829)

PR #772, opened 2019-12-05
chore(merge): update master with the latest 1.x card fixes

scottdickerson:
Add latest fixes to master release, here's an overview in order of the file in the PR:
- refactor: Extracted the CardToolbar component from the Card 
- fix(Card): cached the onCardAction callback
- test(card): fixed the expanded story
- refactor: Extracted the CardRangePicker component from the Card  and simplified the Range Picker UX
- fix(CardRenderer): show the setupError on ImageCards and always load the card
- fix(Table): allow boolean values
- feat(TimeSeriesCard): support lazyLoading for large cards
- fix(PropTypes): support quarter intervals
- fix(validators): support additional validation


package.json
```diff
@@ -199,7 +199,7 @@
     "rollup-plugin-replace": "^2.1.0",
     "rollup-plugin-uglify": "^6.0.1",
     "sass-loader": "^7.1.0",
-    "semantic-release": "^16.0.0-beta.19",
+    "semantic-release": "^16.0.0-beta.39",
```
tay1orjones: Why update the semantic-release deps? These can be a bit unstable as we're running the beta versions.

src/components/Card/Card.story.jsx
```diff
@@ -100,7 +100,7 @@ storiesOf('Watson IoT|Card', module)
           size={size}
           isLoading={boolean('isLoading', false)}
           isEmpty={boolean('isEmpty', false)}
-          isEditable={boolean('isEditable', true)}
+          isEditable={boolean('isEditable', false)}
           isExpanded={boolean('isExpanded', false)}
```
tay1orjones: I know you didn't change the `isExpanded` knob here, but it seems backwards in storybook? When isExpanded is checked, the card is not expanded visually.

| - | - |
| - | - |
| image | image |

package.json
scottdickerson: this was to try and fix that weird semantic-release issue we had where it just lost the previous tag

src/components/Card/Card.story.jsx
scottdickerson: can I fix that in my next track?  I probably shouldn't have fixed the other bug since this is supposed to just be a plain merge...

PR #773, opened 2019-12-06
refactor(dashboard): split the rendering decisions from the data decisions

### LINKED ISSUE #653: [Dashboard]: Refactor the components to split the rendering and data awareness
Need to better split the rendering responsibilities of the Dashboard and data fetching components so that @stuckless can reuse them in other products with different backends.

<sub>[Internal issue #89](https://github.ibm.com/Watson-IoT/pal-tracking/issues/89)</sub>

scottdickerson:
Closes #653

**Summary**

- Downstream consumers that want to build their own dashboard with custom card types can't reuse our Dashboard component as it only supports a fixed set of card types.

- refactor(Card): move the isExpanded decision into the Card component so downstream teams can share the isExpanded behavior
- refactor(Dashboard): split the rendering into a new component (DashboardGrid) so downstream teams can render their own card types.  
- refactor(Dashboard): Remove the ability to pass custom row heights, gutters, dimensions to the component to standardize our paddings across the whole IoT Dashboard organization
-refactor(CardRenderer): move the isExpanded styles out of the CardRenderer component so downstream teams can share the isExpanded behavior.  

- You can verify that I didn't break anything by running the existing Dashboard and card stories


scottdickerson:
@stuckless you can do either though the second is better as it will support the isExpanded bit automatically.  I'll add a DashboardGrid story to serve as the README

stuckless:
@scottdickerson that sounds good.  The sooner we can get this mergedt he better, so that I can start on the maximo integration pieces.  thx.

src/components/Card/Card.jsx
```diff
@@ -250,6 +255,8 @@ const Card = ({
       )}
     </VisibilitySensor>
   );
+
+  return isExpanded ? <div className="bx--modal is-visible">{card}</div> : card;
```
tay1orjones: The `bx` prefix is user configurable, we'll need to import and use the prefix variable from `settings`. We do this in `Button.jsx` as an example

src/components/Dashboard/DashboardGrid.jsx
```diff
@@ -0,0 +1,173 @@
...
+      >
+        {cards}
+      </StyledGridLayout>
+    </div>
+  );
+};
+
+DashboardGrid.propTypes = DashboardGridPropTypes;
+DashboardGrid.defaultProps = defaultProps;
+export default DashboardGrid;
```
tay1orjones: Does this need to be added to the exports in `index.js`?

src/components/Card/Card.jsx
scottdickerson: thanks for catching, will fix

src/components/Dashboard/DashboardGrid.jsx
scottdickerson: thanks for catching, will fix

PR #774, opened 2019-12-09
720 table editing

### LINKED ISSUE #720: [Table] In-line editing of Data Table
**Description**
Extend the IoT Data Table to allow for editing of lines row-by-row

**Name of Pattern or Asset**
Data Table

**User outcome**
Use can edit the data of visible columns using appropriate component, and save multiple row edits at once

**Proposal**
https://jsw.ibm.com/browse/GRAPHITE-3571
screen shot 2019-10-29 at 4 48 13 pm
screen shot 2019-11-14 at 9 41 18 am

**Additional context**
https://jsw.ibm.com/browse/GRAPHITE-3571
- Add an icon to toggle in-line edit mode on in the top-right of the data table
-  Replace other content in the table header with Save and Cancel buttons at top when edit mode is on
-  Programmatically display the correct component for editing the fields shown, depending on the data source. Date fields are displayed with date picker, numeric with numeric spinner, alphanumeric with text entry, etc.  
    - Issue to add Date picker and Numeric spinner to design library: https://github.ibm.com/Watson-IoT/pal-tracking/issues/6
- Components should display with proper padding and alignment within the width of the row/column
- Fields should be validated on entry, before save (Install date cannot be in the future, End of Life cannot be in the past, Expected life must be greater than or equal to 1, etc) 
    - Issue for error state: https://github.ibm.com/Watson-IoT/pal-tracking/issues/32
- Display toast with undo option https://ibm.invisionapp.com/d/main#/console/15350624/319220019/preview (Use light version of toast. Dark version does not match PAL update)

karlinglee93:
Closes #720 

**Summary**

- Add an icon to toggle in-line edit mode on in the top-right of the data table
- Replace other content in the table header with Save and Cancel buttons at top when edit mode is on
- Table "freeze" when the table in edit mode
- Display the TextInput component for editing the fields shown
- Save or undo edited data when click Save or Cancel buttons
- Display toast with undo option 

- items_here

- tests here


src/components/Table/Table.jsx
```diff
@@ -68,6 +68,7 @@ const propTypes = {
   columns: TableColumnsPropTypes.isRequired,
   /** Row value data for the body of the table */
   data: TableRowPropTypes.isRequired,
+  oldData: PropTypes.arrayOf(PropTypes.object),
```
stuckless: I have a hard time understand what `oldData` is and how it is used.  Can you explain what `oldData` is doing here in `Table`?

src/components/Table/Table.jsx
```diff
@@ -134,6 +136,7 @@ const propTypes = {
           columnId: PropTypes.string.isRequired,
           /* Visibility of column in table, defaults to false */
           isHidden: PropTypes.bool,
+          editDataFunction: PropTypes.bool,
```
stuckless: Am I missing something?  Shouldn't editDataFunction be a function?

src/components/Table/TableBody/TableBody.jsx
```diff
@@ -55,15 +55,18 @@ const propTypes = {
...
     onRowExpanded: PropTypes.func,
+    onApplyEdit: PropTypes.func,
   }).isRequired,
   /** What column ordering is currently applied to the table */
   ordering: PropTypes.arrayOf(
     PropTypes.shape({
       columnId: PropTypes.string.isRequired,
       /* Visibility of column in table, defaults to false */
       isHidden: PropTypes.bool,
+      editDataFunction: PropTypes.bool,
```
stuckless: why do we need editDataFunction boolena in the ordering section?

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -442,19 +455,49 @@ const TableBodyRow = ({
...
+                    row: values,
+                  })
+                ) : (
+                  <TableCellRenderer>{values[col.columnId]}</TableCellRenderer>
+                )}
+              </StyledNestedSpan>
+            ) : (
+              <StyledNestedSpan nestingOffset={offset}>
+                {col.editDataFunction ? (
+                  <CarbonTextInput
```
stuckless: No, we cannot just use a CarbonTextInput in all cases.  This is why when we talked we said that an `editDataFunction` should be like the `renderDataFunction` and when there is an `editDataFunction` then we use that.  Otherwise you can default to using carbon text input if you want, but, we need to be able to supply our own editor.  eg, booleans will use a checkbox/switch, dates should have a picker, etc.  If the table isn't going to manage all the possible types, etc, then it just needs to delegate to something that will.

stuckless:
I'm also seeing that while the select rows is disabled, the entire table pager is still there.  It appears that we can jump to other rows, etc, and we should not be able to do that in edit mode... or at the very least it needs to be an option.  ie, in a table with 1,000,000 rows you just go into edit mode and edit page 1, and page 100,000 and then hit save.  That's fine for in-memory tables of a few items but, in a real world case if we change pages, then we need to issue a save, or prevent it.  @JordanWSmith15  @lukefirth any designs around the behaviours here.

tay1orjones:
Commit messages should follow the conventional commit format. [Link to contributing.md](https://github.com/IBM/carbon-addons-iot-react/blob/master/.github/CONTRIBUTING.MD#submitting-pull-requests-and-commits)

No need to change the commits you've already pushed, but future commits should be in format

```
feat(component): short description
```

To ensure we increment the version appropriately, please use `feat` for your next commit.

tay1orjones:
The PR body should link to the issue that should be closed/resolved by this PR.

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
tay1orjones: Yeah, the original discussion is here https://github.ibm.com/Watson-IoT/pal-tracking/issues/20

I agree and think it's clearer to not provide any `Input` by default, but instead rely on the `editDataFunction` entirely. Then, in the story we can show how to use this with different input types.

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -2,7 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import IconColumnSelector from '@carbon/icons-react/lib/column/20';
 import IconFilter from '@carbon/icons-react/lib/filter/20';
-import { DataTable, Button } from 'carbon-components-react';
+import IconEdit from '@carbon/icons-react/lib/edit/20';
```
tay1orjones: There is now an updated entrypoint for the icons library to eliminate the long import paths.

```diff
- import IconEdit from '@carbon/icons-react/lib/edit/20';
+ import IconEdit20 from '@carbon/icons-react';
```

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -70,6 +71,10 @@ const StyledTableBatchActions = styled(TableBatchActions)`
   }
 `;
 
+const StyledCarbonButton = styled(Button)``;
+
+const StyledCarbonToastNotification = styled(ToastNotification)``;
```
tay1orjones: These should be removed.

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -182,24 +198,69 @@ const TableToolbar = ({
...
+        caption=""
+        hideCloseButton={false}
+        iconDescription="undo changes and close"
+        kind="success"
+        lowContrast
+        notificationType="toast"
+        onCloseButtonClick={onUndoEditAction}
+        role="alert"
+        style={{
+          marginBottom: '.5rem',
```
tay1orjones: These styles should be extracted into a Sass partial for the component. [link to contributing.md](https://github.com/IBM/carbon-addons-iot-react/blob/master/.github/CONTRIBUTING.MD#styling-new-components) 
Values should be carbon design token variables instead of hardcoded values.

karlinglee93:
@stuckless @tay1orjones There were some misunderstands between the architectural conversation and me. After reading your comments, I understand a lot and will update my codes.
If I have any other questions, will discuss with you on github, thank you.
@tay1orjones I clicked 'requested a review from tay1orjones' by mistake, apologize.

tay1orjones:
@karlinglee93 All good 🙂 When a PR is created, it automatically requests a review from all [code owners](https://github.com/IBM/carbon-addons-iot-react/blob/master/.github/CODEOWNERS#L20)

src/components/Table/Table.jsx
karlinglee93: I have changed `oldData` to 'preData' and removed `oldData` in `Table`. `preData` is a state and means previous table data. It is used to save current table data after user clicked the edit icon. After the saving, table can do undo function when user clicked cancel button or the toast. It only shows up in `AsyncTable` and `tableReducer` now.

src/components/Table/Table.jsx
karlinglee93: Yes, it should be a function. In the past, I wrote the contents of editDataFunction in `TableBodyRow` with a hard-code `<TextInput>` component, and if editDataFunction is set `true` in `tableColumns` of `Table.story`, the corresponding column/rows/cells will re-render into `<TextInput>` after clicking edit icon. But now, I move the contents of editDataFunction into `Table.story` and change its props as `func`, besides user can customize the contents of editDataFunction now. BTW you can see the default contents of editDataFunction is `editDefaultData` with a `<TextInput>` in `Table.story`.

src/components/Table/TableBody/TableBody.jsx
karlinglee93: My fault, in the beginning, I thought there was a connection between `hasRowSelection` and `hasEdit`, i.e. when table does the row-selection action and hide some columns, edit action on the above columns should be hiden too. But now, I remove editDataFunction from `ordering`, and find it is OK.

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
karlinglee93: Yes, I understood that at first the time. But I thought we could supply our own editor in `TableBodyRow` not in `Table.story`, besides `CarbonTextInput` I wrote is just a demo. But now, I move the contents of editDataFunction into Table.story, besides user can customize the contents of editDataFunction - I have wrote two functions for demo, the one is `editInputData` which will render `<TextInput>`, the other is `editSelectData` which will render `<Select>`. User can alse add some checkbox/switch, picker, etc. they want.

src/components/Table/StatefulTable.jsx
```diff
@@ -124,6 +137,24 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
...
         dispatch(tableSearchApply(string));
       },
+      onCancelEditAction: () => {
+        dispatch(tableEditCancel());
+        callbackParent(onCancelEditAction);
+      },
+      onSaveEditAction: () => {
+        dispatch(tableEditSave());
+        dispatch(tableToastToggle('undo'));
+        window.hideToastTimeoutID = setTimeout(function() {
```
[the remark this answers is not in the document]
karlinglee93: Yes, I changed that. In `setTimeout` and `clearTimeout` function, I use `useEffect` of Hook to handle this.

src/components/Table/StatefulTable.jsx
```diff
@@ -124,6 +137,24 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
...
+          dispatch(tableToastToggle());
+        }, 4000);
+        callbackParent(onSaveEditAction);
+      },
+      onUndoEditAction: () => {
+        clearTimeout(window.hideToastTimeoutID);
```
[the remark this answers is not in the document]
karlinglee93: Yes, I've updated my codes and followed the way you said. BTW, it's really helpful, thank you!

src/components/Table/TableToolbar/TableToolbar.jsx
karlinglee93: Something bad happened when I tried the updated entrypoint way, so I undid it and waiting for your comments.
image

src/components/Table/TableToolbar/TableToolbar.jsx
karlinglee93: yes, removed

src/components/Table/TableToolbar/TableToolbar.jsx
karlinglee93: Yes, I've extracted the styles into `_table.scss`. BTW, what is carbon design token variables? Am I right now?

karlinglee93:
@stuckless @tay1orjones hi team, I've updated my codes and changed a lot as your comments, meanwhile, replied all your questions. 
Now table can be into/out-of edit mode by clicking edit icon, cancel/save button or the toast. All the functions above are written by Hook. 
Besides, table can also save/undo data by `onSaveCurData`/`onUndoEditData`, these functions are called by clicking edit icon, cancel/save button or the toast too, and I used stateful table(i.e. Reducer/Redux) to update the datas. 
Finally, user can customize `editDataFunction` in story, and I wrote two demos(<TextInput> and <Select>). 
More further details you can go into branch `720-table-editing` and have a try. If something need to improve, leave me a message, thank you!

BTW, something went wrong when I pushed my codes in Travis - "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."
Could you please help me out?

karlinglee93:
BTW, something went wrong when I pushed my codes in Travis: 
"No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."
Could you please help me out?
image

karlinglee93:
More further details you can go into branch `720-table-editing` and have a try. If something need to improve, leave me a message, thank you!

tay1orjones:
I pulled in the latest from `master` and resolved the conflicts in attempt to fix the Travis build timing out. Recent changes in `master` reconfigured our testing setup and config to prevent timeouts from jest.

src/components/Table/tableReducer.js
```diff
@@ -22,6 +22,8 @@ import {
   TABLE_COLUMN_ORDER,
   TABLE_REGISTER,
   TABLE_SEARCH_APPLY,
+  TABLE_DATA_SAVE,
+  TABLE_DATA_UNDO,
```
tay1orjones: I'm on the fence about building this into the table and exporting these stateful reducers. Would these be used? It feels like this would be implemented at the product level to hook up to a data source or central state (redux, mobx, apollo, etc) via their own reducers or hooks. 

In it's place, we could provide any necessary callbacks or props to enable this behavior (including the "undo" state).

@bryancboyd @davidicus Do you have some thoughts here?

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -94,6 +94,8 @@ const propTypes = {
   learnMoreText: PropTypes.string,
   /** I18N label for dismiss */
   dismissText: PropTypes.string,
+  /** table toolbar */
+  editBar: PropTypes.oneOf(['buttons', 'toast']),
```
tay1orjones: This name is confusing to me. Is it a state of the toolbar? I'm unclear what buttons vs toast is. I think there may be a more appropriate name and value options.

src/components/Table/AsyncTable/AsyncTable.jsx
```diff
@@ -182,7 +182,6 @@ const AsyncTable = ({ fetchData, id }) => {
       onRowSelected: (rowId, isSelected) => {
         dispatch(baseTableActions.tableRowSelect(rowId, isSelected));
       },
-
```
tay1orjones: Let's revert this change to keep the file changelog/history clean

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
karlinglee93: Hi, 'buttons' means toolbar come to save/cencel buttons status, besides 'toast' means toolbar come to toast status. It's a state of the toobar. Maybe we can change these to 'edit' and 'undo' or whatever, what do you recommend?
image
image

karlinglee93: And, we use callback (hook) to write the codes, so I didn't put editBar with activeBar together.

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -132,26 +138,41 @@ const propTypes = {
...
+  getUndoEdit: PropTypes.func,
 };
 
 const defaultProps = {
   i18n: {
     ...defaultI18NPropTypes,
   },
+  getActiveEditBar: () => {},
+  getSaveCurData: null,
+  getUndoEdit: null,
```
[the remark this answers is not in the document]
karlinglee93: done, now it looks like this.
image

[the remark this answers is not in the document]
karlinglee93: done, now the default props for functions are a noop `( ) => { }`.

src/components/Table/AsyncTable/AsyncTable.jsx
karlinglee93: Done.

src/components/Table/TableBody/TableBody.jsx
```diff
@@ -64,6 +64,7 @@ const propTypes = {
       isHidden: PropTypes.bool,
     })
   ).isRequired,
+  editBar: PropTypes.oneOf(['buttons', 'toast']),
```
stuckless: I don't understand why "toast" is an option here.  A toast is something that the application may show in response to the save (or even the cancel) action, but, I don't think the table itself will manage that.  @tay1orjones @davidicus am I wrong here?

src/components/Table/Table.jsx
```diff
@@ -164,6 +165,8 @@ const propTypes = {
       onApplyBatchAction: PropTypes.func,
       /** Apply a search criteria to the table */
       onApplySearch: PropTypes.func,
+      onSaveCurData: PropTypes.func,
```
stuckless: Not a big fan of using `Cur` there are many other props larger than this one, and none of them are using short forms.  Personally i'd just use `saveData` why do we need `Cur` or `Current`.  Is there any other kind of data that is being saved?

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
stuckless: I don't think that 'toast' is something that the table action bar should manage.  A toast is something that happens in response to the action (save or cancel) and that is generally up to the application to supply.   I'm not ever sure how I'd use this prop value... are you suggesting that I first set it to buttons and then hook into the on save and then set it to `taost`??

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -426,19 +439,33 @@ const TableBodyRow = ({
...
+                  <TableCellRenderer>{values[col.columnId]}</TableCellRenderer>
+                )}
+              </StyledNestedSpan>
+            ) : (
+              <StyledNestedSpan nestingOffset={offset}>
+                {col.editDataFunction ? (
+                  col.editDataFunction({
+                    value: values[col.columnId],
+                    columnId: col.columnId,
+                    rowId: id,
```
stuckless: Give the editDataFunction the same signature of the renderDataFunction (ie, same 4 paramters).  Otherwise I'll be having to read the code to figure out why, in 3 months, the editDataFunction is not accepting a row.

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -196,30 +262,74 @@ const TableToolbar = ({
...
+          <Button className="buttons" onClick={onSaveEditAction}>
+            {i18n.batchSave}
+          </Button>
+        </StyledTableToolbarContent>
+      ) : activeEditBar === 'toast' ? (
+        <ToastNotification
+          className="toast"
+          caption=""
+          hideCloseButton={false}
+          iconDescription="undo changes and close"
```
stuckless: Not sure where the toast stuff is coming from.  Is there something in the spect that says that the table is manage the showing and hiding of the toast??  This seems completely unneccary.  Not to mention that all the descriptions and labels  now need to be translated (which they are hard coded strings here).  If everyone agrees... let's just move showing a toast as part of the story.

src/components/Table/tableReducer.js
stuckless: Does this affect my ability to hook into the save and undo events coming from the table?  ie, I need to be able to register listeners that are called when save and undo are clicked.  The whole messaging system of the table is extremely confusing to me, so, I have no idea if this is needed to achience the simple callback or not.  But, we do need to get notified when these things happen in the table as a result of the user clicking save or cancel or undo.  I'm not using the stateful table, so not sure if this is ONLY for the stateful implementation.

src/components/Table/TableBody/TableBody.jsx
karlinglee93: In my opinion, the toast showed up when I clicked save button according to the design. And the toast blocked the toolbar (i.e. when the toast shows up, the toolbar icons or save/cancel buttons are hidden ). That's why I think the toast belongs to the toolbar and is an option here.

src/components/Table/Table.jsx
karlinglee93: I agree, I'll change `onSaveCurData ` to `saveData` or `onSaveData`. Do you think `saveData` is better than `onSaveData `? Because I named the prop according to `onApplySearch ` which is also a table toolbar action.

stuckless: I'm ok with 'onSaveData' since 'on' properties are self describing as callbacks/events.

src/components/Table/TableBody/TableBody.jsx
stuckless: Does the design say that the toast should show up and cover the action bar?  That seems odd.  Something like a toast should be consistent within an application and not just random show anywhere on the screen.  I agree that the "story" needs to show this, I just think that the table needs to manage it.  How long does it stay on the screen, can I dismiss it, how do I translate it, how do i hook into undo (based on the message)?  Can I disable the "undo" (not every table in the world will support it) ... if you are going to manage a toast then you need to answer all these kinds of questions and provide solutions.  This is why I think that the application needs to manage the toast.  This can be demostrated in the story, but, not sure I agree that the table should be showing the toast.  I would expect that when I click save the edit toolbar is disabled (can't click it twice) or it goes away... but I would think there is a pattern for saving / pending save.  @tay1orjones @davidicus @JordanWSmith15

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
karlinglee93: In my opinion, the default value of `editBar` is `null`, and when clicked edit icon in toolbar, set `editBar` to `buttons` (i.e. save and cancel buttons showed up). When clicked save button, set `editBar ` from `buttons ` to `toast` (i.e. the toast showed up). Or when clicked cancel button, set `editBar ` from `buttons ` to `null` (i.e. toolbar showed up with search, filter, edit icons, etc.). Besides, when clicked the toast, set `editBar ` from `toast ` to `null` (i.e. toolbar showed up with search, filter, edit icons, etc.).

src/components/Table/TableBody/TableBody.jsx
JordanWSmith15: @stuckless @tay1orjones  I would think that any toast is not managed by the table, but rather triggers from the product's implementation of a warning message (whatever that is), rather than a one-off toast built directly into the table. That way, the message format and location is consistent with all the other messages in the product.

src/components/Table/TableToolbar/TableToolbar.jsx
karlinglee93: I'm here to listen to everyone's opinions😄. I wrote the toast stuff entirely based on the design.

src/components/Table/Table.jsx
karlinglee93: got it.

src/components/Table/TableBody/TableBody.jsx
stuckless: @karlinglee93 yeah, i see the PAL note has

> Display toast with undo option https://ibm.invisionapp.com/d/main#/console/15350624/319220019/preview (Use light version of toast. Dark version does not match PAL update)

But I do think that this is something that needs to be managed at the application level, but the story should show this behaviour so that someone can implement it.  ie, the Pattern says that a toast needs to be shown, and as such when implementing a table, the app will need to implement that pattern.  I just feel there is so much configuration around a toast that we'd end up with a huge 'toast' configuration node on the table just account for all this.

src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
karlinglee93: got it, I'll give the editDataFunction the same signature of the renderDataFunction.

src/components/Table/TableBody/TableBody.jsx
karlinglee93: @stuckless , I understood that, but I have no ideas about how to code when we don't manage the toast at the application level. ie, how the app will need to implement that pattern when implementing a table, how the toast showed up when I clicked save button and how we undid datas when I clicked the toast. I mean I just see one way we can solve the table editing features according to the design. Do you have some suggestions?

karlinglee93:
hi @stuckless , let me clarify. Do you suggest to move the codes of toast from `TableToolbar` to `Table.story`? If so, it may take me some time to rethink/redesign how to show/hide the toast, how/where to undo data (At the moment, we undo data by calling `onUndoEditData` when click cancel button or the toast. If move the toast to the story, I think it becomes different from click cancel button). Thank you!

JordanWSmith15:
@karlinglee93 I think what he is suggesting is to simply trigger an event that can be interpreted buy the application to trigger _something_... a toast, a dialog, whatever. There is a lot of logic in our application regarding error messages and alerts that we have had to do via extending the toast component. In order for this to work for us, we would have to put that logic into this component, which would be duplicate code and not useful for other teams.

karlinglee93:
@JordanWSmith15 got it, thank you! I'll remove the option `toast` from `editBar` and trigger an event to show the toast after clicking save button if my understanding is right : ). @stuckless

karlinglee93:
hi @stuckless , I've separate `toast` from `editBar`, ie. the propType of `editBar` is bool now. Besides, I added `showAlert` to trigger an event to show the toast. But the toast component is still in `TableToolbar`, anything need to be improved about that?

src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -196,30 +266,74 @@ const TableToolbar = ({
...
+          kind="success"
+          lowContrast
+          notificationType="toast"
+          onCloseButtonClick={onAlertClose}
+          role="alert"
+          subtitle="Click to undo changes"
```
stuckless: If you are going to leave the Toast in the Toolbar, then, you'll need to add i18n string for all the text that you are using.  Personally, I don't think the "table" or "toolbar" should be firing a "toast" message.  I think the story can simulate this simply by hooking into the onSave and then showing it's own Toast from the story.

stuckless:
@tay1orjones @davidicus I'm pretty close to just accepting this, as is.  If the toast is left in, then it just needs to make sure that the i18n strings are defined properly, otherwise @karlinglee93 should just remove the toast from the table/toolbar.  Other than that... I'm ok with everything.

src/components/Table/TableToolbar/TableToolbar.jsx
davidicus: You need to wrap it in curly braces `import { IconEdit20 } from '@carbon/icons-react';` I believe.

src/components/Table/TableToolbar/TableToolbar.jsx
davidicus: I agree that it seems weird to have a toast as part of this component. The popup placement of the toast also seems visually off. @lukefirth @amymillett any comments on this from a visual/UX perspective? I second Sean's recommendation to just show the toast as part of the story to reinforce the pattern but not make a part of the Table component.

src/components/Table/_table.scss
```diff
@@ -1 +1,23 @@
...
+.toast {
+  margin-bottom: 0.5rem;
+  min-width: 30rem;
+  position: fixed;
+  top: 0;
+  right: 0;
+  z-index: 1;
+}
+
+.buttons {
```
davidicus: Class names such as `toast` and `buttons` are too generic and could clash in the global namespace. We should be using BEM syntax for our classnames or at the very least qualified as a button of the tableheader or wherever these classes are used.

src/components/Table/_table.scss
```diff
@@ -1 +1,23 @@
...
+  width: 88px;
+  height: 32px;
+  border-color: #0062ff;
+  border-width: 1px;
+  min-height: auto;
+  margin: auto 8px;
```
davidicus: we should be using carbon tokens for spacing and we should not be using px values.

src/components/Table/_table.scss
```diff
@@ -1 +1,23 @@
...
+  text-align: center;
+  padding: 0px;
```
davidicus: zero values do not need to have units here

src/components/Table/tableReducer.js
davidicus: @tay1orjones I agree with your comment. Would like to hear what @bryancboyd thinks. @stuckless These reducers are for the stateful implementation only I believe.

karlinglee93:
hi @davidicus @stuckless , I've updated the scss style and move the toast to the story. Please code review again, thank you!

@tay1orjones There are some conflicting files between my branch and the master. Could you please help me to merge again? Thank you!

src/components/Table/_table.scss
```diff
@@ -1 +1,23 @@
 @import 'TableHead/table-head';
+@import '~carbon-components/scss/components/data-table/data-table';
+
+.#{$prefix}--toast {
```
tay1orjones: This should be scoped to the table via a class applied in edit mode, otherwise this is a globally applied style.

Values, colors, etc should be tokens

src/components/Table/_table.scss
```diff
@@ -1 +1,23 @@
...
+  margin-bottom: 0.5rem;
+  min-width: 30rem;
+  position: fixed;
+  top: 0;
+  right: 0;
+  z-index: 1;
+}
+
+.#{$prefix}--toolbar-button {
```
tay1orjones: This should be scoped to the table via a class applied in edit mode, otherwise this is a globally applied style.

Values, colors, etc should be tokens

src/components/Table/_table.scss
```diff
@@ -1 +1,23 @@
 @import 'TableHead/table-head';
+@import '~carbon-components/scss/components/data-table/data-table';
```
tay1orjones: This import can be removed, Table relies on DataTable, which includes this import in it's sass partial

karlinglee93:
@tay1orjones , I've fixed and updated the scss, please review again. Thank you!

davidicus:
closing this to revisit and a new PR will be opened.

PR #790, opened 2019-12-15
refactor(center-story): stories now centered with official package #479

### LINKED ISSUE #479: [Storybook] Refactor how stories are centered
Storybook has an [official addon for centering stories.](https://github.com/storybookjs/storybook/tree/next/addons/centered)

We should move to this and deprecate our wrapper that does this. Should help us retain the ability for stories to be full width and still center, without having custom css.

sstone2423:
Closes #479

**Summary**

- @storybook/addon-centered --dev package added as the new standard for
centering stories. https://github.com/storybookjs/storybook/tree/next/addons/centered

- imported the new package and added the decorator in config.js

- Removed in-line styling from .storybook/Container.js as the new package handles it

- Updated the StoryShot snapshots because the new styling added new lines which shifted the rest of the stories down

- refactor(center-story): stories now centered with official package #479

- Check the built Storybook to see that each story is indeed centered. https://deploy-preview-790--carbon-addons-iot-react.netlify.com/


.storybook/Container.js
```diff
@@ -5,18 +5,6 @@ export default class Container extends Component {
   render() {
     const { story } = this.props;
 
-    return (
-      <div
-        className="storybook-container"
-        style={{
-          padding: '3rem',
```
tay1orjones: The stories that are larger than the viewport height are put right on the top/bottom edge.  Left this PR, right production:

image

Would like to keep the breathing room via the padding if possible.

sstone2423: Re-added the padding. Thanks!

stuckless:
One of the challenges that I see with this 'centering' approach is that it is very hard to tell what padding/margin the component is exporting by looking at it.  Some components have padding, some don't.  This makes it hard to tell.

sstone2423:
@stuckless While I understand your point, Carbon also centers their components in stories. It seems right to continue their style

davidicus:
@sstone2423 we just need to resolve this conflict and we can merge this.

PR #805, opened 2020-01-03
refactor(relative-import) #534

### LINKED ISSUE #534: Refactor carbon-components-react imports to relative imports
We shouldn't be importing components from `carbon-components-react` in the components. If a component in this library relies on another component in this library (Carbon proxy or not), it should be imported from a relative path, not `carbon-components-react`.

We need to have a `./src/components/ComponentName/index.js` for each component instead of importing and reexporting directly from `carbon-components-react` in `./src/index.js`.

Then, all existing imports from `carbon-componets-react` need to be refactored to use the new relative locations.

Basically looking to mirror the strategy that ibm-security has put forth in their addons repo. Their [Checkbox](https://github.com/carbon-design-system/ibm-security/blob/44a91d37eb8e1714c2ba963950cd34cba027069e/src/components/Checkbox/index.js#L6) export is a good example because it's a component that only deviates in style, not in function (js).

### LINKED ISSUE #543: [SideNav] icon height are not in line with pattern
The Sidenav menu heights found in Carbon are not in line with our [pattern](https://pages.github.ibm.com/Watson-IoT/IoT-Design-Site/components/ui-shell-side-nav/usage#anatomy) . Update to be 48px tall as spec. 

image

sstone2423:
Closes #534

**Summary**

These files follow the standard set by IBM Security and Carbon on how to import carbon components. More info can found in #534

Each component now uses relative imports vs importing directly from `carbon-components-react`.

The `src/index.js` has also been refactored and reorganized to reflect the new import strategy, as well as alphabetized for easier reading.

**Why did I split this up into 2 PRs?**

After some thought on how best to break this issue up, I have divided the work into 2 PR's. The first PR will be adding the index.js files to each carbon component, and changing the import paths in the main src/index.js. The second PR will change each component's import paths to use relative instead of carbon-components-react.

My reasoning is that if I were to split this task into more PRs, 5 components per PR for example, there would be a mismatch of how components are being imported across the library which could lead to confusion. By changing ALL components at once, per subtask, it helps to keep consistency.

- `src/index.js` 
- All IoT extension components that are currently using `carbon-components-react` components

- Check built web app below to ensure that all IoT extension components render and function correctly in Storybook.


davidicus:
@sstone2423 I think you have linked to the wrong issue in the "closes" part of your description. Can you update with the right issue number?

sstone2423:
@davidicus Good catch! I updated the numbers.

@scottdickerson I agree that it obscures it a bit. Especially on the components such as Button and ComposedModal, as this library has its own version, but we should also be exporting the Carbon versions

I can swap it back if needed

sstone2423:
@davidicus @tay1orjones @scottdickerson Should we go ahead and close this one?

scottdickerson:
can we kill this one, I really don't like the idea behind this one, and think it makes our components harder to understand..

PR #807, opened 2020-01-06
feat(tilecatalog): isSelectedByDefault prop that determines whether to force a selection or not

### LINKED ISSUE #809: [TileCatalog]  allow tile catalogs to start off unselected
### Summary

Add prop that allows tile catalogs to start off unselected

scottdickerson:
Closes #809 

**Summary**

- Added new property that allows tile catalogs to start off unselected

- feat(StatefulTileCatalog): add new isSelectedByDefault prop and only invoke onSelection if set true (true by default)
- feat(tileCatalogReducer): only select by default (in initial state) if isSelectedByDefault is true

- Check the new story and testcases and confirm the tile catalog is actually unselected


scottdickerson:
@bryancboyd @davidicus can I get a review on this one?

PR #808, opened 2020-01-06
feat(pagewizard): add onClearError prop

scottdickerson:
Closes #

**Summary**

- add onClearError prop

- add new onClearError prop so that the parent components can clear the error
- add Close button label to the i18n prop

- See new testcases


scottdickerson:
@davidicus can I get a review on this one?

PR #822, opened 2020-01-15
feat(csvdownload): export sharable csv download handler

scottdickerson:
Closes #

**Summary**

- should export the csv download handler in case others want to use it with their table card


stuckless:
What's the shareable csv handler?  Is there a story for this?  I checked the Table stories but nothing popped out at me that suggested that there was a download button, etc.

PR #825, opened 2020-01-15
fix(table): added overflow scroll and text nowrap

### LINKED ISSUE #824: [Table] overflow-x is not being hidden
### Describe the bug

image

Column overflows are not being hidden, resulting in the columns flowing past the table header and pagination.

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--simple-stateful-example
2. Reduce window size until a few columns are hidden
3. Scroll to the right on table
4. See styling issue

### Expected behavior

Fits to table

** Will fix myself**

sstone2423:
Closes #824

Fixes: 
Screen Shot 2020-01-15 at 1 59 56 PM

Screen Shot 2020-01-15 at 2 00 06 PM

**Summary**

- Wrapped the table in a div with styling to handle overflow in Table
- Added styling to nowrap the text as it got messed up when overflow was fixed
- Updated snapshots

- src/components/Table/Table.jsx
- src/components/Table/_table.scss

1. Go to Simple Stateful Table example
2. Reduce window size until a few columns are hidden
3. Scroll to the right on table
4. See that there are no longer styling overflow issues


src/components/Table/_table.scss
```diff
@@ -1 +1,9 @@
 @import 'TableHead/table-head';
+
+.table-container {
```
scottdickerson: let's get a little more specific here to prevent accidental collisions

src/components/Table/_table.scss
```diff
@@ -1 +1,9 @@
...
+  overflow-x: auto;
+}
+
+table.bx--data-table {
```
scottdickerson: prefix import for the bx

PR #829, opened 2020-01-16
Revised bread crumb

### LINKED ISSUE #719: [Breadcrumbs] Breadcrumb on Mobile

davidicus:
Closes #719 

**Summary**
Add ability to collapse breadcrumbs when container is too small

- created Breadcrumb.jsx

- Go to breadcrumb story and look at the resizeObserver one.


davidicus:
Test need to be written. Low threshold is failing the travis build.

davidicus:
Need to open issue on Carbon's overflow  menu for focus trap.

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,119 @@
...
+  const breakingWidth = useRef([]);
+  const { ref: breadcrumbRef } = useResizeObserver({
+    useDefaults: false,
+  });
+  const [overflowItems, setOverflowItems] = useState([]);
+  const [afterOverflowItems, setAfterOverflowItems] = useState(children.slice(1));
+
+  /** update breadcrumbs  */
+  useEffect(() => {
+    if (hasOverflow) {
```
stuckless: I was really consfused about this name `hasOverflow`, but I see now it's an indication that the component will actually check for overflows.

davidicus: > In RTL mode it doesn't appear to be rendering correctly.
> 
> I'm trying to understand how this breadcrumb would be used. In a graphite app, the breadcrumb will be rendered as a link but, we need to hook into the onClick of the item and transition the page using an even on the item instead of a url. Is that even possible here?

I will fix RTL defects. Thanks for the catch. 

For the naming I was trying to stay in line with the `has-*` convention for denoting features that the component has like `Table` component's `hasRowSelection`. 

As for the functionality, I needed to pass the href prop from `BreadCrubmItem` to `OverflowItem`. Adding this now.

davidicus: RTL and onClick items should now be addressed.

davidicus:
@stuckless RTL and onClick items should now be addressed

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -0,0 +1,124 @@
...
+          breakingWidth.current[breakingWidth.current.length - 1] &&
+        overflowItems.length > 0
+      ) {
+        // Move the item to the visible list
+        setAfterOverflowItems([overflowItems.pop(), ...afterOverflowItems]);
+        setOverflowItems([...overflowItems]);
+        breakingWidth.current.pop();
+      }
+    }
+  });
```
scottdickerson: hey I thought we were gonna protect this so that it only reexecuted if the clientWidth and scrollWidth changed

davidicus: I had added your deepCompare hook in place but the package threw an error because of the type of  object it was. And if I add `breadcrumbref.current` to the dependency array it breaks the functionality.

PR #833, opened 2020-01-16
fix(header): an extra click was required to expand the header menu

scottdickerson:
Closes #

**Summary**

- For some reason, an extra click was required to expand the header menu

- refactor(Header): split the header into multiple subcomponents and move the state to the HeaderActionGroup component

- Compare clicking and keyboard activating the header between this version and the old one.  The old one loses many of the clicks because of the fixes


davidicus:
There was weird opening and closing behavior which is why the onFocus handler was placed on the top level menu. Perhaps there is a better solution but this behavior is now back and present in the preview link. 

Click on icon on the far right. panel opens. Click on the icon directly to the left and the dropdown menu overlays the header panel. There are others instances where one menu does not shut the other.

PR #843, opened 2020-01-21
feat(statefultable): add sortFunction prop for custom column sorts

scottdickerson:
Closes #https://github.com/IBM/carbon-addons-iot-react/issues/251

**Summary**
Allows the StatefulTable consumers to pass a custom sortFunction prop on their column.

- feat(StatefulTable): pass the columns to the reducer so it can custom sort if a custom sort function is passed for that column
- test(Table.story): add a customSortFunction on the Status field of the table so we can integration test it
- feat(TablePropTypes): new sortFunction property
- feat(tableReducer): take the custom sort function into account in the filterSearchAndSort (which resorts the table when a new filter is applied)
- feat(tableReducer): use the custom sort function instead of our default getSortedData if it exists on the column

- Verify the custom sort works on the Status column in the Table story.  Broken should be sorted to the top, then Running, then Not_Running and vice versa on descending sort

### DIFF
--- a/src/components/Table/Table.story.jsx
+++ b/src/components/Table/Table.story.jsx
@@ -62,7 +62,6 @@ const renderStatusIcon = ({ value: status }) => {
 };
 // Example custom sort method for the status field.  Will sort the broken to the top, then the running, then the not_running
 const customColumnSort = ({ data, columnId, direction }) => {
-  console.log(`custom sort triggered on ${columnId} for ${direction}`);
   // clone inputData because sort mutates the array
   const sortedData = data.map(i => i);
   sortedData.sort((a, b) => {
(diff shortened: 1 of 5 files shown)


src/components/Table/Table.story.jsx
```diff
@@ -60,6 +60,31 @@ const renderStatusIcon = ({ value: status }) => {
       );
   }
 };
+// Example custom sort method for the status field.  Will sort the broken to the top, then the running, then the not_running
+const customColumnSort = ({ data, columnId, direction }) => {
+  console.log(`custom sort triggered on ${columnId} for ${direction}`);
```
sstone2423: Should this be removed?

PR #851, opened 2020-01-27
fix(header): moved state lower for event handling

### LINKED ISSUE #850: [Header] HeaderActions require 2 clicks to activate the panel
### Describe the bug

If a HeaderAction is already focused, it takes 2 clicks to focus / click another HeaderAction. In Storybook, it can be difficult to reproduce so try to change focus between the different actions very quickly multiple times. This primarily showed up in our production instance.

### To Reproduce

1. Go to Header story
2. Click on 1 HeaderAction
3. Rapidly click another HeaderAction
4. Repeat 2 & 3 until bug occurs

### Expected behavior

HeaderActions should expand on single click

### Environment/versions:

- OS: MacOS
- Browser: Chrome
- `carbon-addons-iot-react` version: [e.g. `v2.7.1`]

sstone2423:
Closes #850 

Old behavior: 

**Summary**

- The `onBlur` and `onClick` events were both being fired if another HeaderAction was already in focus. This behavior was unpredictable and it is believed that the `onBlur` was causing a re-render, which would "throw away" the second event
- The `isExpanded` state was moved into the HeaderAction rather than the HeaderActionGroup in an effort to avoid the event where the `expandedItems` object was getting overwritten by other events. By moving it to the action itself, it became unaware of other action's events.

- src/components/Header/HeaderAction/HeaderAction.jsx
- src/components/Header/HeaderActionGroup.jsx
- src/components/Header/HeaderAction/HeaderActionPanel.jsx

1. Go to Header story
2. Click on 1 HeaderAction
3. Rapidly click another HeaderAction
4. Repeat 2 & 3 multiple times
5. See that the bug no longer occurs

**Note:** This primarily occured when the Header was implemented in our actual project

### DIFF
--- a/src/components/Header/HeaderAction/HeaderActionPanel.jsx
+++ b/src/components/Header/HeaderAction/HeaderActionPanel.jsx
@@ -41,7 +41,7 @@ const HeaderActionPanel = ({ item, index, onToggleExpansion, isExpanded, focusRe
         aria-label={item.label}
         aria-haspopup="menu"
         aria-expanded={isExpanded}
-        onClick={onToggleExpansion}
+        onClick={() => onToggleExpansion()}
         ref={focusRef}
       >
         {item.btnContent}
--- a/src/components/Header/HeaderAction/HeaderAction.jsx
+++ b/src/components/Header/HeaderAction/HeaderAction.jsx
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react';
+import React, { useRef, useState } from 'react';
 import { settings } from 'carbon-components';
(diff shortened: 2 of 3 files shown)


src/components/Header/HeaderAction/HeaderActionPanel.jsx
```diff
@@ -41,7 +41,7 @@ const HeaderActionPanel = ({ item, index, onToggleExpansion, isExpanded, focusRe
         aria-label={item.label}
         aria-haspopup="menu"
         aria-expanded={isExpanded}
-        onClick={onToggleExpansion}
+        onClick={() => onToggleExpansion()}
```
scottdickerson: I think you probably don't need this change

PR #856, opened 2020-01-28
feat(list): List component

### LINKED ISSUE #706: [Component Proposal] SimpleList
### Summary

Basic implementation of a `List` component that supports:

* Search
* Actions (Header buttons)
* Pagination

Search and pagination will be handled in component state (async search/paging would use the base `<List />` component)

**Additional context**

image

image

image

### You will consider this component complete when...

- [ ] It's available as an export of this library

### Available extra resources

@jessieyan

jessieyan:
Closes #706 

**Summary**
The experimental List component and Simple List component.

SimpleList has search, header actions and pagination. Support large row, row actions, expansion and select.


src/components/List/List.jsx
```diff
@@ -0,0 +1,139 @@
...
+  content: PropTypes.shape({
+    value: PropTypes.string,
+    icon: PropTypes.node,
+  }),
+  children: PropTypes.arrayOf(PropTypes.any), // TODO: make this recursive
+  isSelectable: PropTypes.bool,
+};
+
+const propTypes = {
+  title: PropTypes.string.isRequired,
```
davidicus: Would love some comments as to these props are for.

src/components/List/List.jsx
```diff
@@ -0,0 +1,139 @@
...
+        onExpand={toggleExpansion}
+        selected={isSelected}
+        expanded={isExpanded}
+        isExpandable={hasChildren}
+        isLargeRow={isLargeRow}
+        isCategory={isCategory}
+        isSelectable={isSelectable}
+      />,
+      ...(hasChildren && isExpanded
+        ? item.children.map(child => renderItemAndChildren(child, level + 1))
```
davidicus: We may want to use `React.Children.map` to loop over in case there is only one child? depending on if this is just pulling in Child elements or you are passing this as an object.

src/components/List/ListHeader/_list-header.scss
```diff
@@ -0,0 +1,44 @@
+.list-header {
```
davidicus: we have added an IoT prefix that we should use to differentiate our styles from Carbon and the consuming application. 

https://github.com/IBM/carbon-addons-iot-react/blob/master/src/globals/_vars.scss

https://github.com/IBM/carbon-addons-iot-react/blob/caa120186687098c100639e226cc9c3cceb59449/src/components/Button/Button.jsx#L9

src/components/List/ListHeader/_list-header.scss
```diff
@@ -0,0 +1,44 @@
...
+    display: flex;
+    padding: $spacing-03;
+    align-items: center;
+    & > * {
+      margin-left: $spacing-03;
+    }
+  }
+
+  &--search {
+    padding: 0 1rem;
```
davidicus: should use spacing token here for this rem value

src/components/List/ListItem/ListItem.story.jsx
```diff
@@ -0,0 +1,176 @@
...
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import { text, select, boolean } from '@storybook/addon-knobs';
+import { Edit16, Star16, StarFilled16 } from '@carbon/icons-react';
+
+import { Button, OverflowMenu, OverflowMenuItem } from '../../..';
+
+import ListItem from './ListItem';
+
+storiesOf('Watson IoT Experimental|ListItem', module)
```
davidicus: i see we have the first story with knobs. Can we eliminate some of these stories if they can be represented with Knobs?

src/components/List/ListItem/_list-item.scss
```diff
@@ -0,0 +1,105 @@
+.list-item {
```
davidicus: same comment about the prefix

src/components/List/SimpleList/SimpleList.jsx
```diff
@@ -0,0 +1,110 @@
...
+    const currentItemsOnPage = filteredItems.slice(rowUpperLimit - rowPerPage, rowUpperLimit);
+    setCurrentPageNumber(page);
+    setItemsToShow(currentItemsOnPage);
+  };
+
+  const pagination = {
+    page: currentPageNumber,
+    onPage: onPage,
+    maxPage: Math.ceil(numberOfItems / rowPerPage),
+    pageOfPagesText: page => `Page ${page}`,
```
davidicus: Does this need to be translated?

src/components/List/SimpleList/SimpleList.story.jsx
```diff
@@ -0,0 +1,320 @@
...
+  <Add16 onClick={action('add')} />,
+  <Close16 onClick={action('close')} />,
+];
+
+const rowActionsOverFlowMenu = [
+  <OverflowMenu flipped>
+    <OverflowMenuItem itemText="Edit" />
+    <OverflowMenuItem itemText="Add" />
+    <OverflowMenuItem itemText="Delete" />
+    <OverflowMenuItem itemText="Danger option" hasDivider isDelete />
```
davidicus: Translation?

src/components/List/SimpleList/SimpleList.story.jsx
```diff
@@ -0,0 +1,320 @@
+import React from 'react';
```
davidicus: Can we limit the amount of stories needed here with the use of Knobs?

src/components/List/_list.scss
```diff
@@ -1 +1,33 @@
-@import '~carbon-components/scss/components/list/list';
+.list {
```
davidicus: def need a prefix for class names that are this generic.

davidicus:
We will need to add this to the index.js file to export

src/components/List/ListHeader/_list-header.scss
bryancboyd: hey, cool!  didn't know this.

src/components/List/ListHeader/_list-header.scss
bryancboyd: 1 rem = $spacing-05

src/components/List/SimpleList/SimpleList.story.jsx
bryancboyd: That's all story stuff -- the user can provide whatever text they'd like, there (we'd assume it's translated)

src/components/List/SimpleList/SimpleList.story.jsx
bryancboyd: @davidicus do we have a ideal range for # stories?

src/components/List/ListHeader/ListHeader.jsx
```diff
@@ -0,0 +1,55 @@
...
+      {search && (
+        <div className={`${iotPrefix}--list-header--search`}>
+          <Search
+            placeHolderText={i18n.searchPlaceHolderText}
+            onChange={search.onChange}
+            size="sm"
+            small
+            className=""
+            value={search.value}
+            labelText="Search"
```
stuckless: needs to be in the i18n structure

stuckless:
One other little thing... when you mouse over the buttons in the header/row the title obstructs the button and it appears immediately.  Compared to when you hover over the overflow menu the title doesn't show until after a second and when it does show it's below button so that it doesn't obstruct it (ie, good).

bryancboyd:
> One other little thing... when you mouse over the buttons in the header/row the title obstructs the button and it appears immediately. Compared to when you hover over the overflow menu the title doesn't show until after a second and when it does show it's below button so that it doesn't obstruct it (ie, good).

This seems to be the default behavior for carbon `iconOnly` buttons; seems a bit odd to me too.  If we want to move the positioning of the accessible hover text below the button, we should make this change in our IoT button.  @davidicus have you seen this?

(Also, the buttons provided in header/rows are determined by the application rendering the `List`, passed in via props; the List itself doesn't import the button, it just renders whatever node is passed in on the prop)

src/components/List/List.jsx
bryancboyd: Just an object 👍

src/components/List/ListItem/ListItem.story.jsx
bryancboyd: Could you take a pass at this @davidicus?  Would be best I think to have the component library caretakers drive consistency in how we do our IoT stories, at least until we have a solid pattern to follow 👍

src/components/List/SimpleList/SimpleList.jsx
bryancboyd: yep fixed

jessieyan:
> The list appears to have different "sizes" for the number of items visible... Is there a size that simply fills its parent container? ie, if I put this list in a dialog and the dialog content is 200px high the list just fills that space.
> 
> For the data... Does this list support incrementally loading the data? ie the classic scroll to fetch more data? Similarly, does the child expansion support async loading of children?
> 
> Much of the list work ok in RTL mode but there are some definate issues in the header and some other cases that forces extra scrolling, etc. Each story should be tested in RTL mode.

Hi Sean, 

SimpleList component has pagination. In SimpleList, if pagesize not specified, it will fill the parent container with all the rows. And if rows overflow the parent container, there is scroll overflow.  List component itself does not have pagination. 

In SimpleLIst component, we do not support incremental loading or async loading. For List, maybe we could compute data before pass in props items to List? We are trying to make List component simple and flexible so that user can use it in their own way.

src/components/List/ListItem/ListItem.jsx
```diff
@@ -0,0 +1,190 @@
...
+  onExpand,
+  expanded,
+  isSelectable,
+  onSelect,
+  selected,
+  value,
+  secondaryValue,
+  rowActions,
+  icon,
+  iconPosition = 'left', // or "right"
```
davidicus: we already set this as a default in the default props. Do we need here as well?

src/components/List/SimpleList/SimpleList.story.jsx
davidicus: < 5. I think if stories are needed we should show the functionality but we also do not want stories to get lost like they do for components like the Table or valuecards which have huge list of stories to go through and you can not easily make out what they are. 

For these stories I can not tell the difference between the last two. Are both needed?

src/components/List/SimpleList/SimpleList.story.jsx
davidicus: sorry. True that.

src/components/List/_carbon_list.scss
```diff
@@ -0,0 +1 @@
+@import '~carbon-components/scss/components/list/list';
```
davidicus: Is there a reason this needs to be in a separate file. Is Carbon's list unrelated to this component?

src/components/List/_list.scss
```diff
@@ -1 +1,35 @@
...
+    overflow-y: auto;
+  }
+
+  &__full-height {
+    height: 100%;
+  }
+
+  &--page {
+    position: relative;
+    div[class^='SimplePagination__StyledContainer'] {
```
davidicus: do we need to qualify these selectors with elements?

src/components/List/_carbon_list.scss
jessieyan: We move inherent carbon _list.scss to _carbon_list.scss so that we could have our own List style.

src/components/List/_list.scss
bryancboyd: I think it's already qualified, right?  div w/ class attribute, inside of the `.list--page` should get that style applied.

PR #861, opened 2020-01-29
fix(SimplePagination): styling now uses correct vars

### LINKED ISSUE #477: [SimplePagination] Incorrect background color
image

This background should be white, as it is on the primary pagination component. I suspect we're not using a theme variable for this, which is why it didn't change from the v9-v10 upgrade.

sstone2423:
Closes #477 

**Summary**

- Migrated styling to scss #370 
- Changed colors to use correct Carbon vars

Screen Shot 2020-01-28 at 11 20 35 PM

- Check SimplePagination story to see that colors are correct


src/components/SimplePagination/_simple-pagination.scss
```diff
@@ -0,0 +1,28 @@
+@import '~carbon-components/scss/globals/scss/vars';
+
+div.simple-pagination-container {
```
davidicus: do we have to use these element selectors?

src/components/SimplePagination/_simple-pagination.scss
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+  height: 3rem;
+  justify-content: flex-end;
+  width: 100%;
+  border: 1px solid $ui-03;
+  background-color: $ui-background;
+  align-items: center;
+
+  span.simple-pagination-page-label {
+    padding-right: 1rem;
```
davidicus: should use the Carbon spacing token here

src/components/SimplePagination/_simple-pagination.scss
davidicus: We should use our new iot prefix for our styles moving forward.

davidicus:
Looking at the storybook I still see the gray background. 

image

src/components/SimplePagination/_simple-pagination.scss
```diff
@@ -0,0 +1,28 @@
+@import '~carbon-components/scss/globals/scss/vars';
+
+div.simple-pagination-container {
+  display: flex;
+  height: 3rem;
+  justify-content: flex-end;
+  width: 100%;
+  border: 1px solid $ui-03;
+  background-color: $ui-background;
```
davidicus: I believe this needs to be switched to `$ui-01`

sstone2423: > Looking at the storybook I still see the gray background.
> 
> image

I noticed Carbon uses gray for their pagination as well. I'm thinking it may have switched in v10?

sstone2423: http://react.carbondesignsystem.com/?path=/story/pagination--pagination

src/components/SimplePagination/_simple-pagination.scss
sstone2423: Nope, changed them over to regular classnames

src/components/SimplePagination/_simple-pagination.scss
sstone2423: Fixed

src/components/SimplePagination/_simple-pagination.scss
sstone2423: Fixed

src/components/SimplePagination/_simple-pagination.scss
davidicus: yes, our gray10 theme though has white color for these elements that are on a gray background

davidicus: that being said. The theme should be handling the change in color..🤔

sstone2423: Swapped to `$ui-01`

src/components/SimplePagination/_simple-pagination.scss
```diff
@@ -0,0 +1,29 @@
+@import '~carbon-components/scss/globals/scss/vars';
+@import '../../globals/vars';
+
+.#{$iot-prefix}-simple-pagination-container {
+  display: flex;
+  height: $spacing-09;
+  justify-content: flex-end;
+  width: 100%;
+  border: 1px solid $ui-03;
```
jessieyan: Do we usually wrap px with rem()?

PR #863, opened 2020-01-29
feat: Piglets GRAPHITE-5346: #Health widget extension with donut graph

psdixit:
Closes #

Summary

Allow to display donut graph in valuecard

src/components/Card/_card.scss 
src/components/Dashboard/CardRenderer.jsx 
src/components/Dashboard/Dashboard.story.jsx 
src/components/Dashboard/__snapshots__/Dashboard.story.storyshot 
src/components/ValueCard/Attribute.jsx 
src/components/ValueCard/GraphRenderer.jsx 
src/components/ValueCard/ValueCard.jsx 
src/components/ValueCard/ValueCard.story.jsx 
src/components/ValueCard/ValueRenderer.jsx 
src/components/ValueCard/__snapshots__/ValueCard.story.storyshot 

Allow to show donut graph in side the valuecard by passing cardType,
The valuecard getting tow parameters ( value , thresholdRange) as attributes.
On the behalf of value and thresholdRange, card display as donut chart

- tests here

### DIFF
--- a/src/components/Dashboard/CardRenderer.jsx
+++ b/src/components/Dashboard/CardRenderer.jsx
@@ -169,7 +169,7 @@ const CardRenderer = React.memo(
     };
 
     return type === CARD_TYPES.VALUE ? (
-      <ValueCard {...commonCardProps} />
+      <ValueCard {...commonCardProps} cardType="DONUT" />
     ) : type === CARD_TYPES.IMAGE ? (
       <ImageCard {...commonCardProps} error={card.setupError || card.error} />
     ) : type === CARD_TYPES.TIMESERIES ? (
--- a/src/components/Dashboard/CardRenderer.jsx
+++ b/src/components/Dashboard/CardRenderer.jsx
@@ -169,7 +169,7 @@ const CardRenderer = React.memo(
     };
 
     return type === CARD_TYPES.VALUE ? (
-      <ValueCard {...commonCardProps} cardType="DONUT" />
+      <ValueCard {...commonCardProps} />
     ) : type === CARD_TYPES.IMAGE ? (
       <ImageCard {...commonCardProps} error={card.setupError || card.error} />
     ) : type === CARD_TYPES.TIMESERIES ? (
--- a/src/components/ValueCard/Attribute.jsx
+++ b/src/components/ValueCard/Attribute.jsx
@@ -92,7 +92,7 @@ const propTypes = {
   ),
   renderIconByName: PropTypes.func,
   precision: PropTypes.number,
-  cardType: PropTypes.string,
+  cardType: PropTypes.oneOf(['VALUE', 'DONUT']),
 };
 
 const defaultProps = {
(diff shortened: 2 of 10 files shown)


psdixit:
@davidicus , I have created a PR #863 , Please review this code and provide you comments on this PR

src/components/ValueCard/Attribute.jsx
```diff
@@ -81,21 +82,32 @@ const propTypes = {
...
+    PropTypes.shape({
+      lower: PropTypes.number,
+      upper: PropTypes.number,
+      color: PropTypes.string,
+      rangename: PropTypes.string,
+    })
+  ),
   renderIconByName: PropTypes.func,
   precision: PropTypes.number,
+  cardType: PropTypes.string,
```
davidicus: This should probably be a `oPropTypes.oneOf`

src/components/Dashboard/CardRenderer.jsx
```diff
@@ -169,7 +169,7 @@ const CardRenderer = React.memo(
     };
 
     return type === CARD_TYPES.VALUE ? (
-      <ValueCard {...commonCardProps} />
+      <ValueCard {...commonCardProps} cardType="DONUT" />
```
scottdickerson: please don't force all value cards to be DONUT here :)

davidicus:
image
 The text and donut do not seem to be rendering at the proper size in the storybook. I also think these value cards are supposed to always have a title on them. @scottdickerson is this correct?

davidicus:
Please fill out the template with the issue that this will close.

scottdickerson:
Donut cards should be a totally new type of card I think, and instead of writing your own custom Graph donut, you should really be using the donut available here https://github.com/carbon-design-system/carbon-charts

davidicus:
image

JordanWSmith15:
@scottdickerson @stuckless So, this goes against all our assumptions for this story that we worked on for the past two weeks. We investigated the Circle chart from Carbon, and it was looking more difficult to adapt it to our use case than to just make our own (it is not structured aa 100% gauge). Second, we needed the trend indicator, which the current value card supports.

stuckless:
@scottdickerson  The premise here is that a value card, should have an option to render the value in a donut.  We don't want to create a completely new card, with trend information, and all the other features of value card just to show a donut.   A donut card is not card the just shows a donut... it's a value card that happens to 100% of the stuff that value card provides but also allows you to renderthe value in a donut, the donot can change color, etc, etc.  This is not a new component... it's an iteration of the existing component.

If the value vard was more composable and supported a render function for the value... then maybe this could be done entirely as something that just passes a donut renderer to the value card, but, the cards are not designed that way.

I agree that the default behaviour of a value card should be no donut and the standard rendering that we see with the default value card.  It's not until we set the do nut properties that the donut would appear.

psdixit:
I have made all changes please review again..

sstone2423:
FYI: This new PR will probably close this issue and PR https://github.com/IBM/carbon-addons-iot-react/pull/885

scottdickerson:
this has been superceded by the new PR

PR #864, opened 2020-01-30
test(table): wip for pre-set toolbar search value

### LINKED ISSUE #429: [Table]: Search default value and clear filters not working as expected
If I set the `toolbar.search.value` state with a value, the Table's Search does not reflect that value.  It does appear to apply that search but it does not reflect it in the search text box.

Likewise clear all filters action appears to only clear the filters and not the search text (that might be by design, which is fine).  This is where i noticed the issue, since I don't allow both a filter and a search to co-exist, so once a filter is applied, I clear the search box, but the table never actually updates that search text box field.

The other use case is that our search state comes from a datasource which holds the current search that is applied.  When restoring state to the table from the datasource, and setting the `toolbar.search.value` to a value, the textbox does not reflect that value.

## Expected Behaviour
When we set or clear toolbar.search.value the table's Sarch textbox should visually update with that value.

You can easily test this in storybook, by, quickly, setting the default props value for `view.toolbar.search.value` to some text, and see if the storybook table shows that value (it doesn't).

enricoberti:
DO NOT MERGE

Closes #429

**Summary**
There's a new test on the story for the stateful available at ?path=/story/watson-iot-table--stateful-example-with-initial-search-value where you can see that setting it (to 'option-A') actually filters the table

In the stateless table there's a new knob to test it as well

### DIFF
--- a/.storybook/__snapshots__/Welcome.story.storyshot
+++ b/.storybook/__snapshots__/Welcome.story.storyshot
@@ -2528,6 +2528,30 @@ exports[`Storybook Snapshot tests and console checks Storyshots Getting Started|
                       <div
                         className="bx--structured-list-td"
                       >
+                        EmptyTable
+                      </div>
+                    </div>
+                    <div
+                      className="bx--structured-list-row"
+                    >
+                      <div
+                        className="bx--structured-list-td"
+                      />
+                      <div
+                        className="bx--structured-list-td"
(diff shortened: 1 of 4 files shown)


davidicus:
closing as issue is fixed in Carbon's upcoming release candidate. Will open an issue to track.

PR #865, opened 2020-01-30
fix(dashboardgrid): if card is missing from layout regenerate the layout

scottdickerson:
Closes #https://github.ibm.com/wiotp/monitoring-dashboard/issues/717

This is to fix this bug where if the user hasn't been to a layout breakpoint before (i.e. they added a card but didn't visit all the possible layout sizes to generate the layout), the cards were assuming that they were SMALL instead of using their true size.

image

**Summary**

- refactor(DashboardGrid): extract the layout generation to another function
- fix(DashboardGrid): add validation of layout to make sure all cards exist
- test(DashboardGrid): added testcases to validate the layout gets regenerated if a card is missing

### DIFF
--- a/src/components/Dashboard/DashboardGrid.jsx
+++ b/src/components/Dashboard/DashboardGrid.jsx
@@ -2,11 +2,11 @@ import React, { useMemo, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { Responsive, WidthProvider } from 'react-grid-layout';
 import styled from 'styled-components';
+import some from 'lodash/some';
 import find from 'lodash/find';
 
 import { getLayout } from '../../utils/componentUtilityFunctions';
 import {
-  CARD_SIZES,
   GUTTER,
   ROW_HEIGHT,
   CARD_DIMENSIONS,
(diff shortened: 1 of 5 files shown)


src/components/Dashboard/DashboardGrid.jsx
```diff
@@ -2,11 +2,11 @@ import React, { useMemo, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { Responsive, WidthProvider } from 'react-grid-layout';
 import styled from 'styled-components';
+import some from 'lodash/some';
```
sstone2423: This is native in Javascript now: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/utils/componentUtilityFunctions.js
```diff
@@ -2,6 +2,8 @@ import delay from 'lodash/delay';
 import moment from 'moment';
 import { sortStates } from 'carbon-components-react/lib/components/DataTable/state/sorting';
 import fileDownload from 'js-file-download';
+import find from 'lodash/find';
+import some from 'lodash/some';
```
sstone2423: Same as before,  this is native in JS https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/utils/componentUtilityFunctions.js
```diff
@@ -145,6 +153,14 @@ export const printGrid = grid => {
   console.log(result); // eslint-disable-line
```
sstone2423: Remove this? Unless its only being used by tests / storybook

src/utils/componentUtilityFunctions.js
```diff
@@ -145,6 +153,14 @@ export const printGrid = grid => {
...
+/**
+ *
+ * @param {*} x  the current x location of a card
+ * @param {*} y  the current y location of a card
+ * @param {*} w  current width of a card
+ * @param {*} h  current height of a card
+ * @param {*} grid nested array of rows and columns with the current card index that is occupying them. for example, if the entire top row was taken by the first card it would look like this [[1, 1, 1, 1], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]]
+ */
 export const canFit = (x, y, w, h, grid) => {
   // console.log(`canFit? x=${x}, y=${y}, w=${w}, h=${h}`);
```
sstone2423: Remove this? Unless its only being used by tests / storybook. Its also commented out

src/utils/componentUtilityFunctions.js
scottdickerson: in this case I actually want the simpler predicate to make the code more readable

src/components/Dashboard/DashboardGrid.jsx
scottdickerson: in this case I want the simpler predicate to make the code a little more readable 
image

PR #870, opened 2020-02-03
fix(floating-menu): add floating menu container to Storybook

### LINKED ISSUE #576: [Table] Data Table component with overflow menu option does not close on a tablet when scrolling
Steps:
We see an issue with the opened menu in maximo while using the IoT Data Table overflow menu option
This issue happens only on ipad. The overflow menu option gets displaced when the records in the table are scrolled. 

Open the IoT Storybook
http://react.carbondesignsystem.com/?path=/story/datatable--with-overflow-menu
Click on DataTable -- overflow-menu option on the left panel
Click on the 3 dot menu
Note- Menu option gets expanded
When scrolling the menu option remains opened which cause the options to be displaced in the consumed product

IoT_dataTable

maximo_DataTable

enricoberti:
Closes #576

**Summary**

To get the floating menu to scroll together with the page, it should be attached to its parent element instead of `<body>` itself. To do that, you just need to add `data-floating-menu-container` to the element you want the menu to attach to (in this case the storybook container)

Verify it on ?path=/story/watson-iot-table--stateful-example-with-secondary-title

Please not that our storybook has a wrapping element with the styling `position: fixed` wrapping the stories, so the menu offset calculation fails sometimes. It's not a problem in normal apps.

### DIFF
--- a/src/components/Table/Table.story.jsx
+++ b/src/components/Table/Table.story.jsx
@@ -483,7 +483,10 @@ storiesOf('Watson IoT|Table', module)
         }}
       />
     </FullWidthWrapper>
-  ))
+  ),
+  {
+    centered: { disable: true },
+  })
   .add(
     'Stateful Example with every third row unselectable',
     () => (


stuckless:
If normal apps don't use `position: fixed` then maybe we shouldn't use it by default in storybook?  

I'm seeing an issue when I scoll the page and then click on overflow menu that the overflow menu is not positioned correctly.  Now this could be caused by what you described here... but, how can we be sure, and can we validate this?  Should we try to turn off the position fixed in the story so that we can validate the correct behaviour will happen?

enricoberti:
@stuckless thanks for going through this.

In our case the 'position:fixed' on storybook is from the 'centered' addon. My recommendation is to remove that, and with that at the same time we get the same experience of Carbon (they don't center the components in the stories) and also avoid positioning problems like this. (@davidicus ).

I updated this PR removing the centered addon for the specific story and it fixes the offset problems.
You can test it here: https://deploy-preview-870--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-table--stateful-example-with-secondary-title
If we implement it storybook-wise, this also eliminates the need of adding `data-floating-menu-container` to our storybook container.

Also, as you said, this doesn't fix the problem downstream but it helps with fixing it. I don't have access to the downstream implementation to check for it.
It is not really well documented on [Carbon's component](https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/OverflowMenu/OverflowMenu.js#L444) but it should be the choice of who's developing an app wether to attach an overflow menu to the body (should be 99.9% of the cases) or to another element so the positioning works correctly.
Also, given how the floating menu positioning is implemented, it cannot be arbitrarily added to the best available parent, like the table itself in our case.

stuckless:
@enricoberti Thanks for looking into this.  So, I can confirm it appears to work ok in my browser, I guess the original bug was opened up against this only happening on on an ipad.  I don't have an ipad (here) for testing... have you verified that even with your most recent change that it behaves the same on ipad vs desktop.

Nothing to do with your change :)  but these kinds of behaviours scare me, a little, since it's like a mine field for the downstream consumer.  ie, when do I set this, where, how, etc :(

enricoberti:
@stuckless yes, verified on the ipad too :) i know how you feel with that - it should actually be documented better also on the Carbon side of things

PR #883, opened 2020-02-04
[List] HierarchyList stateful component

### LINKED ISSUE #882: [List] Exportable stateful hierarchical list
### Summary

A common use-case for the List is to have nested list items categorized in some way. Instead of leaving it up to the user to make their own implementation of the different states, we should create and export our own version for the simplest use-cases

End of February release

### Available extra resources

Will create myself

sstone2423:
Closes #882

**Summary**

- Stateful list that supports nested lists categorized in some way
- Supports recursive 1-n search
- Supports pagination
- Adds Multi-select, nested Checkbox Story to base List component

- HierarchyList component added
- HierarchyList tests added
- HierarchyList story added

- See HierarchyList story to test functionality


sstone2423:
@stuckless Updated. Good catch

package.json
```diff
@@ -78,6 +78,7 @@
     "carbon-icons": "7.0.7",
     "classnames": "^2.2.5",
     "core-js": "2.6.5",
+    "deepdash": "^4.4.2",
```
scottdickerson: since we're only using deepdash in the storybook, it should be in devDependencies, not dependencies

src/components/List/List.jsx
```diff
@@ -10,13 +10,13 @@ import ListHeader from './ListHeader/ListHeader';
...
 
-const itemPropTypes = {
+export const itemPropTypes = {
   id: PropTypes.string,
   content: PropTypes.shape({
     value: PropTypes.string,
     icon: PropTypes.node,
   }),
-  children: PropTypes.arrayOf(PropTypes.any), // TODO: make this recursive
+  children: PropTypes.arrayOf(PropTypes.any), // TODO: make this recursive on items?
```
scottdickerson: I don't think you can , so you can remove the comment

src/components/List/ListItem/ListItem.jsx
```diff
@@ -101,7 +101,10 @@ const ListItem = ({
         onClick={handleExpansionClick}
         onKeyPress={({ key }) => key === 'Enter' && handleExpansionClick()}
       >
-        <Icon icon={expanded ? iconChevronUp : iconChevronDown} />
+        <Icon
+          icon={expanded ? iconChevronUp : iconChevronDown}
+          description={expanded ? 'Expand' : 'Close'}
```
scottdickerson: these strings need to be passed in as translatable props

package.json
sstone2423: Moved to devDeps

src/components/List/List.jsx
sstone2423: I didn't think so. Changed it to simply `object`

src/components/List/ListItem/ListItem.jsx
sstone2423: Updated with i18n props

PR #885, opened 2020-02-06
Gauge card

davidicus:
Closes ##849

**Summary**

Implementation of Gauge card component 

- create GaugeCard.jsx, _gauge-card.scss GaugeCard.story.jsx, GaugeCard.test.jsx
- updated CardRender.jsx and Dashboard.story.jsx to include GaugeCard

- Go to GaugeCard story and go to Dashboard story and make sure it renders properly


src/components/GaugeCard/GaugeCard.jsx
```diff
@@ -0,0 +1,201 @@
...
+
+export const getColor = (gauge, value) => {
+  let { color } = gauge;
+  let grade = '';
+  const comparisons = {
+    '<': (a, b) => a < b,
+    '>': (a, b) => a > b,
+    '=': (a, b) => a === b,
+    '<=': (a, b) => a <= b,
+    '>=': (a, b) => a >= b,
```
stuckless: Imagine how much more powerful the comparisons could be if we were allowed to supply our own custom comparsion functions.  Doing ranges would be breeze.

src/components/GaugeCard/GaugeCard.jsx
```diff
@@ -0,0 +1,201 @@
...
+  };
+  if (gauge.thresholds) {
+    gauge.thresholds.forEach(thresh => {
```
stuckless: do we need to iterate all of them?  could we use something like `every()` instead?

src/components/GaugeCard/GaugeCard.jsx
```diff
@@ -0,0 +1,201 @@
...
+  values,
+  data,
+  isLoading,
+  hasMoreData,
+  size,
+  className,
+  ...others
+}) => {
+  const [loadedState, setLoadedState] = useState(false);
+  useEffect(() => {
```
stuckless: I'm confused by this.  What are we doing here?  Seems like we are setting our loaded state to true after 1 second??

src/components/GaugeCard/GaugeCard.jsx
```diff
@@ -0,0 +1,201 @@
...
+  isLoading: false,
+  description: null,
+  content: {
+    gauges: [
+      {
+        dataSourceId: null,
+        units: '%',
+        minimumValue: 0,
+        maximumValue: 100,
+        renderValueFunction: null,
```
stuckless: I like this... but I didn't see this used anywhere.

src/components/GaugeCard/GaugeCard.jsx
davidicus: This is to trigger the load animation of the gauge

src/components/GaugeCard/GaugeCard.jsx
davidicus: I wanted to get this simple implementation out to satisfy the current use case. But we have plans to extend this to accept a threshold function, different types of gauges, etc.

src/components/GaugeCard/GaugeCard.jsx
davidicus: I believe that would still need to iterate over the entire array, no? I think we need to run each comparison to be sure it doesn't trigger .

src/components/GaugeCard/GaugeCard.jsx
davidicus: I will address these changes and comments in our upcoming sprint. [Issue](https://github.com/IBM/carbon-addons-iot-react/issues/888) to track

src/components/GaugeCard/GaugeCard.jsx
davidicus: I will address these changes and comments in our upcoming sprint. [Issue](https://github.com/IBM/carbon-addons-iot-react/issues/888) to track

davidicus:
I will address these changes and comments in our upcoming sprint. [Issue](https://github.com/IBM/carbon-addons-iot-react/issues/888) to track

PR #891, opened 2020-02-07
feat(imagecard): support small size

scottdickerson:
Closes #

**Summary**

- Just add SMALL as a supported size for an Image Card

- Test out the SMALL card size in the Image Card story

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,14 +65,13 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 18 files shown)


CynZhang:
If we can manage to support 2 types of image card, one being a static image with no hotspot allowed, and the other could be interactive. Then we could limit the small size to only be used for static image.

PR #803, opened 2020-01-02
6277 dashboard grid is editable tile jump

sunilkotiyaibm:
Closes #780

**Summary**

- Dashboard Grid (dashboard,isEditable) : Tiles on dashboard grid is jump (animated) while rending the page. 

**Change List (commits, features, bugs, etc)**

src/components/Dashboard/DashboardGrid.jsx
src/components/Dashboard/__snapshots__/DashboardGrid.story.storyshot

**Acceptance Test (how to verify the PR)**

- Click on Dashboard (dashboard,isEditable) link, Dashboard tiles should be rendering static without animation

PR #894, opened 2020-02-10
fix(dashboardgrid): 6277-Dashboard-Grid-Tile-Jump

### LINKED ISSUE #780: [DashboardGrid] isEditable causes tiles to jump on initial render
<sub>[Internal tracking issue #134](https://github.ibm.com/Watson-IoT/pal-tracking/issues/134)</sub>

### Describe the bug

On initial render, the tiles jump/move. They should be static from initial render.

resize rendering

### To Reproduce

1. [View the story in the storybook](https://carbon-addons-iot-react.com/?path=/story/watson-iot-dashboard-grid--dashboard-is-edtiable)
1. Flip between the different dashboard grid stories, you'll see the reflow only on the `isEditable` story

### Expected behavior

The cards are static on initial render, no moving. They move/flow only when the browser is resized to effect the layout of the cards.

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: `v2.18.0`

psdixit:
Closes #780 

**Summary**

Dashboard Grid (dashboard,isEditable) : Tiles on dashboard grid is jump (animated) while rending the page.

src/components/Dashboard/DashboardGrid.jsx
src/components/Dashboard/snapshots/DashboardGrid.story.storyshot

Click on Dashboard (dashboard,isEditable) link, Dashboard tiles should be rendering static without animation

### DIFF
--- a/src/components/Dashboard/DashboardGrid.jsx
+++ b/src/components/Dashboard/DashboardGrid.jsx
@@ -163,11 +163,14 @@ const DashboardGrid = ({
   );
 
   const [animationState, setAnimationState] = useState(false);
-  useEffect(() => {
-    setTimeout(() => {
-      setAnimationState(true);
-    }, 1000);
-  }, []);
+  useEffect(
+    () => {
+      requestAnimationFrame(() => {
+        setAnimationState(isEditable);
+      });
+    },
+    [isEditable]
+  );
 
   return (
     <div style={{ flex: 1 }}>
--- a/src/components/Dashboard/DashboardGrid.jsx
+++ b/src/components/Dashboard/DashboardGrid.jsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useCallback } from 'react';
+import React, { useMemo, useCallback, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Responsive, WidthProvider } from 'react-grid-layout';
 import styled from 'styled-components';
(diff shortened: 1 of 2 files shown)


psdixit:
Hi @davidicus @stuckless 

Please review the PR #894 related the defect #780 .


Thanks,
P S Dixit

tay1orjones:
Is this a duplicate of #803? They both have the same graphite issue number.

src/components/Dashboard/DashboardGrid.jsx
```diff
@@ -162,6 +162,13 @@ const DashboardGrid = ({
     [childrenArray]
   );
 
+  const [animationState, setAnimationState] = useState(false);
+  useEffect(() => {
+    setTimeout(() => {
```
tay1orjones: Is the timeout necessary here? useEffect is ran after the first render and after every update.

psdixit: > Is the timeout necessary here? useEffect is ran after the first render and after every update.

Hi @tay1orjones ,

The element animation depends on the property "shouldAnimate". It is initially set to false and animation should not be effect the element.
I'll change the value of "shouldAnimate" property to true and it should animate,
But it doesn’t as all of this happens really fast with setState , and browser decides to render changes on the end of result, 
That's why i used the settimeout, this way we can achive this animation effect.

tay1orjones: Yes, the timeout works because it prevents the batching of setState component updates. A better way to achieve this will be to use `requestAnimationFrame` instead of `setTimeout`.

Additionally, with this as-is, we have lost configurability of the animation via the `isEditable` prop.

I will push an update to this branch to make clear these two suggestions.

PR #899, opened 2020-02-11
fix(tableCard): add size validation

### LINKED ISSUE #755: [StatefullTable] and [TableCard] Pagination component icons mis-aligned.
### Describe the bug

I am using [StatefullTable] and [TableCard] components and both has mis-aligned icons in the pagination in Chrome and Safari browsers
image
image

I am using "carbon-addons-iot-react": "2.17.0"

JoelArmendariz:
Closes #755 

**Summary**

Added a size validation to not allow card sizes smaller than 'LARGE'.

Here is the 'TALL' size:
TableCardTall

Here is 'TALL' size after size validation:
TableCardSizeNotSupported

src/components/Card/CardToolbar.jsx
src/components/TableCard/TableCard.jsx

Test any size other than the supported 'LARGE' and 'XLARGE'

### DIFF
--- a/src/components/TableCard/TableCard.jsx
+++ b/src/components/TableCard/TableCard.jsx
@@ -16,6 +16,12 @@ import { generateTableSampleValues } from '../TimeSeriesCard/timeSeriesUtils';
 import { csvDownloadHandler } from '../../utils/componentUtilityFunctions';
 import CardToolbar from '../Card/CardToolbar';
 
+const ContentWrapper = styled.div`
+  height: 100%;
+  max-height: 100%;
+  padding: 0 16px 16px 16px;
+`;
+
 const StyledOverflowMenu = styled(OverflowMenu)`
   &&& {
     margin-left: 10px;
@@ -689,6 +695,12 @@ const TableCard = ({
     ? columnsToRender.find(item => item.priority === 1)
     : columnStartSortDefined;
 
+  const supportedSizes = [
+    CARD_SIZES.LARGE,
+    CARD_SIZES.XLARGE,
+  ];
+  const supportedSize = supportedSizes.includes(size);
+
   const cardToolbar = (
     <CardToolbar
       availableActions={{ expand: isExpandable, range: true }}
@@ -711,8 +723,9 @@ const TableCard = ({
       i18n={i18n}
       hideHeader
       {...others}
-    >
-      {({ height }) => {
+    > 
+      {supportedSize ?
+      ({ height }) => {
         const numberOfRowsPerPage = !isNil(height) ? Math.floor((height - 48 * 3) / 48) : 10;
         return (
           <StyledStatefulTable
@@ -771,7 +784,10 @@ const TableCard = ({
             i18n={i18n} // TODO: add Card defaultprops ?
           />
         );
-      }}
+      } :
+      <ContentWrapper>
+        <p>Size not supported.</p>
+      </ContentWrapper>}
     </Card>
   );
 };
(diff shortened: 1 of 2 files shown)


src/components/TableCard/TableCard.jsx
```diff
@@ -16,6 +16,12 @@ import { generateTableSampleValues } from '../TimeSeriesCard/timeSeriesUtils';
 import { csvDownloadHandler } from '../../utils/componentUtilityFunctions';
 import CardToolbar from '../Card/CardToolbar';
 
+const ContentWrapper = styled.div`
```
tay1orjones: It looks like `ImageCard` also uses a very similar component. Could we abstract this out into one shared component for both instead of duplicating here?

Also we're discouraging use of styled-components and asking that new styles be put within a sass partial for reusability and ease of in-application overrides

lukefirth:
I think if we take a step back and make a strategic judgement on this:

- Right now it's a possible problem for anyone.
- In future, the UI will be able to stop most users from reaching this issue.
- At that time, advanced users will still be able to use JSON and reach this state.
- When this happens, we can either give the user an imperfect result (terrible 1 column table) or a blank card to convey the message (this PR).
- If we deprecate like this, anyone (I don't think there is anyone) using it in tall would be broken.

My recommendation is that we approach this differently and instead validate all cards against their 'supported sizes'. In this case we should throw a validation warning, rather than failure validation that states which cards have a problem (eg: table card is not supported in tall, you may find imperfect results using it). 

This way the experiences now will set us up for how we'd want it to work in the future:
- UI users will not get the chance to go wrong (as the UI builder would only allow supported sizes to be chosen).
- JSON upload users will get a validation warning upon upload that tells them that this is not a supported size, so beware the results.
- Users with it already would get the same validation when they next enter the edit mode.
- If we do design support for more sizes in future, this allows us to quietly turn off the validation warning for that card at that size.


Thoughts?

davidicus:
> I think if we take a step back and make a strategic judgement on this:
> 
> * Right now it's a possible problem for anyone.
> * In future, the UI will be able to stop most users from reaching this issue.
> * At that time, advanced users will still be able to use JSON and reach this state.
> * When this happens, we can either give the user an imperfect result (terrible 1 column table) or a blank card to convey the message (this PR).
> * If we deprecate like this, anyone (I don't think there is anyone) using it in tall would be broken.
> 
> My recommendation is that we approach this differently and instead validate all cards against their 'supported sizes'. In this case we should throw a validation warning, rather than failure validation that states which cards have a problem (eg: table card is not supported in tall, you may find imperfect results using it).
> 
> This way the experiences now will set us up for how we'd want it to work in the future:
> 
> * UI users will not get the chance to go wrong (as the UI builder would only allow supported sizes to be chosen).
> * JSON upload users will get a validation warning upon upload that tells them that this is not a supported size, so beware the results. This is the mechanism they already get feedback about their JSON upload, so would be an expected channel.
> * Users with it already would get the same validation when they next enter the edit mode.
> * If we do design support for more sizes in future, this allows us to quietly turn off the validation warning for that card at that size.
> 
> Thoughts?

I agree with @lukefirth sentiment here.

scottdickerson:
@JoelArmendariz so I think a prop-type validation rather than an actual UI Change at the limited sizes would do the trick here.

JoelArmendariz:
@davidicus @tay1orjones @scottdickerson I have opened a new PR to resolve this issue:
https://github.com/IBM/carbon-addons-iot-react/pull/970

davidicus:
closing this PR in favor of #970

PR #902, opened 2020-02-12
feat(table): toolbar only renders with children

### LINKED ISSUE #729: Stateful table: Any settings to disable data table toolbar section?
### Summary

As I don't have any toolbar items for the stateful table component that I am using, so I don't want the table toolbar section to be shown. Please suggest if there is any porp to be set to get rid of it. Below image has the highlighted section which needs to be removed, and this screen capture is from the storybook. https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--simple-stateful-example

image

### LINKED ISSUE #900: [Table] toolbar should not appear if there are no items in the toolbar
### Summary

If there is nothing being rendered in TableToolbar, it should be hidden.

**Additional context**
Example:
Screen Shot 2020-02-11 at 3 29 34 PM

### Available extra resources

Will fix myself

sstone2423:
Closes #900 
Closes #729

**Summary**

- TableToolbar only renders if there are toolbar options, secondaryTitle, or tooltip

- Added conditional rendering to `Table` for the `TableToolbar`
- Removed the secondary title story and added the `secondaryTitle` knob to all `StatefulTable` stories to help reduce the number of stories

- Check the first 'Simple Stateful Table' story to see that the toolbar is hidden (that story doesn't have any toolbar options)


src/components/Table/Table.jsx
```diff
@@ -352,57 +352,65 @@ const Table = props => {
...
-            'customToolbarContent',
-            'isDisabled'
-          ),
-        }}
-      />
+      {/* If there is no items being rendered in the toolbar, don't render the toolbar */
+      options.hasFilter ||
+      options.hasSearch ||
+      options.hasRowActions ||
+      options.hasRowCountInHeader ||
```
scottdickerson: I think you also need to check the onDownloadCSV action

scottdickerson: and the hasColumnSelection option

sstone2423:
> I think we need a few more checks on the conditional

Added the 2 conditionals

src/components/Table/Table.jsx
sstone2423: Added

src/components/Table/Table.jsx
```diff
@@ -357,6 +357,8 @@ const Table = props => {
       options.hasSearch ||
       options.hasRowActions ||
       options.hasRowCountInHeader ||
+      options.hasColumnSelection ||
+      actions.toolbar.onDownloadCSV ||
```
scottdickerson: do we need to conditionally check actions and toolbar for null?  Can actions or toolbar ever be null?
actions?.toolbar?.onDownloadCSV

sstone2423: Good catch. `options` can also be null

PR #904, opened 2020-02-13
[HierarchyList] Test coverage, conditional rendering, and defaultSelected expansion

### LINKED ISSUE #892: [HierarchyList] If I start the list with a child element selected, the parent is collapsed
### Describe the bug

Create the hierarchy list with defaultSelectedId.  The parent is not expanded by default and if the element is offscreen in a scroll pane, it's not visible.  We should probably scroll it into view if it's offscreen (even if it's in it's own pane)

sstone2423:
Closes #892 

**Summary**

- Coverage was failing as it was under 80%
- The `secondaryValue` div was still being rendered, even when it was null
- The `defaultSelectId`'s parent elements are now expanded by default
- List scrolls to the `defaultSelectedId`

- Additional tests added to rise to >80%
- Conditional rendering added to `ListItem`. If `secondaryValue` is null, it will not render, freeing up the other 50% of the width
- Additional functions for recursively searching were added to find items by `id`
- New story added to show `defaultSelectedId` behavior
- `toBeTruthy()` functions renamed to `toBeInTheDocument()` where applicable to give a better idea of what I am actually testing for
- Ref added to `defaultSelectedId`'s DOM element
  - Custom hook magic to determine if the node is set before trying to scroll

- See the 'with defaultSelectedId' HierarchyList story and see that an item is pre-selected, its parent is expanded, and the element is scrolled into view

### DIFF
--- a/src/components/List/HierarchyList/HierarchyList.jsx
+++ b/src/components/List/HierarchyList/HierarchyList.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import cloneDeep from 'lodash/cloneDeep';
 import debounce from 'lodash/debounce';
 
+import { caseInsensitiveSearch } from '../../../utils/componentUtilityFunctions';
 import List from '../List';
 
 const propTypes = {
@@ -53,32 +54,6 @@ const defaultProps = {
 };
 
 /**
- * Searches an item for a specific value
- * @param {Object} item to be searched
- * @param {string} searchTerm
- * @returns {Boolean} found or not
- */
-export const searchForItemValue = (item, searchTerm) => {
-  // Check that the value is not empty
-  if (item.content.value !== '' && item.content.value !== undefined) {
-    if (item.content.value.toLowerCase().search(searchTerm.toLowerCase()) !== -1) {
-      return true;
-    }
-    // Check that the secondary value is not empty
-    if (
-      item.content.secondaryValue !== '' &&
-      item.content.secondaryValue !== undefined &&
-      // Check if the value or secondary value has a match
-      item.content.secondaryValue.toLowerCase().search(searchTerm.toLowerCase()) !== -1
-    ) {
-      return true;
-    }
-    return false;
-  }
-  return false;
-};
-
-/**
  * Assumes that the first level of items is not searchable and is only uses to categorize the
  * items. Because of that, only search through the children. If a child is found while filtering,
  * it needs to be returned to the filtered array. Deep clone is required because spread syntax is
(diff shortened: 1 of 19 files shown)


src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -87,25 +87,98 @@ export const searchItem = (item, searchTerm) => {
...
+
+/**
+ * The ref should only be set if the item is selected by default. By using a callback,
+ * we can check if a node is properly attached to the ref, then scroll to it within the list
+ */
+const useHookWithRefCallback = () => {
+  const ref = useRef(null);
+  const setRef = useCallback(
+    node => {
+      if (node) {
```
scottdickerson: So functionally this is really scrollIntoViewAndSetRef?

scottdickerson: do you really need the ref at all at this point?

src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -87,25 +87,98 @@ export const searchItem = (item, searchTerm) => {
...
+/**
+ * Searches an item for a specific id
+ * @param {Object} item to be searched
+ * @param {string} searchId
+ * @returns {Boolean} found or not
+ */
+export const searchForItemId = (item, searchId) => {
+  if (
+    // Check if the value or secondary value has a match
+    item.id.toLowerCase().search(searchId.toLowerCase()) !== -1
```
scottdickerson: Can't this function and searchForItemValue be refactored into one to take a list of content keys to search for so the same codebase can be shared?

src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -87,25 +87,98 @@ export const searchItem = (item, searchTerm) => {
...
+    node => {
+      if (node) {
+        // Check if a node is actually passed. Otherwise node would be null.
+        // node is the text of the item
+        // parentNode is the container div with the button role
+        // parentNode.parentNode is the list content, which is also the element to be scrolled
+        // the offsetHeight needs to be multiplied by 3 to be able to view the whole element
+        const offset =
+          node.offsetTop -
+          node.parentNode.offsetHeight -
```
scottdickerson: need to optionally null check these keys

src/components/List/HierarchyList/HierarchyList.jsx
sstone2423: I renamed the function to a more suited name.

The ref is still needed as its the pointer to the DOM node that needs to be scrolled to

src/components/List/HierarchyList/HierarchyList.jsx
sstone2423: Fixed

src/components/List/HierarchyList/HierarchyList.jsx
sstone2423: Refactored and made a general util search function

PR #906, opened 2020-02-13
test(table): add story for sticky header and tooltip positioning

### LINKED ISSUE #818: [Stateful Table] Using Sticky header prop in stateful table component brings issue with tooltip when redered as custom cell.
### Describe the bug

I am having Tooltip as custom render in the Stateful table cell. The table has stickyHeader prop set to true. Tooltip is clicked the tooltip text remains at the same position when table content is scrolled.

image

upcoming release?
Looking for a fix for Jan2020 release.

enricoberti:
Closes #818

**Summary**

The positioning calculation of anything that uses in Carbon (Tooltip, Dropdown, etc) has some serious limitations (and it's [known and on the roadmap](https://github.com/carbon-design-system/carbon/issues/2907)).
This is especially true when needed inside a table with sticky header.
As a workaround, one can use `menuOffset` as explained [here](https://github.com/carbon-design-system/carbon/issues/4755#issuecomment-558131945)

A new story has been added: ?path=/story/watson-iot-table--with-sticky-header-and-cell-tooltip-calculation

### DIFF
--- a/src/components/Table/Table.story.jsx
+++ b/src/components/Table/Table.story.jsx
@@ -8,6 +8,7 @@ import Arrow from '@carbon/icons-react/lib/arrow--right/20';
 import Add from '@carbon/icons-react/lib/add/20';
 import Delete from '@carbon/icons-react/lib/delete/20';
 import { Add20 } from '@carbon/icons-react';
+import { Tooltip } from 'carbon-components-react';
 
 import { getSortedData, csvDownloadHandler } from '../../utils/componentUtilityFunctions';
 import FullWidthWrapper from '../../internal/FullWidthWrapper';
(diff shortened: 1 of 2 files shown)


davidicus:
@enricoberti can we fix existing stories that have drop downs like https://deploy-preview-906--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-table--stateful-example-with-expansion ?

enricoberti:
let me have a look if we can apply the same there @davidicus

enricoberti:
@davidicus so, i've done a bit of experimenting on this and the answer is 'sort-of' :/

We can set the `data-floating-menu-container` property to the RowActionsCell so the menu is attached there and then play with `menuOffsetFlip` of the StyledOverflowMenu but the calculation of the offset BUT I haven't found a way that consistently positions it perfectly right, not to talk about magic correction numbers. And again, if any of the parents has `position: fixed`, you can throw away the offset calculation as a whole.

My suggestion is not to implement it ourselves at component level, but expose though the `menuOffsetFlip` prop up to the Table component and apply it on RowActionsCell as needed. If you agree, let's create another issue to track that.

stuckless:
@davidicus @enricoberti A component that requires that the app developer have some intrinsic knowlege about whether or not some parent somewhere used a position:relative, doesn't offer any value to me :(  This is just a customer issue magnet.

Imagine if the HTML spec said that if the `title` attribute on an element was somehow affected by changes to some parent's change in position type.  It would be a disaster and people would just refuse to use it.  Or imagine if the select dropdown behaved the same way.

It's OK to ask 1 app developer building 1 application to spend 3-4 hours making sure their tooltips are rendering correctly, but, we can't ask 100s of app developers writing 100s of applications to do that work.  Sure, the effort for us is greater if we can automagically do this, but, otherwise we are just pushing our defects down to the app developers, and in turn will likley just come back to us via customers.

Personally though, what I'd reather see when it comes to Tooltips and the problematic way of how carbon does with these "popup" kind of components is simply to use some sort of TooltipFactory pattern we we delegate the rendering and management of tooltips to "soemthing else".  In graphite we tend to do this now, since, the tooltip needs to be managed.  When it renders it needs to reposition itself not only relative to a component, but, also so that it's not flowing off the screen.  Sometimes a tooltip pops up, sometimes left, depending on where it is trying to be positioned.  So allowing something else to control the rendering, means that an app developer is not writing the same code over and over and over as they are trying to position a tooltip.  This code is not trvial and we certainly don't want 10 different app developers wirting 10 different versions of the positioning code.

tay1orjones:
@enricoberti I think snapshots may need to be updated here? Status checks are failing

enricoberti:
@tay1orjones @stuckless beside the snapshots that I just fixed, I'm actually trying to find a way to create a nice "positioner" for all kind of situations: with or without a position:fixed parent up the tree, inside or outside a table, in a scrolling div etc

That said though, I don't know how big the incidence of devs having the position:fixed situation (and having to write their offset calculation) is gonna be. Imho positioning as is or just by using the `[data-floating-menu-container]`to tell Carbon where to append the menu, should cover 99% of the cases out there.

enricoberti:
@sstone2423 fixing it on all the stories that use any dropdown/tooltip would require to disable the centered plugin on them.
My stand on this is to just disable the story centering plugin once and for all so everything would work ootb.

sstone2423:
> @sstone2423 fixing it on all the stories that use any dropdown/tooltip would require to disable the centered plugin on them.
> My stand on this is to just disable the story centering plugin once and for all so everything would work ootb.

Good point. Maybe we can open a separate issue for that?

tay1orjones:
I opened #962 for the work to remove the centering addon

PR #908, opened 2020-02-13
889 value card error empty

### LINKED ISSUE #889: [ValueCard] Error & empty state for cards

We need a pattern to represent, in a card (or other places), that there is a data collection error state.  In Health, the scheduled data collection task can error for various reasons.  We want an indicator that allows to user to click to see more information about the source of the error.

Need a state that allows the user to see that there was an error in the data.  This is not the same as missing data. 

Not clear what the difference is between Null data and No data, so providing a single "no data" state for now.

**Value card**
- No data state
- Error state

image

- Double dash where result would have been.
- No data **icon** is displayed, left-aligned.
- No data **label** is displayed to the right of the icon. It could vary based on the nature of the no data problem. 
- No data **description** is displayed in the card if there's room.  
- Icon, label, and description are provided by app.
- User can click / hover on icon or message to see a tooltip with additional information.  Tooltip contains anything not displayed in the card and any additional content provided by the application. (Question: should it include the icon, label and description regardless of what's show in the card?)
- User can click Learn more to get more information about the problem, if available.

Content adaptations for small cards
- If it's an extra small card (or too small to show the full label), label does not display (or should it just be truncated?)
- If it's an extra small card, the description only appears in the tooltip.

**Table card**
TBD

**Timeline card**
TBD

**Donut card**
TBD

Being worked on in APM https://github.ibm.com/asset-performance/apm-design/issues/438

@KarenGski 
> We did some review with Luke in [asset-performance/apm-design#438](https://github.ibm.com/asset-performance/apm-design/issues/438) and settled on the following design for value cards.
> 
> image
> 
> Open questions:
> 
> * Should the tooltip always contain all information? Or only the things that don't fit in the card?
> * If the label doesn't fit, should it truncate or disappear?
> 
> Charts will just follow Carbon charting standards (I think).
> 
> @lukefirth can you share your thoughts on the open questions and then confirm if good to go?

@lukefirth 
> Open questions:
> 
> **Should the tooltip always contain all information? Or only the things that don't fit in the card?**
> I'd relent to ID on this, but my recommendation would be just the info not shown. (Although I'm easy on this because the tooltip the info when open).
> 
> **If the label doesn't fit, should it truncate or disappear?**
> Labels as in 'No data available' and 'data error' and/or their text below?
> I think @msperli s proposal works with just the icon and the tooltip suffices.
> 
> Good to go from my perspective.

@KarenGski 
> @lukefirth Claudia is working on the content question of what goes in/out of the tooltip. We'll go with whatever the official Content recommendation is.
> Just the icon works for me.
> @msperli We're good to go. You can close the APM issue for this.
> @smithjor FYI
> @rdconner This is approved so I think we can move out of design and into dev backlog.

bjornalm:
Closes #889

**Summary**
Error & empty state for cards
Added a new prop to the ValueCard to show states for no-data and error.

1. Supports icons of all sizes and always aligns with the label
2. Supports overflow ellipsis for label and multiline message
3. Added the "learn more" text in addition to the "learn more URL" since that is how the RowErrorAction was implemented
4. Tooltip is always available, even if all information was displayed in the card (implementation tradeoff)
5. Default icons supplied by the component if not supplied by the app
6. Supports tooltip activation by clicking on either icon, label or description. This was deemed needed since the label and description can both be shortened with ellipsis. Because of this the component is using 3 separate tooltip instances. I initially tried an "uncontrolled" tooltip to limit it to 1 instance, but that would mean re-implementing a lot of control functionality (aria, click outside to close, esc to close etc).

There are 4 data states stories added to the ValueCard with knobs for all props, verify that they work according to spec.


src/components/ValueCard/ValueCard.test.jsx
```diff
@@ -37,4 +40,31 @@ describe('ValueCard', () => {
...
+      />
+    );
+    expect(wrapperWithoutDataState.find(`.${iotPrefix}--data-state-container`)).toHaveLength(0);
+
+    const wrapperWithDataState = mount(
+      <ValueCard
+        title="Health score"
+        content={{ attributes: [{ label: 'title', dataSourceId: 'v' }] }}
+        size={CARD_SIZES.SMALL}
+        dataState={{
```
stuckless: does this work for multi-value cards?

src/components/ValueCard/DataStateRenderer.jsx
```diff
@@ -0,0 +1,107 @@
...
+  VALUE_CARD_DATA_STATE,
+} from '../../constants/LayoutConstants';
+
+const { iotPrefix, prefix: carbonPrefix } = settings;
+const dsPrefix = `${iotPrefix}--data-state`;
+
+// Testing the tooltip content outside carbon proved difficult.
+// So this is exposed for testing purpose.
+export const TooltipContent = ({ tooltipContent }) => {
+  const { label, description, extraTooltipText, learnMoreURL, learnMoreText } = tooltipContent;
```
stuckless: learnMoreURL doesn't seem to have much value in a SPA.  This seems fine if you are building a website, but, I'd much rather see an onClick handler for the button/link so that an app can register what they need to do when the user does a "learn more" action.   ie, show a dialog with more information?  Jump to a page in the app?

src/components/ValueCard/ValueCard.test.jsx
bjornalm: Yes, if you add the dataState property it will be activated regardless of how many values you added. It overrides all other rendering.

src/components/ValueCard/DataStateRenderer.jsx
bjornalm: Yes, that makes sense. I will change it.  I chose this implementation since that is how the conceptually related learnMoreText from the RowActionError was implemented. Perhaps we should add an onClick to that one as well, so that we stay consistent?

davidicus: Shouldn't this just be defined as content by the user. So they can pass in whatever they want? @stuckless Would that be preferable?

bjornalm: Now it is possible to pass a `readMoreElement` which can be a link or button with onClick or whatever is needed.

davidicus:
I notice that the tooltip gets scrolled up away from its target. 
image

bjornalm:
> I notice that the tooltip gets scrolled up away from its target.
> image

I think that is just how the tooltip behaves and is independent of this feature. See screenshot below:
image

src/components/ValueCard/DataStateRenderer.jsx
stuckless: @davidicus Yeah, I think that using a node/element here is better, since we can wire that up however we need/want.

stuckless:
> > I notice that the tooltip gets scrolled up away from its target.
> > image
> 
> I think that is just how the tooltip behaves and is independent of this feature. See screenshot below:
> image

That doesn't seem to be right.   The tooltip should not be anchoring away from my content.  It maybe that tooltip should open upwards instead of downwards, etc, but not like this :(  This is where it's almost preferable if the tooltip management can be offloaded, since, our tooltips respect the screen area and will open differently depending on where they are triggered.

src/components/ValueCard/ValueCard.test.jsx
stuckless: Yeah, but the dataState property is set as a single object for the entire Card.  So what happens if I can load 2 out 3 pieces of data?  How each each value rendering get the error state, since, 2 will be good and show a value and the third might have an error?  Or in the case of a value card is it's an error state for all values even if 1 fails.  ie, it's the card that has an error state and shows and error vs each value in a multi-value card.

bjornalm:
> > > I notice that the tooltip gets scrolled up away from its target.
> > > image
> > 
> > 
> > I think that is just how the tooltip behaves and is independent of this feature. See screenshot below:
> > image
> 
> That doesn't seem to be right. The tooltip should not be anchoring away from my content. It maybe that tooltip should open upwards instead of downwards, etc, but not like this :( This is where it's almost preferable if the tooltip management can be offloaded, since, our tooltips respect the screen area and will open differently depending on where they are triggered.

I've implemented a fix for this using `data-floating-menu-container` and the `menuOffset` function so now the tooltip scrolls with the card. You can try it here
https://deploy-preview-908--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-valuecard--data-state-no-data-small-scroll-page

But like you say, it doesn't position itself in a smart way.

src/constants/PropTypes.js
```diff
@@ -2,7 +2,15 @@ import PropTypes from 'prop-types';
...
 
-import { CARD_SIZES, CARD_LAYOUTS, DASHBOARD_SIZES, TIME_SERIES_TYPES } from './LayoutConstants';
+import {
+  CARD_SIZES,
+  CARD_LAYOUTS,
+  DASHBOARD_SIZES,
+  TIME_SERIES_TYPES,
+  VALUE_CARD_DATA_STATE,
+} from './LayoutConstants';
+// import { VALUE_CARD_DATA_STATE } from '../components/ValueCard/ValueCard';
```
sstone2423: Looks like these need to be removed?

src/components/ValueCard/ValueCard.test.jsx
bjornalm: It is implemented as as error-state (or empty-state) for the ValueCard itself, not for specific values. It will be up to the app to decide how to handle the case of 1/3 items error, if you want to show the 2 items that didn't error you cannot use the dataState in its current implementation.  Error state per value would become much more complex, it is doable, but not how I interpreted the spec. @davidicus What is your take on this?

src/constants/PropTypes.js
bjornalm: Fixed

PR #910, opened 2020-02-14
fix(simple-pagination): de-activate nav buttons at first and last page

### LINKED ISSUE #866: [SimplePagination] prevPage and nextPage buttons are not disabling
### Describe the bug

When the component is on Page 1, the prevPage button is still enabled, giving the indication that you could click backwards.

When on the last page, the nextPage is still enabled, giving the indication that you could keep going to the next page.

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-simplepagination--default
2. Click on the previous page button
3. See that its enabled

### Expected behavior

When on Page 1, prevPage should be disabled. When on last page, nextPage should be disabled

### Environment/versions:

- OS: MacOS
- Browser: Chrome
- `carbon-addons-iot-react` version: [e.g. `v2.7.1`]

### Additional context

First page: 

Screen Shot 2020-01-30 at 1 42 15 PM

Second page:

Screen Shot 2020-01-30 at 1 42 22 PM

JoelArmendariz:
Closes #866 

**Summary**

De-activates the navigate-left button on page 1 and the navigate-right button on the last page.

src/components/SimplePagination/SimplePagination.jsx
src/components/SimplePagination/_simple-pagination.scss

On the SimplePagination component from page one try to navigate farther left, and on the last page try to navigate farther right. The respective button should be de-activated.
PaginationNavFix


davidicus:
Hi @JoelArmendariz. Can we match the styles we see in the stories with the regular pagination? https://deploy-preview-910--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-table--stateful-example-with-expansion

JoelArmendariz:
@davidicus thanks for pointing out the different styles. I have updated the SimplePagination component to mirror those from Pagination.

src/components/SimplePagination/_simple-pagination.scss
```diff
@@ -19,11 +19,23 @@
...
+  }
+  cursor: not-allowed;
 }
 
 .#{$iot-prefix}-simple-pagination-caret {
   fill: $icon-01;
 }
+
+.#{$iot-prefix}-simple-pagination-caret-disabled {
+  fill: #c6c6c6;
```
davidicus: Should be a color token? `$disabled-02`

PR #919, opened 2020-02-17
feat(modal): support hiding primary button prop

### LINKED ISSUE #918: [ComposedModal] Custom footer story is missing custom footer
### Describe the bug

This is because passiveModal is set to true by default.  To be honest, I'm not sure that we need a passiveModal prop at all.  If no onSubmit is passed and no footer is passed to the component, then no footer will be rendered anyway.

### To Reproduce

https://carbon-addons-iot-react.com/?path=/story/watson-iot-composedmodal--custom-footer

image

Should look like this:
image

scottdickerson:
Closes #918

**Summary**

- feat(Modal): only show the primary button if new prop isPrimaryButtonHidden isn't true

- fix(passiveModal): removed the passiveModal prop as the same thing can be accomplished by passing a null footer and lack of an onSubmit action.  It also caused bug 918 above and was breaking us downstream.
- test(ComposedModal.story): remove the extraneous translation story and update the i18n story to show both translatable strings

- Verify all the ComposedModal stories


src/components/ComposedModal/ComposedModal.jsx
```diff
@@ -131,8 +133,6 @@ export const ComposedModalPropTypes = {
   invalid: PropTypes.bool, // eslint-disable-line react/boolean-prop-naming
   /** Callback to submit the dialog/form */
   onSubmit: PropTypes.func,
-  /** switchc to render footer buttons or not (carbon consistency) */
-  passiveModal: PropTypes.bool,
```
davidicus: This will cause a breaking change. Is there any way we can deprecate this first, then remove in next major release?

davidicus: Otherwise this looks good to go.

scottdickerson: ok repushed with false rather than removing

PR #914, opened 2020-02-17
fix(card-sizes): update dashboard layout card sizes

JoelArmendariz:
Closes #

**Summary**

Update card sizes and breakpoints

**Change List (commits, features, bugs, etc)**

src/components/Card/Card.jsx
src/constants/LayoutConstants.js

**Acceptance Test (how to verify the PR)**

In the dashboard grid, see the cards are updated to newest layout and respond to breakpoints accordingly.

PR #920, opened 2020-02-17
fix(ci): reconfigure coveralls to not swallow error exit codes

tay1orjones:
Related to #904 , #898 

**Summary**

- In #898 the testing thresholds weren't being met, yet CI didn't catch this. #904 updated the thresholds to get the build passing locally. 
- I suspect that CI builds passed while local builds failed because the testing script ran by travis piped the text results of `jest` directly to `yarn run coveralls`. Due to this, error codes from `jest` were being swallowed and not causing build failures when coverage thresholds were not met.

- Modify jest config to output lcov coverage reporter file
- Modify travis script to pipe the lcov coverage reporter file to coveralls, instead of directly piping lcov text output from jest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
 
 script:
   - yarn lint
-  - travis_wait yarn test --ci --maxWorkers=2 --verbose --coverage --coverageReporters=text-lcov | yarn run coveralls # report coveralls status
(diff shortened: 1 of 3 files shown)


tay1orjones:
Some related discussion from others with the same problem on the coveralls tracker
https://github.com/nickmerwin/node-coveralls/issues/230
https://github.com/nickmerwin/node-coveralls/issues/126

tay1orjones:
Just pushed an updated snapshot file, #914 updated with latest master before merging which needed snapshots regenerated. Failing snapshots slipped through to `master` due to this same bug

PR #923, opened 2020-02-18
feat(tablerowaction): add unique id supporting testing automation

### LINKED ISSUE #912: [Table] Row Actions need a unique identifier to assist with automation testing
This is part bug, part enhancement.  But the table's row actions do not have a reliable way to target the row action button from an automation testing point of view.  I am seeing that that overflow actions have `tableid_rowid_actionid` ad the id and that is perfect.   Could we apply this pattern and add IDs or (data- attribute) to the buttons so that we can effectively do selenium automation testing on the UI and find buttons for the row actions.

I realize that we can use xpath but we are trying to create tests that are less brittle and not going to fail because someone made trivial dom changes to the components, etc.

@enricoberti If you are looking for an easy one that has tremendous value... this is it ;)

tay1orjones:
Closes #912

**Summary**

- Add ~`id`~ `data-testid` to row action button for consumers to target with testing automation
- Also improved the button's `key` uniqueness by adding `tableId`

- View the story, you should see the ID present in the DOM
- @stuckless I think this is the correct element you wanted the identifier on, could you confirm?


tay1orjones:
Updated to use data attribute instead of id

stuckless:
> LGTM  I will let @stuckless confirm Taylor's question before merging.

@davidicus this works for me.  I just want to be able to target the buttons in the row and this does that.  You can merge whenenver.

PR #926, opened 2020-02-18
feat(modal): add new isPrimaryButtonDisabled property

scottdickerson:
**Summary**

- Allows dialog users to disable the primary button with a new isPrimaryButtonDisabled property

- 

- Test out the primary button hidden story and see that the isPrimaryButtonDisabled can be set to disable the button


scottdickerson:
@tay1orjones nah the storyshots don't need to update because I didn't make a story that specifically started with isPrimaryButtonDisabled, I just make it an optional knob in the story

Issue #872, opened 2020-02-03
[Charts] Bar chart card for dashboards


davidicus:
**Description**
Bar Chart Card displays a bar chart inside a card in the dashboard.
Similar to other chart cards in iot-react: https://carbon-addons-iot-react.com/?path=/story/watson-iot-timeseriescard--medium-single-point-interval-hour

**Name of Pattern or Asset**
https://www.carbondesignsystem.com/data-visualization/basic-charts/#bar-chart

**Additional context**
https://carbon-design-system.github.io/carbon-charts/

Quote from @matsgothe 

> Bar chart cards elaborated in this issue
> [wiotp/monitoring-dashboard#382](https://github.ibm.com/wiotp/monitoring-dashboard/issues/382)
> 
> Bar Chart card requirements
> 
> * Based on Carbon Bar charts
> * Supporting Vertical and Horizontal bar orientation
> * Primarily supporting bars by time-series grain sum. For example, sum energy consumption by grain. The x-axis is a time axes.
> * Secondarily supporting bars by value count by time series time range. For example alert count by severity, Count by user, Count by Alert type. The x-axes is categories / values.
> * Stacked bar charts, for multi categories by % in one 100% stack
> * Side bar charts, for comparison cards. One bard for current time anchor, one for comparison time anchor.
> 
> Concepts - https://ibm.ent.box.com/notes/530770619155
> UX Design - https://ibm.box.com/s/b185xveyvx4defyr7x1xazfd7gqu8bpd
> JSON updates - https://ibm.box.com/s/bq0uqegb4pa2hn3z6w8nrkwz0kmm72oi

quote from @bryancboyd 
> A few screenshots from the UX design link above:
> 
> ![image](https://media.github.ibm.com/user/699/files/1c100c00-0d06-11ea-97f8-b5fb1f2b26f5)
> 
> ![image](https://media.github.ibm.com/user/699/files/14506780-0d06-11ea-9bd3-0430b9e3fc12)
> 
> ![image](https://media.github.ibm.com/user/699/files/229e8380-0d06-11ea-94ef-44ec45585518)

davidicus:
@scottdickerson I saw that in the box note for the JSON you have asked Mats to deprecate. Can you give some clarity for Enrico on how to proceed?

davidicus:
Closed by #936

PR #936, opened 2020-02-19
feat(bar-chart-card): first pass at barChart card

enricoberti:
#872

image

**Summary**

First pass at the BarChartCard component. Added a few stories for simple, grouped and stacked bar charts.

There is no data transformation from the original 'data' definition used by CarbonCharts.

?path=/story/watson-iot-experimental-barchartcard--simple-bar and consequent ones


src/components/BarChartCard/_bar-chart-card.scss
```diff
@@ -0,0 +1,29 @@
+@import '../../globals/vars';
+
+.#{$iot-prefix}--bar-chart-container {
+  padding-left: 16px;
```
davidicus: we need to use the carbon spacing tokens here. This i believe is `$spacing-05`. WE should consolidate all these padding styles as well. `padding: 0 $spacing-05 $spacing-05

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -0,0 +1,114 @@
...
+  },
+  size,
+  locale,
+  i18n,
+  isExpanded,
+  isLazyLoading,
+  className,
+  ...others
+}) => {
+  const isAllValuesEmpty = data
```
davidicus: is a tertiary necessary here? I am not too familiar with this lodash function but it looks to return true or false based off the check.

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -0,0 +1,114 @@
+import React, { Fragment } from 'react';
+import SimpleBarChart from '@carbon/charts-react/bar-chart-simple';
+import StackedBarChart from '@carbon/charts-react/bar-chart-stacked';
+import GroupedBarChart from '@carbon/charts-react/bar-chart-grouped';
+import '@carbon/charts/styles.css';
```
davidicus: do we not include this somewhere already? All styles should be brought into our main sass partial. @scottdickerson I see we are bringing this in TimeSeriesCard like this as well. Do they not offer a sass partial for each chart?

stuckless:
Check out the RTL tab, there are definate issues there.  

Does carbon charts support any way for us to format the data before they render it?  or is that what this statement is meaning and the answer is no?
> There is no data transformation from the original 'data' definition used by CarbonCharts.

enricoberti:
@stuckless @davidicus regarding RTL, it's a common problem for charting in general. I think we should keep it as a separate issue there so I can fix it in the barchart and the timeseries too

src/components/BarChartCard/_bar-chart-card.scss
enricoberti: thanks, fixed!

src/components/BarChartCard/BarChartCard.jsx
enricoberti: there are no partials for the single charts, but I moved this out and created a globals/_charts to import this + having common things for the charts, like RTL behavior

src/components/BarChartCard/BarChartCard.jsx
enricoberti: true, simplified, thanks!

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -0,0 +1,113 @@
+import React, { Fragment } from 'react';
+import SimpleBarChart from '@carbon/charts-react/bar-chart-simple';
+import StackedBarChart from '@carbon/charts-react/bar-chart-stacked';
+import GroupedBarChart from '@carbon/charts-react/bar-chart-grouped';
+import classnames from 'classnames';
+import every from 'lodash/every';
```
sstone2423: This function is in base JS https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

src/utils/sample.jsx
```diff
@@ -2626,3 +2626,205 @@ export const tableData = [
     },
   },
 ];
+
+export const barChartData = {
```
sstone2423: It would be nice to get this file split up into different categories / sections. 2630 lines of sample data is difficult to navigate if someone is looking for examples. Maybe all of the "chartable data" in their own file?

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -0,0 +1,113 @@
...
+    ChartComponent = StackedBarChart;
+  }
+
+  const scaleType = isTimeSeries ? 'time' : 'labels';
+
+  return (
+    <Card
+      title={title}
+      className={`${iotPrefix}--bar-chart-card`}
+      size={size}
```
sstone2423: The `XSMALL` size is essentially useless. Should a warning be added here to not use it?

src/components/BarChartCard/BarChartCard.story.jsx
```diff
@@ -0,0 +1,209 @@
...
+  BAR_CHART_ORIENTATION,
+} from '../../constants/LayoutConstants';
+import { getCardMinSize } from '../../utils/componentUtilityFunctions';
+import { barChartData } from '../../utils/sample';
+
+import BarChartCard from './BarChartCard';
+
+storiesOf('Watson IoT Experimental|BarChartCard', module)
+  .add('Simple bar', () => {
+    const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.WIDE);
```
sstone2423: The `XSMALL` size is essentially useless as its too small to see any of the data. Should a warning be added here to not use it? Or maybe not show it as an option at all?

scottdickerson: I'd vote for a prop-types validation on the XSMALL size...  Also needs to be updated to match the newer card sizes that @JoelArmendariz just delivered to dev

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -0,0 +1,113 @@
...
+const { iotPrefix } = settings;
+
+const BarChartCard = ({
+  title,
+  content: {
+    xLabel,
+    yLabel,
+    orientation = BAR_CHART_ORIENTATION.VERTICAL,
+    chartType = BAR_CHART_TYPES.SIMPLE,
+    data,
```
scottdickerson: I'd pass this in as a 'values' prop to be consistent with the other card types and then map it internally to the barchart card data.

scottdickerson: in addition, we should order the "groups" of data as a new prop of the content object.  I.e. separate the "groups" and the values into two different properties to be consistent with the other card types.

src/components/BarChartCard/BarChartCard.jsx
enricoberti: true that, force of habit. changed!

src/utils/sample.jsx
enricoberti: done

src/components/BarChartCard/BarChartCard.story.jsx
enricoberti: done

src/components/BarChartCard/BarChartCard.jsx
enricoberti: refactored to be compatible with the other cards

Issue #939, opened 2020-02-19
Dashboard within accordion rendering as a stack


davidicus:
copied from PAL tracking [issue](https://github.ibm.com/Watson-IoT/pal-tracking/issues/259)

Occuring in APM  here: http://mahi-docker09.swg.usma.ibm.com:3000/#/assetDetail?ids=1588

The dashboard is automatically adjusting itself on the screen when rendering. 
1: The dashboard inside of the Predictions accordion (and any other accordion) do not properly render when opened.  On screen re-size, they "snap" into the proper position.

![stacking](https://media.github.ibm.com/user/17018/files/aae87c80-4e4e-11ea-9ba0-d8a8e20dbc33)

2: The dashboard causes a horizontal scroll on the screen. When you re-size the browser, the horizontal scroll disappears. 
![horizontal scroll](https://media.github.ibm.com/user/17018/files/36640c80-4e53-11ea-814a-0b98d8a87fe0)

stuckless:
Here's a CRA that demonstrates the dashboard overflow issue.  In short this is caused because of page animations.  ie, we want to show a little bump when you navigate to the page.  The dashboard never updates itself based on available space so it overflows.  When you build this app, it starts on a main page, just click the dashboard on the side menu to show the dashboard view with an overflow.

[dashboard-overflow.zip](https://github.com/IBM/carbon-addons-iot-react/files/4232483/dashboard-overflow.zip)

davidicus:
@stuckless that is what I needed. 

![image](https://user-images.githubusercontent.com/1590966/74978686-2b811d00-53fb-11ea-8b73-17ec76598258.png)

This should be applied to the container and it allows DashboardGid to calculate properly. The only other way is to force trigger a render but then you would lose the animation in.

davidicus:
Closing issue since there is no component change. @stuckless please feel free to reopen if necessary.

PR #940, opened 2020-02-19
Accordion enhancement

### LINKED ISSUE #938: [AccordionItem] child progressive loading UX
Copied from PAL tracking [issue](https://github.ibm.com/Watson-IoT/pal-tracking/issues/257)

The Carbon accordion component currently renders all children, no matter if they are opened or closed. https://carbon-addons-iot-react.com/?path=/story/addcard--handles-click

After rendering, it then hides the rendered content. This is a large performance hit, in that content the user has not requested to see has been loaded. In APM, the accordion is used to display entire dashboards: https://ibm.invisionapp.com/share/APNZ6UW9Z5B#/319351669_Asset_Details_-_Navigation_Outline_-_Experimental5

In the above InVision, all accordions are open, but in the coded version, we will want to close these accordions on page-load in order to improve performance. See current implementation: http://mahi-docker09.swg.usma.ibm.com:3000/#/assetDetail?ids=1586

As more widgets are added to each section, performance will degrade without an update to the accordion. Progressive loading is generally a better UX for any use case of the accordion.

davidicus:
Closes https://github.ibm.com/Watson-IoT/pal-tracking/issues/257 and closes #938 and closes part of #939

**Summary**

Conditionally load children of AccordionItem when item is open

- created AccordionItem.jsx, AccordionItem.test.jsx, AccordionItemStory.jsx
- updated snapshots for welcomepage
- updated index.js to export our local AccordionItem instead of Carbon

- Please double check that this item is exporting by running a build and searching for our variation. You can test functionality by opening story and your dev tools to see that the content is only rendered when Item is open.


src/components/Accordion/AccordionItem.jsx
```diff
@@ -0,0 +1,79 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { AccordionItem as CarbonAccordionItem } from 'carbon-components-react';
+
+// same default used in carbon. adding to avoid prop type validation issues.
+const defaultRenderExpando = props => <button type="button" {...props} />;
```
sstone2423: Small typo in variable name `Expando`

src/components/Accordion/AccordionItem.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+const AccordionItem = ({ children, ...props }) => {
+  const [openState, setOpenState] = useState(false);
```
stuckless: will this get updated if the incomming open prop changes?  I can't really tell from the code, but I don't really see anything that would cause it to update... and should the default value be props.open instead of false?

src/components/Accordion/AccordionItem.jsx
davidicus: Hey @sstone2423. Is the typo `Expando`? I think this is intentional?

src/components/Accordion/AccordionItem.jsx
davidicus: yes. good catch

PR #941, opened 2020-02-20
fix(datatable): make th break-all conditional for resize

### LINKED ISSUE #930: [DataTable] header columns auto wrap by default past their text width on render
### Describe the bug
The alerts table in our dashboard renders with column headers that are wrapping to multiple lines by default.

image

### To Reproduce

1.Because the word-break is set to true (even if a column is not resizable), when tables are being sized to fit within smaller containers, it automatically starts wrapping the columns on first render.
2. We need to either conditionally turn off the word-break wrapping if resize isn't turned on, or we need to auto-size the header column to at least contain the text width by default to stop it from word-break wrapping by default

bjornalm:
Closes #930

**Summary**
fix to make the table headers only use `word-break: break-all` when the table is resizable 

tests here


bjornalm:
> we actually need to also make the td.data-table word-break:break-all conditional on resize as well to preserve existing table behavior when resize is off

Alright, will add that as well.

PR #942, opened 2020-02-20
fix(tablehead): column resizing bug when columns initially empty

### LINKED ISSUE #933: [Table] Column Resizing throws an error
### Describe the bug

Enable hasResize, then resize a column, it dies.

### To Reproduce

Enable hasResize, then resize a column, it dies.

Codesandbox is just showing an error on the button example
> https://codesandbox.io/s/codesandbox-yxbkg

### Expected behavior

Should resize the column and not crash

### Additional context

image

My yarn lock has the following...

image

I'm a little concerned here, since I thought that we were fixing the carbon component versions but, it appears that we will be getting diffrernent version of carbon components for different projects that use iot components at a fixed version.  I have no idea if that is causing this issue or not.

I've also been told that there is this PR in carbon and not sure if it's related to this issue or not.

https://github.com/carbon-design-system/carbon/pull/5084

tay1orjones:
Closes #933

**Summary**

- Fixes a bug where if the columns array was initially empty, and then subsequently updated with new data - I was able to write a unit test against this bug
- Also refactored a bit while I was in there

- combination of some null checks and adding appropriate dependencies to the dependency array of a few hooks
- refactor a couple functions to useCallback, since they have dependencies and can be memoized without affecting functionality
- conditionally apply mouse event handlers for only when `useResize=true`
- refactor an abstracted function that was only used in `useLayoutEffect` to be inline there

- Check out the `with resize and no initial columns` story

### DIFF
--- a/src/components/Table/TableHead/TableHead.jsx
+++ b/src/components/Table/TableHead/TableHead.jsx
@@ -1,6 +1,6 @@
 /* eslint-disable jsx-a11y/no-static-element-interactions */
 
-import React, { useState, useEffect, useLayoutEffect, createRef } from 'react';
+import React, { useState, useEffect, useLayoutEffect, createRef, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { DataTable, Checkbox } from 'carbon-components-react';
 import isNil from 'lodash/isNil';
@@ -169,24 +169,12 @@ const TableHead = ({
     });
   };
 
-  const getRenderedWidths = () => {
-    return columnRef.map(ref => ref.current && ref.current.getBoundingClientRect().width);
-  };
-
-  const getRenderedWidthsMap = () => {
-    const widths = getRenderedWidths();
-    const widthsMap = {};
-
-    ordering.forEach((orderedColumn, index) => {
-      widthsMap[orderedColumn.columnId] = {
-        width: widths[index],
-        index,
-        id: orderedColumn.columnId,
-        visible: !orderedColumn.isHidden,
-      };
-    });
-    return widthsMap;
-  };
+  const getRenderedWidths = useCallback(
+    () => {
+      return columnRef.map(ref => ref.current && ref.current.getBoundingClientRect().width);
+    },
+    [columnRef]
+  );
 
   const updateColumnWidthsAfterResize = modifiedColumnWidths => {
     setCurrentColumnWidths(prevColumnWidths => {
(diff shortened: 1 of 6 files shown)


src/components/Table/TableHead/TableHead.jsx
```diff
@@ -169,24 +169,12 @@ const TableHead = ({
     });
   };
 
-  const getRenderedWidths = () => {
```
bjornalm: Were you able to detect a performance improvement with the introduction of the `useCallback`? I'm not seeing where it has any effect in this scenario. This function is only called once on init and once per resize drop even without the `useCallback`. And I guess that when the table data changes, the `useLayoutEffect` won't be called unless any of the now explicit dependencies changes, meaning that `getRenderedWidths` won't be called unless needed. But perhaps I'm missing something?

src/components/Table/TableHead/TableHead.jsx
```diff
@@ -214,18 +202,31 @@ const TableHead = ({
 
   useLayoutEffect(
     () => {
-      if (hasResize) {
-        setCurrentColumnWidths(getRenderedWidthsMap());
+      if (hasResize && columns.length && isEmpty(currentColumnWidths)) {
```
bjornalm: Nice!

src/components/Table/TableHead/TableHead.test.jsx
```diff
@@ -120,4 +120,36 @@ describe('TableHead', () => {
     const tableHeaders = wrapper.find(TableHeader);
     expect(tableHeaders).toHaveLength(2);
   });
+
+  test('header renders with resizing columns when columns are empty on initial render', () => {
```
bjornalm: Great!

bjornalm:
Nice work! Thanks for catching this. I left a question in the code review.

src/components/Table/TableHead/TableHead.jsx
scottdickerson: if we don't use useCallback on getRenderedWidth, the useLayoutEffect will trigger every render since the function pointer will change for getRenderedWidth

bjornalm: @scottdickerson, but in practice if I count the calls to `useLayoutEffect` (using console.log) I see no difference with or without `useCallback` here. I tested a range of things like loading the component, resizing, sorting, row selection, loading more data etc. 

Sorry, just trying to understand. Please bear with me :-)

PR #946, opened 2020-02-21
fix(table): reducer should not reset the page if the data changes

scottdickerson:
Closes:
Internal bug
https://github.ibm.com/wiotp/monitoring-dashboard/issues/769

**Summary**

- the bug was that the table page would reset if the data changes.  This causes problems if the user has made a paging decision and chooses to refresh the table data
This PR prereqs the th header wrap word-break:
https://github.com/IBM/carbon-addons-iot-react/pull/941

- fix(tableReducer): use the page size from state, don't reset to the original every time the TABLE reregisters its data
- test(sidenav): added a page title bar story to the sidebar
- refactor(Table): remove the styled components from the base table and replace with scss as it was causing a strange failure to hide the pagination dropdown
- fix(TableCard): prevent whitespace warp on the table card because we need to dynamically size the height and assume a given row size to measure

- Verify the table card stories
- Verify the table stories


.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
 
 script:
   - yarn lint
-  - travis_wait yarn test --ci --maxWorkers=2 --verbose --coverage && cat ./coverage/lcov.info | yarn run coveralls && rm -rf ./coverage # report coveralls status
+  - travis_wait yarn test --ci --maxWorkers=2 --coverage && cat ./coverage/lcov.info | yarn run coveralls && rm -rf ./coverage # report coveralls status
```
tay1orjones: Is the `--verbose` flag causing issues?

src/components/Table/_pagination.scss
```diff
@@ -0,0 +1,21 @@
...
+.#{$iot-prefix}--pagination {
+  .bx--pagination__left {
+    margin: auto auto auto 0;
+  }
+  .bx--select .bx--select-input ~ .bx--select__arrow {
+    align-self: center;
+  }
+  .bx--pagination__left,
+  .bx--pagination__text {
+    display: var(--pagination-text-display);
```
tay1orjones: Where is this variable declared/set? I can't find a reference to it

tay1orjones:
Do we have design sign off that the pagination should auto-hide in this way based on the width of the component at specifically `500px`? cc @lukefirth 

It seems that this auto-hide logic should be included in the base pagination component, rather than a non-exported variant only used within the table. Doesn't this also effectively render SimplePagination obsolete?

lukefirth:
Agree with @tay1orjones here, it seems like what should happen here is that at lower sizes (where things won't fit anymore), the pagination should responsively swap to the simple paginator.

The one thing this does that simple paginator does not currently support is the page number dropdown. This could be added pretty simply however and I'm not against it in concept (either as a knob on simple paginator or just a flat update).

src/components/Table/_pagination.scss
davidicus: i believe this is set in `Pagination.jsx`.

tay1orjones: Ah, yes nevermind

.travis.yml
scottdickerson: yeah it was causing intermittent travis errors

scottdickerson: https://blog.m157q.tw/posts/2018/03/30/travis-ci-stdout-write-error-and-resource-temporarily-unavailable-workaround/

scottdickerson:
they kept the items per page just to show the record count

scottdickerson:
luke's gonna open a separate issue to unify the paginator between Table and SimplePagination

tay1orjones:
👍 If we hide the `Items per page` text when the dropdown is hidden, this should be good to go.

PR #948, opened 2020-02-21
fix(hierarchylist): changed scrollto to scrollintoview

### LINKED ISSUE #947: [HierarchyList] not scrolling to element and horizontal scroll bar
### Describe the bug

When `defaultSelectedId` is set, the selected id is not being properly scrolled to.

If an list item's text is wider than the list content width, a horizontal scroll bar appears

### To Reproduce

1. Go to local version of HierarchyList story
2. Set the defaultSelectedId on many different id's (the scrolling sometimes works)
3. See that the element is not properly scrolled to

1. Go to local version of HierarchyList story
2. Change one of the ListItem values to a very long string (longer than the width)
3. See that a horizontal scroll bar appears

### Environment/versions:

- OS: MacOS
- Browser: Chrome, Safari

sstone2423:
Closes #947 

**Summary**

- Fixed scrollTo `defaultSelectedId` bug and horizontal scrollbar appearing when listItem's are longer than the list content
- Fixed multiple expanded parents for a single `defaultSelectedId`
- Fixed search function when there is only 1 column of values (no `secondaryValue`)

- Updated search function to only search the `secondaryValue` if it is not nil
- Updated search for `defaultSelectedId` to only look for exact matches, not `includes()`
- Updated `scrollTo` function to use `scrollIntoView` as measurements were not working correctly
- Updated 1 `HierarchyList` story to have 1 very long player name, added 1 of the player names to another team (to test multiple id , for further testing of the above issues
- Added `overflow-x: hidden` to the list content container
- Added keys to a few stories to remove warnings

- Check both `HierarchyList` stories and ensure that searching works for both
- Check the `defaultSelectedId` to ensure the element is scrolled to properly
- Check that the horizontal scroll bar does not appear on either story


jessieyan:
Hey Stephen, I just pulled your branch. For the search bar, what is the expected behavior when you put space in it?  For now, one of the child item was selected. I think it might be the same problem in SimpleList too.

sstone2423:
> Hey Stephen, I just pulled your branch. For the search bar, what is the expected behavior when you put space in it? For now, one of the child item was selected. I think it might be the same problem in SimpleList too.

If its a single space, it will match any list item that contains a space, which would be all of the players

PR #953, opened 2020-02-25
fix(table): allow default values to be set

### LINKED ISSUE #429: [Table]: Search default value and clear filters not working as expected

enricoberti:
#429

Closes #429

**Summary**
While we wait for Carbon 10 to be released and referenced, this is a port of their fix: https://github.com/carbon-design-system/carbon/commit/e97a7ed9be2911707066c406544310f6cb76f273

The default value can now be set with `view.toolbar.search.defaultValue`

- Added a wrapper around TableToolbarSearch

- Added a new story on ?path=/story/watson-iot-table--with-pre-filled-search

### DIFF
--- a/src/components/Table/Table.test.jsx
+++ b/src/components/Table/Table.test.jsx
@@ -357,25 +357,9 @@ describe('Table', () => {
     expect(wrapper.find('.bx--search-input')).toHaveLength(1);
     expect(wrapper.find('.bx--search-input').prop('value')).toEqual('toyota');
 
-    const wrapper2 = mount(
-      <Table
-        columns={tableColumns}
-        data={tableData}
-        actions={mockActions}
-        options={{
-          hasSearch: true,
-        }}
-        view={{
-          toolbar: {
-            search: {
-              value: 'ferrari',
-            },
-          },
-        }}
-      />
-    );
+    wrapper.setProps({ view: { toolbar: { search: { defaultValue: 'ferrari' } } } });
+    wrapper.update();
 
-    expect(wrapper2.find('.bx--search-input')).toHaveLength(1);
-    expect(wrapper2.find('.bx--search-input').prop('value')).toEqual('ferrari');
+    expect(wrapper.find('.bx--search-input').prop('value')).toEqual('ferrari');
   });
 });
--- a/src/components/Table/Table.test.jsx
+++ b/src/components/Table/Table.test.jsx
@@ -334,4 +334,48 @@ describe('Table', () => {
     fireEvent.keyDown(filterButton, { keyCode: keyCodes.ENTER });
     expect(mockActions.toolbar.onToggleFilter).toHaveBeenCalledTimes(1);
   });
+
+  test('toolbar search should render with default value', () => {
+    const wrapper = mount(
+      <Table
+        columns={tableColumns}
+        data={tableData}
+        actions={mockActions}
+        options={{
+          hasSearch: true,
+        }}
+        view={{
+          toolbar: {
+            search: {
+              defaultValue: 'toyota',
+            },
+          },
+        }}
+      />
+    );
+
+    expect(wrapper.find('.bx--search-input')).toHaveLength(1);
+    expect(wrapper.find('.bx--search-input').prop('value')).toEqual('toyota');
+
+    const wrapper2 = mount(
+      <Table
+        columns={tableColumns}
+        data={tableData}
+        actions={mockActions}
+        options={{
+          hasSearch: true,
+        }}
+        view={{
+          toolbar: {
+            search: {
+              value: 'ferrari',
+            },
+          },
+        }}
+      />
+    );
+
+    expect(wrapper2.find('.bx--search-input')).toHaveLength(1);
+    expect(wrapper2.find('.bx--search-input').prop('value')).toEqual('ferrari');
+  });
 });
--- a/src/components/Table/TablePropTypes.js
+++ b/src/components/Table/TablePropTypes.js
@@ -211,5 +211,5 @@ export const defaultI18NPropTypes = {
 };
 
 export const TableSearchPropTypes = PropTypes.shape({
-  value: PropTypes.string,
+  defaultValue: PropTypes.string,
 });
--- a/src/components/Table/TablePropTypes.js
+++ b/src/components/Table/TablePropTypes.js
@@ -1,5 +1,7 @@
 import PropTypes from 'prop-types';
 
+import deprecate from '../../internal/deprecate';
+
 export const RowActionPropTypes = PropTypes.arrayOf(
   PropTypes.shape({
     /** Unique id of the action */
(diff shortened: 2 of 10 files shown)


src/components/Table/TableToolbarSearch/_table-toolbar-search.scss
```diff
@@ -0,0 +1,12 @@
+@import '~carbon-components/scss/globals/scss/vars';
```
tay1orjones: This is not needed, it is imported in the local global vars file that you import on the following line
```suggestion

```

src/components/Table/TablePropTypes.js
```diff
@@ -211,5 +211,5 @@ export const defaultI18NPropTypes = {
 };
 
 export const TableSearchPropTypes = PropTypes.shape({
-  value: PropTypes.string,
+  defaultValue: PropTypes.string,
```
tay1orjones: Renaming this prop would be a breaking change. Can we leave it as `value`?

src/components/Table/TableToolbarSearch/TableToolbarSearch.jsx
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  const onClick = e => {
+    setFocusTarget(toolbarSearch);
+    handleExpand(e, true);
+  };
+
+  const onChange = e => {
+    if (onChangeProp) {
+      onChangeProp(e);
```
tay1orjones: Is the event `e` not passed if you just directly use onChange from props, instead of renaming it `onChangeProp`?

src/components/Table/TableToolbarSearch/TableToolbarSearch.jsx
```diff
@@ -0,0 +1,101 @@
+/* istanbul ignore file */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
tay1orjones: We should fix the a11y violations, rather than ignore them.

src/components/Table/TableToolbarSearch/TableToolbarSearch.jsx
```diff
@@ -0,0 +1,101 @@
...
+import classnames from 'classnames';
+
+import { settings } from '../../../constants/Settings';
+
+const { iotPrefix } = settings;
+
+const { TableToolbarSearch: CarbonTableToolbarSearch } = DataTable;
+
+const propTypes = {
+  defaultValue: PropTypes.string,
```
sstone2423: Prop comments would be helpful

src/components/Table/TableToolbarSearch/TableToolbarSearch.jsx
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * ----------------------------- WARNING ----------------------------
+ * This is a temporary solution to fix a bug in Carbon 9,
+ * to be removed when we import Carbon 10
+ * ------------------------------------------------------------------
+ */
+const TableToolbarSearch = ({
+  defaultValue,
+  defaultExpanded,
+  onChange: onChangeProp,
```
sstone2423: Looks like this is unneeded as the only time `onChange` is used, is when passing it to the `onChange` event handler

src/components/Table/TablePropTypes.js
sstone2423: It looks Carbon is using it in their commited fix. Maybe it could be aliased as `value: defaultValue` ?

src/components/Table/TableToolbarSearch/TableToolbarSearch.jsx
```diff
@@ -0,0 +1,101 @@
...
+  defaultValue: '',
+  defaultExpanded: false,
+  onChange: null,
+  onExpand: null,
+};
+
+/**
+ * ----------------------------- WARNING ----------------------------
+ * This is a temporary solution to fix a bug in Carbon 9,
+ * to be removed when we import Carbon 10
```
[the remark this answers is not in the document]
enricoberti: Thanks for this @tay1orjones 
Speaking with @davidicus we cannot just blindly upgrade to 10.10.0 for the MVP downstream consumers so we need to do this fix ourselves

enricoberti:
@sstone2423 @tay1orjones  thanks for the reviews,  i've rebased on master and added a couple of tests too to see that the prop aliasing works correctly

src/components/Table/Table.test.jsx
```diff
@@ -334,4 +334,48 @@ describe('Table', () => {
...
+            },
+          },
+        }}
+      />
+    );
+
+    expect(wrapper.find('.bx--search-input')).toHaveLength(1);
+    expect(wrapper.find('.bx--search-input').prop('value')).toEqual('toyota');
+
+    const wrapper2 = mount(
```
stuckless: This test is basically the same as the one above it... ie, create a table and verify the value exists (which is a good test, but no need to do it twice).  A better test here for this second one would be use the first wrapper and update the props and then see if the that table instance now reflects the new default search value as set from changing the props.

The use case for how this used is as follows.
We use the table to render the state of a "datasource" which is basically some stateful component that holds things like our data, responds to searches, etc.  That table instance could dynamically load a new "view" which means that it might have a default search, etc, and when that happens we need to the UI to reflect the state of our datasource.  Today, everything about the table, including filters, columns, etc, all comes from our datasource, and it all works woo, with the exception that the "search" value cannot be set in the toolbar.  So we load a new view and show it, and the data is filtered, but, the UI doesn't show that search value.  In essence we have a single table and we just keep updating the props on that table to reflect the state of the data from the datasource.

enricoberti: Gotcha! The duplication was just to make sure aliasing works but, given `value` is being deprecated, I removed it and put something like this in place: https://github.com/IBM/carbon-addons-iot-react/pull/953/commits/f16851c5bbdabe46d120327146f41c11308e5871#diff-f47894e32d4dbf684d2ac3f89df148c9R360

PR #955, opened 2020-02-25
Breadcrumb knobs

### LINKED ISSUE #924: IoT Breadcrumb Overflow implementation has issues with Knobs for BreadcrumbItems
### Describe the bug

Using knobs in graphite storybook for Breadcrumb (specifically for breadcrumb items) will not be updated when the "hasOverFlow" is set. In my IoT development environment, I implemented similar knobs and has the same issues. After a quick troubleshooting in the code, it seems that the overflow items does not get updated with the latest breadcrumb item attribute changes.

### To Reproduce

1. Go to https://pages.github.ibm.com/maximo-app-framework/graphite/react/storybook/?path=/story/components-%E2%98%A2-breadcrumb--breadcrumb
2. Go to knobs
3. modify an value AFTER "Breadcrumb option 1" (the first breadcrumb item is the Only one that can be modified)
4. See that the displayed value does not get updated

### Expected behavior

The display in the story should be updated when knob values are changed.

### Additional context

Make a simple update to your Breadcrumb.story.jsx to add simple text knobs, then try modifying the text knob for Breadcrumb option 2:

```
storiesOf('Watson IoT | Breadcrumb', module)
  .addDecorator(withKnobs)
  .add(
    'default',
    () => {
      return <Breadcrumb {...props()} hasOverflow>
          <BreadcrumbItem href="#">
            {text('Breadcrumb option 1', '1 Homexxxxxxxxxxxxxxxxxxxxxx')}
          </BreadcrumbItem>
          <BreadcrumbItem href="#">
            {text('Breadcrumb option 2', 'breadcrumb 2')}
          </BreadcrumbItem>
          <BreadcrumbItem href="#">Breadcrumb 3</BreadcrumbItem>
        </Breadcrumb>;
    },
    {
      info: {
        text: `
          Breadcrumb enables users to quickly see their location within a path of navigation and move up to a parent level if desired.
        `,
      },
    }
  )
```

upcoming release?

date, the better!_

davidicus:
Closes #924

**Summary**

This allows for children to update component. 

- change breadcrumb to update when Children prop changes
- changed stories to have knobs to change text. 

checkout story for resize and change knobs. Text should update.

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.story.jsx
+++ b/src/components/Breadcrumb/Breadcrumb.story.jsx
@@ -10,7 +10,7 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
-import { withKnobs, boolean } from '@storybook/addon-knobs';
+import { withKnobs, boolean, text } from '@storybook/addon-knobs';
 import { BreadcrumbSkeleton, BreadcrumbItem } from 'carbon-components-react';
 
 import Breadcrumb from './Breadcrumb';
@@ -41,7 +41,7 @@ storiesOf('Watson IoT | Breadcrumb', module)
         <Breadcrumb {...props()}>
           <BreadcrumbItem href="#">Breadcrumb 1</BreadcrumbItem>
           <BreadcrumbItem href="#">Breadcrumb 2</BreadcrumbItem>
-          <BreadcrumbItem href="#">Breadcrumb 3</BreadcrumbItem>
+          <BreadcrumbItem href="#">{text('Breadcrumb 3 text', 'Breadcrumb 3')}</BreadcrumbItem>
         </Breadcrumb>
       );
     },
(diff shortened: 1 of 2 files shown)


src/components/Breadcrumb/Breadcrumb.story.jsx
```diff
@@ -106,11 +106,15 @@ storiesOf('Watson IoT | Breadcrumb', module)
...
+                {text('Breadcrumb 4 text', '4 Another page')}
               </BreadcrumbItem>
-              <BreadcrumbItem href="#" isCurrentPage title="5th level page">
-                5th level page
+              <BreadcrumbItem
+                href="#"
+                isCurrentPage
+                title={text('Breadcrumb 5 text', '5th level page')}
+              >
+                {text('Breadcrumb 5 text', '5th level page')}
```
sstone2423: I don't see this knob, only the knob at line 115. I'm guessing its because they have the same title.

src/components/Breadcrumb/Breadcrumb.story.jsx
```diff
@@ -106,11 +106,15 @@ storiesOf('Watson IoT | Breadcrumb', module)
               <BreadcrumbItem href="#" title="3 A really long page name">
                 3 A really long page name
               </BreadcrumbItem>
-              <BreadcrumbItem href="#" title="4 Another page">
-                4 Another page
+              <BreadcrumbItem href="#" title={text('Breadcrumb 4 text', '4 Another page')}>
+                {text('Breadcrumb 4 text', '4 Another page')}
```
sstone2423: I don't see this knob, only the knob at line 109. I'm guessing its because they have the same title.

src/components/Breadcrumb/Breadcrumb.story.jsx
davidicus: it is supposed to be the same knob. This populates the title with the same value

src/components/Breadcrumb/Breadcrumb.story.jsx
davidicus: same reply as above.

src/components/Breadcrumb/Breadcrumb.story.jsx
davidicus: image

PR #957, opened 2020-02-26
952 value card wrap units for small sizes

### LINKED ISSUE #952: [ValueCard] Wrap units instead of hiding at small screen sizes.
Currently the ValueCard will hide all tangential information other than the value, so units and thresholds,  at smaller sizes. Card needs to have these values wrap instead of hiding them. 

Must check all Value card stories to be sure content is not pushed off the card.

bjornalm:
Closes #952

**Summary**
Instead of hiding units and thresholds icons at smaller sizes the ValueCard now allows them to wrap  for attributes with a measured width <= 100px. For vertical cards with wrapping enabled and more than 2 items listed the size of the value and the units are decreased to fit the available height.

- Added SCSS files `_value-card.scss`, `_attribute.scss`, `_unit-renderer.scss` and `_value-renderer.scss` to house new CSS needed
- Refactored styled components for Unit-renderer info the _unit-renderer.scss
- Added a new story that shows wrapping and modified story  "tall / vertical / 6" to show compact wrapping.

_Note_
Attribute alignments in Attribute.jsx are achieved by different margins and we should probably use flex instead, but I didn't dare to refactor that code. Also, I think there is an over use of wrapping divs but I didn't dare to change that either due to fear of unknown downstream effects.

Tests the different card stories with a wrapper width of 120 to trigger the value and threshold icon wrapping.


src/components/ValueCard/Attribute.jsx
```diff
@@ -84,6 +78,7 @@ const propTypes = {
   ),
   renderIconByName: PropTypes.func,
   precision: PropTypes.number,
+  attributeCount: PropTypes.number.isRequired,
```
sstone2423: Prop comments would be helpful

src/components/ValueCard/ValueCard.story.jsx
```diff
@@ -72,6 +72,31 @@ storiesOf('Watson IoT|ValueCard', module)
       </div>
     );
   })
+
+  .add('small / wrapping', () => {
```
sstone2423: There is a ton of stories already. Is there any way to combine this with another story?

src/components/ValueCard/ValueRenderer.jsx
```diff
@@ -17,6 +21,8 @@ const propTypes = {
   size: PropTypes.string.isRequired,
   color: PropTypes.string,
   isVertical: PropTypes.bool,
+  allowedToWrap: PropTypes.bool.isRequired,
```
sstone2423: Prop comments would be helpful

src/components/ValueCard/_unit-renderer.scss
```diff
@@ -0,0 +1,25 @@
...
+  padding-left: 0.25rem;
+  font-weight: lighter;
+  white-space: nowrap;
+  padding-bottom: 0.25rem;
+  font-size: var(--default-font-size);
+}
+
+.#{$iot-prefix}--value-card__attribute-unit--mini {
+  padding-bottom: 0;
+  font-size: 1rem;
```
sstone2423: Needs carbon styling variables

src/components/ValueCard/_value-renderer.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../globals/vars';
+
+.#{$iot-prefix}--value-card__attribute-value--wrappable {
+  margin-right: 0.25rem;
```
sstone2423: Needs styling variables

src/components/ValueCard/_unit-renderer.scss
bjornalm: Done. 
Can I just assume that it is always safe translate 1rem into $carbon--spacing-05 or are there any special considerations needed given that the 1rem value of $carbon--spacing-05 is defined as a SCSS default and can be overridden?

src/components/ValueCard/ValueRenderer.jsx
bjornalm: done

src/components/ValueCard/ValueCard.story.jsx
bjornalm: Ok, merged it into the "small / units" story that now shows to cards

src/components/ValueCard/Attribute.jsx
bjornalm: Good point

src/components/ValueCard/_value-renderer.scss
bjornalm: done

src/components/ValueCard/_unit-renderer.scss
sstone2423: As far as I've seen / know, its safe to stick to those variables. I've personally never seen them change, but @davidicus or @tay1orjones could maybe give a better answer

tay1orjones: Yes, typically it's a safe bet to use `$carbon--spacing-05` in lieu of `1rem` on layout properties (`padding`, `margin`, `top`, etc). Though it's worth considering if the variable value should change, how that may impact or break the component. By and large we should be using the variables and avoid hardcoded values.

In this case though, font settings should generally be defined by `@include`ing the [typography variables](https://www.carbondesignsystem.com/guidelines/typography/productive), not spacing variables. In this case we might want to use `productive-heading-02`?

`@include type-style('productive-heading-02');`

bjornalm:
> Looks good pending that comment about potentially needing typography styles via the Carbon variable.
> 
> This story in particular seemed to be a good demonstration of the difference at 120px wide:
> 
> https://deploy-preview-957--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-valuecard--medium-vertical-3
> 
> https://carbon-addons-iot-react.com/?path=/story/watson-iot-valuecard--medium-vertical-3

Yes indeed

PR #965, opened 2020-02-26
feat(storybook): remove centered addon

### LINKED ISSUE #962: [Storybook] Remove centering addon
### Summary

We've run into a few issues with this addon, notably in #906 that the usage of `position: fixed` doesn't work well with tooltip positioning

We should remove this addon and ensure that the rendering of our stories mimics the Carbon storybook for consistency's sake.

enricoberti:
Closes #962

**Summary**

- Removed the 'centered' addon that was causing a lot of problems with positioning of the tooltips/dropdowns. This also bring visual consistency with Carbon's way of displaying stories in Storybook

- removed from the Storybook config, package.json and updated all the snapshots

- check that all the stories render correctly


yarn.lock
```diff
@@ -3046,13 +2986,6 @@
   dependencies:
     core-js "^3.0.1"
 
-"@storybook/core-events@5.2.8":
```
sstone2423: Were these all dependencies of the addon centered plugin? Seems like a lot

enricoberti: that came with `yarn remove @storybook/addon-centered`, most of them are different version of what's imported/cross-referenced by the other storybook plugins
should I just manually remove the dependency on package.json instead?

sstone2423: Nope, sounds good. Didn't realize the version difference

PR #971, opened 2020-02-27
feat(tile-catalog): tilecatalog for marketplace

### LINKED ISSUE #967: [TileCatalog] Simple Tile Canvas
Simple Tile canvas includes tile canvas, sort, search and pagination.

image

jessieyan:
Closes #967 

**Summary**
TileCatalog with Simple Tile Canvas ( Deprecate old TileCatalog)

Includes:
1. A canvas with specified number of columns and rows. 
2. Optional sort function
3. Optional search function
4. Pagination
5. React Tests


src/components/FileDrop/FileDrop.jsx
```diff
@@ -63,7 +63,7 @@ const propTypes = {
   /** file type either TEXT or BINARY to determine the correct encoding */
   fileType: PropTypes.oneOf(Object.values(FILE_TYPES)),
 };
-
+/* istanbul ignore next */
```
sstone2423: Instead of ignoring coverage, tests should be written

src/components/Table/AsyncTable/MockApiClient.js
```diff
@@ -2,6 +2,7 @@
  * Simulates a client capable of asynchronously fetching
  * paginated, filtered and sorted data from some resource (e.g. an HTTP API)
  */
+/* istanbul ignore file */
```
sstone2423: Instead of ignoring coverage, tests should be written

src/components/TileCatalog/TileCatalog.jsx
```diff
@@ -1,189 +1,176 @@
...
-          )}
-        </StyledEmptyTile>
-      )}
-      {!isLoading && tiles.length > 0 && !error && pagination ? (
-        <SimplePagination {...pagination} maxPage={Math.ceil(totalTiles / pageSize)} />
-      ) : null}
-    </StyledContainerDiv>
+    <div>
+      <div>
+        {/* {persistentSearch ? (
```
sstone2423: Is this suppose to be removed?

src/components/TileCatalog/TileCatalog.jsx
```diff
@@ -1,189 +1,176 @@
...
+          <div className="tile-catalog--persistent-search">
+            <Search placeHolderText="" onChange="'" size="sm" value="" labelText="" />
+          </div>
+        ) : null} */}
+        <div />
+      </div>
+
+      <div className={`${iotPrefix}--tile-catalog--canvas-container`}>
+        <div className={`${iotPrefix}--tile-catalog--tile-canvas`}>
+          {/* {featuredTile ? (
```
sstone2423: Is this suppose to be removed?

src/components/TileCatalog/TileCatalog.jsx
```diff
@@ -1,189 +1,176 @@
...
+          <div className={`${iotPrefix}--tile-catalog--tile-canvas--content`}>{renderGrid()}</div>
+          <div className={`${iotPrefix}--tile-catalog--tile-canvas--bottom`}>
+            <TilePagination
+              page={currentPage}
+              numPages={Math.ceil(tiles.length / (numRows * numColumns))}
+              onChange={newPage => setCurrentPage(newPage)}
+            />
+          </div>
+        </div>
+        {/* {hasFilter ? (
```
sstone2423: Is this suppose to be removed?

src/components/TileCatalog/TileCatalog.test.jsx
```diff
@@ -1,18 +1,145 @@
...
+        title="Test Tile Catalog"
+        numRows={1}
+        numColumns={2}
+      />
+    );
+    expect(getByText('Tile 1')).toBeTruthy();
+    expect(getByText('Tile 2')).toBeTruthy();
+  });
+
+  test('TileCatalog placeholder tile are rendered', () => {
```
sstone2423: Is this test any different than the above test?

src/components/TileCatalog/TileCatalog.test.jsx
```diff
@@ -1,18 +1,145 @@
...
+        title="Test Tile Catalog"
+        numRows={2}
+        numColumns={2}
+      />
+    );
+    expect(getByText('Tile 1')).toBeTruthy();
+    expect(getByText('Tile 2')).toBeTruthy();
+  });
+
+  test('TileCatalog to have default ', () => {
```
sstone2423: This is checked by PropTypes already so a test is redundant

src/components/TileCatalog/TilePagination/TilePagination.jsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const propTypes = {
+  page: PropTypes.number.isRequired,
```
sstone2423: Needs prop comments

src/components/TileCatalog/TilePagination/TilePagination.jsx
```diff
@@ -0,0 +1,101 @@
...
+  i18n: {
+    ariaLabelPreviousPage: 'Previous page',
+    ariaLabelNextPage: 'Next page',
+    ariaLabelPagination: 'pagination',
+    ariaLabelPage: 'page',
+  },
+};
+
+const TilePagination = ({ page, numPages, onChange, i18n }) => {
+  const prevButton = (
```
sstone2423: Since there's not much logic in these buttons, it's easier to read if you put them all into the return below, but thats just a personal preference

src/components/TileCatalog/TitleCatalog.story.jsx
```diff
@@ -0,0 +1,129 @@
+import React, { useState } from 'react';
```
sstone2423: This file has a typo. `TitleCatalog` vs `TileCatalog`

src/components/TileCatalog/TitleCatalog.story.jsx
```diff
@@ -0,0 +1,129 @@
...
+        numRows={number('numRows', 2)}
+        hasSort={boolean('hasSort', true)}
+        sortOptions={sortOptions}
+        onSort={action('sort', () => {})}
+        selectedSortOption={selectedSortOption}
+      />
+    </div>
+  ))
+  .add('StatefulTileCatalog', () => {
+    const StatefulTileCatalog = () => {
```
sstone2423: Should we export this for others to use?

src/components/FileDrop/FileDrop.jsx
davidicus: This is ignoring the default no op functions that do not need to be tested

src/components/Table/AsyncTable/MockApiClient.js
davidicus: This is a mock api file used to create data for the stories. This does not need to be included in our test coverage.

src/components/TileCatalog/TilePagination/TilePagination.jsx
bryancboyd: I actually disagree in this particular case, since we're using vanilla JS from the carbon simple pagination (which doesn't have a react wrapper yet) it's a bit verbose to keep all in the render method.  I prefer keeping the render method pretty sparse so that you can more easily see the layout.  (but it's also a personal preference 👍)

src/components/TileCatalog/TitleCatalog.story.jsx
bryancboyd: I don't think so; the new TileCatalog doesn't include a data model for the individual tiles (just the node to be rendered) so any sort and search logic has to be implemented by the user directly.  TileCatalog does handle pagination in component state, so it's kind of a "half-stateful" component.

src/components/TileCatalog/TileCatalog.jsx
bryancboyd: These are additions to the component coming soon, I'm fine with leaving them in as comments.

src/components/TileCatalog/TileCatalog.jsx
bryancboyd: These are additions to the component coming soon, I'm fine with leaving them in as comments.

src/components/TileCatalog/TileCatalog.jsx
bryancboyd: These are additions to the component coming soon, I'm fine with leaving them in as comments.

src/components/TileCatalog/TileCatalog.test.jsx
jessieyan: It is checked by PropTypes. The jest test coverage still asks me to cover the default function. So I added this test. Will update the name a bit.

src/components/TileCatalog/TitleCatalog.story.jsx
jessieyan: This is just an example not for others to use.

src/components/TileCatalog/TilePagination/TilePagination.jsx
sstone2423: Good point. Code on!

jessieyan:
Hold on to merge. Needs to deprecate old TileCatalog.

bryancboyd:
@jessieyan in fact, I wonder if we should make the new catalog be a separate component `TileCatalogNew` and keep the original as `TileCatalog`, until our next major release?  That way we don't have a breaking change.  Then as part of the next major release cleanup, we can remove the old TileCatalog and replace it with the new.

PR #978, opened 2020-02-28
fix(card): added proper padding for title

### LINKED ISSUE #969: [Dashboard] Info icon and long headings on cards
When the headline is long, the info icon on dashboard tiles tiles lack the standard 1em padding between the edge of the tile. Similarly, if a heading is too long without the icon it has the same visual defect. 
 
info-icon-padding

Create padding for card component header and verify on different card stories.

bjornalm:
Closes #969

**Summary**
Made sure the title has correct padding on the right side. Corner cases included presens of tooltip, toolbar, custom render prop using external heading etc.

image

image

1. Added padding to the card header so that both sides have the same padding. 
2. Only render CardToolbar if there are any toolbar actions instead of rendering empty zero space divs.
3. Since the CardToolbar buttons manage their own width which includes the proper whitespaces, the toolbar will get a negative padding when placed in a card header. This does not effect the TableCard that also uses the CardToolbar.
4. Added unit test 

Verify that the padding of the title text and title tooltip icon looks okay when the title is really long for all the available Card stories.


src/components/Card/CardToolbar.jsx
```diff
@@ -85,9 +88,10 @@ const CardToolbar = ({
   availableActions,
   timeRange,
   onCardAction,
+  className,
 }) => {
   return isEditable ? (
-    <div className="card--toolbar">
+    <div className={classNames(className, 'card--toolbar')}>
```
sstone2423: Maybe go ahead and add the prefix while you're here?

src/components/Card/_card.scss
```diff
@@ -1,6 +1,8 @@
-@import '~carbon-components/scss/globals/scss/vars';
+@import '../../globals/vars';
 @import '~carbon-components/scss/globals/scss/layout';
 
+$iot-header-padding: $spacing-05;
+
 .card {
```
sstone2423: Add the prefix while you're here

src/components/Card/Card.story.jsx
```diff
@@ -51,6 +51,27 @@ storiesOf('Watson IoT|Card', module)
       </div>
     );
   })
+  .add('with explicit tooltip and long title', () => {
```
sstone2423: Does this actually need a story? I think this could be achieved by making the tooltip a text knob in the other stories that use it

src/components/Card/_card.scss
bjornalm: I did consider that, but this "card" string itself is a prefix for a bunch of other classes (like "card--toolbar-action") due to the Sass ampersand construct: `.card { &--toolbar-action {...} }`. These classes are in turn used used by other components and _Styled Components_ so such a refactoring would spread out and I did not want to touch them at this time. Is it okay address this in a separate issue? 

I personally don't think this ampersand naming construct is a good strategy since it makes gradual refactoring more difficult.

src/components/Card/CardToolbar.jsx
bjornalm: see comment below

src/components/Card/Card.story.jsx
bjornalm: In my opinion the explicit tooltip should appear in at least one story and I don't think it can be a knob since it isn't a string but a ReactElement. I've now merged this story with the story  _with ellipsed title and tooltip_ so that it contains both tooltip scenarios.

src/components/Card/_card.scss
tay1orjones: We can import the prefix to the js files and prepend the prefix on the styled components definition as well as the className. Searching globally in the project this is the only `.card--` usage I could find in styled components https://github.com/IBM/carbon-addons-iot-react/blob/5a0689ae45cd47a7e5dd2a787f5881ba6ea17639/src/components/Card/CardToolbar.jsx#L44-L63

There's ~10 other spots we'd need to update the className, as well as 1 spot in a selector in Card.text.jsx

Probably beyond the scope here, but it would be nice to have the prefix so consumers can reliably override the styles if they need to. If we don't include it here, we should open a new issue to track this and address when we can.

tay1orjones: Also I agree with you on the ampersand construct - we should avoid it's use and nesting as well to provide an easier path for maintenance and refactoring.

davidicus: I would argue that the use of ampersand makes it easier to rename elements. Change the overarching class and the rest of them are changed. When would you add a prefix to one but not all of the classes that fall under that component?

bjornalm: @tay1orjones Yes, I could add the prefix to the Styled Component as well, but even better would be to move the Styled Component over to the scss file. And then this refactor would grow even bigger.

@davidicus I agree that we would want to change everything related to the card naming to be consistent. My point was that we are making it more difficult to do it gradually step by step with this construct which might lead a developer to continue using the old naming scheme instead of renaming everything if it leads to a bigger refactor that the task itself would call for. Anyway, it is not really a big deal here. Should I just go ahead and rename it all in this PR or should we do that in a separate issue and include removing the Styled Components as well?

tay1orjones: A separate issue is fine from my POV

davidicus:
I have created an [issue](https://github.com/IBM/carbon-addons-iot-react/issues/983) to track the prefix and refactor work

PR #980, opened 2020-03-02
feat(tests): add snapshot test for public component api interface

tay1orjones:
**Summary**

- Adds a test to snapshot the public API interface of our exported components to help adhere to semantic versioning and catch inadvertent breaking changes in PRs.
- [Borrowed directly from Carbon](https://github.com/carbon-design-system/carbon/pull/5495), thanks @joshblack!

- Add new testing file and required dependency

- Verify the newly generated snapshot makes sense


joshblack:
🔥 

Let us know how it works out for you all! We're trying it now and 🤞 it's more helpful than it is added noise.

We're also considering adding a similar test for our styles as well if you all are interested in sync'ing up on that work 👀

PR #989, opened 2020-03-08
fix(card): refactored styled components to scss

### LINKED ISSUE #983: [Card] Refactor off styled components and add prefix to styles
Remove the styled components and move them to a `_card.scss` stylesheet. Add prefixes to class declaration in Card.jsx and use prefix in the stylesheet.

bjornalm:
Closes #983

**Summary**
Remove the styled components and move them to a _card.scss stylesheet. Add prefixes to class declaration in Card.jsx and use prefix in the stylesheet.

Beyond removing the styled components I also fixed

1. Created new scss files for the card-range-picker and the card-toolbar components
2. Removed ToolbarDateRangeWrapper for edit scenario where it was wrapping the overflow menu that does not have a date range.  @scottdickerson  is it safe to remove that wrapper there?
3. The Styled component code `span#timeRange { display: ${props => (props.cardWidthSize < 230 ? `none` : `flex`)}; }` was not working, fixed in the CardRagePicker instead.
4. The CardWrapper expanded style was defined 2 times, where one was never used, removed that one.
5. Changed hardcoded color `#0062ff` to `$focus`  and  `#e5e5e5` to `$hover-ui`. Could not find any other matching color variables so I assumed these were the ones to be used.
6. The cardwrapper explicitly defines the event handler props that are attached by the Dashboard grid when "isEditable" is active. 
7. Since I can't be sure that other applications aren't adding additional HTML attributes (this was allowed by Styled Components) I've created a component utility filter method to filter all remaining props and only pass forward valid HTML attributes. Our own custom props that "accidentally" have been forwarded will be filtered out.
8. Added the 'data-testid' to the card and the wrapper
9. Added role=presentation since a role was needed and it is the only accepted role for a static element with event handlers.

Test all versions of cards with different configurations


src/components/Card/Card.jsx
```diff
@@ -27,17 +28,37 @@ const { prefix, iotPrefix } = settings;
...
+  dimensions,
+  isExpanded,
+  showOverflow,
+  id,
+  style,
+  className,
+  ...others
+}) => {
+  // TODO: Solve this better by either
+  // a) explicitly define and pass along the HTML attributes needed for the wrapper
```
sstone2423: I'd prefer this route, but we can see what @davidicus and @scottdickerson think. We have similar issues in quite a few components where we somewhat blindly pass those `...other` props

src/components/Card/Card.jsx
```diff
@@ -47,34 +68,37 @@ export const CardHeader = (
...
-export const SkeletonWrapper = styled.div`
-  padding: ${CARD_CONTENT_PADDING}px;
-  width: 80%;
-`;
+const CardContent = props => {
+  const { children, dimensions, isExpanded } = props;
+  const height = `${dimensions.y - CARD_TITLE_HEIGHT}px`;
+  return (
+    <div
+      style={{ [`--card-content-height`]: height }}
```
sstone2423: Should this be its own class?

src/components/Card/Card.jsx
```diff
@@ -284,15 +306,18 @@ const Card = props => {
                   {!isVisible && isLazyLoading ? ( // if not visible don't show anything
                     ''
                   ) : isLoading ? (
-                    <SkeletonWrapper>
+                    <div
+                      style={{ ['--card-content-padding']: `${CARD_CONTENT_PADDING}px` }}
```
sstone2423: Should this be its own class?

src/components/Card/CardRangePicker.jsx
```diff
@@ -3,6 +3,11 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import EventSchedule from '@carbon/icons-react/lib/event--schedule/20';
 import { ToolbarItem, OverflowMenu, OverflowMenuItem } from 'carbon-components-react';
+import classNames from 'classnames';
+
+import { settings } from '../../constants/Settings';
+
+const { iotPrefix } = settings;
 
 const TimeRangeLabel = styled.div`
```
sstone2423: Looks like this one got missed

src/components/Card/Card.jsx
bjornalm: Do you mean an HTML class for the height? The height is a JS calculated value so that is why it is set as a custom property (CSS variable) and not prop in a class like the other fixed values.

sstone2423: Ah I see. I read that as a boolean 👍

src/components/Card/Card.jsx
sstone2423: Disregard this one as well 👍

src/components/Card/CardRangePicker.jsx
bjornalm: Great catch! Fixed now

src/components/Card/Card.jsx
tay1orjones: Yeah, it would be nice to have an explicit API defining what's required for positioning.

davidicus: Do we know of any reasons to pass on additional attributes for  the `Card` container `div`? Perhaps we just add a `data-testid` for the div and not spread any props at all. I like the flexibility of spreading props but it's definitely error prone when being applied to a DOM element.

scottdickerson:
@bjornalm it's tough to say, these are all the react-grid-layout props that need to be passed through to the CardWrapper to correctly position and size the cards in the grid

src/components/Card/Card.jsx
bjornalm: Now I'm explicitly adding the event handlers that are attached by the Dashboard grid when "isEditable" is active. **But, since I can't be sure that other applications aren't adding additional HTML attributes** (this was allowed by Styled Components) I've created a component utility filter method to filter all remaining props and only pass forward valid HTML attributes or react library props like 'ref' etc. Our own custom props that "accidentally" have been forwarded will be filtered out.

The source of my list of valid HTML attributes is the Styled Components source code. My idea behind this is that we can use this new `filterValidAttributes` when we refactor Styled Components and face a similar situations.

Also Adding the 'data-testid' to the card and the wrapper

tay1orjones:
`react-grid-layout` has documented their props - https://github.com/STRML/react-grid-layout#grid-layout-props

bjornalm:
> `react-grid-layout` has documented their props
> 
> * https://github.com/STRML/react-grid-layout#grid-layout-props
> * https://github.com/STRML/react-grid-layout#responsive-grid-layout-props
> * https://github.com/STRML/react-grid-layout#grid-item-props
> 
> I'm not opposed to just spreading these props for `react-grid-layout`. I meant we should just clearly define the api for the unrelated prop(s) that broke (id, data-testid, etc). The helper is nice, but is an additional overhead we'd have to remember to consider moving forward. This feels like a use case for an eslint rule or a retweaking of one we already have enabled.

I looked at those props earlier, but I think they describe what props we can use when configuring the grid and items, they do not represent the actual HTML attributes that the library attaches to the grid items DOM elements. For instance, the Card wrapper will get `onMouseDown` and `onMouseUp` attached (not described as props in the docs) and if we prevent those the repositioning in the dashboard grid won't work.

scottdickerson:
exactly right @bjornalm unfortunately the react-grid-layout documentation is very poor in this regard, we should open an issue on them to update their docs

bjornalm:
> exactly right @bjornalm unfortunately the react-grid-layout documentation is very poor in this regard, we should open an issue on them to update their docs

Agree, I opened an issue https://github.com/STRML/react-grid-layout/issues/1166  
The author says he is only fixing bugs, so I'm not convinced the docs will be updated anytime soon.

PR #990, opened 2020-03-09
fix(tablecard): render data function now working

### LINKED ISSUE #766: [TableCard] renderDataFunction is not working.
Hi Team,

I am using ``` <TableCard> ``` component for implementing a card as shown below in design reference. The cell data needs to be customized so that it behaves like a navigation link. For this I am trying to use renderDataFunction but it is not working and seems to be hardcoded to renderThresholdIcon only.

Design reference : 
image

Storybook reference:
I am using https://carbon-addons-iot-react.com/?path=/story/watson-iot-tablecard--size-tall
carbon-addons-iot-react: v2.17.2

We need this fix for our Jan 2020 release.

<sub>[Internal tracking issue #131](https://github.ibm.com/Watson-IoT/pal-tracking/issues/131)</sub>

bjornalm:
Closes #766

**Summary**
The objects in the column array can now contain a render-function called renderDataFunction.

Prop was incorrectly called `renderer` but should have been `renderDataFunction` to work with the underlying Table component. Renamed prop and added unit test.

Add a renderDataFunction function to a TableColumnsProp in content.columns[colItem] the same way you would do for a table.

### DIFF
--- a/src/constants/PropTypes.js
+++ b/src/constants/PropTypes.js
@@ -116,7 +116,8 @@ export const TableCardPropTypes = {
         width: PropTypes.number,
         label: PropTypes.string.isRequired,
         priority: PropTypes.number,
-        renderer: PropTypes.func,
+        /** See the renderDataFunction for TablePropTypes */
+        renderDataFunction: PropTypes.func,
         type: PropTypes.string,
       })
     ).isRequired,
--- a/src/components/TableCard/TableCard.test.jsx
+++ b/src/components/TableCard/TableCard.test.jsx
@@ -137,4 +137,58 @@ describe('TableCard', () => {
     const totalColumns = tableColumns.filter(item => item.priority === 1 || item.priority === 2);
     expect(wrapper.find('TableHeader').length).toBe(totalColumns.length + 1); // +1 for action column
   });
+  test('Columns should use custom render fuction when present', () => {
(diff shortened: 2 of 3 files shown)


src/constants/PropTypes.js
```diff
@@ -116,7 +116,8 @@ export const TableCardPropTypes = {
         width: PropTypes.number,
         label: PropTypes.string.isRequired,
         priority: PropTypes.number,
-        renderer: PropTypes.func,
```
davidicus: Is this prop something that worked at all? So, is this a breaking change here? If the feature is completely broken in its current state then we can probably safely remove this prop but if there are possible users out there we will need to add a deprecation warning and alias it to the correct prop name.

bjornalm: From what I could see this did not work at all, it was simply passed down to the Card component where it was ignored. Not sure about if it ever worked in the past.

bjornalm: Went back as far as commit 9e3490b7 on 5 June 2019 when the `TableCardPropTypes` were added to the `PropTypes.js` file but even then the 'renderer' did not work. Replacing the `renderer` with `renderDataFunction` also did the trick back then though.

tay1orjones: Great! In that case we can forego a breaking change.

PR #996, opened 2020-03-10
feat(tablecard): support custom icon names and colors

scottdickerson:
Closes #https://github.ibm.com/wiotp/monitoring-dashboard/issues/773

**Summary**

- Change to allow passing custom icon names and colors to configured TableCard thresholds. here's a custom Icon and Color for a TableCard threshold:
image

- refactor(CardIcon): share all Icon rendering code across all Cards so the same logic is followed
- refactor(Attribute): refactor to use shared CardIcon
- refactor(Hotspot): refactor to use shared CardIcon
- fix(Hotspot): also pass the 'title' attribute to the svg rendered so that when we mouseover the threshold shows
- fix(PropTypes): list the renderIconByName function in the CardPropTypes (code was already there, just documentation was missing)
- feat(TableCard): add ThresholdIcon component that support custom icons and color to be passed to a TableCard threshold
- fix(table): sorting breaks if some fields are undefined inside the table data
- fix(table): support filtering on numeric values without throwing a prop types error

- Verify the custom icons shows correctly in the story called 'watson-iot-tablecard--table-with-thresholds-precision-and-expanded-rows'


src/components/TableCard/TableCard.story.jsx
```diff
@@ -164,6 +167,15 @@ storiesOf('Watson IoT/TableCard', module)
           values={tableData}
           onCardAction={(id, type, payload) => action('onCardAction', id, type, payload)}
           size={size}
+          renderIconByName={(name, props = {}) =>
```
sstone2423: Should a description of how to use this prop be added?

src/utils/componentUtilityFunctions.js
```diff
@@ -2,6 +2,7 @@ import delay from 'lodash/delay';
 import moment from 'moment';
 import { sortStates } from 'carbon-components-react/lib/components/DataTable/state/sorting';
 import fileDownload from 'js-file-download';
+import isNil from 'lodash/isNil';
```
sstone2423: Fun fact, just found this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

PR #998, opened 2020-03-11
feat(tablecard): do not require the threshold column be shown in a table

scottdickerson:
Closes #

**Summary**

- Table card users want to be able to add icon threshold columns without showing the value column

- feat(TableCard): allow the table to only show the threshold column
- docs(PropTypes): add additional info to the prop type
- fix(TableCard): fix several prop types errors in table

- Notice the new 'table card thresholds only with icon'

### DIFF
--- a/src/components/TableCard/TableCard.jsx
+++ b/src/components/TableCard/TableCard.jsx
@@ -370,6 +370,7 @@ const TableCard = ({
     label: uniqueThresholds[index].label
       ? uniqueThresholds[index].label
       : `${capitalize(columnId)} ${strings.severityLabel}`,
+    width: uniqueThresholds[index].width,
     isSortable: true,
     renderDataFunction: renderThresholdIcon,
     priority: 1,
--- a/src/components/TableCard/TableCard.jsx
+++ b/src/components/TableCard/TableCard.jsx
@@ -9,7 +9,7 @@ import cloneDeep from 'lodash/cloneDeep';
 import capitalize from 'lodash/capitalize';
 import OverFlowMenuIcon from '@carbon/icons-react/lib/overflow-menu--vertical/20';
 
-import { CardPropTypes, TableCardPropTypes } from '../../constants/PropTypes';
+import { CardPropTypes, TableCardPropTypes } from '../../constants/CardPropTypes';
 import Card, { defaultProps as CardDefaultProps } from '../Card/Card';
 import { CARD_SIZES } from '../../constants/LayoutConstants';
 import StatefulTable from '../Table/StatefulTable';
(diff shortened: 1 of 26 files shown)


src/components/TableCard/TableCard.jsx
```diff
@@ -623,6 +642,7 @@ const TableCard = ({
           <StyledStatefulTable
             columns={columnsToRender}
             data={tableDataWithTimestamp}
+            id={`table-for-card-${id}`}
```
tay1orjones: I assume this is for testing - could we use `data-testid` instead?

scottdickerson: we could but this is primary just to fix the duplicate ids issue in the DOM when multiple tablecards were in the same view.  since they can configure the id of the card they'll still be able to locate the correct DOM elements

src/styles.scss
```diff
@@ -77,6 +77,50 @@ $carbon--theme: $carbon--theme--g10;
 @import 'globals/grid';
 @import 'globals/charts';
 
+// Font families needed for Charts (not included by carbon by default)
+@font-face {
+  font-family: 'IBM Plex Sans Condensed';
+  font-style: italic;
+  font-weight: 300;
+  src: url(~@ibm/plex/IBM-Plex-Sans-Condensed/fonts/complete/woff2/IBMPlexSansCondensed-LightItalic.woff2)
```
tay1orjones: If we're going to self host the plex files we should update all weights to use the self hosted version to avoid version mismatching between the default weights provided by `@carbon/type` and these additional weights. 

Also, this is going to increase our bundle size by a decent amount. What about [using unpkg instead](https://github.com/carbon-design-system/carbon/blob/master/docs/guides/ibm-plex.md#unpkg)?

tay1orjones: I worry that we're forcing a bloated bundle on consumers by adding plex as self hosted. 

unpkg is a good option, as is google fonts. Either of these allow consumers to decide how best to load the fonts (their own CDN or self host), with the google fonts fallback/default that falls in line with `@carbon/type`

tay1orjones:
```diff
- src: url(~@ibm/plex/IBM-Plex-Sans-Condensed/fonts/complete/woff2/IBMPlexSansCondensed-LightItalic.woff2)
+ src: url(//unpkg.com/@ibm/plex/IBM-Plex-Sans-Condensed/fonts/complete/woff2/IBMPlexSansCondensed-LightItalic.woff2)
```

tay1orjones:
I've opened #1007 to follow on from this to remove our fix once it's fixed upstream in either `@carbon/type` or `@carbon/charts`

PR #1000, opened 2020-03-12
fix(table): remove padding from icon-only row actions

### LINKED ISSUE #937: [Table]: Row actions have extra padding between them
### Describe the bug

Create two row actions right next to each other, you'll notice some extra padding between the two buttons

### To Reproduce

image

enricoberti:
Closes #937 

**Summary**

- Row actions have left padding on the icons regardless of the presence of their label

- Removed padding from the SVG and the button itself when there's no label text specified

Screenshot 2020-03-12 12 45 31


src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
```diff
@@ -0,0 +1,8 @@
+@import '../../../../globals/vars';
+
+.#{$iot-prefix}--btn--icononly {
+  margin-left: 0 !important; // !important needed to override the styled component value
```
tay1orjones: Could we instead increase the scope here to attempt to migrate off of styled-components entirely for the `RowActionsCell` component?

src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
```diff
@@ -0,0 +1,8 @@
+@import '../../../../globals/vars';
+
+.#{$iot-prefix}--btn--icononly {
```
tay1orjones: Would be best to scope this to a block level identifier of the component

```suggestion
.#{$iot-prefix}--row-actions-cell-btn--icononly {
```

enricoberti: Makes sense, thanks!

src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
enricoberti: I kept it to a minimum on this PR and opened https://github.com/IBM/carbon-addons-iot-react/issues/1005 to do the refactoring

scottdickerson: importants are bad usually, can we increase the specificity in a different way to force the override?

enricoberti: @scottdickerson @tay1orjones thanks! i updated a bit the css to avoid !important (feels better now)

Issue #1001, opened 2020-03-12
[testing] add data-testid attribute to all components


tay1orjones:
<sub>[Internal issue #320](https://github.ibm.com/Watson-IoT/pal-tracking/issues/320)</sub>

### Summary

To enable support of testing automation in applications that consume this library, we should provide a dedicated `data-testid` data attribute for each component. It should be placed on the containing element of each component as well as any pertinent sub sections. This attribute would be used in place of class or id selectors in testing automation for applications that consume this library.

| Prop Name | Attribute Name |
| - | - |
| `testId` | `data-testid` |

[Here's a link to Kent C. Dodd's blog with some reasoning for this naming convention](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change). Additionally, it follows precedent set by React Native Web and the [React Testing Library that we use](https://testing-library.com/docs/dom-testing-library/api-queries#overriding-data-testid)

**Additional context**
For **exported, non-internal components** this should be configurable via a prop with a default prop of the component name.

```jsx
function ExportedComponent({ testId }) {
  return (
    <>
      <span data-testid={testId}>Example</span>
    </>
  );
}

ExportedComponent.propTypes = {
  testId: PropTypes.string,
};
ExportedComponent.defaultProps = {
  testId: 'ExportedComponent',
};
```

**Internal components** (such as those used internally in the table or elsewhere) should utilize any available id props to attempt to provide a unique value.

```jsx
function InternalTableCell({ tableId, columnId, rowId }) {
  return <td data-testid={`${tableId}-${columnId}-${rowId}`}>Example</td>;
}
```

High level checklist:
- [ ] Add attribute to all components
- [ ] Fix any usages we might have of improper naming, `test-ID` instead of `testId` or similar
- [ ] Update testing documentation to no longer state that "we working on this" but instead simply say "it's provided"
- [ ] Write test to ensure that these sections of the UI are always selectable  and do not break with changes.

davidicus:
Things to consider when addressing Table test ids

[...]

PR #1002, opened 2020-03-12
fix(docs): add versioning and testing documentation

tay1orjones:
**Summary**

- After recent breaking changes that slipped into the production build, @davidicus and I have pulled together a draft of the versioning contract we'd like to adhere to as a project. Inspired by Carbon's documentation, it lays out specific types of changes with associated change levels major/minor/patch.
- Also adds documentation regarding the `data-testid` attribute - I opened an issue to implement this everywhere, #1001 
- <sub>[Internal issue #315](https://github.ibm.com/Watson-IoT/pal-tracking/issues/315)</sub>

- Add/change documentation `.md` files

- Verify that you agree with the statements provided here. @stuckless @scottdickerson @bryancboyd I'm particularly interested in any feedback you may have, please let me know how we can improve this.


docs/guides/versioning.md
```diff
@@ -0,0 +1,312 @@
...
+ExportedComponent.defaultProps = {
+  message: '',
+- testID: 'ExportedComponent',
++ testID: 'ExportedComponentNewName',
+};
+```
+
+#### The DOM node that an `id` corresponds to is changed
+
+semver bump: **minor**
```
stuckless: I argue that in the event that is no existing data-test-id then this is a major change and would be like removing a data-test-id

docs/guides/versioning.md
```diff
@@ -0,0 +1,312 @@
...
+  message: PropTypes.string,
++ testID: PropTypes.string,
+};
+ExportedComponent.defaultProps = {
+  message: '',
++ testID: 'ExportedComponent',
+};
+```
+
+#### A `data-testid` attribute default prop value is changed or removed
```
stuckless: While not clearly specified here, moving a data-test-id to some other element should be considered a breaking change.

stuckless:
> Looks good. I'd consider adding a comment that consumers shouldn't rely on the library's internal constants either. There are times such as with the `CARD_SIZES` issue where constants need to be changed or moved to other files. In the case that the library wants to export specific constants, those should be explicitly exported in the index. All other constants should be open to changes for internal development

While this is a good practice, you should find another constants example since, CARD_SIZES, CARD_TYPES, etc, are all exposed in the index as a means for downstream consumers to be aware of the known types, sizes, etc.  But, it brings up a good point that changing a constant by removing it, or changing its meaning should be considered a major change.  (that may already be in the doc, not sure)

This triggers another thing that likely should be in the document.   If we don't change a property but we change how we interpret that property then that should be documented in scenarios as breaking or non-breaking changes.  ie, Like the CARD_SIZES where SMALL constant didn't change but how we interpret small changed (ie, small means something different now in terms of sizing), then that is likely a major change.

stuckless:
As stated above, I'm not sure the "as is" interpretation of semver is good for UI libraries, because, we focus too heavily on the API changing as a means of something major happening.  ie, we consider MAJOR to be when something changes that may result in your application breaking, but we can do lots of things without changing an API that can 'break' an application.  And it's that understanding of what is "breaking change" that not everyone agree on.

For example, let's say you create a table v1.1.0 and in that table it can just render static information in a grid.  Then you add a cool new feature where you can drag columns around and re-order them, and you decide that you store that view in the browswer local storage so that when a user comes back it is remembered.  You decide that you are going to enable this by default, and have a flag that can be used to turn it off. Is that a version v2.0.0 or v1.2.0?  Most developers would say it's a feature, it's new, it doesn't "break" anything (ie my code compiles), so, it's v1.2.0.  But what if I'm a highly regulated company, like an energy company, and this change is accepted into our product and then shipped to them, and all of sudden their users can start to rearrange their screen, which, they should not be able to do.  That is a breaking change for them, and for us.

And this is where semver for UI components is tricky.  It works great for non-ui libraries, but, it doesn't really have much value in a UI component library, when a v1.1.1 still requires that I have to think very carefully before accepting the update because from a UI/Application point of view, there may be lots of changes that happened where the API didn't break, and the code compiles, but the change will 'break' an application.

So all that being said, how do we convey this information.  ie, how you let me know that you've added a feature that by default enables something in the UI that wasn't there before.  Since I can add a feature that is off by default, and as such won't change the UI or I can add a feature that is on default and it does change the UI, and under this model, they are both a MINOR version update.

In addition to versioning, should there be a living developer document that describes these kinds of changes.  ie, RELEASE NOTES.   We don't want to generate this from the commits, since those are  generally not very good in details.  But I think there needs to be some way to convey "what has changed" in every minor and major release.

docs/guides/testing.md
```diff
@@ -0,0 +1,50 @@
+# Testing
+
+Enabling projects that consume this library to accurately test applications using the components from this library is an important use case that we intend to support as best we can. If you have suggestions as to how we can improve, please [open an issue](https://github.com/IBM/carbon-addons-iot-react/issues/new?assignees=&labels=%3Ahammer%3A++Enhancement&template=feature-request-or-enhancement.md&title=%5BComponentName%5D+request_title).
+
+## Automated Testing Support
+
+To enable support of testing automation in applications that consume this library, we are working to provide a dedicated `data-testid` data attribute for each component. [Here's a link to the issue to follow our progress](https://github.com/IBM/carbon-addons-iot-react/issues/1001).
```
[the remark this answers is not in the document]
davidicus: No,  we will add data-testid's in addition to the current id prop. We want to standardize and not make breaking changes unnecessarily like you mention.

davidicus:
@stuckless your point of features on by default is a good one but one we are mitigating by following the opt in model. We haven't been consistent with this in the past but something we check for moving forward. This is something that we should document.

tay1orjones:
Please do not merge until https://github.com/IBM/carbon-addons-iot-react/issues/1062 is resolved.

docs/guides/versioning.md
tay1orjones: @stuckless I'm not opposed, but Carbon doesn't consider this a breaking change. So if we pull in a new minor version of Carbon and the DOM node for an ID has changed, we would be forced to issue a new major/breaking version.

If we kept this as a minor change, we could mitigate the problem by working to ensure the DOM nodes for IDs are included in a snapshot, so we can provide accurate release/migration notes for when these things change. Thoughts?

docs/guides/versioning.md
tay1orjones: @stuckless Similar to IDs on DOM nodes, would it be more helpful if we could provide release notes listing these changes? A potential stretch goal could be to include codemods/scripts to assist in automating the migration.

docs/guides/testing.md
```diff
@@ -0,0 +1,50 @@
...
+ExportedComponent.defaultProps = {
+  testID: 'ExportedComponent',
+};
+```
+
+Internal components will utilize any available id props to attempt to provide a unique value.
+
+```jsx
+function InternalTableCell({ tableId, columnId, rowId }) {
+  return <td data-testid={`${tableId}-${columnId}-${rowId}`}>Example</td>;
```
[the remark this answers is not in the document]
tay1orjones: @stuckless Yes `ComponentName` would be a good addition. I'll update this example.

Should we attempt to provide a way to partially customize the `testid` via a prop? Maybe a cheap way would be to insert the `className`?

```diff
- data-testid={`${tableId}-${columnId}-${rowId}`}
+ data-testid={`${componentName}-${tableId}-${columnId}-${rowId}-${className}`}
```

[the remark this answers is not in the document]
stuckless: The problem with using className is you might inadvertently change the ID by harmlessly adding/removing the classname.  Not sure className is needed unless you are thinking of a specific use case.

docs/guides/versioning.md
stuckless: Yeah, I'm just trying to show the difference in the meaning of "breaking" when it comes to APIs vs UI LIbraries.   Carbon (and yourself) really are only looking at the API side of things, but there are lots of things that break an application from the UI side.  And moving the ID to a new element would be one of them. (ie, the app side has a test selector based off the id of the component and now all the tests fail because the id moved to a different location)   Now, if this is a case where there is an existing data-test-id then I agree, we should be using the data-test-id and moving the id should be minor.   But the same logic would apply to moving a data-test-id to a new element, and that would be a breaking change.

docs/guides/versioning.md
stuckless: This is where semantic versioning works well for a library.  I know that if move to version 1.1 to version 1.2 or 1.1.1 there are no breaking changes.  My code will run just fine, and you just added new features or fixed a bug.   But in a UI component library you would need to knwo the same thing, and if you are introducing changes that will break testing, then that is a breaking change.  If as a consumer I have to assume that 1.1.1 and 2.0 needs the same amount of regression since while you having broken any APIs you may have broken testing.  So in planning an upgrade you always have to plan for the worse, since the semantic versioning means very little if you can't trust it.  One of the purposes of semantic versioning is to help a downstream consumer identify the risk involved in upgrading to a newer version of the library.

docs/guides/versioning.md
tay1orjones: 👍 I've made a note of this in #1024

docs/guides/versioning.md
tay1orjones: 👍 I've added a new section for `The DOM node that an `data-testid` corresponds to is changed` describing this as a breaking change.

tay1orjones:
I've moved/added both `Internal constants and helpers` as well as `New features should be off by default` to the `Code` style guide section of #1024

PR #1020, opened 2020-03-18
fix(breadcrumb): disable hasOverflow functionality when ResizeObserver not supported

### LINKED ISSUE #1017: [Breadcrumb] fails to render in browsers without ResizeObserver support
### Describe the bug

The use-resize-observer hook does not check for ResizeObserver support, so is failing to render Breadcrumb stories even when the `hasOverflow` feature is not enabled. Also the `useResizeObserver` story title should mirror the prop name `hasOverflow`.

### To Reproduce

1. In Safari go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-breadcrumb--default

### Expected behavior

This story should render even without ResizeObserver support as `hasOverflow` is false.

### Environment/versions:

- OS: MacOS
- Browser: safari
- `carbon-addons-iot-react` version: 2.57.3

### Additional context

image

tay1orjones:
Closes #1017

**Summary**

- Breadcrumb stories weren't rendering at all in Safari because we don't polyfill ResizeObserver and we also weren't disabling `hasOverflow` in environments with no ResizeObserver support

- Conditionally assign the ref used for `hasOverflow` based on ResizeObserver support
- In documentation, refactor usage of `useResizeObserver` prop to `hasOverflow`
- Add dev warning for no ResizeObserver support when using `hasOverflow`
- Add test, Update snapshot

- Breadcrumb stories don't work in Safari here (existing previous build): 
  - https://deploy-preview-1008--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-breadcrumb--default
- Whereas now this should work in Safari (new build of this PR):
  -  https://deploy-preview-1020--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-breadcrumb--default

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.jsx
+++ b/src/components/Breadcrumb/Breadcrumb.jsx
@@ -7,6 +7,7 @@ import { Breadcrumb as CarbonBreadcrumb } from 'carbon-components-react';
 import warning from 'warning';
 
 import { OverflowMenuItem, OverflowMenu } from '../../index';
+import { browserSupports } from '../../utils/componentUtilityFunctions';
 
 const propTypes = {
   /**
@@ -52,6 +53,7 @@ const Breadcrumb = ({ children, className, hasOverflow, ...other }) => {
   const [afterOverflowItems, setAfterOverflowItems] = useState(childrenItems.slice(1));
   const [prevChildren, setPrevChildren] = useState([]);
 
+  const browserSupportsResizeObserver = browserSupports('ResizeObserver');
   useResizeObserver({
     useDefaults: false,
     ref: breadcrumbRef,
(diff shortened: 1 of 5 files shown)


src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -98,7 +109,7 @@ const Breadcrumb = ({ children, className, hasOverflow, ...other }) => {
       className={classnames('breadcrumb--container', {
         'breadcrumb--container__overflowfull': afterOverflowItems.length === 1,
       })}
-      ref={breadcrumbRef}
+      ref={typeof ResizeObserver !== 'undefined' ? breadcrumbRef : null}
```
bjornalm: nit: We check this twice now, perhaps put in a function and call it like `browserSupports(ResizeObserver)` to make it  clear that we are dealing with feature detection.

tay1orjones: 👍 Great suggestion, done. 

I took a trip down memory lane to see if we could/should use Modernizr but they don't support `esm` and also don't have a feature detection check for `ResizeObserver`. The function is extensible for future detection.

bjornalm: Sweet!

PR #1022, opened 2020-03-19
fix(tablecard): thresholds render regardless of order

### LINKED ISSUE #1021: [TableCard] Custom threshold icon displays beside the wrong column in Table views
### Describe the bug

When definining multiple thresholds and are not order in accordance with the column headers ordering, they render in the incorrect columsn, with the incorrect headers.

Internal issue: https://github.ibm.com/wiotp/monitoring-dashboard/issues/873

### Expected behavior

Threshold columns should render in the correct positions with the correct labels regardless of order

sstone2423:
Closes #1021 

**Summary**

- Threshold columns were being render based off the `uniqueThreshold` array's index, rather than the values within. This meant that if the thresholds were defined in the wrong order, the columns would render in the incorrect position which was extremely fragile

**With Bug**
Screen Shot 2020-03-19 at 5 43 25 PM

**With Fix**
image

- Changed `generateThresholdColumn` to return based on dataSourceId index
- Added more comments and documentation to story and code
- Re-ordered the thresholds in the story to show that it can be done
- Added a test for this bug

1. Go to https://deploy-preview-1022--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-tablecard--table-with-thresholds-precision-and-expanded-rows
2. See that the columns are correctly rendered. ('Count Severity' and 'Pressure Sev' column exists, but 'Pressure Severity' column does not exist)

### DIFF
--- a/src/components/TableCard/TableCard.jsx
+++ b/src/components/TableCard/TableCard.jsx
@@ -365,33 +365,44 @@ const TableCard = ({
   // filter to get the indexes for each one
   const columnsUpdated = cloneDeep(columns);
 
-  const generateThresholdColumn = (columnId, index) => ({
-    id: `iconColumn-${columnId}`,
-    label: uniqueThresholds[index].label
-      ? uniqueThresholds[index].label
-      : `${capitalize(columnId)} ${strings.severityLabel}`,
-    width: uniqueThresholds[index].width,
-    isSortable: true,
-    renderDataFunction: renderThresholdIcon,
-    priority: 1,
-    filter: {
-      placeholderText: strings.selectSeverityPlaceholder,
-      options: [
-        {
-          id: 1,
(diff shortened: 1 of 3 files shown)


tay1orjones:
Based on the testing guidance you gave - I'm seeing the pressure severity column exists in the `tablecard--table-with-thresholds-precision-and-expanded-rows` story

image

sstone2423:
@tay1orjones When the bug is present, the column `Count Severity` would actually be called `Pressure Severity` and in addition, the `Pressure Sev` column would still be present. I'll rename that custom column to make it more obvious.

The issue was that it was taking the attributes from the Pressure column, but applying them to the Count Severity column instead. That sounds confusing as I write it. I'll try to get a screenshot of the bug existing in this same story with the old code

PR #1025, opened 2020-03-19
fix(table): column resizing breaks row action button container

### LINKED ISSUE #1004: [Bug] Column resizing breaks row action button container
Tables with column re-sizing seem to ignore the in-line actions container, causing overflow
https://pages.github.ibm.com/maximo-app-framework/graphite/reference-app/#/tables
Screen Shot 2020-03-12 at 3 39 04 PM

bjornalm:
Closes #1004

**Summary**
Tables with column re-sizing gets table-layout:fix by default which means the <td> cell content is not considered for the width of the columns. Since the 'row actions' do not wrap and the <th> of the column is empty, the content is overflowing into the column to the left.

image

Already existing fix
The table-layout can be made dynamic setting the prop `option: { useAutoTableLayoutForResize: true }`

MIN WIDTH FIX:

1. Allowing a min-width property for the row action column so we can force it to never shrink beyond that and hence never overflow (given proper minWidth).
2. Adding a min-width default of 40 px
3. Adding and updating unit tests

TRANSPARENT-TO-BACKGROUND GRADIENT FIX:
image
1.  Added a container with gradient background that matches the hover color behind the buttons, to avoid the visual overlapping.
2. The background of this container is only visible on row hover.
3. Added a line-height to the RowActionError since the height of the background is controlled by the containers content

### Note!
This fix will show a background container with the height of its content, i.e. the action buttons it contains. If the columns to the left have more than 2 lines of text in their table cells the result will be less visually appealing since its content will appear above and below this gradient background container. 
image

MIN WIDTH FIX:
Find/create a table with both resizing and row actions enabled. Set a row action width as in the example below:
```
...
view={{
   table: {
      rowActionColumn: { minWidth: '200px' },
},
...
```
Make sure the row action will not grow smaller than the width you set, but that it can grow larger if needed.

TRANSPARENT-TO-BACKGROUND GRADIENT FIX:
Find/create a table with both resizing and row actions enabled.  If needed, resize the table so that the row actions don't fit their column. Hover a row with actions to make sure you get the desired effect shown above. You can use this story
https://deploy-preview-1025--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-table--stateful-example-with-expansion

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -116,6 +116,7 @@ const propTypes = {
       ),
       /** what is the current state of the row actions */
       rowActions: RowActionsStatePropTypes,
+      rowActionColumn: PropTypes.shape({ minWidth: PropTypes.string }),
       expandedIds: PropTypes.arrayOf(PropTypes.string),
       emptyState: EmptyStatePropTypes,
       loadingState: PropTypes.shape({
@@ -202,6 +203,7 @@ export const defaultProps = baseProps => ({
       isSelectAllSelected: false,
       selectedIds: [],
       rowActions: [],
+      rowActionColumn: { minWidth: '40px' },
       sort: {},
       ordering: baseProps.columns && baseProps.columns.map(i => ({ columnId: i.id })),
       loadingState: {
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -116,7 +116,6 @@ const propTypes = {
       ),
       /** what is the current state of the row actions */
       rowActions: RowActionsStatePropTypes,
-      rowActionColumn: PropTypes.shape({ minWidth: PropTypes.string }),
       expandedIds: PropTypes.arrayOf(PropTypes.string),
       emptyState: EmptyStatePropTypes,
       loadingState: PropTypes.shape({
@@ -203,7 +202,6 @@ export const defaultProps = baseProps => ({
       isSelectAllSelected: false,
       selectedIds: [],
       rowActions: [],
-      rowActionColumn: { minWidth: '40px' },
       sort: {},
       ordering: baseProps.columns && baseProps.columns.map(i => ({ columnId: i.id })),
       loadingState: {
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -341,6 +341,7 @@ const Table = props => {
       options.hasColumnSelection ||
       actions.toolbar.onDownloadCSV ||
       secondaryTitle ||
+      view.toolbar.customToolbarContent ||
       tooltip ? (
         <TableToolbar
           tableId={id}
--- a/src/components/Breadcrumb/Breadcrumb.jsx
+++ b/src/components/Breadcrumb/Breadcrumb.jsx
@@ -4,8 +4,10 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { OverflowMenuHorizontal20 } from '@carbon/icons-react';
 import { Breadcrumb as CarbonBreadcrumb } from 'carbon-components-react';
+import warning from 'warning';
 
 import { OverflowMenuItem, OverflowMenu } from '../../index';
+import { browserSupports } from '../../utils/componentUtilityFunctions';
 
 const propTypes = {
   /**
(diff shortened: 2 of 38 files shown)


bjornalm:
See discussions in internal repo
https://github.ibm.com/Watson-IoT/pal-tracking/issues/321

src/components/Table/Table.jsx
```diff
@@ -116,6 +116,7 @@ const propTypes = {
       ),
       /** what is the current state of the row actions */
       rowActions: RowActionsStatePropTypes,
+      rowActionColumn: PropTypes.shape({ minWidth: PropTypes.string }),
```
tay1orjones:
```suggestion

```

src/components/Table/Table.jsx
```diff
@@ -202,6 +203,7 @@ export const defaultProps = baseProps => ({
       isSelectAllSelected: false,
       selectedIds: [],
       rowActions: [],
+      rowActionColumn: { minWidth: '40px' },
```
tay1orjones:
```suggestion
```

src/components/Table/Table.jsx
bjornalm: fixed

src/components/Table/Table.jsx
bjornalm: fixed

bjornalm:
> > This fix will show a background container with the height of its content, i.e. the action buttons it contains. If the columns to the left have more than 2 lines of text in their table cells the result will be less visually appealing since its content will appear above and below this gradient background container.
> 
> To remedy this could we make the button height flexible to accommodate rows that may not be the default height? I suspect @lukefirth and @JordanWSmith15 would prefer the button background overlay to not look the way it does in the screenshot.

Not sure I understand what you mean by "make the button height flexible". As I understand it, we would probably like the action button container (with the gradient  background) to fill height of the table cell. It seems the only CSS solution available requires the height of the table to have a value, e.g. '0' or '100%'. This works (thanks @enricoberti), but since we don't actually understand why the table-height value is needed or if it will have any other side effects I'm hesitant to adding it.

Another option would be a JS based where we measure the height of each action cell, possibly only on hover, and use that value for the height. None of these two options felt great so that I why I went for the visual tradeoff. Feedback based on this comment is much appreciated.

bjornalm:
Based on PR feedback:

1. Removed the prop rowActionColumn.minWidth
2. Removed related unit tests including the test for default min-width since it is now CSS based
3. Kept the class iot--table-header-row-action-column for app CSS modifications and added the default min-width of 40px there

tay1orjones:
> As I understand it, we would probably like the action button container (with the gradient background) to fill height of the table cell.

Yes, that's what I meant - sorry for the confusion.

> It seems the only CSS solution available requires the height of the table to have a value, e.g. '0' or '100%'. This works (thanks @enricoberti), but since we don't actually understand why the table-height value is needed or if it will have any other side effects I'm hesitant to adding it.

If the other stories have no adverse effects, I think it's safe to add it. I spoke with @lukefirth today and while he'd prefer to have the background gradient fill the table cell height, but if it ends up not being feasible due to technical limitations, we can merge this as-is.

src/components/Table/Table.jsx
tay1orjones: I'm still seeing these present in the diff?

bjornalm: Hmm.. not sure how that got back in there. Should be gone now. Thanks :-)

bjornalm:
> > It seems the only CSS solution available requires the height of the table to have a value, e.g. '0' or '100%'. This works (thanks @enricoberti), but since we don't actually understand why the table-height value is needed or if it will have any other side effects I'm hesitant to adding it.
> 
> If the other stories have no adverse effects, I think it's safe to add it. I spoke with @lukefirth today and while he'd prefer to have the background gradient fill the table cell height, but if it ends up not being feasible due to technical limitations, we can merge this as-is.

Ok, I will try it and then check all stories containing a table

bjornalm:
> If I resize the table to get this effect, them sometimes the table goes into a horizontal scroll. When that happens the hover actions is confusing, since I see part of a button, but I don't get the option click other buttons, or the overflow menu. not sure it should work this way. Should the row actions columns be a "locked" column? are we ok with this behaviour as is?

I assume this is more of a question for @lukefirth . If it was up to me I would recommend that additional improvements to this new action column feature would go into another issue with proper design specs and I could pick it up again from there.

lukefirth:
Agreed - let's keep deliveries small and numerous so we can seperate concerns.

What we have in this works for the exact problem listed, and for any tables with a fixed width (no horizontal scroll).

I'd recommend we open another issue specifically for how this should work for horizontally scrolling tables, including the 'frozen column' for actions. There we can also discuss other expectations for that, such as if the first column, or at least the checkboxes (when present) should be frozen too, etc etc.

tay1orjones:
@lukefirth @bjornalm @stuckless I've created https://github.com/IBM/carbon-addons-iot-react/issues/1043 for the additional items

PR #1028, opened 2020-03-20
fix(value-card-wrap): wrap long strings to two lines

JoelArmendariz:
**Summary**

This PR references these issues:

https://github.ibm.com/wiotp/monitoring-dashboard/issues/693

https://github.com/IBM/carbon-addons-iot-react/issues/1027

Long strings in ValueCards were not very accessible with only one line of text before the ellipsis. This PR allows long strings to wrap to a max of 2 lines.

Added styling to ValueRenderer
Created new ValueCard story with a long string

Navigate to ValueCard 'small / long string' and see the text wrap to two lines before the ellipsis.

Before:
Screen Shot 2020-03-20 at 11 57 31 AM
 
After:
Screen Shot 2020-03-20 at 11 57 05 AM

### DIFF
--- a/src/components/ValueCard/ValueCard.story.jsx
+++ b/src/components/ValueCard/ValueCard.story.jsx
@@ -160,6 +160,31 @@ storiesOf('Watson IoT/ValueCard', module)
       </div>
     );
   })
+  .add('small / long string', () => {
+    const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.SMALL);
+    return (
+      <div style={{ width: text('cardWidth', `${getCardMinSize('lg', size).x}px`), margin: 20 }}>
+        <ValueCard
+          title="Tagpath"
+          id="facilitycard"
+          content={{
+            attributes: [
+              {
+                label: 'Tagpath',
+                dataSourceId: 'footTraffic',
+              },
+            ],
+          }}
+          breakpoint="lg"
+          size={size}
+          values={{
+            footTraffic:
+              'rutherford/rooms/northadd/ah2/ft_supflow/eurutherford/rooms/northadd/ah2/ft_supflow/eu',
+          }}
+        />
+      </div>
+    );
+  })
   .add('small / trend down', () => {
     const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.SMALL);
     return (
--- a/src/components/ValueCard/ValueCard.story.jsx
+++ b/src/components/ValueCard/ValueCard.story.jsx
@@ -160,31 +160,42 @@ storiesOf('Watson IoT/ValueCard', module)
       </div>
     );
   })
-  .add('small / long string', () => {
-    const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.SMALL);
-    return (
-      <div style={{ width: text('cardWidth', `${getCardMinSize('lg', size).x}px`), margin: 20 }}>
-        <ValueCard
-          title="Tagpath"
-          id="facilitycard"
-          content={{
-            attributes: [
-              {
-                label: 'Tagpath',
-                dataSourceId: 'footTraffic',
-              },
-            ],
-          }}
-          breakpoint="lg"
-          size={size}
-          values={{
-            footTraffic:
-              'rutherford/rooms/northadd/ah2/ft_supflow/eurutherford/rooms/northadd/ah2/ft_supflow/eu',
-          }}
-        />
-      </div>
-    );
-  })
+  .add(
+    'small / long string',
+    () => {
+      const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.SMALL);
+      return (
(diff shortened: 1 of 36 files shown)


src/components/ValueCard/ValueCard.story.jsx
```diff
@@ -160,6 +160,31 @@ storiesOf('Watson IoT/ValueCard', module)
...
+            attributes: [
+              {
+                label: 'Tagpath',
+                dataSourceId: 'footTraffic',
+              },
+            ],
+          }}
+          breakpoint="lg"
+          size={size}
+          values={{
```
sstone2423: Could we make this a knob so users could try can more immediately see the difference with a shorter string?
There's an example in the above stories

src/components/ValueCard/ValueCard.story.jsx
```diff
@@ -160,6 +160,31 @@ storiesOf('Watson IoT/ValueCard', module)
...
+            footTraffic:
+              'rutherford/rooms/northadd/ah2/ft_supflow/eurutherford/rooms/northadd/ah2/ft_supflow/eu',
+          }}
+        />
+      </div>
+    );
+  })
```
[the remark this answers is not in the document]
JoelArmendariz: Added.

src/components/ValueCard/ValueCard.story.jsx
JoelArmendariz: Added.

bjornalm:
Nice work, I implemented something similar for the ValueCard's data state earlier and created a sass mixin with a fallback. Perhaps you could reuse that if you move the style over to scss.
https://github.com/IBM/carbon-addons-iot-react/blob/master/src/globals/_mixins.scss

JoelArmendariz:
@bjornalm thanks for the info! That looks great. I do think it would be beneficial to use that mixin here, although this specific implementation of this fix works on all browsers that I have tested. (all 'relevant' browsers, ie. Chrome, Firefox, Safari, Edge..) I tinkered a little with applying this mixin multiline-text-overflow and was having trouble getting the same result.
Seeing as though the fallback is not necessarily needed in this case, would it still be worth extracting to scss?

bjornalm:
> @bjornalm thanks for the info! That looks great. I do think it would be beneficial to use that mixin here, although this specific implementation of this fix works on all browsers that I have tested. (all 'relevant' browsers, ie. Chrome, Firefox, Safari, Edge..) I tinkered a little with applying this mixin multiline-text-overflow and was having trouble getting the same result.
> Seeing as though the fallback is not necessarily needed in this case, would it still be worth extracting to scss?

Totally up to you :-) I just wanted to let you know the mixin existed.

JoelArmendariz:
> > @bjornalm thanks for the info! That looks great. I do think it would be beneficial to use that mixin here, although this specific implementation of this fix works on all browsers that I have tested. (all 'relevant' browsers, ie. Chrome, Firefox, Safari, Edge..) I tinkered a little with applying this mixin multiline-text-overflow and was having trouble getting the same result.
> > Seeing as though the fallback is not necessarily needed in this case, would it still be worth extracting to scss?
> 
> Totally up to you :-) I just wanted to let you know the mixin existed. Just out of curiosity, what was failing with the mixin?

The words were not breaking to wrap like they were supposed to. For some reason the word that was nearest the wrap break on the second line would just disappear, despite the use of `overflow-wrap: break-word`. Didn't do a whole lot of digging into it, though.

scottdickerson:
@JoelArmendariz do you know why if I expand the card width a little to 400 px there's all that empty white space after the wrap?  Can we make sure it goes full-width before wrap in this case?
image

JoelArmendariz:
> See strange 'early wrapping' in the deployment preview

Fixed.

scottdickerson:
Can we do anything to improve the case where we do have units?  Seems like a lot of wasted space to the right of the units causing an unnecessary truncation.  Also, shouldn't it go two lines before truncating?

image

JoelArmendariz:
> Can we do anything to improve the case where we do have units? Seems like a lot of wasted space to the right of the units causing an unnecessary truncation. Also, shouldn't it go two lines before truncating?
> 
> image

Can I propose that we only let the values that have units wrap to two lines at the same breakpoint that the units appear to the right of the value instead of under? In other words, in the case that the value has units, lets truncate on one line, until the card gets wide enough to trigger the previously implemented breakpoint that moves the unit onto the same row as the value.
Namely: at 133px width in the `small / wrapping` story. The value wrapping to two lines on the card when it's thinner than that looks bad, and it doesn't seem necessary to try and fit just a few more characters when the user should probably just make the card a bit wider anyway.

JoelArmendariz:
Looks like some changes that I didn't make just got pushed into this branch somehow..

JoelArmendariz:
Looks like some rogue commits from various authors made it into this PR. Closing in favor of a fresh one at: https://github.com/IBM/carbon-addons-iot-react/pull/1042

PR #1033, opened 2020-03-23
[TableCard] Thresholds labels are customizable and optionally hidden

### LINKED ISSUE #1032: [TableCard] Threshold labels should be customizable or hidden
### Describe the bug

Threshold labels should have customizable strings and the ability to hide.
Currently there is no way to do either.

### To Reproduce

1. Open a table with threshold labels in a development environment
2. Change the `label` property to an empty string or other custom string
3. See that only the threshold column header changes

### Expected behavior

Expect to either change the label or hide it - this is the label within the rows and not the column header label

### Environment/versions:

- OS: MacOS
- Browser: chrome

### Additional context

Current:

Screen Shot 2020-03-19 at 12 19 08 PM

sstone2423:
Closes #1032 

**Summary**

- There was previously no way to customize the label strings or hide the label strings completely. This PR enables both of those use-cases

Screen Shot 2020-03-23 at 1 19 58 PM

- added `showSeverityLabel` boolean prop that optionally hides the label text
- added `severityLabel` string prop that allows for custom strings to be rendered in place of the default *Critical, Moderate, Low* labels
- added tests for each use case
- added knobs for both props to all threshold stories
- added universal knobs to all TableCard stories
- removed 2 redundant stories
- converted `ThresholdIcons` styled-component to scss

- Go to each threshold story and use the knobs to verify that both use-cases work

### DIFF
--- a/src/components/TableCard/TableCard.jsx
+++ b/src/components/TableCard/TableCard.jsx
@@ -332,7 +332,8 @@ const TableCard = ({
         } ${matchingThresholdValue.value}`}
         {...matchingThresholdValue}
         strings={strings}
-        showLabel={matchingThresholdValue.showLabel}
+        showSeverityLabel={matchingThresholdValue.showSeverityLabel}
+        severityLabel={matchingThresholdValue.severityLabel}
         renderIconByName={others.renderIconByName}
       />
     ) : null;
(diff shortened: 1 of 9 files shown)


swallacertp:
Thanks - will look for this in beta (Monitor-Demo) and test in my dashboards.

PR #1034, opened 2020-03-23
fix(table): column ordering slow

### LINKED ISSUE #1006: [table] column ordering slow
### Describe the bug
The table updates “realtime” while the user is dragging a column and it can cause sluggish and buggy behaviour because of all the processing that is going on per render. See the attached gif for a table with 100 rows.

column-reorder-issue-1

It is also possible to end up in a state where the columns position keep jumping back and fourth.
bug2

With the additional complexity/processing needed for table column resize this behaviour gets worse when `hasResize` is true. 

### To Reproduce
1. Choose a table story with `hasColumnSelection:true`
2. Click the column reorder button
3. Drag a column slowly across the as shown in the GIFs

### Expected behavior
The best in my opinion would be to update the table on drop instead of on drag when rearranging the columns.

### Environment/versions:

- OS: MacOS
- Browser: chrome

bjornalm:
Closes #1006

**Summary**
The table updates “realtime” while the user is dragging a column and it can cause sluggish and buggy behaviour because of all the processing that is going on per render.

- Update the columns on drop instead of hover
- Updated unit tests

- Try to reorder columns and verify that nothing changes (visually nor onChangeOrdering) until the column is dropped.

### DIFF
--- a/src/components/Table/TableHead/ColumnHeaderRow/ColumnHeaderRow.test.jsx
+++ b/src/components/Table/TableHead/ColumnHeaderRow/ColumnHeaderRow.test.jsx
@@ -44,7 +44,7 @@ describe('TableHead', () => {
     const dropTarget = wrapper.find("DropTarget(ColumnHeaderSelect)[columnId='col2']").instance();
     backend.simulateBeginDrag([dragSource.getHandlerId()]);
     backend.simulateHover([dropTarget.getHandlerId()]);
-    backend.simulateEndDrag();
+    backend.simulateDrop();
     expect(commonTableHeadProps.onChangeOrdering).toHaveBeenCalled();
   });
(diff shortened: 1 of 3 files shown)


tay1orjones:
Is there any way we could improve the usability for knowing where to drag an item? The drop targets can be a bit difficult to find. 

I have trouble trying to drag this to the end, then drag it back to it's original location but it goes to the very first column spot:
dragdrop

bjornalm:
> Is there any way we could improve the usability for knowing where to drag an item? The drop targets can be a bit difficult to find.
> 
> I have trouble trying to drag this to the end, then drag it back to it's original location but it goes to the very first column spot:
> dragdrop

By design this is a bit confusing, since the drop zones are not before or after other column headers, the drop zones are the headers themselves. However I did add a drop indication for the column headers  now.

image

PR #1036, opened 2020-03-24
fix(treeshaking): fix treeshaking and add esm and umd versions

### LINKED ISSUE #993: Is tree-shaking working?
I see my issue https://github.com/IBM/carbon-addons-iot-react/issues/793 is already closed. I tried the steps mentioned in it with latest version 2.55.2, but I don't think tree-shaking is working. Then I tried 'npx create-iot-react-app' to create a hello world app, and run 'yarn build', I got the following after running source map analyzer against the biggest bundle files generated.

image

In App.js, only 5 components are imported, but the bundled index.js is 1.34MB. So I doubt if tree-shaking is really working, or I need apply any additional steps in the created simple app. 

```
import {
  HeaderContainer,
  Header,
  SideNav,
  Button,
  PageTitleBar
} from "carbon-addons-iot-react";
```

Per my understanding, to support tree-shaking, carbon-addons-iot-react should be bundled in es module, i.e. there should be "module" entry in package.json, and "sideEffects:false" as well. But I didn't find them.

And I see lots of unused components are bundled from carbon-component-react even they are not used in App.js. If I directly import components from carbon-component-react, tree-shaking works well.  

I really need to get the answer since after importing carbon-addons-iot-react, our bundle becomes extremely bigger than before and I am still only using Stateful component.  Thanks!

### LINKED ISSUE #1011: Tree shaking is not working in the PAL addons repo
Components that are not used by consuming applications are being included in the build. We need to optimized our package to enable this for consumers.

Issue is complete when:
- [x] we are able to build a es, cjs, and umd version of the project.
- [x] sass files are copied over along with main sass entry point
- [ ] we have a compiled css file
- [ ] documentation is updated to reflect this new package structure.

davidicus:
Closes #993 closes #1011

**Summary**

Changes to our build to allow for es and umd versions and to allow for treeshaking in consuming applications. 

- items_here

yarn build should output a lib, es, umd, and scss folder.


scottdickerson:
@davidicus so the sideEffects false really was required for treeshaking to work?  Sorry for the bad info about that?

tay1orjones:
@scottdickerson [Yep, here's an example from webpack](https://github.com/webpack/webpack/tree/master/examples/side-effects)

From the link: (I subbed the example package name with ours)
> The `"sideEffects": false` flag in `carbon-addons-iot-react` `package.json` indicates that the package's modules have no side effects (on evaluation) and only expose exports. This allows tools like webpack to optimize re-exports. In the case `import { a, b } from "carbon-addons-iot-react"` is rewritten to `import { a } from "carbon-addons-iot-react/a"; import { b } from "carbon-addons-iot-react/b"`.

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -61,36 +23,6 @@ Map {
       "kind": "primary",
       "loading": false,
     },
-    "propTypes": Object {
```
stuckless: I'm seeing most (if not all) propTypes removed from the api snapshot.  Is this intentional?  I thought the purpose of this file was to compare the APIs from build to build, release to release, etc.

stuckless:
I have a couple questions regarding this, but these questions should not impact the approval of the PR.
1. Have you validated what the savings are?  ie, have you tried creating a single app and include, say, Button only in your app, and compared what the before/after sizes are for the entire build/ output tree.  ie, I'd be curious to know a concrete savings number.
2. What is required from a downstream consumer to enable this?
3. I assume if a downstream consumer does nothing, then there is no impact to their build.
4. The webpack example is pretty lean with index.js and files all in the same directory.  Does this work if your src repo has sub directories...ie, does webpack convert `import {a} from 'module'` to something like `import {a} from 'module/component/a'` if `a` is within some other directory?
5. Does enabling this speed up the consuming build or slow it down... or no difference?

babel.config.js
```diff
@@ -17,32 +18,23 @@ module.exports = function generateConfig(api) {
     ],
     ignore: ['__mocks__'],
     plugins: [
+      ['transform-react-remove-prop-types', { removeImport: true }],
```
tay1orjones: We need to wrap the prop types, not strip them. If we strip them, applications consuming this library won't get prop type warnings when using our components.

We should use [`mode: 'unsafe-wrap'`](https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types#mode):
> The wrap modes are targeting React libraries like material-ui or react-native-web. They are not intended to be used by application authors.

[See the material-ui babel config here](https://github.com/mui-org/material-ui/blob/master/babel.config.js#L34-L40)

```suggestion
      ['transform-react-remove-prop-types', { removeImport: true, mode: 'unsafe-wrap', }],
```

babel.config.js
```diff
@@ -9,6 +9,7 @@ module.exports = function generateConfig(api) {
             browsers: ['last 2 version', 'ie >= 11'],
           },
           useBuiltIns: 'usage',
+          // modules: false,
```
tay1orjones:
```suggestion

```

package.json
```diff
@@ -1,22 +1,27 @@
...
     "url": "https://github.com/IBM/carbon-addons-iot-react/issues"
   },
   "homepage": "https://carbon-addons-iot-react.com",
   "scripts": {
-    "build": "yarn build:pre && rollup -c",
-    "build:pre": "rimraf lib storybook-static",
+    "build": "yarn build:pre && yarn build:main",
+    "build:pre": "rimraf lib es umd scss storybook-static",
     "build:storybook": "NODE_OPTIONS='--max-old-space-size=16384' build-storybook -s public/production",
+    "build:main": "rollup -c ",
```
tay1orjones: Minor nit: move `build:main` up two lines to keep them alphabetical. 
```json
    "build": "yarn build:pre && yarn build:main",
    "build:main": "rollup -c ",
    "build:pre": "rimraf lib es umd scss storybook-static",
    "build:storybook": "NODE_OPTIONS='--max-old-space-size=16384' build-storybook -s public/production",
```

src/index.js
```diff
@@ -4,7 +4,7 @@ import sizeMe from 'react-sizeme';
 sizeMe.noPlaceholders = true;
 
 // Styles
-export styles from './styles.scss';
+// export styles from './styles.scss';
```
tay1orjones:
```suggestion

```

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
tay1orjones: Yes this is intended, but we shouldn't be stripping them completely. They should be wrapped in a conditional, [see my review](https://github.com/IBM/carbon-addons-iot-react/pull/1036#discussion_r398787376) so consuming applications still get proptype warnings.

There is [a bug with Terser](https://github.com/terser/terser/issues/293), with [a decent workaround synopsis here](https://github.com/alex996/react-css-spinners/issues/1#issuecomment-513634336), where tree shaking breaks when 2 or more static properties are assigned to a function. Our components typically have `defaultProps` and `propTypes`. [Carbon ran into the same issue recently](https://github.com/carbon-design-system/carbon/issues/5442) causing them to revert an update to their build.

tay1orjones:
@stuckless I'll step in and answer as I've occasionally been pairing with @davidicus on this for the past week or so.

> 1. Have you validated what the savings are?  ie, have you tried creating a single app and include, say, Button only in your app, and compared what the before/after sizes are for the entire build/ output tree.  ie, I'd be curious to know a concrete savings number.

Yes, originally the entire library was included no matter which components were imported. About ~2.5mb as I recall. After testing in a `create-iot-react-app` @davidicus saw this drop to ~200k with the bare boilerplate generated by `create-iot-react-app` (mainly just some UIShell components). I think he has some screencaps he could share to show exacts. Bundle size savings for applications will depend on how many components are imported/used.

> 2. What is required from a downstream consumer to enable this?

[To enable with Webpack as your module bundler:](https://webpack.js.org/guides/tree-shaking) (below from the link)
- Use ES2015 module syntax (i.e. import and export).
- Ensure no compilers transform your ES2015 module syntax into CommonJS modules (this is the default behavior of the popular Babel preset @babel/preset-env - see the documentation for more details).
- Add a "sideEffects" property to your project's package.json file.
- Use the production mode configuration option to enable various optimizations including minification and tree shaking.

> 3. I assume if a downstream consumer does nothing, then there is no impact to their build.

Yep, shouldn't be any change. Do note that webpack will likely automatically begin importing the esm version by default instead of the lib version because of the addition of `"module": "es/src/index.js"` in `package.json`. [A bit more info on this from webpack here](https://webpack.js.org/guides/author-libraries/#final-steps)

> 4. The webpack example is pretty lean with index.js and files all in the same directory.  Does this work if your src repo has sub directories...ie, does webpack convert `import {a} from 'module'` to something like `import {a} from 'module/component/a'` if `a` is within some other directory?

Yes

> 5. Does enabling this speed up the consuming build or slow it down... or no difference?

Likely no difference to your app's build times, but If your app only uses a small subset of components, tree shaking will provide smaller bundle(s) delivered in production environments.

stuckless:
@tay1orjones thx... that great info.  btw, is Carbon doing this as well?

tay1orjones:
@stuckless Yes, [Carbon provides an esm build](https://github.com/carbon-design-system/carbon/blob/master/packages/react/package.json#L7) that supports tree shaking. 

The bug referenced in my comment above was when they tried to ship a single entrypoint generated by Rollup for their ESM and CJS bundles.

src/globals/_charts.scss
```diff
@@ -1,4 +1,9 @@
-@import '~@carbon/charts/styles.css';
+@import '~@carbon/charts/styles/styles.scss';
+@import './vars';
+
+.#{$iot-prefix}--card .#{$prefix}--chart-holder {
```
tay1orjones: This should be moved to the `_card.scss` partial, no?

davidicus:
closing in favor of #1045. Will address all comments in other PR.

src/index.js
davidicus: this was removed in the refactor and we are not compiling styles by using sass entry point.

package.json
davidicus: using same commands since using rollup for all versions.

babel.config.js
davidicus: this was a typo and is different in final solution .

babel.config.js
davidicus: this was used in new solution as well.

src/globals/_charts.scss
davidicus: agreed. Will address in other PR

PR #1038, opened 2020-03-24
Migrate TileCatalog to scss

JoelArmendariz:
**Summary**

Migrate TileCatalog to scss from styled components.

Created an scss file for TileCatalog
Removed StyledComonents from TileCatalog
Updated test to accommodate for new className.

See that TileCatalog is still styled as expected.

### DIFF
--- a/src/components/TileCatalog/TileCatalog.jsx
+++ b/src/components/TileCatalog/TileCatalog.jsx
@@ -92,7 +92,7 @@ const TileCatalog = ({
   const totalTiles = pagination && pagination.totalItems ? pagination.totalItems : 10;
 
   return (
-    <div className={classNames(`${iotPrefix}--tile-catalog`, className)}>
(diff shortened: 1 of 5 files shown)


src/components/TileCatalog/TileCatalog.jsx
```diff
@@ -128,8 +92,8 @@ const TileCatalog = ({
   const totalTiles = pagination && pagination.totalItems ? pagination.totalItems : 10;
 
   return (
-    <StyledContainerDiv className={className}>
-      <StyledCatalogHeader>
+    <div className={classNames(`${iotPrefix}--tile-catalog`, className)}>
```
sstone2423: Swap the ordering of these 2 as consumers should be able to override our class if needed

JoelArmendariz: Makes sense. Will do.

PR #1039, opened 2020-03-25
fix(table): column resizing doesn't work after column reordering

### LINKED ISSUE #1016: [table] column resizing doesn't work after column reordering
### Describe the bug
When you drag to re-order columns and then attempt to re-size, the table seems to "remember" where the columns originally were, and takes width from that area, which feels different than the intended UX.

See movie
https://ibm.ent.box.com/s/q36zlk7mfdc56xlh8par51mnan0jdlue/file/631476825899

### To Reproduce

1. Open a table that allows both column resizing and column reordering
2. Resize a column (works well)
3. Change the order of a column
4. Try to resize a column again

### Expected behavior
Resize should work as before reordering

### Environment/versions:
- Any OS
- Any browser

bjornalm:
Closes #1016

**Summary**
Solves two issues originally described in internal [312](https://github.ibm.com/Watson-IoT/pal-tracking/issues/312)  

1. Moved Columns
When you drag to re-order columns and then attempt to re-size, the table seems to "remember" where the columns originally were, and takes width from that area, which feels different than the intended UX.

2. Hidden columns
When you hide columns, the right-hand side of the table is still considered occupied, and you cannot drag the furthest-right column to fill the space.
(This is issue is more complicated and have corner cases not solved with this fix, they are tracked here [1019](https://github.com/IBM/carbon-addons-iot-react/issues/1019) )

- Updated table stories so that the two issues and fixes could be verified 
- Updated ColumnResize.jsx with improved naming and correct identification of affected column sibling
- Created `columnWidthUtilityFunctions.js` for dealing with the calculations of column widths. Previously the widths where measured instead of calculated after column toggle which meant that the row action column kept growing (point 2). Hopefully this strategy is also faster compared to the previous since it doesn't require an extra render.
- Removed the useEffect in TableHead since `onColumnResize(updatedColumns)` now can be called without measuring the new widths.
- Added px to the values exported by the onColumnResize (is that a breaking change?)
- Created unit tests for point 2

**Left Todo**
- Working on creating meaningful unit tests for point 1 and other resize functionality 

_Point 1_
1. Go to https://deploy-preview-1039--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-table--with-resize-hascolumnselection-and-initial-column-widths
2. Click the column selection button i the top right corner 
3. Rearrange the columns (this might be slow, there is another PR for that here [1034]( https://github.com/IBM/carbon-addons-iot-react/pull/1034) )
4. Resize the columns and make sure it works as expected
5. Also make sure that the emitted values for actions `onColumnResize` and `onChangeOrdering` are correct 

_Point 2_
1. Go to https://deploy-preview-1039--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize
2. Click the column selection button i the top right corner 
3. Hide a couple of columns and make sure that the action column doesn't grow in width
4. Show the hidden columns and verify that the action column width stays the same


src/components/Table/TableHead/TableHead.jsx
```diff
@@ -323,13 +327,15 @@ const TableHead = ({
       )}
       {activeBar === 'column' && (
         <ColumnHeaderRow
+          erRow
```
davidicus: is this a typo?

bjornalm: Yes, must have been a burglar. It is a thing, they break into people's homes and add typos everywhere.

bjornalm:
> First just wanted to say, this should be the gold star for table related stories/defects :) The acceptance criteria has a direct link to the story in storybook with instructions. Given how many stories there are in Table it's almost impossible to find one when you are looking for a specific story, so kudos for adding direct links.
> 
> That being said... there is a bug.
> 
> 1. Open https://deploy-preview-1039--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize
> 2. Open column resize/move
> 3. Don't click anything, but hover over React Node.  Notice how there is no right resize bar (this is neither right or wrong, just an observation, at this point)
> 4. Hide React Node
> 5. Hover over boolean column.  Notice that there is a right side slider.
> 6. Drag the right side slider to the right and release (it doesn't move, which is ok)
> 7. Drag the left side of the boolean slider to the left and release.
> 8. Notice the the boolean's right side slider is still visible and in the wrong place.

Thanks. Good catch! I've fixed the bug, now the last column will never have the resize handle on the right side. I've also added a unit test for the fix.

PR #1045, opened 2020-03-29
[treeshaking]: fix treeshaking and add esm and umd versions

### LINKED ISSUE #993: Is tree-shaking working?

### LINKED ISSUE #1011: Tree shaking is not working in the PAL addons repo

davidicus:
Closes #993 closes #1011

**Summary**

Changes to our build to allow for es and umd versions and to allow for treeshaking in consuming applications. 

- items_here

yarn build should output a lib, es, umd, and scss folder. 

Using the `create-iot-react-app` cli I was able to test the baseline build. Like the issue opened on this reports the original build was around `2.74 MB`. With tree shaking and dead code elimination this new build is now able to deliver the same app in under `314 KB` which is around an 88.5% savings.

Before tree shaking            |  After tree shaking
image  |  image


rollup.config.js
```diff
@@ -9,121 +9,205 @@ import copy from 'rollup-plugin-copy';
...
+//   'core-js/modules/es6.array.find',
+//   'core-js/modules/es6.array.fill',
+//   'core-js/modules/es6.array.find-index',
+//   'core-js/modules/es6.array.iterator',
+//   'core-js/modules/es7.array.includes',
+//   'core-js/modules/es6.regexp.replace',
+//   'core-js/modules/es6.regexp.split',
+//   'core-js/modules/es6.regexp.search',
+//   'core-js/modules/es6.regexp.to-string',
+// ];
```
tay1orjones:
```suggestion
```

src/globals/_charts.scss
```diff
@@ -1,4 +1,9 @@
-@import '~@carbon/charts/styles.css';
+@import '~@carbon/charts/styles/styles.scss';
+@import './vars';
+
+.#{$iot-prefix}--card .#{$prefix}--chart-holder {
```
tay1orjones: This should be moved to the `_card.scss` partial

src/index.js
```diff
@@ -3,9 +3,6 @@ import sizeMe from 'react-sizeme';
 
 sizeMe.noPlaceholders = true;
 
-// Styles
-export styles from './styles.scss';
```
tay1orjones: Storybook relied on this for the styles to be included I believe. Styles are missing/invalid in the storybook deploy preview currently:

image

I think to fix you can import the styles.scss entrypoint to `.storybook/story-styles.scss`. Styles in that file are applied to all stories.

babel.config.js
```diff
@@ -17,6 +20,9 @@ module.exports = function generateConfig(api) {
     ],
     ignore: ['__mocks__'],
     plugins: [
+      ...(BABEL_ENV === 'production'
```
tay1orjones: Will this file ever be read _not_ in production? Storybook ships with it's own babel config, and there's an outstanding bug that `babel.config.js` files aren't sourced - https://github.com/storybookjs/storybook/issues/6633#issuecomment-553397732

I wonder if we could just move the production conditionals to be the default? We can additionally change the script in `package.json` to not set the `BABEL_ENV` to `production`

.eslintignore
```diff
@@ -2,5 +2,7 @@ node_modules/
 coverage/
 config/
 lib/
+es/
+umd/
```
tay1orjones: Should we include tmp here as well just in case?
```suggestion
umd/
tmp/
```

jest.config.js
```diff
@@ -113,5 +113,7 @@ module.exports = {
     // https://github.com/storybookjs/storybook/pull/9795#issuecomment-587808019
     // either way, this can be removed in the future
     'react-syntax-highlighter/dist/esm/(.*)': 'react-syntax-highlighter/dist/cjs/$1',
+    '@carbon/icons-react/es/(.*)': '@carbon/icons-react/lib/$1',
```
tay1orjones: Some commenting to notate the history of what's/why happening here
```suggestion
    // rewrite carbon-components(-react) es imports to lib/cjs imports because jest doesn't support es modules
    // https://github.com/facebook/jest/issues/4842
    '@carbon/icons-react/es/(.*)': '@carbon/icons-react/lib/$1',
```

babel.config.js
davidicus: When the bug in storybook is fixed it will use our bable config in which case we will need to specify then environment

src/index.js
davidicus: added to storybook styles

.eslintignore
davidicus: added

babel.config.js
tay1orjones: Good point, plus as you mentioned to me it's being used for jest, which is a non production env.

davidicus:
> A couple small things here.
> 
> Additionally, the builds are being generated in a `src` folder. Is there any way to remove that unecessary second level of nesting so we publish just `lib/index.js`, `es/index.js`, `es/components/ComponentName/ComponentName.js`, etc?

I believe this is the case because of the preserve modules flag. We can only dictate a directory with this property set to true.

rollup.config.js
```diff
@@ -9,121 +9,168 @@ import copy from 'rollup-plugin-copy';
...
+            dest: ['es/node_modules', 'lib/node_modules'],
+          },
+          // Copy CSS
+          {
+            src: ['lib/css/'],
+            dest: ['./', 'umd'],
+          },
+        ],
+        verbose: env !== 'development', // logs the file copy list on production builds for easier debugging
+      }),
```
tay1orjones: Does this need to be part of the UMD build or could it be moved up into the `Compile styles` build section after the compilation?

rollup.config.js
```diff
@@ -9,121 +9,168 @@ import copy from 'rollup-plugin-copy';
...
+            src: [
+              'node_modules/react-resizable/css/**/*.css',
+              'node_modules/react-grid-layout/css/**/*.css',
+            ],
+            dest: ['es/node_modules', 'lib/node_modules'],
+          },
+          // Copy CSS
+          {
+            src: ['lib/css/'],
+            dest: ['./', 'umd'],
```
tay1orjones: Does the `umd` build folder need a `css` folder in it?

davidicus: no actually, I meant to take that out

rollup.config.js
davidicus: It needs to be separate from styles build unfortunately. Even though these plugins are supposed to run sequentially we are unable to copy the compiled file that is created from the postcss plugin. I put it in the umd build as to not have to have an extra build process.

tay1orjones: Bummer on that - makes sense though. Something to revisit for a future major version bump.

davidicus:
after adding changes to CLI to add d3 as dep for our esm module version we can see that d3 is still  not shipped with the application 
image

PR #1049, opened 2020-03-30
[TimeSeriesCard] enable hiding zero

### LINKED ISSUE #1048: [TimeSeriesCard] Hide zero from either axis
### Summary

If a user has data that is much higher and focused away from 0, the graph stretches so that 0 is always included. By hiding zero, the data is more focused and easier to read.

**Additional context**
Carbon charts added this feature here: https://github.com/carbon-design-system/carbon-charts/pull/515

date, the better!_

sstone2423:
Closes #1048 

**Summary**

- Allows user to hide zero, meaning the data can be focused and be easier to read

**Before**
Screen Shot 2020-03-30 at 12 37 06 PM

**After**
Screen Shot 2020-03-30 at 12 37 12 PM

- Upgraded carbon version to pick up the `showZero` prop
- Added `includeZeroOnXaxis` and `includeZeroOnYaxis` props which default to true
- Added knobs to each TimeSeriesCard story

- Go to any TimeSeriesCard story that has data high above 0 and change the `content` JSON object's `includeZeroOnYaxis` prop to false to see the change. The stories' data is randomized so this may take a few clicks between stories or refreshes.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -64,7 +64,7 @@ module.exports = {
       statements: 70,
       branches: 50,
       lines: 69,
-      functions: 77,
+      functions: 76.47,
     },
     './src/components/Page/EditPage.jsx': {
       statements: 63,
--- a/package.json
+++ b/package.json
@@ -65,8 +65,8 @@
     ]
   },
   "dependencies": {
-    "@carbon/charts": "^0.29.0",
-    "@carbon/charts-react": "^0.29.0",
+    "@carbon/charts": "^0.30.5",
+    "@carbon/charts-react": "^0.30.5",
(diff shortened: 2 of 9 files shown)


tay1orjones:
@sstone2423 it looks like test coverage dropped a slight amount with this addition

```
Jest: "./src/components/TimeSeriesCard/TimeSeriesCard.jsx" coverage threshold for functions (77%) not met: 76.47%
```

We can modify the test threshold exception for this component in the jest.config.js as your change didn't actually produce any new untested lines or branches. If you're feeling charitable there are some outstanding functions like `handleFillColor` that could be a quick win for coverage. Either way is fine with me 👍

PR #1054, opened 2020-03-31
Allow tooltip content to contain button and links

### LINKED ISSUE #1010: Tooltip needs to support links and buttons
Edit by @rdconner 
There are a few instances where we compose components that include the tooltip. We need to make sure all of the tooltips capabilities are surfaced in the API. Examples are where we include an "additional info" tooltip like the Card. 

This is an implementation of a component feature that exists in Carbon.
Attn: @rdconner @smithjor 

In the Carbon tooltip component, there is the option to include a link and a button.  We need these to work for the infopops in offerings.

_[Cody to provide more info here]_

Carbon examples: https://www.carbondesignsystem.com/components/tooltip/usage
IoT Storybook: https://carbon-addons-iot-react.com/?path=/story/tooltip--default-bottom

image

davidicus:
Closes #1010 

**Summary**

Tooltip should have the ability to render links and buttons inside. Most of our implementation allows for this. The only exceptions in PageTitleBar. This PR changes the markup to allow for any type of content and not just text. 

- Added check for description type in PageTitleBar.jsx to support old markup and the new custom node. 
- Changed the story to show off the rendering node as tooltip. 

- go to PageTitleBar story at ...
- Should see tooltip render with Button inside.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -110,6 +110,5 @@ module.exports = {
     // rewrite carbon-components(-react) es imports to lib/cjs imports because jest doesn't support es modules
     '@carbon/icons-react/es/(.*)': '@carbon/icons-react/lib/$1',
     'carbon-components-react/es/(.*)': 'carbon-components-react/lib/$1',
-    'react-syntax-highlighter/dist/esm/(.*)': 'react-syntax-highlighter/dist/cjs/$1',
   },
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -110,5 +110,6 @@ module.exports = {
     // rewrite carbon-components(-react) es imports to lib/cjs imports because jest doesn't support es modules
     '@carbon/icons-react/es/(.*)': '@carbon/icons-react/lib/$1',
     'carbon-components-react/es/(.*)': 'carbon-components-react/lib/$1',
+    'react-syntax-highlighter/dist/esm/(.*)': 'react-syntax-highlighter/dist/cjs/$1',
   },
 };
--- a/src/components/PageTitleBar/PageTitleBar.jsx
+++ b/src/components/PageTitleBar/PageTitleBar.jsx
@@ -60,6 +60,12 @@ const defaultProps = {
   content: undefined,
 };
 
+// do typecheck to support old markup with surrounding 'p' tag
+// eslint-disable-next-line
+export const TooltipDescription = ({ description }) => {
+  return typeof description === 'string' ? <p>{description}</p> : description;
+};
+
 const PageTitleBar = ({
   title,
   description,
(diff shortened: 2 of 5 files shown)


src/components/PageTitleBar/PageTitleBar.jsx
```diff
@@ -60,6 +60,12 @@ const defaultProps = {
   content: undefined,
 };
 
+// do typecheck to support old markup with surrounding 'p' tag
+// eslint-disable-next-line
```
tay1orjones, line 64: Scope the ignore to specific rule(s) so there's less exposure to potential errors.

I assume this is because only one export is allowed per file or something similar? You could inline the logic in the jsx and test the `p` tag is/isn't rendered appropriately by passing the node through PageTitleBar instead of directly mounting and testing `TooltipDescription`.

jest.config.js
```diff
@@ -110,5 +110,6 @@ module.exports = {
     // rewrite carbon-components(-react) es imports to lib/cjs imports because jest doesn't support es modules
     '@carbon/icons-react/es/(.*)': '@carbon/icons-react/lib/$1',
     'carbon-components-react/es/(.*)': 'carbon-components-react/lib/$1',
+    'react-syntax-highlighter/dist/esm/(.*)': 'react-syntax-highlighter/dist/cjs/$1',
```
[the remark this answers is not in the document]
davidicus: I have that change pulled into this branch and still needed this to resolve build error.

src/components/PageTitleBar/PageTitleBar.jsx
davidicus: I applied this because I am using the same description that is used in the component. It already has prop-type validation but eslint is expecting prop type validation again for this as well.

davidicus:
> Should the PropType get updated to node OR string?
> 
> Looks good otherwise

I am only supporting the string for legacy uses with the <p> tag but that is a good point. It probably should be.

jest.config.js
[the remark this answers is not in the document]
davidicus: I tested again and it passed. I have removed this from jest config

src/components/PageTitleBar/PageTitleBar.story.jsx
```diff
@@ -24,6 +24,13 @@ export const commonPageTitleBarProps = {
   ),
 };
 
+const NodeTooltip = () => (
+  <div>
+    Descriptive text about this page and what the user can or should do on it{' '}
+    <button type="button">Even Buttons</button>
```
tay1orjones, line 30: Could we use the Carbon button to show a more visually valid example?

Current storybook deploy preview:
image
Design guidance from root issue:
image

davidicus:
For whatever reason my local dev environment gets blown up when trying to pull this branch down to make changes. I have redone the changes and will open a new PR

Issue #1024, opened 2020-03-19
[RFC] codeowner acceptance standards


tay1orjones:
### Summary

We should document our standards of acceptance. What we're looking to see in a contribution. It's currently pretty opaque today as to what we're looking for. As example, our browser support is virtually undocumented as we're currently just using the default browserlist config for autoprefixer. There's a set of must-haves. Is there a set of nice-to-haves that we can defer into separate issues if needed?

Additionally the definition/responsibilities of a codeowner should be codified. Anyone can review a PR, but codeowners are the people relied on to ensure submissions are properly evaluated against the standard of acceptance. 

**Must-have items for all PRs**
- Code
  - Code changes must be properly isolated. Pull requests should be laser focused to one single fix or feature. Without isolated changes we can not reliably cherry pick or back-merge fixes into old versions as hotfixes.
  - For helper functions (not components) intended for internal use only we should attempt to omit it from the final build or make it clear via a prefix with `__internal`. This could include some constants. Experimental _components_ will be flagged via the final decision in #1137 
  - New features should be off by default to preserve the original behavior where possible. Occasionally this principle may be overridden by guidance from the design team, particularly when a change to default behavior increases the usability substantially such as the change in #1063
- Documentation
  - Has the DOM node that an `id` corresponds to changed? This should be visible in the snapshots. If so, provide a migration note that could be included in the release notes.
- Code covered in unit tests
- `data-testid` on outermost container of component
- list of components modified with description of how to utilize
- RTL
- i18n, no hardcoded strings
- ...?

**Deferrable items for PRs**
- a11y?
- ...?

**Codeowner responsibilities**
- Review PRs and sign off on all must-have items
- ...?

**To Do**
- [ ] ~implement [`browserslist-config-carbon`](https://github.com/carbon-design-system/carbon/blob/master/packages/browserslist-config-carbon/index.js#L10) to adhere to the same set of browser support as Carbon.~ new `browserslist` will be implemented in v3 #956 
- [ ] Document contribution guidelines, PR must haves, what an ideal contribution looks like
- [ ] Document codeowner responsibilities, provide PR review template of checklist items
- [ ] Update Pull Request Template to include a link to this documentation
- [ ] Fill in and round out additional related documentation (testing, code documentation, should we provide a code style guide beyond what we get automagically from prettier?)

tay1orjones:
Related reading
[...]

Issue #979, opened 2020-03-02
Upgrade Carbon deps to v10.10.x


davidicus:
There are several fixes needed in this upgrade. This work will have to be done post MAS MVP release. 

<sub>[Internal issue #301](https://github.ibm.com/Watson-IoT/pal-tracking/issues/301)</sub>

Complete when:
- [x] update all carbon deps
- [x] test our stories for defects

cgirani:
MAS team depends on the latest carbon updates, `Notification` component

Carbon:
![image](https://user-images.githubusercontent.com/22034430/77328627-8e681b80-6cfb-11ea-927f-687d78e9595a.png)


Addons:
![image](https://user-images.githubusercontent.com/22034430/77328640-94f69300-6cfb-11ea-8245-e8d047f16845.png)

tay1orjones:
We will need too take this into account: https://github.com/carbon-design-system/carbon/pull/5379

davidicus:
`carbon-components@10.10.3`
>Bug fixes 🐛
fix(tag): define default tag style (#5677) (f8b485066)
fix(button): prevent hover on disabled buttons in g90/g100 theme (#5678) (04ae1e92d)
fix(tile): change expandable tile to button (#5609) (fbab6c0bd)
fix(components): remove @carbon/icons from peerDependencies (#5656) (39e612536)
fix(button): prevent icon and tooltip jumping for tertiary button (#5618) (f32c1a611)
fix(code-snippet): set correct width for feedback container in Safari (#5601) (0917bb27e)
fix(progress-indicator): use circle--filled icon (#5585) (7ae99155c)
fix(tile): add non link styling to clickable tile (#5591) (20517eb72)
fix(SideNav): disregard hover on non-rail SideNav when closed (#5562) (b494e528d)
Bug fixes 🐛
fix(button): prevent text jump in disabled button set (#5556) (38fb161af)
fix(dependencies): update icons dependency in vanilla (#5574) (f793a156a)
fix(toggle): change spacing to match style guide (#5462) (bbe9f0d50)
fix(link): use hover color token for visited links (#5559) (8b03644ec)
fix(tooltip): wrap icon tooltip content (#5544) (dfc695617)
fix(modal): change bottom margin (#5510) (3b7314442)
fix(tabs): update tab styles (#5482) (2c14be25b)
fix(modal): set half width single buttons (#5502) (80903ac57)
[...]

PR #1072, opened 2020-04-07
feat(datetimepicker): new DateTimePicker component

### LINKED ISSUE #854: [datepicker] Time range / date-time picking for dashboards
Copied from PAL tracking [issue](https://github.ibm.com/Watson-IoT/pal-tracking/issues/26)

**Description**
Time range picker is a common dashboard component requested by many product teams. The component gives user control over the default range, as well as allowing them to filter the data that is being visualized in dashboard by time.

**Name of Pattern or Asset**
_Add the name and a link to the pattern you are referencing in IoT or Carbon_

- Carbon's date range picker: http://react.carbondesignsystem.com/?path=/story/datepicker--range-with-calendar

- Carbon's time picker: http://react.carbondesignsystem.com/?path=/story/timepicker--default

**User outcome**
_Is this related to a user outcome you are working on?_

Users need to be able to change the time window of the data in dashboard. 

**Proposal**
_Do you have a proposed solution? or something you have tried already?_

Proposed design 
Proposal
- the picker provides preset ranges that are relative to Now, this helps user to quickly filter their data
- ability to set a custom relative range 
- ability to set an absolute range 
- ability to inline edit the range that conforms to a specific format
- only commit the change upon clicking "Apply" 

Additional comments from thread:

@cynthiaz
>Here's the latest design exploration 
>image

@Laura-Dohle
>@cynthiaz I think that looks awesome! 
Only two comments:
>- I have doubts about is calling it "custom range". When I think about custom range I would expect to add a custom relative range, e.g. last 31 days. I would propose we call it absolute range instead.
>- Instead of using the calendar icon plus dropdown, why don't we use the same visual style carbon is using for their calendar component for aligment:  screenshot 2019-11-20 at 15 03 03

@cynthiaz
>@Laura-Dohle 
thanks for the feedback!
The reason we didn't use the exact carbon's calendar component is because we wanted to hint user that it's a dropdown with options, rather than suggesting it's only a calendar for picking date.
>
>The default behavior that we expect is to open a dropdown with a list of common relative ranges. Then if the user is not finding anything, they can set an absolute range, which uses the carbon's calendar and behaves exactly the same way as here: http://react.carbondesignsystem.com/?path=/story/datepicker--range-with-calendar
>The only differences are: 
>1. merged two input fields into one; 
>2. added input field to set time
>3. support inline edit with x button to clear the range 

@tay1orjones 
>As this firms up and we begin to look at implementation, it will be important to have a clear path of keyboard navigation through the interactions of this component. There are [a few items open on Carbon for a11y of their DatePicker](https://github.com/carbon-design-system/carbon/issues?q=is%3Aissue+datepicker+label%3A%22type%3A+a11y+%E2%99%BF%22), as well as [the library under the hood, `flatpickr`](https://github.com/flatpickr/flatpickr/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+keyboard).
>
>Some things I'm thinking about:
>- The dropdown tooltip showing absolute range will need to be available on focus in addition to hover.
>- Between the relative/absolute "modes" of the input, it will impact if we can show the tooltip
>  - When a dropdown is focused, it does not open. This works with the tooltip fine. 
>  - Although, when datepicker is focused, it does open. That doesn't work with the tooltip. We'd be opening the calendar as well as showing the tooltip.
>- When editing a range on the calendar, pressing enter applies the selection. This may need to be prevented so the user can select a date, then navigate to the time pickers.

@cynthiaz
>Iteration 3:
>image
>- Default to show any applicable relative ranges (preset)
>- Select "Absolute range" option to open the calendar view
>
>@Laura-Dohle TBI's requirement of being able to choose the reference date instead of Today [Now], maybe a feature that the team can extend based on this pattern. Conceptually, I think once the reference date/time is changed, it becomes an absolute range with a specific Start and End date. The difference is around the experience of setting a custom range. Both options below achieve the same result, but I think what you requested is option 1, which is less burden for user to set the duration.
>1. Last `[ 5 ]` `[hours]` ending at `[Oct 8 15:00] ` 
>2. From `[Oct 8 10:00]` to `[Oct 8 15:00] `

>Saw @Laura-Dohle comment in #39 
>`select an end date and time as either a fixed end time or a relative end time of yesterday, 12AM`
>
>So this requires an option to also change the reference time for a relative range. We could use the custom range option to set a relative or an absolute range: 
>image
>
>Relative end time could include Today/Yesterday/End of this week... so on

@lukefirth 
>I like this, although my question now is - 'when i'm in custom, how do I get back to regular? 😄 '
>
>Also, are my eyes playing tricks on me, or is the 'custom range' button in the dropdown on the left, a smaller text size than the other options?

@Cynthiaz
>sharp eyes! @lukefirth 😄 
>
>wrt your first question, I thought about it and made a trade off for simplicity.. 
So you are able to click anywhere outside or "Cancel" to close it, and click the dropdown again to get back to the default (2 clicks). We could add a content switcher or a button to get down to just 1 click. 

@lora_dohle
>I like the flow @cynthiaz. Love how we are keeping the dropdown simple while still allowing for full customisability. 
>
>Regarding @lukefirth 's point: How about changing the Cancel button on the custom range modal to a back button? They could still cancel by clicking outside of the modal.

### LINKED ISSUE #1114: [DateTimePicker] Fix issues with sass partial
There are a few items that were noted on review that we need to clean up but didnt want to hold up the component for. 
`src/components/DateTimePicker/_date-time-picker.scss`

- [x] There are places were we need to use Carbon tokens
- [x] There are places were we are using px values
- [x] remove vender prefixes

enricoberti:
Closes #854 closes #1114 

**Summary**
- Supports presets, relative and absolute times
- Presets and relative options are configurable

image

image

image

- Known item to be addressed after this PR is through: RTL support

- ?path=/story/watson-iot-experimental-datetime-picker--default and sibling stories


davidicus:
@enricoberti this is looking great! I did see one small bug. If you click on your start and end date for custom range. Then click on any date under the selected. It will start to increment the end date by months.

JordanWSmith15:
@enricoberti Looking great! A few scenarios to consider checking that will cause problems that Graphite has had to fix for the Dropdown component:
1. When positioned near the bottom of a page, the editing dialog should expand upwards, rather than below the page fold. Example at bottom of this storybook: https://pages.github.ibm.com/maximo-app-framework/graphite/react/storybook/?path=/story/components-%E2%9C%93-dropdown--dropdown
2. Plan for issues with z-index. It would be good to demonstrate what this does inside of a dialog, for instance. The editing dialog of this component should overlap the modal buttons and expand outside of a modal, rather than cause an internal scroll: See dropdown 2: https://pages.github.ibm.com/maximo-app-framework/graphite/react/storybook/?path=/story/components-%E2%9C%93-dropdown--datasource-dropdown-on-a-dialog

enricoberti:
thanks @davidicus and @JordanWSmith15, valid feedback! fixing the issues now

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+  /** show the picker in the expanded state */
+  expanded: PropTypes.bool,
+  /** disable the input */
+  disabled: PropTypes.bool,
+  /** show the relative custom range picker */
+  showRelativeOption: PropTypes.bool,
+  /** triggered on input click */
+  onCancel: PropTypes.func,
+  /** triggered on input click */
+  onApply: PropTypes.func,
```
stuckless, line 119: What am I getting back?

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+            endDate.minutes(value.relative.relativeToTime.split(':')[1]);
+          }
+        }
+        const startDate = endDate
+          .clone()
+          .subtract(
+            value.relative.lastNumber,
+            value.relative.lastInterval ? value.relative.lastInterval : INTERVAL_VALUES.MINUTES
+          );
+        readableValue = `${moment(startDate).format('YYYY-MM-DD HH:mm')} to ${moment(
```
stuckless, line 261: translations.  either this needs to a formatted string `{0} to {1}`, or a function.
I would think that that the datetime mask would be a prop.  it won't always be that format.

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+          endDate
+        ).format('YYYY-MM-DD HH:mm')}`;
+        break;
+      }
+      case PICKER_KINDS.ABSOLUTE:
+        readableValue = `${moment(value.absolute.start).format('YYYY-MM-DD')} ${
```
stuckless, line 267: date mask should be a prop

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+          value.absolute.startTime ? value.absolute.startTime : '00:00'
```
stuckless, line 268: time mask shold be a prop (what if I want AM/PM, etc)

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+    }
+    if (focusOnFirstField && datePickerRef.current.inputField) {
+      datePickerRef.current.inputField.focus();
+    } else {
+      datePickerRef.current.toInputField.focus();
+    }
+
+    const newAbsolute = { ...absoluteValue };
+    [newAbsolute.start] = range;
+    newAbsolute.startDate = moment(newAbsolute.start).format('MM/DD/YYYY');
```
stuckless, line 328: date mask

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+      onApply(value);
+    }
+  };
+
+  const getIntervalValue = () => {
+    if (currentValue) {
+      if (currentValue.kind === PICKER_KINDS.PRESET) {
+        return `${moment()
+          .subtract(currentValue.preset.offset, 'minutes')
+          .format('YYYY-MM-DD HH:mm')} to Now`;
```
stuckless, line 418: not sure how this is used but if it's displayed then we need translation and custom date/time mask

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+            <TooltipDefinition
+              align="start"
+              direction="bottom"
+              tooltipText={getIntervalValue()}
+              triggerClassName=""
+            >
+              {humanValue}
+            </TooltipDefinition>
+          ) : null}
+          <Calendar16 aria-label="Calendar" className={`${iotPrefix}--date-time-picker__icon`} />
```
stuckless, line 485: translation (aria-label)

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                      {preset.label}
+                    </ListItem>
+                  );
+                })}
+              </OrderedList>
+            ) : (
+              <div>
+                {showRelativeOption ? (
+                  <FormGroup
+                    legendText="Custom range"
```
stuckless, line 531: translation (legendText)

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                  >
+                    <RadioButtonGroup
+                      defaultSelected={customRangeKind}
+                      onChange={onCustomRangeChange}
+                      name="radiogroup"
+                    >
+                      <RadioButton
+                        value={PICKER_KINDS.RELATIVE}
+                        id="relative"
+                        labelText="Relative"
```
stuckless, line 542: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                      />
+                      <RadioButton
+                        value={PICKER_KINDS.ABSOLUTE}
+                        id="absolute"
+                        labelText="Absolute"
```
stuckless, line 547: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                      />
+                    </RadioButtonGroup>
+                  </FormGroup>
+                ) : null}
+                {customRangeKind === PICKER_KINDS.RELATIVE ? (
+                  <div>
+                    <FormGroup
+                      legendText="Last"
```
stuckless, line 555: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                      className={`${iotPrefix}--date-time-picker__menu-formgroup`}
+                    >
+                      <div className={`${iotPrefix}--date-time-picker__fields-wrapper`}>
+                        <NumberInput
+                          id="last-number"
+                          invalidText="Number is not valid"
```
stuckless, line 561: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                          {intervals.map((interval, i) => {
+                            return (
+                              <SelectItem key={i} value={interval.value} text={interval.label} />
+                            );
+                          })}
+                        </Select>
+                      </div>
+                    </FormGroup>
+                    <FormGroup
+                      legendText="Relative to"
```
stuckless, line 585: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                        />
+                      </div>
+                    </FormGroup>
+                  </div>
+                ) : (
+                  <div>
+                    <div className={`${iotPrefix}--date-time-picker__datepicker`}>
+                      <DatePicker
+                        datePickerType="range"
+                        dateFormat="m/d/Y"
```
stuckless, line 617: custom format

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                      </DatePicker>
+                    </div>
+                    <FormGroup
+                      legendText=""
+                      className={`${iotPrefix}--date-time-picker__menu-formgroup`}
+                    >
+                      <div className={`${iotPrefix}--date-time-picker__fields-wrapper`}>
+                        <TimePickerSpinner
+                          id="start-time"
+                          labelText="Start time"
```
stuckless, line 643: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+                          value={absoluteValue ? absoluteValue.startTime : '00:00'}
+                          onChange={onAbsoluteStartTimeChange}
+                          spinner
+                          autoComplete="off"
+                        />
+                        <TimePickerSpinner
+                          id="end-time"
+                          labelText="End time"
```
stuckless, line 651: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+          </div>
+          <div className={`${iotPrefix}--date-time-picker__menu-btn-set`}>
+            {isCustomRange ? (
+              <Button
+                kind="secondary"
+                className={`${iotPrefix}--date-time-picker__menu-btn ${iotPrefix}--date-time-picker__menu-btn-back`}
+                {...others}
+                onClick={toggleIsCustomRange}
+              >
+                Back
```
stuckless, line 672: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+              </Button>
+            ) : (
+              <Button
+                kind="secondary"
+                className={`${iotPrefix}--date-time-picker__menu-btn ${iotPrefix}--date-time-picker__menu-btn-cancel`}
+                onClick={onCancelClick}
+                {...others}
+              >
+                Cancel
```
stuckless, line 681: translation

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,702 @@
...
+              </Button>
+            )}
+            <Button
+              kind="primary"
+              className={`${iotPrefix}--date-time-picker__menu-btn ${iotPrefix}--date-time-picker__menu-btn-apply`}
+              {...others}
+              onClick={onApplyClick}
+            >
+              Apply
```
stuckless, line 690: translation

enricoberti:
thank you for the feedback @stuckless ! i pushed a few changes:
- a bit more documentation on the returned object
- i18n
- better cancelation flow
- date time mask
- ellipsis on overflowing text
- filter out non HH:mm characters on the timepickerspinner

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,865 @@
...
+        break;
+    }
+
+    /* istanbul ignore else */
+    if (onApply) {
+      onApply(value);
+    }
+  };
+
+  const getIntervalValue = () => {
```
tay1orjones, line 565: Another spot to add some code comments documenting the function's purpose

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,865 @@
...
+  showRelativeOption: true,
+  onCancel: null,
+  onApply: null,
+  i18n: {
+    toLabel: 'to',
+    toNowLabel: 'to Now',
+    calendarLabel: 'Calendar',
+    presetLabels: 'Last 30 minutes|Last 1 hour|Last 6 hours|Last 12 hours|Last 24 hours',
+    intervalLabels: 'minutes|hours|days|weeks|months|years',
+    relativeLabels: 'Yesterday|Today',
```
tay1orjones, line 224: An array of strings would be more appropriate for these. I don't think it would conflict with i18n libraries in applications? That or they could be a function that returns an array of strings.

src/components/DateTimePicker/DateTimePicker.story.jsx
```diff
@@ -0,0 +1,88 @@
...
+      <div style={{ width: `${getCardMinSize('lg', size).x}px`, margin: 20 }}>
+        <DateTimePicker dateTimeMask={text('dateTimeMask', 'YYYY-MM-DD HH:mm')} />
+      </div>
+    );
+  })
+  .add('Selected preset', () => {
+    const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.MEDIUMWIDE);
+    return (
+      <div style={{ width: `${getCardMinSize('lg', size).x}px`, margin: 20 }}>
+        <DateTimePicker defaultValue={customPresets[3]} onApply={action('onApply')} />
```
tay1orjones, line 57: Include the onCancel callback as well
```suggestion
        <DateTimePicker defaultValue={customPresets[3]} onApply={action('onApply') onCancel={action('onCancel')} />
```

src/components/DateTimePicker/DateTimePicker.story.jsx
```diff
@@ -0,0 +1,88 @@
...
+      </div>
+    );
+  })
+  .add('Selected relative', () => {
+    const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.MEDIUMWIDE);
+    return (
+      <div style={{ width: `${getCardMinSize('lg', size).x}px`, margin: 20 }}>
+        <DateTimePicker defaultValue={defaultRelativeValue} onApply={action('onApply')} />
```
tay1orjones, line 65: Include the onCancel callback as well
```suggestion
        <DateTimePicker defaultValue={defaultRelativeValue} onApply={action('onApply') onCancel={action('onCancel')} />
```

src/components/DateTimePicker/DateTimePicker.story.jsx
```diff
@@ -0,0 +1,88 @@
...
+      </div>
+    );
+  })
+  .add('Selected absolute', () => {
+    const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.MEDIUMWIDE);
+    return (
+      <div style={{ width: `${getCardMinSize('lg', size).x}px`, margin: 20 }}>
+        <DateTimePicker defaultValue={defaultAbsoluteValue} onApply={action('onApply')} />
```
tay1orjones, line 73: Include the onCancel callback as well
```suggestion
        <DateTimePicker defaultValue={defaultAbsoluteValue} onApply={action('onApply') onCancel={action('onCancel')} />
```

src/components/DateTimePicker/DateTimePicker.story.jsx
```diff
@@ -0,0 +1,88 @@
...
+    );
+  })
+  .add('Without a relative option', () => {
+    const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.MEDIUMWIDE);
+    return (
+      <div style={{ width: `${getCardMinSize('lg', size).x}px`, margin: 20 }}>
+        <DateTimePicker
+          defaultValue={defaultAbsoluteValue}
+          showRelativeOption={false}
+          onApply={action('onApply')}
```
tay1orjones, line 84: Include the onCancel callback as well
```suggestion
          onApply={action('onApply')}
          onCancel={action('onCancel')
```

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,865 @@
...
+    absolute: {
+      startDate: null,
+      startTime: null,
+      endDate: null,
+      endTime: null,
+    },
+  };
+
+  useEffect(() => {
+    window.setTimeout(() => {
```
tay1orjones, line 300: Doesn't this `setTimeout` need to be assigned to a variable and cleaned up via the useEffect return function to prevent infinite loops and memory leaks?

src/components/DateTimePicker/_date-time-picker.scss
```diff
@@ -0,0 +1,183 @@
...
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+      .#{$iot-prefix}--date-time-picker__icon {
+        position: absolute;
+        top: 0;
+        right: $spacing-05;
+        bottom: 0;
+        height: 100%;
+        transition: -webkit-transform 70ms cubic-bezier(0.2, 0, 0.38, 0.9);
```
tay1orjones, line 106: We should use [the easing functions from `@carbon/motion`](https://github.com/carbon-design-system/carbon/tree/master/packages/motion#sass)

You can `yarn add` it to `dependencies` and **pin the version to 10.5.0.**

@davidicus once this is added, you'll need to update it as well in your WIP branch updating to Carbon v10.10.3 #979

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,865 @@
...
+        </div>
+      </div>
+    </div>
+  );
+};
+
+DateTimePicker.propTypes = propTypes;
+DateTimePicker.defaultProps = defaultProps;
+
+export default DateTimePicker;
```
tay1orjones, line 865: Based on some recent conversations with maintainers, @davidicus, #1024 around the experimental section, I think we should make clear that this component is not finalized and prefix this export with `__unstable`
```suggestion
export default __unstableDateTimePicker;
```

You'll need to modify the other places above to utilize the prefix as well.

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,865 @@
...
+    },
+    absolute: {
+      start: Date // the start point in time
+      end: Date // the end point in time
+      startDate: String // start date in the mask or default format
+      startTime: String // in the HH:MM format
+      endDate:  // end date in the mask or default format
+      endTime: String // in the HH:MM format
+    },
+  } */
```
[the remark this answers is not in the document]
tay1orjones: If it is already being logged out in the action panel, we can delete that code comment section

davidicus:
Minor RTL issue when picking a custom range 

image

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,865 @@
...
+      startDate: null,
+      startTime: null,
+      endDate: null,
+      endTime: null,
+    },
+  };
+
+  useEffect(() => {
+    window.setTimeout(() => {
+      /* istanbul ignore next */
```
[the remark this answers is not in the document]
enricoberti: Yes, better like that!

src/components/DateTimePicker/DateTimePicker.jsx
enricoberti: imho in this case it's not needed, the dom reparenting or calendar opening do not trigger a rendering cycle on the component itself, so the effect is not called recursively by itself

src/components/DateTimePicker/_date-time-picker.scss
enricoberti: perfect! changed to use the duration var :)

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -0,0 +1,881 @@
...
+    },
+  } */
+  onApply: PropTypes.func,
+  /** All the labels that need translation */
+  i18n: PropTypes.shape({
+    toLabel: PropTypes.string,
+    toNowLabel: PropTypes.string,
+    calendarLabel: PropTypes.string,
+    presetLabels: PropTypes.string,
+    intervalLabels: PropTypes.string,
```
davidicus, line 153:
```suggestion
    intervalLabels: PropTypes.arrayOf(PropTypes.string),
```

Issue #956, opened 2020-02-25
[Release] v3.0.0


tay1orjones:
A running list of breaking changes to be contained within the v3 release.

#### Must Haves
- [ ] ...
- [ ] Write migration guide documentation

#### Nice to Haves
- [ ] All `warning()` deprecations addressed and warnings removed
- [ ] Address remaining v2 nice-to-have items that were not closed. Evaluate the validity of these - we may choose to close them or restructure the work into smaller pieces.

davidicus:
Final list of Carbon tokens: https://www.notion.so/Final-List-of-new-tokens-fc5985c7331341baa6546064399d36da

PR #1078, opened 2020-04-09
1051 cells can truncate content show tooltip if truncated

### LINKED ISSUE #1051: [Table] cells can truncate content, show tooltip if truncated
Copied from internal repo from @KarenGski 
> 
> In data tables, we need to allow the content to truncate so that the rows don't become very very tall if the content is longer that the set column widgth. If the content is trunctated, the user should be able to click on it to see the full, untruncated content.
> 
> Include mockups of any fidelity, the more spec the better as this will enable higher quality feedback and evaluation.
> 
> Include for accessibility
> 
> * Examples of the elements in various states
> * Expected keyboard behavior
> * Alt text
> 

bjornalm:
Closes #1051

**Summary**

> In data tables, we need to allow the content to truncate so that the rows don't become very very tall if the content is longer that the set column widgth. If the content is trunctated, the user should be able to click on it to see the full, untruncated content.

_Implementation_
image

Cell text wrapping is configurable via an option.wrapText prop.

1. If wrapText is 'auto' we apply wrapping only for tables with normal columns*. For tables with fixed or resizable columns we use truncation.
2. if wrapText is 'always' we wrap in all scenarios including tables with fixed or resizable columns (i.e. current situation).
3. if wrapText is 'never' we never wrap and as a consequence the tables with normal columns* get wide cells and more often horizontal scrolling. For tables with fixed or resizable columns we use truncation.

*) table with normal columns is here a table with the default table-layout:auto and with neither column resize nor fixed column widths.

See further discussion [here](https://github.com/IBM/carbon-addons-iot-react/issues/1051)

- added an external prop called wrapCellText
- added an internal prop called truncateCellText which is calculated in Table.jsx and passed down to the affected table components.
- wrapping like normal (carbon tables) is the default
- Truncation will only happen if wrapText:auto or wrapText:never for `hasResize:true` or if *any* of the columns has a set width.
- Changed min-width of columns so that truncated column header text and sort arrows fit together
- The tooltip is only added for table body cells, not the table header cells. Table header cells have the title showing on hover, but a click to show the carbon tooltip would be strange since click is used for sorting.
- The tooltip is only added if actual truncation is occurring, therefore all cells are measured for that.
- The tooltip will adopt to column width changes made by the user, i.e. if a previously truncated cell will have enough space after the user resizes the column the tooltip will be removed.
- Updated the relevant stories to have a knob for wrapCellText
- Added unit tests for the above described functionality when possible/reasonable

- Fixed a bug where the minimal width of the last column was not respected if the table needed horizontal scrollbars.

For all cases adjust your browser width to be around 1300px to see the effects of wrapping and truncation.

- [x] Table with column resize

https://deploy-preview-1078--carbon-addons-iot-react.netlify.com//?path=/story/watson-iot-table--with-resize-and-initial-column-widths-on-simple-stateful-and-row-selection

0. Set knob `wrapCellText` to `auto` 
1. Verify that the cell texts are being truncated where needed.
2. Click on the truncated texts to check that the tooltip is showing (also try scrolling when the tooltip is showing)
3. Verify that non truncated texts do not have a tooltip 
4. Resize a column that isn't truncating, e.g Secret Information, and verify that truncation and tooltip is added, also the the reverse.
5. Select `always` for the knob wrapCellTex and make sure the cell texts are now wrapping

- [x] Table with fixed column widths

https://deploy-preview-1078--carbon-addons-iot-react.netlify.com//?path=/story/watson-iot-table--with-fixed-column-width-and-no-resize

0. Set knob `wrapCellText` to `auto` 
1. Verify that the cell texts are being truncated where needed.
2. Select `always` for the knob wrapCellTex and make sure the cell texts are now wrapping

- [x] Table with no column resize nor any fixed width columns (i.e. dynamic column widths)

https://deploy-preview-1078--carbon-addons-iot-react.netlify.com/?path=/story/watson-iot-table--simple-stateful-example

1. Check that column texts are wrapping and that the column widths are distributed proportionally based on the content of the cells.
2. Select `never` for the knob wrapCellText and verify that the table no longer wrap the texts. The table should grow in size and get horizontal scrolling.


src/components/Table/Table.jsx
```diff
@@ -291,6 +293,28 @@ const Table = props => {
     ...others
   } = merge({}, defaultProps(props), props);
 
+  const [, forceUpdateCellTextWidth] = useState(0);
```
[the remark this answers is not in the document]
tay1orjones: I had to look this one up - the opening comma will skip/ignore the first array element. [mdn link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Ignoring_some_returned_values)

src/components/Table/Table.jsx
```diff
@@ -183,6 +184,7 @@ export const defaultProps = baseProps => ({
     hasResize: false,
     useAutoTableLayoutForResize: false,
     shouldLazyRender: false,
+    wrapCellText: undefined,
```
tay1orjones, line 187: From the PR body you mentioned
> If wrapText is not explicitly defined we apply it only for tables with normal columns*. For tables with fixed or resizable columns we use truncation.

`undefined` as default will cause this PR to be a breaking change. This will modify the existing default behavior of the table. Applications with automated testing around the table may have tests fail based on this prop. When upgrading they will have to modify their code to state `wrapCellText: true` in order to preserve the existing behavior.

If I understand correctly -  to preserve the existing default functionality, we could just have the default be `true` instead?

src/components/Table/Table.jsx
```diff
@@ -62,6 +62,7 @@ const propTypes = {
     hasResize: PropTypes.bool,
     /** If true removes the "table-layout: fixed" for resizable tables  */
     useAutoTableLayoutForResize: PropTypes.bool,
+    wrapCellText: PropTypes.bool,
```
tay1orjones, line 65: I think it would be clearer if we codified this into formal props
```
wrapCellText: oneOf['true', 'false', inherit]
``` 
or something similar, again with it defaulting to `true`.

Maybe `auto` is better than `inherit`?

```
wrapCellText: oneOf['true', 'false', auto]
```
Maybe `auto` implies too much though. Thoughts?

davidicus: I think this just ends up being a boolean that applies a class if true using the `classnames` package.

src/components/Table/Table.jsx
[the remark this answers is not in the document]
bjornalm: nope, it is to show that we are not interested in reading this state, we just use it to force a re-render of the table in order to update the width of the cell text

src/components/Table/Table.jsx
bjornalm: Yes, using true would preserve current default. We can do that but I implemented this as a change of default behaviour on purpose since I interpreted this comment that way 
https://github.ibm.com/Watson-IoT/pal-tracking/issues/331#issuecomment-18867586

It is possible that I misunderstood that comment, in that case we should definitively have `wrapCellText: true` as the default value like you say.

src/components/Table/Table.jsx
bjornalm: Yes, that is way better! I think we can make the intent even more clear by using `PropTypes.oneOf(['always', 'never', 'auto'])`

bjornalm: @davidicus there are 3 scenarios for the text here so we can use
'always' =  always use wrapping
'never'  = prevent wrapping and force wide cells
'auto' = truncate where possible and wrap if needed

src/components/Table/Table.jsx
bjornalm: with the latest PR change, default would be `wrapCellText: always`

tay1orjones: Yeah, a default of `wrapCellText: always` will preserve the existing default behavior. I believe the comment you linked is invalid - to my knowledge we've always had wrapping as the default.

src/components/Table/Table.jsx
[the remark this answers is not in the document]
sstone2423: Interesting, never seen that before

src/components/Table/Table.jsx
davidicus: The designers have asked for the truncation to be the new default but I don't think they would be willing to wait for a major  version release to get it. @lukefirth @JordanWSmith15 are we ok with keeping with the opt in so that we can get this feature published?

bjornalm: I've now changed the default to be `wrapCellText: always` and updated tests and stories.

lukefirth: Hey - apologies i'm so late in the game to this stuff. Maximo App Suite is a killer....

There's generally a _lot_ that we'll want to sit back and consider about what a major version update of the tables should contain/consider with all the new capabilities we are building etc and possible changes to default behaviors etc, to the point that I'd like us to maybe start actively considering a table current, vs next major release version and consider that, potentially even putting a date to the next major release version.

In the meantime, I'd like us to make use of the design level to catch and create consistency as long as the implemented components can support the various capabilities. 

Meaning: I'm happy for us to retain the current default behaviors in the component today, as long as when designs require it to be different (eg: truncate over wrap) they can be honored without a lot of code work on the implementer. Once we reach a state where _most_ designs are using truncate by default (via standardisation at the design level), meaning most implementers are making that change, which hopefully from now we will see happen. Then at that point, it will only make logical sense to use the next available major update to change the components behavior to reflect the new default (along with other major changes). 

A timeline to that next major release will help us understand what other major changes should, and can be bundled into it.

tay1orjones: @lukefirth We have an open umbrella issue for v3.0.0 - #956 

Things like this can be added there so we have a running list.

src/components/Table/_table.scss
```diff
@@ -10,6 +11,10 @@ table.#{$prefix}--side-nav--data-table {
   table-layout: fixed;
 }
 
+.#{$iot-prefix}--table__cell--truncate {
+  max-width: 0;
+}
+
```
tay1orjones, line 17: Should this be moved to the `TableCellRenderer/_table-cell-renderer.scss`?

bjornalm: Possibly, to me it isn't obvious where that style belongs. It relates to table cell rendering, but the style is actually for the parent  <TableCell> component and the class is set in the `TableBodyRow.jsx`.

PR #1081, opened 2020-04-10
Allow tooltip content to contain button and links

### LINKED ISSUE #1010: Tooltip needs to support links and buttons

davidicus:
Closes #1010 

**Summary**

Tooltip should have the ability to render links and buttons inside. Most of our implementation allows for this. The only exceptions in PageTitleBar. This PR changes the markup to allow for any type of content and not just text. 

- Added check for description type in PageTitleBar.jsx to support old markup and the new custom node. 
- Changed the story to show off the rendering node as tooltip. 

- go to PageTitleBar story at ...
- Should see tooltip render with Button inside.


davidicus:
@sstone2423 I decided agains the `oneOf`  since that would seem to be redundant according to their Readme
image

davidicus:
@tay1orjones there must have been some funkiness going on because i inlined it again and this time no issues no need for more test.

src/components/PageTitleBar/PageTitleBar.story.jsx
```diff
@@ -30,6 +30,25 @@ export const pageTitleBarBreadcrumb = [
...
+    <p>Descriptive text about this page and what the user can or should do on it </p>
+    <div
+      style={{
+        display: 'flex',
+        'align-items': 'center',
+        'justify-content': 'space-between',
+        'padding-top': '1rem',
+      }}
+    >
+      <a href="/">Link one</a>
```
tay1orjones, line 44: Should we use a `Link` from Carbon so we aren't demo'ing invalid design?

PR #1089, opened 2020-04-14
feat(gaugecard): added data state & error state

### LINKED ISSUE #1055: [GaugeCard] Enable dashboard tile alert and error states on the donut/guage tile
> Enable the dashboard widget alert and error states into the donut card
> Screen Shot 2020-03-30 at 1 37 58 PM
> Screen Shot 2020-03-30 at 1 37 11 PM

bjornalm:
Closes #1055

**Summary**
Enable the dashboard widget alert and error states into the donut card

1. Moved DataStateRenderer (+ test & css) to the Card folder
2. Removed the showOverflow prop since it is no longer needed when the default implementation of the tooltip is used
3. Added the datastate implementation to the GaugeCard
4. Updated the affected stories and added a new story for the GaugeCard with knobs for data/error state testing
5. Changed name from VALUE_CARD_DATA_STATE to the more generic CARD_DATA_STATE

image

Look at the story
`watson-iot-experimental-gaugecard--with-data-state-no-data` and play with the knobs


davidicus:
@sstone2423 Size prop has no effect outside of dashboard

src/components/Card/Card.jsx
```diff
@@ -110,7 +107,6 @@ const EmptyMessageWrapper = props => {
 CardWrapper.propTypes = {
   children: PropTypes.node.isRequired,
   dimensions: PropTypes.shape({ x: PropTypes.number, y: PropTypes.number }).isRequired,
-  showOverflow: CardPropTypes.showOverflow.isRequired,
```
tay1orjones, line 113: [Removal of a prop should result in a breaking change major version bump.](https://github.com/IBM/carbon-addons-iot-react/blob/master/docs/guides/versioning.md#an-existing-prop-is-removed) 

In this case you state there's no functionality being removed but removing the prop could cause application test automation failures. We can [deprecate the prop instead](https://github.com/IBM/carbon-addons-iot-react/blob/master/docs/guides/versioning.md#an-existing-prop-is-deprecated). 

The ``${iotPrefix}--card--wrapper--overflowing`` className is also being removed, which consumers could be targeting. Could we add a `data-testid` to the outermost element and provide it in a deprecation notice? see below:



```
import deprecate from '../../internal/deprecate';

...

showOverflow: deprecate(
  PropTypes.oneOfType([
    PropTypes.shape({
      type: PropTypes.oneOf([Tabs]),
    }),
  ]),
  '\nThe prop `showOverflow` for PageTitleBar has been deprecated. Instead, use the default implementation of the tooltip. The `${iotPrefix}--card--wrapper--overflowing` class has been removed, target `data-testid="card-cardwrapper"` instead.'
),
```

The deprecation message could probably be improved, but you get the idea.

tay1orjones:
@sstone2423 @davidicus [The `guagecard basic` story is wrapped with div that has a static width applied.](
https://github.com/IBM/carbon-addons-iot-react/pull/1089/files#diff-452fc540a786af84058c2c9531fcb258R53)

`<div style={{ width: '142px', margin: 20 }}>`

I think it's somewhat odd to apply wrapper styling like this when, as @davidicus mentioned, the cards have no `size` outside of the dashboard. They're the full width of their container.

I think we either 
1. Remove the width from the `basic` story
2. Add the width to the other story for consistency

I lean towards #1 to prevent confusion.

davidicus:
@tay1orjones I added the width for this as the smaller sizes are the only ones supported with this first pass for the gauge card and this gave the closest approximation to what the designs that were given. This is also the way these sizes were demonstrated in the card story that also is not rendered in a dashboard.

src/components/Card/Card.jsx
bjornalm: Makes total sense, thanks for explaining.

bjornalm: Fixed

sstone2423:
> @sstone2423 Size prop has no effect outside of dashboard

Didn't realize that. Disregard my comment then! 👍

src/constants/CardPropTypes.js
```diff
@@ -320,7 +321,10 @@ export const CardPropTypes = {
   /** should hide the header */
   hideHeader: PropTypes.bool,
   /** sets the CardWrapper CSS overflow to visible */
-  showOverflow: PropTypes.bool,
+  showOverflow: deprecate(
+    PropTypes.bool,
+    '\nThe prop `showOverflow` for Card has been deprecated. It was previously needed for a custom positioned tooltip in the ValueCard, but the ValueCard now uses the default positioning of the tooltip. The `iot--card--wrapper--overflowing` class has been removed, target `data-testid="iot--card--wrapper"` instead.'
```
tay1orjones, line 326:
```suggestion
    '\nThe prop `showOverflow` for Card has been deprecated. It was previously needed for a custom positioned tooltip in the ValueCard, but the ValueCard now uses the default positioning of the tooltip. The `iot--card--wrapper--overflowing` class has been removed. For automated testing, you can target `data-testid="Card"` instead.'
```

PR #1099, opened 2020-04-17
Add dynamic card variables prop

### LINKED ISSUE #1098: [CardRenderer] dynamic variables in card data
### Summary

<sub>[Internal Watson-IoT/pal-tracking#379](https://github.ibm.com/Watson-IoT/pal-tracking/issues/379)</sub>

Give cards the ability to have a prop of variable values that replace corresponding variables in the card data.

example:
{
     size: 'LARGE',
      title: 'Expanded {size} card',
      type: 'IMAGE',
      cardVariables: {
        size: 'large',
      },
      content: { ... }
}

Would populate the card as if it was given a title of `Expanded large card`

**Additional context**

We have a request in Monitor to support this functionality. The user would specify a deviceId as a variable in their JSON file, which would get replaced by the value of that dimension.
Bringing it to the component level allows anyone that uses these cards access to this feature.

I have opened a PR with a non-invasive way to accommodate this feature for anyone that wants to use it.

PR: https://github.com/IBM/carbon-addons-iot-react/pull/1099

JoelArmendariz:
Closes #1098 

**Summary**

- Give cards the ability to have a prop of variable values that replace corresponding variables in the card data.

example:
```
{
  size: 'LARGE',
  title: 'From {manufacturer}',
  type: 'VALUE',
  cardVariables: {
    manufacturer: 'Rentech',
  },
  content: { ... }
}
```

Would populate the card as if it was given a title of `Expanded large card`

Added handleValueCardVariables, handleTableCardVariables, and handleTimeSeriesCardVariables to the cardUtilityFuntions file. 
Unfortunately with the current structure of Image card hotspots we cannot re-assign variable values as the user passes in their own component to the hotspot content.


src/utils/cardUtilityFunctions.js
```diff
@@ -122,3 +122,171 @@ export const getUpdatedCardSize = oldSize => {
...
+    updatedTarget = updatedTarget.replace(variableRegex, values[variable]);
+  });
+  return updatedTarget;
+};
+
+export const handleVariables = card => {
+  if (!card.cardVariables) {
+    return card;
+  }
+  const updatedCard = card;
```
sstone2423, line 150: Need to clone the card first to remove the value by reference

JoelArmendariz:
> Looks good.
> 
>     * Information on how to use these variables should be added to storybook as well. Maybe 1 story for each type of card that variables are enabled with knobs and info/readme of how to use them.
> 
>     * Also need to update `CardPropTypes` to include the new prop

Will do. I need to add support for TimeSeries card as well.

src/utils/cardUtilityFunctions.js
JoelArmendariz: Good catch. I will update.

src/utils/cardUtilityFunctions.js
```diff
@@ -122,3 +124,403 @@ export const getUpdatedCardSize = oldSize => {
   }
   return newSize;
 };
+
+// re-usable variable fetcher that returns an array of variables
+// variables are identified by surrounding curly braces i.e. {deviceid}
+export const getVariables = value => {
```
tay1orjones, line 130: Could you provide comments in jsdoc style for this and the other functions? The logic is quite dense and would be beneficial to have some specifics on return values, param types, etc.

```suggestion
/**
 * Find variables that match a given value, identified by surrounding curly braces, i.e. `{deviceid}`
 * @param {string} value - The description of the value
 * @return {array} variables - The description of the variables
 */
export const getVariables = value => {
```

src/utils/cardUtilityFunctions.js
```diff
@@ -122,3 +124,403 @@ export const getUpdatedCardSize = oldSize => {
...
+  let variables = value && typeof value === 'string' ? value.match(/{[a-zA-Z0-9_-]+}/g) : null;
+  if (variables) {
+    variables = variables.map(variable => variable.replace(/[{}]/g, ''));
+  }
+  return variables;
+};
+
+// replace variables from the list of variables that are found on the target with their corresponding value
+export const replaceVariables = (variables, values, target) => {
```
tay1orjones, line 139: jsdoc here as well

```
/**
 * Replace variables from the list of variables that are found on the target with their corresponding value
 * @param {type?} variables - The description of the variables
 * @param {type?} values - The description of the values
 * @param {type?} target - The description of the target
 * @return {array} updatedTarget - The description of the return value
 */
```

src/utils/cardUtilityFunctions.js
```diff
@@ -122,3 +124,403 @@ export const getUpdatedCardSize = oldSize => {
...
+  variables.forEach(variable => {
+    const variableRegex = new RegExp(`{${variable}}`, 'g');
+    updatedTarget = values[variable]
+      ? updatedTarget.replace(variableRegex, values[variable])
+      : target;
+  });
+  return updatedTarget;
+};
+
+export const handleValueCardVariables = (title, content, values, card) => {
```
tay1orjones, line 150: jsdoc here too - I know this seems redundant, but it's particularly important to highlight the nuance here between the logic for ValueCard variables vs ImageCard variables vs TableCard variables etc.

src/utils/cardUtilityFunctions.js
```diff
@@ -122,3 +124,403 @@ export const getUpdatedCardSize = oldSize => {
...
+    return updatedCard;
+  }
+  const { cardVariables } = updatedCard;
+
+  // check for title variables on all cards and replace them
+  const titleVariables = getVariables(title);
+  if (titleVariables) {
+    const updatedTitle = replaceVariables(titleVariables, cardVariables, title);
+    updatedCard.title = updatedTitle || title;
+  }
```
tay1orjones, line 376: This is reused in many different functions, would it make sense to pull out into it's own function to avoid repetition?

JoelArmendariz:
> I did my best to review the code logic - it's quite dense but looks okay. I'm concerned at a glance about the amount of repetition across the functions. I'm not sure what provides the best experience for code readability, but it might be worth a second critical look to see how we might streamline the logic.
> 
> Otherwise just would like to have jsdoc style comments on all the utility functions.

I have added the jsdoc comments, and a few more in-line, as well as extracted some repeated code into external functions.

src/components/Accordion/__snapshots__/AccordionItem.story.storyshot
```diff
@@ -44,12 +44,17 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Accor
...
+                }
+              }
               viewBox="0 0 16 16"
               width={16}
               xmlns="http://www.w3.org/2000/svg"
             >
-              <polygon
-                points="11,8 6,13 5.3,12.3 9.6,8 5.3,3.7 6,3"
+              <path
+                d="M11 8L6 13 5.3 12.3 9.6 8 5.3 3.7 6 3z"
```
tay1orjones, line 57: I think this is an invalid snapshot generated with an outdated version of Carbon. We recently upgraded Carbon to v10.10.3 in #1095 

To fix, pull in latest from `master`, run `yarn install && yarn test:update`. The polygon is valid - style willChange transform above should be removed. Looks like this entire snapshot change should be effectively discarded.

JoelArmendariz:
> This would have been ideal if cardVarialbes could be either a static map or a function that can resolve variables. I'm trying to understand how this would be used, since, it would appear that the variables are static.

The thought here is that the user would be able to retrieve data dynamically on their end, and send the results of their fetches down to the components.
Example in Monitor: Our user would add these variables in their dashboard JSON file to import. Our code would then parse the variables (dataSourceIds in our case), fetch the values for them, and then add them to the cardVariables prop on the card.
Admittedly, this is a feature we could have handled 100% in our code base, but @scottdickerson and I discussed adding the functionality here at the component level so that others could take advantage of it if they had a use case for it.

stuckless:
> > This would have been ideal if cardVarialbes could be either a static map or a function that can resolve variables. I'm trying to understand how this would be used, since, it would appear that the variables are static.
> 
> @stuckless , The thought here is that the user would be able to retrieve data dynamically on their end, and send the results of their fetches down to the components.
> Example in Monitor: Our user would add these variables in their dashboard JSON file to import. Our code would then parse the variables (dataSourceIds in our case), fetch the values for them, and then add them to the cardVariables prop on the card.
> Admittedly, this is a feature we could have handled 100% in our code base, but @scottdickerson and I discussed adding the functionality here at the component level so that others could take advantage of it if they had a use case for it.

I guess I was looking at your sample and I was thinking that the card variable data is in the same structure as the card title, so, to dynamically update the 'size', I have to to give it a new value, but, since I'm already updating that structure, then the title could have been updated just as easily.  I guess that's the part of the card configuration that seems to confuse me.  ie, the Card's view and the Card's data are linked under the same json parent, so I when I'm updating the data, I'm in effect updating everything about the card(s).  If you compare that to say, the Table, it has a fairly complex view structure, but the data is updated independently of the view.  The dashboard seems to treat them the same, and my guess is that for a typical dashboard the view segement and the data segment is loaded separately and then merged into the single structure.

When I see new features, I am trying to understand how I might leverage it.  I'm not asking for changes, and codewise, this looks fine to me.  One thought that I did have was that if you have a lot of variables, then, this code iterates all variables all the time no matter if there are variable tokens in  the string or not.  Another approach might have been to find your variable tokens in the string and only attempt replacements if a variable token is found.   The current code also doesn't know if I am providing a typo in my variable token in the text. ie iv'e used `{Size}` vs `{size}`.

src/utils/cardUtilityFunctions.js
```diff
@@ -2,6 +2,8 @@ import warning from 'warning';
 
 import { CARD_SIZES } from '../constants/LayoutConstants';
 
+/* eslint-disable no-unused-expressions */
```
sstone2423, line 5: This is disabling the entire file and can probably be fixed. Not sure which expressions are being unused though

JoelArmendariz:
> > > This would have been ideal if cardVarialbes could be either a static map or a function that can resolve variables. I'm trying to understand how this would be used, since, it would appear that the variables are static.
> > 
> > 
> > @stuckless , The thought here is that the user would be able to retrieve data dynamically on their end, and send the results of their fetches down to the components.
> > Example in Monitor: Our user would add these variables in their dashboard JSON file to import. Our code would then parse the variables (dataSourceIds in our case), fetch the values for them, and then add them to the cardVariables prop on the card.
> > Admittedly, this is a feature we could have handled 100% in our code base, but @scottdickerson and I discussed adding the functionality here at the component level so that others could take advantage of it if they had a use case for it.
> 
> I guess I was looking at your sample and I was thinking that the card variable data is in the same structure as the card title, so, to dynamically update the 'size', I have to to give it a new value, but, since I'm already updating that structure, then the title could have been updated just as easily. I guess that's the part of the card configuration that seems to confuse me. ie, the Card's view and the Card's data are linked under the same json parent, so I when I'm updating the data, I'm in effect updating everything about the card(s). If you compare that to say, the Table, it has a fairly complex view structure, but the data is updated independently of the view. The dashboard seems to treat them the same, and my guess is that for a typical dashboard the view segement and the data segment is loaded separately and then merged into the single structure.
> 
> When I see new features, I am trying to understand how I might leverage it. I'm not asking for changes, and codewise, this looks fine to me. One thought that I did have was that if you have a lot of variables, then, this code iterates all variables all the time no matter if there are variable tokens in the string or not. Another approach might have been to find your variable tokens in the string and only attempt replacements if a variable token is found. The current code also doesn't know if I am providing a typo in my variable token in the text. ie iv'e used `{Size}` vs `{size}`.

Ah, I definitely see the confusion with my example. I have updated it to be more in line with what we would actually use it for. Size would not be something we would reference as a card variable. As for iterating over all variables, `replaceVariables` will only iterate through the variables found on that specific string, and only if variables are found by `getVariables`.

src/utils/cardUtilityFunctions.js
JoelArmendariz: The linter doesn't like this syntax: 

thresholds?.forEach()

I think it expects some sort of 'else' statement. I can change to only disable those specific instances instead.

sstone2423: Go ahead and leave it since the rule hasn't been updated to understand that syntax

src/components/Accordion/__snapshots__/AccordionItem.story.storyshot
sstone2423: This is due to the component being changed from `AccordionItem` to `AccordionItemDefer`. This file should just be deleted

JoelArmendariz: I have deleted AccordionItem storyshot.

scottdickerson:
@JoelArmendariz sean's two ideas to optionally accept a callback function per variable key as well as to add a validation check on mismatching the title and the variable key make a lot of sense to me,  Can we add those two ideas, I think it'll make this feature more reusable downstream

src/utils/cardUtilityFunctions.js
tay1orjones: We can isolate the disable to a specific line via one of the following

```
alert('foo'); // eslint-disable-line no-alert

// eslint-disable-next-line no-alert
alert('foo');

alert('foo'); /* eslint-disable-line no-alert */

/* eslint-disable-next-line no-alert */
alert('foo');
```

src/components/ValueCard/ValueCard.story.jsx
```diff
@@ -32,6 +32,45 @@ storiesOf('Watson IoT/ValueCard', module)
...
+    'small / with variables',
+    () => {
+      const size = select('size', Object.keys(CARD_SIZES), CARD_SIZES.SMALL);
+      return (
+        <div style={{ width: text('cardWidth', `${getCardMinSize('lg', size).x}px`), margin: 20 }}>
+          <ValueCard
+            title={text('title', 'Title variable is {not-working}')}
+            id="facilitycard"
+            cardVariables={object('variables', {
+              'not-working': 'working',
```
scottdickerson, line 45: {'not-working': (key, card)=>'working'},   variables: PropTypes.objectOf(PropTypes.oneOfType([PropTypes.string,PropTypes.func])).  typeof value === 'string' ? {value} : typeof value === 'function' ? value(key, card) : undefined

JoelArmendariz:
@stuckless @scottdickerson I have added the feature to provide a callbacks to the cardVariables prop as well as made the variables case in-sensitive in the `replaceVariables` function.

src/utils/cardUtilityFunctions.js
JoelArmendariz: I can change to inline, although there are 21 instances of this in that file. Seemed cleaner to disable the whole file. Thoughts @tay1orjones ?

src/components/TableCard/TableCard.story.jsx
```diff
@@ -152,6 +152,55 @@ storiesOf('Watson IoT/TableCard', module)
...
+            cardVariables={cardVariables}
+          />
+        </div>
+      );
+    },
+    {
+      info: {
+        text: `<p>Dynamic variables can be added to the title, threshold severityLabel, threshold label, and threshold value.</p> 
+              <p>The string value of the variable must be wrapped with curly braces. For example, {variableName}</p>
+              <p>A cardVariable object must be supplied to TableCard where the keys will be the variable name and the value will be
```
scottdickerson, line 197: should update to say the value could be a function that returns the value

src/constants/CardPropTypes.js
```diff
@@ -71,6 +71,9 @@ export const ValueCardPropTypes = {
     extraTooltipText: PropTypes.string,
     learnMoreElement: PropTypes.element,
   }),
+  cardVariables: PropTypes.objectOf(
+    PropTypes.oneOfType([PropTypes.string, PropTypes.func, PropTypes.number])
```
scottdickerson, line 75: card variables might be a boolean as well

src/utils/cardUtilityFunctions.js
```diff
@@ -122,3 +122,86 @@ export const getUpdatedCardSize = oldSize => {
...
+
+  variables.forEach(variable => {
+    const insensitiveVariable = variable.toLowerCase();
+    const variableRegex = new RegExp(`{${variable}}`, 'g');
+    // Need to update the target with all lower-case variables for case-insesitivity
+    updatedTarget = updatedTarget.replace(variableRegex, `{${insensitiveVariable}}`);
+
+    if (
+      typeof insensitiveCardVariables[insensitiveVariable] === 'string' ||
+      typeof insensitiveCardVariables[insensitiveVariable] === 'number'
```
scottdickerson, line 162: could be a boolean as well

src/constants/CardPropTypes.js
scottdickerson, line 75: need to document the function args of the callback function somewhere (maybe in the story) so users of the cardVariables function option know what the callback args are

PR #1102, opened 2020-04-20
1065 action buttons visual defects when parent is expanded

### LINKED ISSUE #1065: [Table] Action buttons visual defects when parent is expanded
### Describe the bug

The new fix we did for actions buttons has a small defect when a row is expanded. Seems like action buttons are always show when expanded but the background color is only on hover. So you can see the rows behind it. 

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize
2. Click on the first row toggle
3. Remove hover of expanded row

### Expected behavior

We should not see action buttons when not being hovered. 

### Additional context

image

Related to #1085

bjornalm:
Closes #1065

**Summary**

> The new fix we did for actions buttons has a small defect when a row is expanded. Seems like action buttons are always show when expanded but the background color is only on hover. So you can see the rows behind it.

Applied fix so that actions are not automatically visible once the roe is expanded. Actions are only visible if the user hovers the row or if the action overflow menu is showing.

It was not possible to add any meaningful unit tests for this fix since it relies on hover and css rather than props and js.

- Row actions not showing unless row is hovered, even if row is expanded
- Made sure the row actions are visible even when not hovering if the action overflowmenu is open
- Fixed color of row-actions-container__background depending on color of the row (normal, hover, selected)
- Added unit test to compensate for fallen threshold as a consequence of the small refactoring

 1. to the PRs stories and choose [watson-iot-table--stateful-example-with-expansion-and-column-resize](https://deploy-preview-1102--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize)
2. Hover the rows to make sure the row actions are showing
3. Expand the row and do not hover the original row, make sure the row actions are not showing.
4. Hover the the original row to verify that the row actions are present

_Additional testing for refactored/improved code_
5. select the row using the checkbox, make sure the row actions background is of the same color as the selected row
6. Combine the above tests with clicking the row-action button with overflow menu and hover the menu alternatives. Make sure the row actions stay visible and with the correct background color.


davidicus:
> Is there a reason some rows wouldn't show the action?
> 
> The expanded hover acceptance test works for me with the exception of these cases
> 
> Screen Shot 2020-04-22 at 12 47 51 PM
> 
> Screen Shot 2020-04-22 at 12 47 59 PM

I believe those cells don't have the actions other than the dropdown 

image

src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
```diff
@@ -18,8 +22,27 @@
...
   align-items: center;
+  opacity: 0;
+  @include backgroundGradient($hover-ui);
 }
 
-.#{$prefix}--data-table tbody tr:hover .#{$iot-prefix}--row-actions-container__background {
-  background: linear-gradient(90deg, rgba(229, 229, 229, 0) 0, #{$hover-ui} 16px);
+.#{$prefix}--data-table tbody tr:hover .#{$iot-prefix}--row-actions-container__background,
+.#{$iot-prefix}--row-actions-container__background--overflow-menu-open {
+  opacity: 1;
```
davidicus, line 31:
```suggestion
  opacity: 1;
  transition: opacity $duration--fast-02 motion(entrance, productive);
```

davidicus: I believe this will help with the slight visibility we see on hover for the overflow.

bjornalm: Great, thanks :-)

src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
```diff
@@ -18,8 +22,28 @@
...
+  transition: opacity $duration--fast-02 motion(entrance, productive);
+}
+
+.#{$prefix}--data-table tbody tr {
+  &:not(:hover) .#{$iot-prefix}--row-actions-container__background--overflow-menu-open {
+    @include backgroundGradient(#ffffff);
```
davidicus, line 37:
```suggestion
    @include backgroundGradient($ui-01);
```

src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
```diff
@@ -18,8 +22,28 @@
...
+  }
+
+  &.bx--data-table--selected:hover .#{$iot-prefix}--row-actions-container__background {
+    // The color #cacaca does not appear to have a name in carbon
+    @include backgroundGradient(#cacaca);
```
davidicus, line 42:
```suggestion
    @include backgroundGradient($hover-selected-ui);
```

PR #1105, opened 2020-04-20
fix(accordionitem): restore original defaults, opt-in deferred content rendering

### LINKED ISSUE #1104: [AccordionItem] conditional loading is not configurable, breaks snapshots
### Describe the bug

https://github.com/IBM/carbon-addons-iot-react/pull/940 introduced new default functionality where accordions do not render their contents until they're opened. This "breaks" snapshots. All accordion contents is not included in a snapshot - [see here on our own welcome story](https://github.com/IBM/carbon-addons-iot-react/pull/940/files#diff-85618dbe19fef60042da086397b79b9eL124-L3220).

#940 replaced the previous default behavior and there doesn't appear to be a way to opt-out of the new. It seems this should be put behind a prop for users to opt-into delayed rendering.

tay1orjones:
Closes #1104

**Summary**

- Restores the original functionality to the `AccordionItem` that was inadvertently removed in #940, while preserving the option to defer rendering of `AccordionItem` content to on expand of the section.
- Adds additional logic to preserve the content when a section is opened, and then closed again. (previously closing would remove the content, a perf issue)

- Restore original `AccordionItem` export from Carbon
- Add `AccordionItemDefer`, can be composed as a normal `AccordionItem` within `Accordion`
- regen welcome story (exports table in the accordion was empty! This is what triggered the bug report)

- see the new `AccordionItemDefer` story - you can see the content not rendered to the DOM initially, but then rendered when you expand the section.


tay1orjones:
Yeah, this is a tricky spot - the original PR overwrote default functionality. It wasn't "opt-in" as we've discussed we're aiming for all new features to be. So it feels to me that the original change #940 introduced a bug. The default behavior from Carbon isn't available at all since the `AccordionItem` was overwritten. So my changes here fix the bug, restoring the original behavior. Additionally, it provides a new way to achieve the deferred rendering of content.

In practice this will mean that if your app requires the use of deferred content rendering in Accordions, you will need to update those usages/imports of `AccordionItem` to `AccordionItemDefer`.

I know this isn't ideal, but it seems appropriate in principle. The alternative would've had us exposing default carbon functionality through a different/new component. That feels backwards and potentially confusing for contributors and consumers.

On the dashboard sizing issue - I don't think the original PR, #940, did anything to fix the dashboard sizing issue #939? I haven't made any changes to styling for any component.

stuckless:
No, this fixed it... for Health because the accordion item by default
renders all content and just "hides" it.  That in itself seems wrong.  But
they do the same thing for tabs,etc, so they are consistently bad at it.

When a dashboard is renderinged in the accordion item and hidden then it
does not get rendered correctly.  In fact, we've seen several issues
related to content in the accordion item.

At least I'm aware they are broken.  When they ask, why, I'll know what to
say ;)

Sean.

stuckless:
btw, since we wrap every component before exposing it, this isn't a big
deal for us, and it's the reason why we don't expose carbon or iot
components downstream.  I just have to remember that this is broken the
next time I do a migration to the latest iot components.

src/components/Accordion/AccordionItemDefer.jsx
```diff
@@ -0,0 +1,31 @@
...
+  );
+  const handleToggle = event => {
+    if (props.onHeadingClick) {
+      props.onHeadingClick(event);
+    }
+    setOpenState(!openState);
+    setHasEverBeenOpened(true);
+  };
+  return (
+    <AccordionItem data-testid="accordion-item-deferred" {...props} onHeadingClick={handleToggle}>
```
sstone2423, line 22: Should we be adding the `iot` prefix for data-testid's?

tay1orjones:
@sstone2423 if we encapsulate the stateful logic behind a prop inside `AccordionItem` we will be overriding Carbon's `AccordionItem` directly. This would cause us to copy/paste component source code, stories, proptypes, etc and then have to ensure it's all kept up to date as things potentially change through updating to new minor versions of Carbon.

Good idea on the info section - I'll add some docs there

src/components/Accordion/AccordionItemDefer.jsx
tay1orjones: We haven't established that pattern as they're inherently scoped to our library, but we could if you think that would be helpful!

sstone2423: Sounds good. I was thinking it could clash with others if they're using `data-testid` as well

PR #1132, opened 2020-04-29
[TimeSeriesCard / BarChartCard]Charts upgrade

### LINKED ISSUE #1133: [Charts] update to v30.18
### Summary

Update to latest version, which entails swapping our data format to the new tabular data format 

https://carbon-design-system.github.io/carbon-charts/?path=/story/tutorials--tabular-data-format

sstone2423:
Closes #1133 
Closes https://github.com/carbon-design-system/carbon-charts/issues/580
Closes https://github.ibm.com/wiotp/monitoring-dashboard/issues/977
Closes https://github.ibm.com/wiotp/monitoring-dashboard/issues/978 (no longer attempts to render anything if the data is empty)

**Summary**

- Updated carbon charts to latest version
  - https://github.com/carbon-design-system/carbon-charts/issues/580 is fixed in the latest version
- Updated TimeSeriesCard and BarChartCard to use new tabular data format required for new version. This new format does not change how the user interacts with the card as the data still comes through the same way.

- carbon charts upgraded to v30.18
- Updated `formatChartData` functions in TimeSeriesCard and BarChartCard to reflect the new data format
- Created a mapping function in BarChartCard that determines what to graph on each axis
- Added a color mapping loop that reflects the new format to each card
- Removed primary and secondary chart keys as they're no longer used in new version
- Added new TimeSeriesCard test for rendering with only 1 color set on a multi-line graph (referenced in https://github.com/carbon-design-system/carbon-charts/issues/580)
- Converted 1 styled component to scss in TimeSeriesCard
- Updated TimeSeriesCard test coverage

- Verify that all TimeSeriesCard and BarChartCard render and function as intended


src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -96,12 +244,16 @@ const BarChartCard = ({
...
+  if (data) {
+    chartData = data;
+  } else {
+    /** This caches the chart value */
+    chartData = formatChartData(series, values, groupDataSourceId, timeDataSourceId);
   }
 
-  const isAllValuesEmpty = chartData.datasets.every(set => set.data.every(value => isNil(value)));
+  const isAllValuesEmpty = isEmpty(chartData);
+  console.log(chartData);
```
scottdickerson, line 256: console.log

sstone2423: Fixed

sstone2423:
> These components are under the experimental section in storybook. Experimental components have no contract of semver or API stability until they're removed from experimental. I think we can do away with the transforms and conditional logic to just use the new format from `@carbon/charts`.
> 
> Additionally I think this is a good time to rename the export of these components as `UnstableComponentName` so the expectation is clear not only from the storybook, but the API itself.
> 
> Once this PR is in and the auto-gen release notes are put up, I can edit it to make a special note of these things.
> 
> Edit - looks like `TimeSeriesCard` isn't experimental. For that one we'll need to provide a data formatter to format the old into the new. Keep the export the same. Add a deprecation warning stating the data format has changed (with a link to docs if you have it!) and will be deprecated in v3. Then we can add this as something to include in v3 in #956

@tay1orjones There will be no deprecations or changes from a consumer perspective. The new format only happens under the covers.

I've discussed the discrepancies between the 2 chart cards with design and they will be reconsidering how the BarChartCard should be implemented. I've moved that prop refactor into its own issue that we can fix once design is ready.

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -113,6 +167,17 @@ const BarChartCard = ({
 
   const scaleType = series && series.timeDataSourceId ? 'time' : 'labels';
 
+  const axes = mapValuesToAxes(series, layout);
+
+  // Set the colors for each dataset
+  const colors = { identifier: 'group', scale: {} };
```
scottdickerson, line 173: would be good to export this and write a testcase around it

src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -93,21 +94,37 @@ export const determinePrecision = (size, value, defaultPrecision) => {
...
-      label,
-      ...(color ? { fillColors: [color] } : {}),
-      data:
-        filter(values, dataFilter) &&
-        filter(values, dataFilter) // have to filter out null values from the dataset, as it causes Carbon Charts to break
-          .filter(i => !isNil(i[dataSourceId]))
-          .map(i => ({ date: new Date(i[timeDataSourceId]), value: i[dataSourceId] })),
-    })),
-  };
+  const timestamps = [...new Set(values.map(val => val[timeDataSourceId]))];
```
scottdickerson, line 103: we really need a testcase to cover this function

src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -277,6 +294,16 @@ const TimeSeriesCard = ({
     [isEditable, series]
   );
 
+  // Set the colors for each dataset
+  const colors = { identifier: 'group', scale: {} };
```
scottdickerson, line 298: can we export and write a testcase to cover this function.  Is this sharable between BarChart and TimeSeries?

src/constants/CardPropTypes.js
```diff
@@ -200,6 +200,19 @@ export const TableCardPropTypes = {
   ),
 };
 
+const BarChartDatasetPropType = {
```
scottdickerson, line 203: this is incorrect, it should have PropTypes.shape wrapped around it

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -26,54 +26,108 @@ const { iotPrefix } = settings;
...
+/**
+ * Maps values to left and bottom axes based on whether layout is vertical
+ * or horizontal, and if the series is grouped or time-based
+ *
+ * @param {Object} series the definition of the plotted series
+ * @param {string} layout vertical or horizontal
+ *
+ * @returns {object} { bottomAxesMapsTo: string, leftAxesMapsTo: string }
+ */
+export const mapValuesToAxes = (series, layout) => {
```
scottdickerson, line 96: can we get a testcase around this function?

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -26,54 +26,108 @@ const { iotPrefix } = settings;
  * @returns {object} with a labels array and a datasets array
  */
 export const formatChartData = (series, values) => {
-  let labels = [];
-  let datasets = [];
+  const data = [];
```
scottdickerson, line 29: Can we get a testcase around this function?  Specifically to cover the two cases

src/components/TimeSeriesCard/TimeSeriesCard.jsx
sstone2423: As the props are different at the moment, I think it would be overly complex if it was shared. Once the props are refactored, they should be able to share the function

src/components/BarChartCard/BarChartCard.jsx
sstone2423: Added

src/components/BarChartCard/BarChartCard.jsx
sstone2423: Added

src/constants/CardPropTypes.js
sstone2423: Fixed

src/components/TimeSeriesCard/TimeSeriesCard.jsx
sstone2423: Added

src/components/BarChartCard/BarChartCard.jsx
sstone2423: Added

src/components/TimeSeriesCard/TimeSeriesCard.jsx
sstone2423: Added testcase

PR #1138, opened 2020-04-30
fix(table): locale format any number values

scottdickerson:
Prereqs: https://github.com/IBM/carbon-addons-iot-react/pull/1134

**Summary**

- Fixes issues with numeric formatting in locales.  I made this feature opt-in so it wouldn't break previous Table users.  If you don't pass a locale prop to the Table, it doesn't locale format your numbers.

- fix(Table): add locale option to table, pass through to TableCellRenderer
- fix(TableCellRenderer): render the numeric value in the current locale with the maximum available amount of decimal points

- Check that the Table i18n story is locale formatting in the thousands correctly.


scottdickerson:
@stuckless can you open a separate issue on what you're thinking about on the formatting platform side, or add some comments if there's an existing one?

PR #1142, opened 2020-05-01
fix(Table): number data formatting fixes

scottdickerson:
**Summary**

- fix(Table): the span titles weren't being formatted correctly
- fix(TableCard): the numbers weren't being formatted correctly because I forgot to pass through the locale prop to the Table

- docs(ValueCard): improve the component-level documentation there

- Verify that if you mouseover the Table with localized values (in the Locale story) that the mouseover shows the same localization
- Verify that the Dashboard stories with different locales in their TableCard show their number formats differently


davidicus:
I am not seeing internationalized labels in the dashboard stories. https://deploy-preview-1142--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-dashboard--i18n-labels. 

Am I missing something?

scottdickerson:
@davidicus yeah this one doesn't add translation, just locale formats the decimal and thousands separator correctly

PR #1144, opened 2020-05-04
feat(imagecard): deliver default hotspot content to addons

scottdickerson:
Closes #

**Summary**

- Adds a default hotspot content to addons, previously all consumers had to build their own custom tooltip content

- Allows the ImageCard to show hotspots based on a standard Content panel
- This new Content panel supports localized numeric metric values
- Continue to support passing in custom hotspot content elements

- Verify the new 'basic' ImageCard story uses the out of the box hotspot example


JoelArmendariz:
I'm getting an error in the `ImageCard` -> `basic` story. 'Can't convert null to object'. Looks like it may have been introduced here.

tay1orjones:
It also might be worthwhile to ask if this could be used outside of an ImageCard? If so, could/should it be renamed away from "Image" and also given it's own story?

scottdickerson:
@tay1orjones I do think the ImageCard itself might need locale at some point to successfully render any numeric or timestamp values in it's top level. Similar to our TableCard.  As for other usages of the content, I can rename it 'HotspotContent' because it's really showing attribute details that could be used in any type of card

scottdickerson:
@JoelArmendariz thanks for catching, I fixed the issue

tay1orjones:
@scottdickerson I think we should resist the idea that the components housed here should handle i18n. Locale formatting will always be handled much better by i18n libraries in application code.

Why would numeric, timestamp, date, etc values passed to `ImageCard` not be formatted for locale by the i18n library in application code? I suppose there's a possibility of `ImageCard` needing to know a value _type_ (`number`, `boolean`, etc) to render values appropriately visually, though I don't see that need right now through this.

scottdickerson:
that's exactly the usecase @tay1orjones , we need to pass the value as a number type so other kinds of logic (like precision formatting, rounding, etc) can take place, but once that number is here, we need to format it with the correct locale to display it.

src/components/ImageCard/HotspotContent.jsx
```diff
@@ -0,0 +1,166 @@
...
+                : thresholdMatch.value
+            }`}
+            width={16}
+            height={16}
+            renderIconByName={renderIconByName}
+          />
+        ) : null;
+      return (
+        <div key={`attribute-${dataSourceId}`} style={{ display: 'flex', alignItems: 'center' }}>
+          <div style={{ flex: 1, paddingRight: 20 }}>
```
davidicus, line 119: is there a reason we are inlining these styles here? These should be added to a sass file.

src/components/ImageCard/HotspotContent.jsx
```diff
@@ -0,0 +1,166 @@
...
+            <span style={{ fontWeight: 'bold' }}>{label}:</span>
+          </div>
+          <div
+            style={{
+              flex: 0,
```
davidicus, line 124: Same as above. We should not have inline styles

src/components/ImageCard/HotspotContent.jsx
```diff
@@ -0,0 +1,166 @@
...
+              whiteSpace: 'nowrap',
+              display: 'flex',
+              alignItems: 'center',
+            }}
+          >
+            {thresholdIcon}
+            <span
+              style={{
+                paddingLeft: thresholdIcon ? 5 : 0,
```
davidicus, line 133: this logic is to be performed to populate a css custom property that can then be used in your sass files

src/components/ImageCard/HotspotContent.jsx
```diff
@@ -0,0 +1,166 @@
...
+                      ? value === 0
+                        ? 0
+                        : 3 // for small decimals give 3 spots
+                      : 1, // otherwise 1 spot if precision isn't set
+                    locale
+                  )
+                : value}
+              {unit && value !== '--' && (
+                <span>
+                  &nbsp;
```
davidicus, line 152: Spacing should be done with styles not in HTML

src/components/ImageCard/_image-hotspot-content.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../globals/vars';
+
+.#{$iot-prefix}--hotspot-content h4 {
+  white-space: nowrap;
+  overflow-x: hidden;
+  overflow-y: hidden;
+  text-overflow: ellipsis;
+  font-size: 1.25rem;
+  padding-bottom: 0.5rem;
```
davidicus, line 9: should use carbon tokens

src/components/ImageCard/_image-hotspot-content.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../globals/vars';
+
+.#{$iot-prefix}--hotspot-content h4 {
+  white-space: nowrap;
+  overflow-x: hidden;
```
davidicus, line 5: should combine these. 
```suggestion
  overflow: hidden;
```

src/components/ImageCard/_image-hotspot-content.scss
davidicus: for padding and color

src/components/ImageCard/Hotspot.jsx
```diff
@@ -128,7 +134,11 @@ const Hotspot = ({
         triggerId={`hotspot-${x}-${y}`}
         tooltipId={`hotspot-${x}-${y}`}
       >
-        {content}
+        {React.isValidElement(content) ? (
+          content
+        ) : (
+          <HotspotContent {...content} locale={locale} renderIconByName={renderIconByName} />
+        )}
```
tay1orjones, line 141: Could we promote composition with `children` instead of passing elements as props? This will eliminate the need to prop drill `locale`

```suggestion
        {content ? content : children}
```

Also add a deprecation warning for `content` so we can remove the conditional at the next major release.

tay1orjones:
I see your point on the locale formatting - it’s a bit of a hard spot I think we can improve upon but it needs a broader strategy that we all can agree on and document.

PR #1151, opened 2020-05-06
[HeaderAction] passing down of onClick handler

### LINKED ISSUE #1150: [HeaderAction] onClick handler not being passed
### Describe the bug

[`HeaderGlobalAction`](https://github.com/IBM/carbon-addons-iot-react/blob/3e50b7ab7555ce042a53f6bdbd9948c3ed683827/src/components/Header/HeaderAction/HeaderAction.jsx#L115) is not being passed the onClick handler that is property of the `actionItems` object. 

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-header--header-no-submenu
2. Open up the knobs panel and go to action tab
3. Click on action item button on right hand side
4. See no action being logged

To resolve we should add `onClick` handler to Header Action button and write unit test to make sure we have this functionality covered. 

### Expected behavior

`onClick` handler should have fired and action logged

rawa-resul:
Closes #1150

**Summary**

Added missing onClick prop to HeaderGlobalAction

- Added onClick prop to HeaderGlobalAction
- Added one test in Header to satisfy jest coverage

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-header--header-no-submenu
2. Open up the knobs panel and go to action tab
3. Click on action item button on right hand side
4. See no action being triggered


src/components/Header/HeaderAction/HeaderAction.jsx
```diff
@@ -116,6 +116,7 @@ const HeaderAction = ({ item, index }) => {
       className={`${carbonPrefix}--header-action-btn`}
       key={`menu-item-${item.label}-global-${index}`}
       aria-label={item.label}
+      onClick={item.onClick || (() => {})}
```
tay1orjones, line 119: Could `onClick` be a non required prop on the `item` object with a noop as a defaultProp, instead of this conditional?

src/components/Header/Header.test.jsx
```diff
@@ -80,6 +80,22 @@ describe('Header testcases', () => {
...
+    user: 'JohnDoe@ibm.com',
+    tenant: 'TenantId: Acme',
+    url: 'http://localhost:8989',
+    className: 'custom-class-name',
+    appName: 'Watson IoT Platform ',
+    skipto: 'skip',
+    actionItems: [
+      {
+        label: 'user',
+        onCLick: undefined,
```
scottdickerson, line 93: you should be able to just delete line 93

src/components/Header/HeaderAction/HeaderAction.jsx
rawa-resul: (Header.jsx) the onClick prop itself is not required:
`export const HeaderActionItemPropTypes = {
 label: PropTypes.string.isRequired,
 hasHeaderPanel: PropTypes.bool,
 btnContent: PropTypes.node.isRequired,
 childContent: PropTypes.arrayOf(PropTypes.shape(ChildContentPropTypes)), 
 onClick: PropTypes.func,
};`
but it the entire prop array is made "required" in (HeaderAction.jsx):
`export const HeaderActionPropTypes = {
 item: PropTypes.shape(HeaderActionItemPropTypes).isRequired,
 index: PropTypes.number.isRequired,
};`

tay1orjones: Looking deeper at this - I think what you have is fine. There’s no way to set default props for a nested object without doing a deep merge comparison which is way more overhead than is necessary for this.

PR #1153, opened 2020-05-06
[Table] add padding removed by Carbon upgrade

### LINKED ISSUE #1152: [Table] TableHeader missing padding from Carbon update
### Describe the bug

When upgrading to Carbon styles were introduced that removed the padding from non sorting Table headers. We need to add back in to avoid layout issues. 

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize
2. Open dev tools
3. examine table header that is not a sorted row
4. See the lack of padding on the `th`

Needed for TBI release in June

### LINKED ISSUE #1157: Stateful table: Row data shifts to right, not aligned to the column name.
### Describe the bug

1. I am using v2.75.2 Stateful table and see that the row data for the columns are not aligned with coulmn title. This issue is more evident when there are more columns in the table, like I have 11+ columns. 

image

2. How to make columns width automatically vary with cell content?
 For eg., The ‘Frequency rate’ column shows a custom content in the cell which is implemented using renderDataFunction. Since its width is more than the cell's default width, its expected that it should automatically expand to fit the progress chart instead of it extending to the next cell. 

image

I suspect when stickyHeader is set to true its causing this alignment issues.

JFYI, I have already tried setting wrapCellText prop but it doesnt fix the issue https://github.com/IBM/carbon-addons-iot-react/issues/1156

We have a release in June 2020 and need this to be fixed before to that.

davidicus:
Closes #1152 closes #1157

**Summary**

Table headers that are in a sortable table have had their padding removed. This PR restores padding.

- added padding to _table-head.scss

- Go to story at ?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize
- open inspector and inspect Table header that does not have sorting
- see that padding is applied to all Table headers


tay1orjones:
@davidicus Will this also close #1157?

davidicus:
@tay1orjones yes. adding to the closing statement

Issue #1137, opened 2020-04-30
[RFC] flagging components experimental/unstable


tay1orjones:
### Summary

Currently, experimental/unstable components are only labeled as such in the storybook. Consumers are not notified in any way that a component they are using is subject to change outside of semver.

How can we communicate this expectation of instability to consumers without them having to know to look at the storybook?

### Initial proposal

I think there are two primary ways we can provide the affordance of instability to consumers:

1. Name components with a prefix of `Unstable`. This will make it hyper clear to consumers on import and usage that the component is experimental.
   - When an unstable component is deemed stable, it would be renamed to remove the `Unstable` prefix. 
   - We could attempt to provide a codemod/script for consumers to run on their codebase that would re-write their imports for them, though a 'find all and replace' in a code editor during upgrade is just as easy.
2. Provide a [`deprecate()` type of message](https://github.com/IBM/carbon-addons-iot-react/blob/master/src/internal/deprecate.js) in the console when a component is used that is unstable.
   - This is sort of annoying, but would completely avoid the import rewrites.
   - Carbon took this route with the UIShell before it 

@stuckless @scottdickerson @sstone2423  @bryancboyd @jm986 which would you prefer here? Any alternative ideas?

sstone2423:
I'd vote for the Unstable prefix. It might also help to add a disclaimer to the stories above the preview. I tend to skim over that Experimental label

scottdickerson:
To be honest, I'm just okay with having a bolded comment in the Storybook, I think that's the starting point for interacting with these widgets @tay1orjones

stuckless:
In our storybook we use Nuke symbol to denote components that are are not fully vetted.  ie, these component might still need UI work, etc.  Checkmark, means they are good.

![image](https://user-images.githubusercontent.com/1056553/80752053-e400c680-8af8-11ea-9efe-4e39f94c7ff3.png)

Now along the lines of the `deprecate()`, you could use an `unstable()` and give an unstable message vs decrecation since they mean 2 different things.  In this way, once you move from Unstable to Stable you just remove the `unstable()` messaging, and the downstream consumers don't need to update anything in their code.

sstone2423:
I'd agree that putting it at the storybook level should be informative enough too. Its hard to imagine that developers are using these components without looking at Storybook

stuckless:
[...]

PR #1162, opened 2020-05-07
[datetimepicker] bug when going from absolute to preset and clicking apply

### LINKED ISSUE #1114: [DateTimePicker] Fix issues with sass partial

davidicus:
Closes #1114 

**Summary**

The bug is when you click on custom ranges and then go back to presets. If you click apply an error is thrown from referencing an object that is undefined.  Additionally I removed the `_unstable` prefix that we discussed for our experimental components in #1137 . 

abp

- removed `_unstable` from DateTimePicker.jsx,  DateTimePicker.test.jsx, DateTimePicker.story.jsx, and index.js
- removed the resetPreset function call when you switch between preset and custom in DateTimePicker.jsx
- updated snapshots. 

- Go to the DateTimePicker story
- click on custom range
- go back to preset and see one selected and if you hit apply see the app does not throw an error. 

I checked various combinations of values and applying and could not see why the function I removed was needed. Can  double check multiple values to be sure those values are being added to the input.


src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { mount } from 'enzyme';
 import moment from 'moment';
 
-import __unstableDateTimePicker, {
+import DateTimePicker, {
```
tay1orjones, line 6: The /* eslint-disable react/jsx-pascal-case */ should be deleted in the first line of this file

tay1orjones:
@davidicus It doesn't look like there is an associated pal-tracking issue?

Issue #1158, opened 2020-05-07
[Breadcrumb] Firefox ESR does not support ResizeObserver: Failed to return correct breadcrumb list and title when the breadcrumb is dynamically updated 


RogerRosado:
### Describe the bug

Firefox ESR does not support ResizeObserver: Failed to return correct breadcrumb list and title when the breadcrumb is dynamically updated (Breadcrumb has the “hasOverflow === true” which is defaulted for both Graphite and Health).

The following three (hasOverflow and browser support for ResizeObserver) combinations seems to work fine: 

1. hasOverflow === true”  and “ResizeObserver !== undefined” ( for browsers that DO support ResizeObserver)

2. hasOverflow === false” and “ResizeObserver === undefined” ( for browsers that do NOT support ResizeObserver)

3. hasOverflow is set to false and and “ResizeObserver !== undefined” ( for browsers that DO support ResizeObserver) 
          **_Note_**… for this combination it throws an incorrect Warning: 
“Warning: You have set hasOverflow to true, but the current browser does not support ResizeObserver. You will need to include a ResizeObserver polyfill for hasOverflow to function properly”

### To Reproduce

This can be validated on the Graphite Reference Application under tables:

Using Firefox ESR browser (which does not support ResizeObserver)
Click on WONUM000 and then click on “Next tab”, or back to table and click another work order. 
The breadcrumb display is incorrect:

![image](https://user-images.githubusercontent.com/56408920/81285749-fc5b6e80-902d-11ea-94a5-d25a756f8ff1.png)

### Expected behavior

![image](https://user-images.githubusercontent.com/56408920/81285786-08dfc700-902e-11ea-80c4-cc29ed32f922.png)

### Environment/versions:

- Browser: FireFox ESR or any non-support ResizeObserver browser
- `carbon-addons-iot-react` latest version: 

### Additional context

I debugged and reviewed the Breadcrumb.jsx in Iot; I believe the simplest fix is to modify line 117 from:
      {hasOverflow ? (

To:
      {hasOverflow && browserSupportsResizeObserver ? ( 

Or maybe:
      {hasOverflow && breadcrumbRef.current ? ( 

Breadcrumb.jsx code snippet:

![image](https://user-images.githubusercontent.com/56408920/81285425-7ccd9f80-902d-11ea-957d-1206ceffd7c6.png)

### Specific timeline issues / requests

Health June/July release

davidicus:
[...]

PR #1163, opened 2020-05-07
fix(breadcrumb): browsers that do not support ResizeObserver

RogerRosado:
#1158

Closes #

**Summary**

- summary_here

- items_here

- tests here

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.jsx
+++ b/src/components/Breadcrumb/Breadcrumb.jsx
@@ -65,17 +65,14 @@ const Breadcrumb = ({ children, className, hasOverflow, ...other }) => {
         browserSupportsResizeObserver,
         'You have set hasOverflow to true, but the current browser does not support ResizeObserver. You will need to include a ResizeObserver polyfill for hasOverflow to function properly.'
       );
-    } else {
-      warning(
-        !browserSupportsResizeObserver,
-        'Your browser supports ResizeObserver. Set hasOverflow to true to leverage mobile functionality.'
-      );
     }
   }
 
+  const useOverflow = browserSupportsResizeObserver && hasOverflow;
+
   useEffect(
     () => {
-      if (hasOverflow && breadcrumbRef.current) {
+      if (useOverflow && breadcrumbRef.current) {
         setOverflowItems([]);
         setAfterOverflowItems(childrenItems.slice(1));
         setPrevChildren(children);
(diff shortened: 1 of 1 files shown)


src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -60,10 +60,17 @@ const Breadcrumb = ({ children, className, hasOverflow, ...other }) => {
...
-    warning(
-      browserSupportsResizeObserver && hasOverflow,
-      'You have set hasOverflow to true, but the current browser does not support ResizeObserver. You will need to include a ResizeObserver polyfill for hasOverflow to function properly.'
-    );
+    if (hasOverflow) {
+      warning(
+        browserSupportsResizeObserver,
+        'You have set hasOverflow to true, but the current browser does not support ResizeObserver. You will need to include a ResizeObserver polyfill for hasOverflow to function properly.'
+      );
+    } else {
```
stuckless, line 68: I don't think we need this warning... really we only need the warning if hasOverflow is set to true, and we don't have a ResizeObserver.  If hasOverflow is false but the browser doesn't has a resizeobserver, we don't care.

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -114,7 +121,7 @@ const Breadcrumb = ({ children, className, hasOverflow, ...other }) => {
       ref={browserSupportsResizeObserver ? breadcrumbRef : null}
       data-testid="overflow"
     >
-      {hasOverflow ? (
+      {hasOverflow && browserSupportsResizeObserver ? (
```
stuckless, line 124: There a several hasOverflow checks in the code, so, what I would do is right after (and outside)  the `if (__DEV__)` check I would just reassign the hasOverflow.

`hasOverflow = browserSupportsResizeObserver && hasOverflow;`

That way all the code checks for hasOverflow will be false when `browserSupportResizeObserver` is false.

RogerRosado: Implemented suggested change

src/components/Breadcrumb/Breadcrumb.jsx
RogerRosado: implemented suggested change

src/components/Breadcrumb/Breadcrumb.jsx
```diff
@@ -60,15 +60,19 @@ const Breadcrumb = ({ children, className, hasOverflow, ...other }) => {
   });
 
   if (__DEV__) {
-    warning(
-      browserSupportsResizeObserver && hasOverflow,
-      'You have set hasOverflow to true, but the current browser does not support ResizeObserver. You will need to include a ResizeObserver polyfill for hasOverflow to function properly.'
-    );
+    if (hasOverflow) {
+      warning(
```
stuckless, line 64: I had to go looking for the documentation on `warning`.  That's one counter-intuitive function :(.  This expression is correct, but, at first , I was going to ask that check be `!browserSupportResizeObserver`, but I see the warning is thrown when the expression doesn't evaluate to `true`.

PR #1165, opened 2020-05-08
720 in line editing of data table

### LINKED ISSUE #720: [Table] In-line editing of Data Table

bjornalm:
Closes #720

**Summary**
_Extend the IoT Data Table to allow for editing of lines row-by-row._

The bulk of the features are implemented outside the actual table as a story with only the minimal required changes to the table and the props:

- New `options.hasRowEdit` prop
- view.toolbar.activeBar prop got new value `rowEdit`
- New `actions.toolbar.onShowRowEdit` callback prop (closing the row edit toolbar is handled by the application)
- New `editDataFunction` prop in the columns objects
- New `i18n.editButtonAria` prop for the table

The table story "Basic table with full rowEdit example" shows a simple but full implementation with different kinds of editable table cells, save/cancel buttons and a toast with an undo-option. 
image

image

- Used explicit prop hasRowEdit instead of relying on the presence of editDataFunction. This is more consistent with how the row actions are implemented. The table toolbar will be shown with an "edit" icon button if option `hasRowEdit` is set to be true. 
image
- A new `rowEdit`-state was added to the 'activeBar' prop which previously only contained  'filter' and 'column'. When activeBar is set to `rowEdit` the other functionality that modifies the data of the table is disabled (except for row actions that stay enabled). When view.toolbar.activeBar = rowEdit the following is disabled: 
	- pagination
	- the row selection
	- the table sort
	- the table column reordering
	- table search
- Refactored the repeated code of a div with tooltip-svg-wrapper class into a component called TableToolbarSvgButton. The classname seems 'iot-tooltip-svg-wrapper' seems wrong, but I did not change it for this PR.
- A toolbar action `onShowRowEdit` was added. 	
- An `editDataFunction` was added and got the same signature of the `renderDataFunction` (ie, same 4 parameters) for consistency even though the fourth parameter was not needed for the examples in the storybook.
- Prop `preventInteraction` added to the Pagination in order to completely disable it
- Moved ActiveTableToolbarPropType to `TablePropTypes.js` to keep it DRY
- Added relevant unit tests
- Created relevant story and story info text that exemplifies the usage 

1. Go to https://deploy-preview-1165--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--basic-table-with-full-rowedit-example
2. Verify RTL is working
3. Verify that there is a "row edit" icon in the toolbar. Click the button.
4. Make sure the table rows becomes editable, including the status columns that uses a custom data renderer function.
5. Verify that cancel/save examples works
6. Verify that undo example works

1. Go to https://deploy-preview-1165--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize
2. Verify that there is a "row edit" icon in the toolbar. Click the button.
3. This is example does not implement thet actual row edit toolbar or editable cells, but verify that pagination, row selection,  sorting, search, column reorder and filter is disabled or hidden.


src/components/Table/TableToolbar/TableToolbarSVGButton.jsx
```diff
@@ -0,0 +1,37 @@
...
+      onClick={onClick}
+      role="button"
+      tabIndex={0}
+      onKeyDown={e => {
+        if (e.keyCode === keyCodes.ENTER) onClick();
+      }}
+      data-testid={testId}
+    >
+      {children}
+    </div>
```
tay1orjones, line 31: What's the reason for not just using an IconOnly button from Carbon? [They spread `...other`](https://github.com/carbon-design-system/carbon/blob/468d6ac9a524b031f920bcb8048b8c4e3e865823/packages/react/src/components/Button/Button.js#L33), so we should be able to put all these props `onKeyDown`, `tabIndex`, `data-testid`, etc. on it.

src/components/Table/_pagination.scss
```diff
@@ -23,3 +23,9 @@
     margin-left: 0rem;
   }
 }
+
+// This is a workaround to hide the select controls since the
+// page size select can't be disabled.
+.#{$iot-prefix}--pagination--hide-select .bx--select {
+  display: none;
```
tay1orjones, line 30: Page size select can be disabled via [the `pageInputDisabled` prop on Pagination](https://github.com/carbon-design-system/carbon/blob/468d6ac9a524b031f920bcb8048b8c4e3e865823/packages/react/src/components/Pagination/Pagination.js#L126-L129) from Carbon

src/components/Table/Pagination.jsx
```diff
@@ -26,9 +28,13 @@ const SizedPagination = sizeMe({ noPlaceholder: true })(
 
 SizedPagination.propTypes = {
   isItemPerPageHidden: PropTypes.bool,
+  // It is currently not possible to completely disable the carbon pagination
+  // therefor we use this prop to combine disable and hiding of controls.
+  preventInteraction: PropTypes.bool,
```
tay1orjones, line 33: Why does the pagination select need to be hidden? The disable via [`pageInputDisable` prop](https://github.com/carbon-design-system/carbon/blob/468d6ac9a524b031f920bcb8048b8c4e3e865823/packages/react/src/components/Pagination/Pagination.js#L126-L129) should be enough imo

src/components/Table/_pagination.scss
bjornalm: Unless this has been fixed in carbon during the last days, the problem is that the `pageInputDisabled` prop on Pagination only disables the input for selecting a predefined page, it does not disable the input for defining a new page size. If the use selects a new page size new data might be asynchronously loaded, which should be prevented here. I consider the fact that the pagination can't be completely disabled a bug and I've filed an [issue](https://github.com/carbon-design-system/carbon/issues/6055) for that.

src/components/Table/Pagination.jsx
bjornalm: See comment above. I discussed with Taylor a few options how to handle this, including trying to disable the control via CSS and JS, but we came to the conclusion that hiding it would be the cleanest and safest way.

src/components/Table/TableToolbar/TableToolbarSVGButton.jsx
bjornalm: I don't know what the original reason was. I simply refactored out some duplicate code into this component and I wanted the HTML output to be the same to avoid introducing new regression issues. 

We can change it, but perhaps it is better to do it in a separate issue?

bjornalm:
@stuckless Do you wanna have a quick look to see if this satisfies the needs of your team?

src/components/Table/Pagination.jsx
stuckless: is there a defect opened back to the carbon site to remedy this?

src/components/Table/TableToolbar/TableToolbarSVGButton.jsx
tay1orjones: Opened #1174

PR #1169, opened 2020-05-08
Row conscious TableCard link variables

### LINKED ISSUE #1167: [TableCard] dynamic variables in links
### Describe the bug

Dynamic variables in links need to handle row specific values

### Expected behavior

In the case of a TableCard that has a column of links that has different values for different rows, we need to be able to properly assign each value to it's row specific variable instance.

Screen Shot 2020-05-08 at 2 36 07 PM

### Additional context

The Manufacturer row above has links that need to be directed according to their own row context.

JoelArmendariz:
Closes #1167 

**Summary**

Makes variables in TableCard links row specific.

updated the `columnsWithFormattedLinks` to search for variable values in context of unique rows.


src/components/TableCard/TableCard.jsx
```diff
@@ -208,6 +209,43 @@ const determinePrecisionAndValue = (precision = 0, value, locale) => {
   return '--';
 };
 
+export const createColumnsWithFormattedLinks = (columns, cardVariables) => {
```
sstone2423, line 212: Needs jsdoc comments

src/components/TableCard/TableCard.jsx
```diff
@@ -208,6 +209,50 @@ const determinePrecisionAndValue = (precision = 0, value, locale) => {
...
+        // fetch variables on the href
+        variables = linkTemplate.href ? getVariables(linkTemplate.href) : [];
+      }
+      return {
+        ...column,
+        // eslint-disable-next-line react/prop-types
+        renderDataFunction: ({ value, row }) => {
+          let variableLink;
+          // if we have variables the value based on its own row's context
+          if (variables && variables.length) {
```
tay1orjones, line 234: How did you validate that this works? `variables` isn't defined as a param, and I think is outside the scope of the function, no?

src/components/TableCard/TableCard.test.jsx
```diff
@@ -54,6 +54,11 @@ describe('TableCard', () => {
     );
     expect(zeroMatchingThreshold2).toHaveLength(0);
   });
+  test('createColumnsWithFormattedLinks adds renderDataFunction to columns with links', () => {
+    const columnsWithFormattedLinks = createColumnsWithFormattedLinks(tableColumnsWithLinks);
+    const columnsWithlinks = columnsWithFormattedLinks.filter(column => column.renderDataFunction);
+    expect(columnsWithlinks).toHaveLength(1);
+  });
```
tay1orjones, line 61: It would be good to add another test with `@testing-library/react` that uses the new function to render to the DOM and then assert that the rendered output is a link as intended.

src/components/TableCard/TableCard.jsx
JoelArmendariz: @tay1orjones That is what I thought at first too, but it seems to work fine. Verified with Monitor's use-case when linking my local addons repo to monitor. Also verified in the new TableCard story I have added. I was not able to break it based on variable scope.

src/components/TableCard/TableCard.test.jsx
JoelArmendariz: Hey @tay1orjones, I have updated the story for TableCard. Notice that the links in the 'Link' column are populated based on the scope of that row.
You can change change the variable (to 'count' for example) to see that the link will update the endpoint based on what variable it has. 
I have also added a RTL test case to show that the link will be updated with its row specific variable value.

src/components/TableCard/TableCard.jsx
tay1orjones: Must be the syntax tripping me up. The arrow function binds the entire lexical context, which includes all the variables in the current scope. Glad you verified via `link` too!

src/components/TableCard/TableCard.test.jsx
tay1orjones: Awesome, thanks - looks great!

PR #1185, opened 2020-05-14
feat(header): add optional subtitle for MAS suite type products

### LINKED ISSUE #1182: [Header] Need an ability to add a subtitle for Maximo Application Suite
# Summary

In addition to the title/Product Name in the header, I need the ability to add a subtitle to the header for the Maximo Application suite

Here's a picture:
image
https://ibm.invisionapp.com/share/GANZVA9PJ9D#/screens/319484300

**Additional context**
Just need styling applied to the sub product name

Yeah we're trying to fit it in the MAS 8.0 of Monitor at the end of the month.

### Available extra resources

We can work on it

scottdickerson:
Closes #1182

**Summary**

- Adds a new optional subtitle to our header, mostly used for Application Suite type usecases

image

- feat(Header): support new subtitle rendering
- fix(Header): remove hardcoded bx in class names

- Verify all the header stories

### DIFF
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -12,7 +12,7 @@ import { settings } from '../../constants/Settings';
 
 import HeaderActionGroup from './HeaderActionGroup';
 
-const { prefix: carbonPrefix } = settings;
+const { prefix: carbonPrefix, iotPrefix } = settings;
 
 /** common proptypes associated with child content for a header action */
 export const ChildContentPropTypes = {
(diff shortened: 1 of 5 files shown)


src/components/Header/Header.jsx
```diff
@@ -121,6 +125,7 @@ const Header = ({
       {hasSideNav && <HeaderMenuButton aria-label="Open menu" onClick={onClickSideNavExpand} />}
       <HeaderName href={url} prefix={prefix}>
         {appName}
+        {subtitle ? <div className={`${iotPrefix}--header__subtitle`}>{subtitle}</div> : null}
```
tay1orjones, line 128: Should we use the `HeaderMenuItem` from Carbon here instead? Might be a good idea so we are able to get any styling updates that might come upstream from Carbon. 

[Carbon story here for reference](https://react.carbondesignsystem.com/?path=/story/ui-shell--header-base-w-navigation)

scottdickerson: no only because this subtitle should be considered 'part of the headername and should be clickable as one.  It shouldn't take us to a separate landing page

PR #1186, opened 2020-05-14
fix(valuecard): always use precision if passed

### LINKED ISSUE #1184: [ValueCard] Shouldn't ignore precision value
### Describe the bug

Inside the ValueCard and TimeSeries Card code if it's a small card we ignore the precision value

```
// If the card is xsmall we don't have room for decimals!
  switch (size) {
    case CARD_SIZES.SMALL:
      return Math.abs(value) > 9 ? 0 : precision;
    default:
  }
```
This is confusing end users, we should leave this decision up to them

scottdickerson:
Closes #1184

**Summary**

- Value Card was ignoring the passed precision for small cards. We should always respect it

- refactor(TimeSeriesCard): migrate the determinePrecision to the shared utils function instead of having a copy in both ValueRenderer and TimeSeriesCard

- Verify that the testcases pass, it's okay that the Dashboard story updated because we have enough room to render the slightly more precise values.

### DIFF
--- a/src/components/Dashboard/__snapshots__/Dashboard.story.storyshot
+++ b/src/components/Dashboard/__snapshots__/Dashboard.story.storyshot
@@ -317,7 +317,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Dashb
                             title="62.1 %"
                             value={62.1}
                           >
-                            62
+                            62.1
                           </span>
                         </div>
                         <span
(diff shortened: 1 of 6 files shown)


tay1orjones:
This could be related to the CI failure with coveralls: https://github.com/lemurheavy/coveralls-public/issues/632

A Travis API outage or something perhaps. Going to override and merge.

PR #1188, opened 2020-05-14
fix(table): empty state button always shows

### LINKED ISSUE #1187: [Table] Clear all filters button shouldn't show in empty state when table empty
### Describe the bug

If the table has column filters applied that clears the table, don't show the Clear All Filter buttons in the Empty State, request from @CynZhang 

image

Remove the second Clear All Filters button

scottdickerson:
Closes #1187

**Summary**

- There's no good way to hide the Create New or Clear All Filters Button in the emptyState of the table.

- fix(Table): change onEmptyStateAction to actually be optional if not passed, so the button doesn't render if it's not passed
- fix(Table): change buttonLabelWithFilters to be optionally set, if it's not set and the table is filtered, don't show the button (I did this for backwards compatibility, I'd love to hide the button for all filtered tables for everybody, but in-case some user really wanted it, I allow them to set buttonLabelWithFilters to render a custom button)

- Verify that if onEmptyStateAction is passed the button shows, but otherwise it doesn't show.  This story shows that button is now missing
http://127.0.0.1:3000/?path=/story/watson-iot-table--with-no-results
Verify custom empty state still works:
http://127.0.0.1:3000/?path=/story/watson-iot-table--with-no-data-and-custom-empty-state
Verify no data with onEmptyStateAction function still shows the button:
http://127.0.0.1:3000/?path=/story/watson-iot-table--with-no-data

### DIFF
--- a/src/components/Table/EmptyTable/__tests__/EmptyTable.test.jsx
+++ b/src/components/Table/EmptyTable/__tests__/EmptyTable.test.jsx
@@ -1,0 +1,73 @@
+import EmptyTable from '../EmptyTable';
+
+import React from 'react';
+import { render } from '@testing-library/react';
+
+const commonTableProps = {
+  id: 'tableid',
+  totalColumns: 3,
+  isFiltered: false,
+};
+
+describe('EmptyTable', () => {
+  it('custom empty state', () => {
+    const { queryAllByText } = render(
+      <EmptyTable {...commonTableProps} emptyState={<span>my custom element</span>} />
+    );
+    expect(queryAllByText('my custom element')).toHaveLength(1);
+  });
+  it('no empty state action does not render button', () => {
+    const { queryAllByText } = render(
+      <EmptyTable
+        {...commonTableProps}
+        emptyState={{ message: 'I am empty', buttonLabel: 'clickme' }}
+      />
+    );
+    expect(queryAllByText('I am empty')).toHaveLength(1);
+    expect(queryAllByText('clickme')).toHaveLength(0);
+  });
+  it('empty state action with custom button', () => {
+    const { queryAllByText } = render(
+      <EmptyTable
+        {...commonTableProps}
+        emptyState={{
+          message: 'I am empty',
+          buttonLabel: 'clickme',
+        }}
+        onEmptyStateAction={jest.fn()}
+      />
+    );
+    expect(queryAllByText('I am empty')).toHaveLength(1);
+    expect(queryAllByText('clickme')).toHaveLength(1);
+  });
+  it('empty state action with custom filtered button', () => {
+    const { queryAllByText } = render(
+      <EmptyTable
+        {...commonTableProps}
+        isFiltered
+        emptyState={{
+          message: 'I am empty',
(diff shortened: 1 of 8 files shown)


src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -293,7 +293,7 @@ Map {
           "onChangeSort": [Function],
           "onColumnResize": [Function],
           "onColumnSelectionConfig": [Function],
-          "onEmptyStateAction": [Function],
+          "onEmptyStateAction": null,
```
tay1orjones, line 296: I think the default we've established elsewhere is a noop `()=>{}` instead of `null`

tay1orjones: This will probably bork your conditional logic for the callback, but I'm not sure we should rely on a null check for that anyway.

`()=>{}` returns `undefined`, maybe that's a more appropriate default? but I'm unsure how that would jive with the proptypes declaration for this.

Thoughts?

src/components/Table/EmptyTable/__tests__/EmptyTable.test.jsx
```diff
@@ -0,0 +1,58 @@
...
+  });
+  it('empty state action with custom button', () => {
+    const { queryAllByText } = render(
+      <EmptyTable
+        {...commonTableProps}
+        emptyState={{
+          message: 'I am empty',
+          buttonLabel: 'clickme',
+        }}
+        onEmptyStateAction={jest.fn()}
```
tay1orjones, line 37: Should we assert that the callback is called?

src/components/Table/EmptyTable/__tests__/EmptyTable.test.jsx
```diff
@@ -0,0 +1,58 @@
...
+  totalColumns: 3,
+  isFiltered: false,
+};
+
+describe('EmptyTable', () => {
+  it('custom empty state', () => {
+    const { queryAllByText } = render(
+      <EmptyTable {...commonTableProps} emptyState={<span>my custom element</span>} />
+    );
+    expect(queryAllByText('my custom element')).toHaveLength(1);
```
tay1orjones, line 17: Using `screen` cleans things up a bit! [Additional context from Kent's blog](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen)

```suggestion
    render(<EmptyTable {...commonTableProps} emptyState={<span>my custom element</span>} />);
    expect(screen.queryAllByText('my custom element')).toHaveLength(1);
```

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
scottdickerson: Unfortunately I'd have to call the function to determine whether to render or not and then it would actually fire the function.  I think the noop on functions by defaults is a kind of strange pattern that keeps us from being able to conditionally callbacks or make render decisions based on functional props, which means that our prop model becomes more complicated than it needs to in several cases.

PR #1190, opened 2020-05-14
fix(table): to be consistent apply column filters when typing

### LINKED ISSUE #1187: [Table] Clear all filters button shouldn't show in empty state when table empty

scottdickerson:
Closes #1187

**Summary**

- Other than inside the table column filters, all of our other search and filters immediately search text without requiring we press enter.

- fix(FilterHeaderRow): change from listening to the enter and blur event to apply the filter to immediately applying the filter
- fix(TableHead): focus was getting lost in a filter row after the filter was applied.  The fix is to remove the key generation on FilterHeaderRow

- Go to a stateful table example and test the column filters
http://127.0.0.1:3000/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize
Note that I found a different bug on filtering the Number and Boolean columns, will fix that separately as it already existed


src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
```diff
@@ -231,13 +228,12 @@ class FilterHeaderRow extends Component {
           .map((c, i) => {
             const column = columns.find(item => c.columnId === item.id);
             const columnStateValue = this.state[column.id]; // eslint-disable-line
-            const filterColumnOptions = options => {
+            const filterColumnOptions = memoize(options => {
```
bjornalm, line 231: I don't understand why memoize is used here. The memoized `filterColumnOptions` is created and called only once in the mapping loop of the `ordering` before it is created again. Removing the memoize results in the same amount of calls to the sort function.

Perhaps the Combobox should be its own component that we pass through the `React.memo` HOC on instead?

src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.test.jsx
```diff
@@ -40,22 +40,6 @@ describe('FilterHeaderRow', () => {
...
-    );
-    const desiredState = { col1: 'option1' };
-
-    wrapper.setState(desiredState);
-    wrapper.find('input').simulate('blur');
-
-    expect(commonFilterProps.onApplyFilter).toHaveBeenCalledWith(desiredState);
-  });
-
   it('text input clear button clears filter', () => {
```
bjornalm, line 43: perhaps replace the deleted blur-test with a test that checks that apply fires on change?

src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
scottdickerson: This one I didn't actually add I just moved, perhaps it doesn't work the way the adder thought it would, I can look into

src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.test.jsx
scottdickerson: there is a testcase already for checking onChange, but I can add one for onKeyPress

scottdickerson: ugh, actually I can't simulate keyPress: https://github.com/enzymejs/enzyme/issues/441

src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
scottdickerson: you're right that the memoize isn't really working here, I just put a TODO to investigate this code in the future and reverted my change

src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.test.jsx
tay1orjones: Yeah, anything keyboard related needs to ideally be tested in browser. I've opened issue https://github.com/IBM/carbon-addons-iot-react/issues/1203 for this

tay1orjones:
@scottdickerson were you going to add the prop condition?

PR #1200, opened 2020-05-18
fix(timeseriescard): multipoint tooltip shows all datapoints

### LINKED ISSUE #1199: [TimeSeriesCard] hover tooltip doesn't show all data points in current timestamp
### Describe the bug

When hovering over a timestamp with multiple data points, only 1 data point's values are shown

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-timeseriescard--large-multi-line-custom-color
2. Hover any of the x-axis points and see that there's no gridline showing there's multiple points on that timestamp

### Expected behavior

Should show all points on that timestamp

### Additional context

image

sstone2423:
Closes #1199 

**Summary**

- Tooltip now shows all points associated with the current x-axis timestamp

Screen Shot 2020-05-18 at 11 46 42 AM

- Updated to latest carbon/charts
- Removed gridline option passed to tooltip as it was the culprit causing this odd bug
- Updated the `handleTooltip` to remove some unneeded logic
- Added `scaleType: 'linear'` to y-axis to comply with Carbon's line example

- Go to any TimeSeriesCard story that has multiple lines and hover any point spot in the graph that has multiple data points for the same x-axis timestamp

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1795,17 +1795,17 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
-"@carbon/charts-react@^0.30.18":
-  version "0.30.18"
-  resolved "https://registry.yarnpkg.com/@carbon/charts-react/-/charts-react-0.30.18.tgz#524f10eacaaa482da2aa7bd1e1ef408e5e497ae8"
-  integrity sha512-bBJFRQVUHlbIF3LBhnaLXT7aqQIc3WnqlfMOiWWsRDoz/fPr+7BHV+O0/FvKOJoo0Mq2uzWiyrKJW9bdl7/XTA==
+"@carbon/charts-react@^0.30.24":
+  version "0.30.24"
+  resolved "https://registry.yarnpkg.com/@carbon/charts-react/-/charts-react-0.30.24.tgz#97e45687154b77eaf443c59d513c6d58edc633f3"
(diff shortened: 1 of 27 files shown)


tay1orjones:
Is the update just to keep us current, or is there a feature/fix included in the update that you need? Either way is fine, just curious. Keeping deps up to date just for the sake of it is valid imo 👍 

One thing that is confusing though - all the carbon charts changelogs are essentially blank from v0.30.18 to v0.30.24. If I pull up [the compare view](https://github.com/carbon-design-system/carbon-charts/compare/v0.30.18...v0.30.24), there are changes. Any clue as to why this is and how we can understand what we're getting in the upgrade?

sstone2423:
@scottdickerson @tay1orjones I actually don't need the upgraded version for this update so I'll remove it until we have use of a newer version

yarn.lock
```diff
@@ -1813,6 +1813,17 @@
...
+  version "0.30.24"
+  resolved "https://registry.yarnpkg.com/@carbon/charts/-/charts-0.30.24.tgz#655875b3e2550ba1e0b229ee330b7502ecbd5d49"
+  integrity sha512-VqWPID2iYrJcS0Kf9o7cGLBZPt/Zc3NKSjLzCJ+HixYDQtfljg4YDjdDJy95UrjaKZCNJA7yeEG3w2oqK9iQ7Q==
+  dependencies:
+    "@carbon/colors" "10.4.0"
+    "@carbon/utils-position" "1.1.1"
+    date-fns "2.8.1"
+    lodash-es "4.17.15"
+    resize-observer-polyfill "1.5.0"
+
```
tay1orjones, line 1826: I don't think this should be here. I think if you run `yarn` again with the new update to `package.json` it'll remove it?

tay1orjones: recap our convo - looks like this is due to charts-react having @carbon/charts as a hard dep.

If we pin these deps, we should remove @carbon/charts or add a `resolutions` field to our package.json.

I think we can leave them as caret ranges though for now

sstone2423: Reverted to the old ^ version so this diff is gone now 👍

tay1orjones:
Status checks aren't reporting properly for some reason, both have passed. Overriding status checks requirement and merging

PR #1201, opened 2020-05-18
[DateTimePicker] Date picker issues

### LINKED ISSUE #1179: DateTime Picker: Absolute option missing when showRelativeOption={false} and one of the Preset values set as defaultValue.
### Describe the bug

For the datetimepicker when we set the showRelativeOptions as false and if we set any preset value as a default value, we are not getting the absolute calendar.

DateTime Picker is showing the relative option only not the absolute one.

Below is the snippet used for the component:

<DateTimePicker
      dateTimeMask="YYYY-MM-DD"
      defaultValue={{
        label: 'Last 1 days',
          offset: 2880
      }}
      disabled={false}
      expanded={false}
      presets={[
        {
          label: 'Last 1 days',
          offset: 2880
        },
        {
          label: 'Last 7 days',
          offset: 10080
        },
        {
          label: 'Last 30 days',
          offset: 43200
        }
      ]}
      showRelativeOption={false}
    />

Here is what it looks with below setting. Absolute option is not available.
image

upcoming release?

The fix is required ASAP as we are using this component in the current sprint for our deliverables.

### LINKED ISSUE #1183: DateTime Picker: In Relative option when 'Yesterday' is selected it takes current date instead of yesterday date..
### Describe the bug

In Relative option of DateTime Picker, if we set 'Yesterday' as Relative to option, it’s not taking the yesterday date instead it’s showing today’s date.

relative_datepicker

```
<DateTimePicker
        dateTimeMask="YYYY-MM-DD"
        defaultValue={{
          label: 'Last 30 days',
          offset: 44640
        }}
        disabled={false}
        expanded={false}
        i18n={{
          absoluteLabel: 'Absolute',
          applyBtnLabel: 'Apply',
          backBtnLabel: 'Back',
          calendarLabel: 'Calendar',
          cancelBtnLabel: 'Cancel',
          customRangeLabel: 'Custom range',
          customRangeLinkLabel: 'Custom Range',
          endTimeLabel: 'End time',
          intervalLabels: [
            'days',
            'weeks',
            'months',
            'years'
          ],
          invalidNumberLabel: 'Number is not valid',
          lastLabel: 'Last',
          presetLabels: [
            'Last 7 days',
            'Last 30 days',
            'Last 90 days'
          ],
          relativeLabel: 'Relative',
          relativeLabels: [
            'Yesterday'
            // 'Today'
          ],
          relativeToLabel: 'Relative to',
          startTimeLabel: 'Start time',
          toLabel: 'to',
          toNowLabel: 'to yesterday'
        }}
        intervals={[
          {
            label: 'days',
            value: 'DAYS'
          },
          {
            label: 'weeks',
            value: 'WEEKS'
          },
          {
            label: 'months',
            value: 'MONTHS'
          },
          {
            label: 'years',
            value: 'YEARS'
          }
        ]}
        onCancel={null}
        presets={[
          {
            label: 'Last 7 days',
            offset: 11520,
            id: "7d"
          },
          {
            label: 'Last 30 days',
            offset: 44640,
            id: "30d"
          },
          {
            label: 'Last 90 days',
            offset: 131040,
            id: "90d"
          }
        ]}
        relatives={[
          {
            label: 'Yesterday',
            value: 'YESTERDAY'
          },
        ]}
        showRelativeOption
      />
```

Need this for June 2020 release.

davidicus:
Closes #1183 closes #1179 

**Summary**

This PR addresses both issues above as well as a bug that exist when you click on custom option and switch back to presests. If you click apply it will apply the preset value even if you had not clicked one.

Another bug is with translations where there could be a mismatch of the translation value and the value that is populated in the select dropdown. 

- Changed default story to use `relatives` prop
- Change the without relative option story to have preset as default value (#1179)
- added presets to base value instead of being null
- when resetting the relative value, setting it to the first item in array instead of blank string
- checked if the current selected `relativeWhen` matches a translation string and either show the appropriate string or the value of the select. 

- Go to DateTimePicker story. 
- Go to last story. Try to go to custom range and see that calendar is rendered and not the relative option. 

For the rest just go through stories and click values and then change to other type (preset/custom)


src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -263,36 +259,34 @@ const DateTimePicker = ({
...
-      offset: null,
+      label: presets[0].label,
+      offset: presets[0].offset,
     },
     relative: {
       lastNumber: null,
-      lastInterval: null,
-      relativeToWhen: null,
+      lastInterval: intervals[0].value,
+      relativeToWhen: relatives[0].value,
```
tay1orjones, line 289: Should the new defaults be set in defaultProps instead of functionally applying defaults here? I think the primary benefit to formally declaring in the defaultProps would be for the automatic documentation we have in storybook.

davidicus: These are not defaults necessarily there is no preset, relative prop. These are arrays and we are setting the first item in the array as the base value.

davidicus: We were initializing everything as null but there are times when we need to calculate the value and an onchange of these values has not yet occurred. Defaulting to the first item in the array to avoid the component blowing up.

PR #1204, opened 2020-05-19
1176 single row edit mode

### LINKED ISSUE #1176: [Table] Single row edit mode
[Internal repo 350](https://github.ibm.com/Watson-IoT/pal-tracking/issues/350)

### Summary
This allows users to quickly, and relatively simply edit individual records in a table, without needing to flip the whole thing into a bulk edit mode. Its mostly a nice UX affordance to reduce errors, but also likely helps performance where entire tables cannot be flipped into edit, then submitted.

**Additional context**

image

image

bjornalm:
Closes #1176

**Summary**
This allows users to quickly, and relatively simply edit individual records in a table, without needing to flip the whole thing into a bulk edit mode. Its mostly a nice UX affordance to reduce errors, but also likely helps performance where entire tables cannot be flipped into edit, then submitted.

Table interaction is suspended while in single row edit mode and since the toolbar is still showing it had to be completely disabled, including any active filters and column reorder that were showing. Simply hiding them like we do for the full table edit mode would not be a good solution here since the rows would jump vertically, moving the editable row away from the mouse pointer. That made this PR grow a bit in size and complexity.

- public `isEdit` prop on RowAction objects
- public `isEditMode` prop on RowActionsState objects
- public `view.table.singleRowEditButtons` prop for app defined save/cancel buttons
- public `options.hasSingleRowEdit` prop added. Needed to set consistent larger width of action row column
- updated stateful table to handle the edit row action
- updated stories with examples and texts
- updated unit tests
- added disable option to the TableToolBar buttons

1. Go to table story [watson-iot-table--basic-table-with-full-rowedit-example](https://deploy-preview-1204--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--basic-table-with-full-rowedit-example)
2. Hover a row and click the row action button
3. Select edit. Make sure the table is disabled except for the row that you want to edit.
4. Try different scenarios of editing/save/cancel/undo
5. Uncheck the hasSingleRowEdit and make sure the row action column gets smaller

6. Go to table story [on-iot-table--stateful-example-with-expansion-and-column-resize](https://deploy-preview-1204--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize) which contains lots of functionality, including a single row edit action.
7. Hover a row and click the row action button
8. Select edit. Make sure all the functionality of the table is disabled (the actual row edit is not implemented here)
9. Try the same story and scenario again but have different toolbars active. Make sure they are all disabled.


bjornalm:
> Is it intentional that while I enable the single row edit, I can't actually edit? What happens if I enable multiple rows for edit, how do I "save' each row... what does the edit look like?

Not sure I understand what you mean by that you can't actually edit. The [watson-iot-table--basic-table-with-full-rowedit-example](https://deploy-preview-1204--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--basic-table-with-full-rowedit-example) will let you edit a row and save your change.

Regarding using multiple single row edits at the same time. I have not built anything to control this, instead it will be up to the application to decide how to deal with that, I think you should have the ability to either only allow one singel row edit at a time or allow multiple simultaneous single row edits and deal with the save/cancel logic in the app. I would suggest to only allow one single row  at a time, since we have the main row edit functionality for batch updates. The story is a simplification that does not handle such cases. If you want I can update the story to only allow a single row edit at a time?

stuckless:
I think this underscores the challenges with the current table stories :(  I went to this link, that you included... clicked edit, and it did nothing except disable the table.  https://deploy-preview-1204--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize

So, in the link that you provided I could enable edit... and I could enable this for multiple rows, but when I clicked cancel on one of the rows, it cancelled everything.  Is that right?  If I have multiple rows in edit mode and click save on 1 rows, it does the save on all rows, is that right?  Seems that if I can edit 1 rows, that the save/cancel should be managed per row.

bjornalm:
> I think this underscores the challenges with the current table stories :( I went to this link, that you included... clicked edit, and it did nothing except disable the table. https://deploy-preview-1204--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize
> 
> So, in the link that you provided I could enable edit... and I could enable this for multiple rows, but when I clicked cancel on one of the rows, it cancelled everything. Is that right? If I have multiple rows in edit mode and click save on 1 rows, it does the save on all rows, is that right? Seems that if I can edit 1 rows, that the save/cancel should be managed per row.

Yes, the stories for the table is a bit of a beast. I did write in the "acceptance test instructions" (point 8) that for this story the actual row edit is not implemented, but it is easy to miss. I added that url in the test instructions since it contains _a lot_ of table functionality and wanted you to get the opportunity to test that it all got disabled when the single row edit action was selected.  :-)

In response to your question, see the answer to your question [above](https://github.com/IBM/carbon-addons-iot-react/pull/1204#issuecomment-632136412), but basically, it is up to the application to decide this, I've just provided a very basic example.

bjornalm:
> Keyboard navigation is broken a bit for me.
> 
> 1. Tab to a row, the focus state of the overflow menu isn't visible.
> 2. Hit `space` it opens the overflow menu
> 3. With it open, I can't get into the overflow menu to select edit/delete/etc.
> 
> If I use the mouse to enter edit mode though, the rest of the navigation works well through the inputs and such.

Hmm, you are right, although it is not due to this PR. That issue was probably introduced when we redesigned the row action actions to float over the other cells on hover. I created a jira [here](https://github.com/IBM/carbon-addons-iot-react/issues/1217) and assigned it to myself.

bjornalm:
@stuckless @tay1orjones Could you have another look at this one?

tay1orjones:
@bjornalm 👍 good to merge once the conflict is resolved

PR #1208, opened 2020-05-20
458 table scalability

### LINKED ISSUE #458: [Table] scalability to millions of rows and up to 500 columns
[Internal Watson-IoT/pal-tracking#409](https://github.ibm.com/Watson-IoT/pal-tracking/issues/409)

We need clarity on the specifics of this ask. From a UX standpoint, usability would dictate that we would never surface this many rows or columns to the user. 

Clarifications of the ask should be posted in the thread and any eventual ux design work should be added here to inform the development of this feature.

davidicus:
Closes #458 

**Summary**

- There is an issue with our pagination that when the table is loaded with a large data set it will try to make select options for the dropdown for each page. This causes the table to crash and never render. 

- Introduction of a `maxPages` prop that is passed along with the `view.pagination` object. This prop will limit the amount of pages that the dropdown will try to render. 
- Updated story to include a maxPages prop so we can see the page limit in the story. 

- Go to [Table story](?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize)
- see the pagination size is limited to 2 because of filters.
- Remove filters and see that maxPages is limiting the amount of pages to 3.

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -369,7 +369,6 @@ const Table = props => {
       style={style}
       className={classNames(className, `${iotPrefix}--table-container`)}
     >
-      {console.log('render fired')}
       {/* If there is no items being rendered in the toolbar, don't render the toolbar */
       options.hasFilter ||
       options.hasSearch ||
@@ -570,7 +569,10 @@ const Table = props => {
       visibleData &&
       visibleData.length ? ( // don't show pagination row while loading
         <Pagination
-          {...paginationProps}
+          pageSize={paginationProps.pageSize}
+          pageSizes={paginationProps.pageSizes}
+          page={paginationProps.page}
+          isItemPerPageHidden={paginationProps.isItemPerPageHidden}
           totalItems={paginationProps.totalItems < maxPages ? paginationProps.totalItems : maxPages}
           onChange={actions.pagination.onChangePage}
           backwardText={i18n.pageBackwardAria}
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -42,6 +42,10 @@ const propTypes = {
   columns: TableColumnsPropTypes.isRequired,
   /** Row value data for the body of the table */
   data: TableRowPropTypes.isRequired,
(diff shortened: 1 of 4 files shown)


src/components/Table/Table.jsx
```diff
@@ -568,7 +572,11 @@ const Table = props => {
       visibleData &&
       visibleData.length ? ( // don't show pagination row while loading
         <Pagination
-          {...view.pagination}
+          pageSize={paginationProps.pageSize}
+          pageSizes={paginationProps.pageSizes}
+          page={paginationProps.page}
+          isItemPerPageHidden={paginationProps.isItemPerPageHidden}
+          totalItems={paginationProps.totalItems < maxPages ? paginationProps.totalItems : maxPages}
```
tay1orjones, line 579: Could we cover this in a test? 1001 pages worth of data is given to the table, then assert the select contains only 1000 items.

davidicus: Done

davidicus:
Test added.

davidicus:
@stuckless

src/components/Table/Table.story.jsx
```diff
@@ -511,11 +511,18 @@ storiesOf('Watson IoT/Table', module)
...
+    'Stateful Example with expansion, maxPages, and column resize',
     () => (
       <FullWidthWrapper>
         <StatefulTable
           {...initialState}
+          view={{
+            ...initialState.view,
+            pagination: {
+              ...initialState.view.pagination,
+              maxPages: 30,
```
bjornalm, line 523: We could add a story knob for this

src/components/Table/Table.jsx
stuckless, line 579: Is the calculation correct?   Should it be
```
totalItems={paginationProps.totalItems < (maxPages * paginationProps.pageSize) ? paginationProps.totalItems : (maxPages * paginationProps.pageSize)}
```
Is `totalItems` the total number of items or the total number of pages?

tay1orjones: Yeah it looks like `totalItems` should be multiplied against the `pageSize`.

PR #1218, opened 2020-05-22
fix(table-card): support links in expanded rows

### LINKED ISSUE #1216: TableCards dont support links in expanded rows
### Describe the bug

TableCards support click-able links in their rows, but we need to support in expanded rows as well.

### Expected behavior

TableCard expanded rows should support linkTemplate hrefs as clickable links.

Screen Shot 2020-05-22 at 2 52 39 PM

JoelArmendariz:
Closes #1216 

**Summary**

Add a handleExpandedItemLinks function to update expanded row data with links

Add HandleExpandedItemLinks function
Add stories and test cases

In the stories `table with row expansion and link variables` and `table with row expansion and row specific link variables`, verify that the links in the expanded rows navigate you to a new page based on the href value in the expanded rows data.

Fill in row specific variables (ie 'count', 'pressure', 'hour', or 'alert') in the href knob to see the link dynamically change based on row context.

Screen Shot 2020-05-22 at 2 52 39 PM

### DIFF
--- a/src/components/TableCard/TableCard.story.jsx
+++ b/src/components/TableCard/TableCard.story.jsx
@@ -617,6 +617,101 @@ storiesOf('Watson IoT/TableCard', module)
       </div>
     );
   })
+  .add(
+    'table with row expansion and link variables',
+    () => {
+      const size = select('size', [CARD_SIZES.LARGE, CARD_SIZES.LARGEWIDE], CARD_SIZES.LARGE);
+
+      return (
+        <div style={{ width: `${getCardMinSize('lg', size).x}px`, margin: 20 }}>
+          <TableCard
+            title="Open Alerts"
+            id="table-list"
+            tooltip="Here's a Tooltip"
+            content={{
+              columns: tableColumns,
+              expandedRows: [
+                {
+                  id: 'long_description',
+                  label: 'Description',
+                  linkTemplate: {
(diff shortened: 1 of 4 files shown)


scottdickerson:
@JoelArmendariz for some reason I'm not seeing the variables in the expanded row in either story in the deploy preview: 

image

JoelArmendariz:
> @JoelArmendariz for some reason I'm not seeing the variables in the expanded row in either story in the deploy preview:
> 
> image

Oh sorry. They're only in certain rows. You'll find them if you click on a couple others.

src/components/TableCard/TableCard.story.jsx
```diff
@@ -617,6 +617,101 @@ storiesOf('Watson IoT/TableCard', module)
       </div>
     );
   })
+  .add(
```
sstone2423, line 620: Do we need 2 different stories for this? I think if we have 1 whole story using all of the different variables, that will suffice

JoelArmendariz: We do need two if we want to show both, unfortunately, because a table with row specific vars needs to omit the cardVariables prop. I agree that there are a ton of stories.

sstone2423: Ah I forgot about that. Nevermind then!

PR #1225, opened 2020-05-26
fix(table): filter dropdown overflow should be visible

### LINKED ISSUE #1108: [Table] z-index issue with dropdowns in table filters
z-index issue with dropdowns in table filters and the table below. Dropdown goes behind the table and you are unable to see the dropdown options

### Steps to reproduce:
- go to [Table story](https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize)
- click on filters
- click on select menu
- see table over dropdown

image

tay1orjones:
Closes #1108

**Summary**

- super small fix, just needed to override an `overflow: hidden` coming from Carbon styles.


tay1orjones:
@davidicus Thanks! Fixed, please review

PR #1226, opened 2020-05-26
Bar chart refactor

### LINKED ISSUE #1135: [BarChartCard] inconsistent props with TimeSeriesCard
### Describe the bug

All chart cards should follow the same prop pattern. Additionally, they have the same functionality overall, but have 2 different ways of doing it.

For example, in TimeSeriesCard, the datasets are explicitly defined by the consumer. In BarChartCard, the datasets are created and defined by the component logic.

Additionally, TimeSeriesCard has the option of being a line or bar chart so duplicate code is being written to accomplish the same task. Also, by defining a BarChartCard specifically, should we not get rid of TimeSeriesCard, and create a LineChartCard that would match the BarChartCard pattern?

### Expected behavior

Props are consistent across all charting components.

sstone2423:
Closes #1135 

**Summary**

- PropTypes were very different from our TimeSeriesCard which made refactoring to the new Carbon Charts tabular data format very difficult a few weeks ago. The Monitor team reviewed our most common use-cases and came up with a new prop model that would benefit us more directly as we begin to implement this over the next couple of weeks

- Functionality stays the same as this card will still support simple, grouped, and stacked time or non-timebased graphs

- As we went through the different use-cases, our team noticed that the prop model could be quite different depending on what the user is graphing. One of the main oddities is the idea of giving custom colors for datasets. Our solution was to give 3 options:
  - Each series' dataset could have its own `color` string defined
  - A single `color` object could be defined with the keys being the dataset's label name
  - An array of `color` strings that would simply map to the order of the datasets

With this PR, we could also deprecate the use of BarCharts in `TimeSeriesCard` if wanted as the functionality is now duplicated

- Prop model changed
- Data formatting functions changed to match new prop model
- Tests and stories changed to reflect the new prop model

- Go to all BarChartCard stories and check that they work as intended.


scottdickerson:
Can we add one story Stephen that looks like this?  Stacked by a category along a time-based axis.

image

PR #1233, opened 2020-05-27
fix(datetime picker): set the current selection programmatically 

navjeets:
Closes  https://github.com/IBM/carbon-addons-iot-react/issues/1231

**Summary**

- As per discussion, "defaultValue" prop is being used for programmatically setting the current value. User can pass the new range object as defaultValue and it will update the DateTime Picker component.
For handling the on cancel button scenario, the last applied value is being passed as an input to the parseDefaultValue() which will be used to restore the current value.

- Issue 1231

- Tested the fix with Predict Asset Timeline widget which uses external previous/next buttons to control the current selection in DateTime Picker.


davidicus:
> Thanks for the PR @navjeets! 🎉
> 
> Unfortunately there isn't a commit in this PR with the proper message format to trigger an automatic release when this is merged. There's more detailed info in [our contributing guidelines here](https://github.com/IBM/carbon-addons-iot-react/blob/master/.github/CONTRIBUTING.MD#submitting-pull-requests-and-commits).
> 
> You can [amend the commit with a new message](https://help.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message) to fix.

good catch!

navjeets:
@tay1orjones Thanks. I've updated the commit and the PR title as well.

PR #1234, opened 2020-05-28
refactor(icons): use @carbon/icons-react instead of carbon-icons

### LINKED ISSUE #1148: [Deps] Migrate off carbon-icons
[Internal Watson-IoT/pal-tracking#401](https://github.ibm.com/Watson-IoT/pal-tracking/issues/401)

### Describe the bug

We're using the legacy `carbon-icons` library in some places of our component library. The library hasn't been updated in 2 years and has open security vulnerabilities.

See https://github.com/carbon-design-system/carbon-icons/issues/95 for more info

### To Reproduce

1. Search for usage of `carbon-icons` in our repo
2. See multiple locations we're using the legacy library

### Expected behavior

We should only be using `@carbon/icons` or `@carbon/icons-react`

tay1orjones:
Closes #1148

**Summary**

- Modify usages of `<Icon/>` to use icons directly from `@carbon/icons-react` while retaining the 10 bundled icons we can use by name.

- Updated usages of `Icon` and `bundledIcons`
- Remove `@carbon/icons` as a dep. It's already in our dependency tree because `@carbon/icons-react` has it listed as a hard dep. 
- We unfortunately have to keep `carbon-icons` as a dependency because it's a required peer dep for `carbon-addons-iot-react`

Check relevant stories to ensure their icons are working as expected:

- [Button icons](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-button--icon-only-buttons)
- [Dashboard custom actions](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-dashboard--custom-actions)
- [Dashboard header custom actions](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-dashboard-header--with-custom-actions)
- [ImageCard hotspot icons](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-imagecard--custom-rendericonbyname)
- [List with single row actions](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-experimental-list--with-row-actions-single)
- [Table row actions](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--with-row-expansion-and-actions)
- [valuecard icons](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-valuecard--mediumthin-vertical-3)
- [valuecard renderByName](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-valuecard--small-thresholds-number-custom-rendericonbyname) 
- [tablecard with single actions](https://deploy-preview-1234--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-tablecard--table-with-single-actions)


src/components/ImageCard/HotspotContent.test.jsx
```diff
@@ -122,7 +122,7 @@ describe('HotspotContent', () => {
       />
     );
     // threshold should be found and french formatted
-    expect(getAllByTitle(/temperature > 0,05/)).toHaveLength(2);
```
bjornalm, line 125: Why did this change, and could we write the tests differently so that it wouldn't have to change?

src/components/Table/TableBody/RowActionsCell/RowActionsCell.jsx
```diff
@@ -240,7 +240,9 @@ class RowActionsCell extends React.Component {
                           action.renderIcon ? (
                             <OverflowMenuContent title={action.labelText}>
                               {typeof action.renderIcon === 'string' ? (
-                                <Icon icon={action.renderIcon} description={action.labelText} />
+                                React.createElement(icons[action.renderIcon], {
+                                  'aria-label': action.labelText,
```
bjornalm, line 244: Was changing from from 'description' to 'aria-label' deliberate?

tay1orjones: Yeah, just mirroring what happens when using `<Icon/>`, `description` is used as the `aria-label`

[Link to carbon-components-react `<Icon/>` implementation](https://github.com/carbon-design-system/carbon/blob/f7fca2be6b552c330a4e3e1341e9056ddde50790/packages/react/src/components/Icon/Icon.js#L146)

src/components/ImageCard/HotspotContent.test.jsx
tay1orjones: `<Icon/>` always renders a `<title>` in the `<svg>`. So the `svg` and the `button` were both getting a title. 

Removing the unnecessary title in the `svg` is why it moved down to expect 1. Ideally we'd use [`getByRole` instead](https://testing-library.com/docs/guide-which-query), but here I wanted to preserve the test as is as much as possible.

package.json
```diff
@@ -81,14 +81,13 @@
...
     "@carbon/colors": "10.8.0",
-    "@carbon/icons": "10.9.3",
     "@carbon/icons-react": "10.10.0",
     "@carbon/layout": "10.7.1",
     "@carbon/motion": "10.6.0",
     "@carbon/themes": "10.10.3",
     "carbon-components": "10.10.3",
     "carbon-components-react": "7.10.3",
-    "carbon-icons": "7.0.7",
+    "carbon-icons": "^7.0.7",
```
davidicus, line 90: Should this be pinned to specific version?

tay1orjones: I don't think we need it to be pinned anymore since we're not using it. If an application is using a different patch version, npm/yarn should de-dupe it as normal with no effect on our components.

Ideally we'd just match `carbon-components-react` and put it as a peer dep, but consumers are already relying on us for a one-shot install.

davidicus: Are we sure this is still needed? https://github.com/carbon-design-system/carbon/blob/5926a5a9911d6aeab0865e9cf62f466dd696798c/packages/icons-react/package.json#L31

davidicus: Never mind, I misread the above comment.

davidicus: I would say for consistency to pin it but since we hope to refactor away from this eventually not that big of deal.

tay1orjones:
@scottdickerson could you review as well for a triple sanity check with dashboard related changes here?

PR #1236, opened 2020-05-28
fix(datetime picker): hide tooltips

### LINKED ISSUE #1232: [DateTimePicker] add prop to hide tooltips
### Summary

From @navjeets [Internal Watson-IoT/pal-tracking#430](https://github.ibm.com/Watson-IoT/pal-tracking/issues/430)

> **Description**
> DateTime Picker component being used in Maximo Predict widget has a custom behavior defined on preset values. Instead of 12 months being defined as last 12 months it is set to 8 months in past and 4 months in future.
> This works fine but the tooltip shows the date range from 8 months in past to today. This is misleading for the user as it doesn't reflect the actual range.
> Having an option to disable the tooltip would help avoid any confusion.
> 
> **Name of Pattern or Asset**
> DateTime Picker
> 
> **User outcome**
> Allows to hide tooltip while customizing the preset values behavior.
> 
> **Proposal**
> Provide a property on the DateTime Picker that allows the user to hide the tooltip.

navjeets:
Closes #1232

**Summary**
Added new "hasTooltip" prop to show/hide tooltip on hover over the DateTime Picker box and also the internal value for selected preset in expanded view.

- https://github.com/IBM/carbon-addons-iot-react/commit/3c814fffb47c49b65231bf4d5aef426224c52d02

- Tested the fix with Predict Asset Timeline widget which uses DateTime Picker component.


navjeets:
> Could you update the default DateTimePicker story with a knob for this prop? That way we we can test/validate functionally in the storybook.

Done

tay1orjones:
Yeah, good point. I think we'd just need to revise the content of the tooltip.

https://github.com/IBM/carbon-addons-iot-react/blob/9a566fd97cb264916dd09a97e742feeb6786fc02/src/components/DateTimePicker/DateTimePicker.jsx#L672

`getIntervalValue()` could refactored to be a function prop with the existing function as a default? This would allow custom parsing/info in the tooltip.

navjeets:
@sstone2423 @tay1orjones That's a great idea. If there's a function prop to pass the internal value for presets, there's no need of hiding the tooltip.

navjeets:
Should I open a new issue for adding a function prop for preset internal value and close this one?

navjeets:
@tay1orjones @sstone2423 Created new issue to replace 1232.
https://github.com/IBM/carbon-addons-iot-react/issues/1251

Closing this PR.

PR #1237, opened 2020-05-28
feat(datetime picker): hide time inputs

### LINKED ISSUE #1235: [DateTimePicker] add prop to hide time input fields
### Summary

https://github.ibm.com/Watson-IoT/pal-tracking/issues/425

> **Description**
> We're using the Date time picker in Health timeline charts, but the hour / minute detail is not relevant for our context.
>
> **Name of Pattern or Asset**
> Date Time picker - experimental
>
> **User outcome**
> In Health, when the user is exploring the Asset timeline chart, if they want to look at a specific date range, they need a custom range picker. The chart does not support the granularity of minute and hour, so it would be better to hide the choice rather than take up real estate with noise and force the user to make a choice they don't need to make.
>
> **Proposal**
> Provide a property on the Date time picker that allows the user to hide the hour / minute option.
> 
> **Additional context**
> Add any other context or screenshots about the request here.

navjeets:
Closes #1235

**Summary**

- Added new "hasTimeInput" prop to show/hide time input fields in relative and absolute range selection boxes.

- https://github.com/IBM/carbon-addons-iot-react/commit/54a38b1d2a40b3f6858ddbe4bb20eb1c2496281d

- Tested the fix with Predict Asset Timeline widget which uses DateTime Picker component.


navjeets:
> Could you please update the default DateTimePicker story with a knob for this new prop so we can test it functionally in the storybook?

Done

PR #1238, opened 2020-05-28
feat(table): add tooltips to buttons and shrink icon size to 16 px, upgrade Carbon to v10.12.0

### LINKED ISSUE #975: [Table] Remove our version of TableToolbarSearch when upgrading to Carbon v10.10.x
### Summary

To fix a table search default value bug in Carbon 10.9.x we had to duplicate the TableToolbarSearch component including [their fix](https://github.com/carbon-design-system/carbon/commit/e97a7ed9be2911707066c406544310f6cb76f273).

The PR for this is here: https://github.com/IBM/carbon-addons-iot-react/pull/953

### What to do

- Remove our component
- Reference Carbon's TableToolbarSearch instead
- Change any reference of the prop `value` to `defaultValue`

### LINKED ISSUE #1080: Update to Carbon v10.11.x
### Summary

We will update to the highest patch version of Carbon `v10.11.x` when `v10.12.0` is released.

We always try to lag behind Carbon by one minor version at the highest patch level.

<sub>[Internal issue #363](https://github.ibm.com/Watson-IoT/pal-tracking/issues/363)</sub>

### LINKED ISSUE #1174: [TableToolbar] icons in buttons are wrong size, refactor to use stock icon only button
[Internal Watson-IoT/pal-tracking#407](https://github.ibm.com/Watson-IoT/pal-tracking/issues/407)

### Describe the bug

The buttons in the table toolbar are too big. They should be using stock buttons with `hasIconOnly`, these use icons of size `16`. We're using size `20`.

@lukefirth I'd like your approval on this one. I opened a similar issue on Carbon about the pagination component: https://github.com/carbon-design-system/carbon/issues/6057

### To Reproduce

1. Go to the [`Table/stateful example with expansion and column resize`](https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-expansion-and-column-resize) story
image

### Expected behavior
Using a stock Button with hasIconOnly prop, these should be using the size 16 icons.

image

### Environment/versions:

- `carbon-addons-iot-react` version: `v2.76.7`

### Items to complete

- [ ] Refactor to remove the `src/components/Table/TableToolbar/TableToolbarSVGButton.jsx` introduced in #1165 (related convo [here](https://github.com/IBM/carbon-addons-iot-react/pull/1165#discussion_r423159362))
- [ ] Utilize a `Button` instead, like:
```
<Button
  kind={'ghost'}
  onClick={onToggleFilter}
  testId="TableToolbar-filter-button"
  renderIcon={Filter16}
  iconDescription={i18n.filterButtonAria}
  hasIconOnly
  tooltipAlignment={'end'}
  tooltipPosition={'bottom'}
/>
```
- [ ] Resolve any style defects of icon coloring
- [ ] Ensure tooltip placement is not effected by table z-index or similar

scottdickerson:
Closes #975 
Closes #1174
Closes https://github.ibm.com/Watson-IoT/pal-tracking/issues/399
Closes #1080 

**Summary**
image

Fixes the toolbar buttons to be 16 px instead of 20, to use the official Carbon icon only buttons and to support tooltips.  Also adds an active state to the button.

- chore(carbon): upgrade carbon to latest version 10.12.0 to get TableToolbarSearch fix
- chore(package.json): increase the heap size for storybook (change for new carbon)
- test(*Snapshots): lots of snapshot updates due to the new carbon
- test(StorySnapshotsTest): needed to exclude two new DatePickerInput propTypes errors (opened a bug on Carbon to fix)
- test(Table.story): fix downloadCSVHandler button to actually work in the story
- test(Table.test): new Button elements no longer listen to onKeyDown, but do work with the enter key
- refactor(TableToolbar): change back to Carbon TableToolbarSearch widget, use default Search icon size. Remove local copy of TableToolbarSearch.
- fix(TableToolbarSVGButton): use official Carbon iconOnly button, support active state
- chore(TileGallerySection): remove the Expand/Collapse i18n (change for new carbon)
- fix(Table): allow overflow so that the tooltips can show at the top of the table
- fix(baseTableReducer): because the search propType was renamed to 'defaultValue' from 'value' in a previous track need to use that in the reducer otherwise we will get a propTypes error after search

- Verify in this story `watson-iot-table--stateful-example-with-expansion-maxpages-and-column-resize` that the tooltips, and active button states work correctly.  Verify the keyboard support of all the toolbar buttons


package.json
```diff
@@ -86,8 +86,8 @@
     "@carbon/layout": "10.7.1",
     "@carbon/motion": "10.6.0",
     "@carbon/themes": "10.10.3",
-    "carbon-components": "10.10.3",
-    "carbon-components-react": "7.10.3",
+    "carbon-components": "^10.12.0",
```
davidicus, line 89: These should be pinned to specific version.
```suggestion
    "carbon-components": "10.12.0",
```

package.json
```diff
@@ -86,8 +86,8 @@
...
+    "carbon-components-react": "^7.12.0",
```
davidicus, line 90:
```suggestion
    "carbon-components-react": "7.12.0",
```

src/components/Table/TableToolbar/_table-toolbar.scss
```diff
@@ -26,6 +20,11 @@ div.#{$prefix}--toolbar-action.#{$prefix}--toolbar-search-container-expandable {
   padding-top: $spacing-01;
 }
 
+// small fix to allow tooltips to show outside the toolbar if the batch action isn't active
```
bjornalm, line 23: Very nice, I was just fighting the clip-path myself, trying to figure out how to work around it without causing a breaking change. So you are sure it was only needed for the batch actions and not for anything else? What about custom toolbar content that might rely on it?

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -14728,21 +14756,22 @@ Map {
...
-      "buttonKind": "primary",
-      "disableLabelChanges": false,
-      "disabled": false,
-      "labelText": "Add file",
-      "multiple": false,
-      "onChange": [Function],
-      "onClick": [Function],
-      "role": "button",
-      "tabIndex": 0,
-    },
```
tay1orjones, line 14745: We should look into Carbon changelogs to see why this has changed.

package.json
```diff
@@ -35,7 +35,7 @@
     "lint:stylelint": "stylelint './src/**/*.scss' --syntax scss --ignorePath .gitignore",
     "prepare": "yarn build",
     "publish-npm": "yarn semantic-release",
-    "start": "yarn test:engines && yarn storybook",
+    "start": "yarn test:engines && NODE_OPTIONS='--max-old-space-size=16384' yarn storybook",
```
tay1orjones, line 38: Could you try a `killall node -9` and then try running this command?

I assume you saw a heap out of memory error and added this. These are typically a result of resource issues on the local machine vs a reproducible issue across other machines. We shouldn't set this flag unless necessary.

src/components/Table/TableToolbar/_table-toolbar.scss
scottdickerson: I believe so @bjornalm I almost opened a bug on carbon to only set the clip-path initially if the batch actions was active to make sure they approved.

package.json
scottdickerson: yeah, it's fine if you wanna take it out, but I couldn't even get my initial build to work without it, and as far as I could tell I didn't have anything else eating local memory

scottdickerson:
> A couple things I noticed on the following story
> https://deploy-preview-1238--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--with-resize-hascolumnselection-and-initial-column-widths
> 
> 1. Do we want the tooltips to appear immediately (as they do), to on a timer on hover (like most tooltips in html).  Is there anything in the design for that.   I can see lots of comments coming back from users with inconsistencies if we do it immediately here on these buttons, but, don't do it immediately on all buttons, etc, or just the annoyance of it being immediate.  Once you know the buttons, having the tooltips is more annoying than helpful.
> 2. The tooltips clip off the screen.  It's pretty much a guarantee that tooltips will will larger descriptions esp in different languages, and that these buttons are going to pushed to the edge of the screen (in both rtl and ltr environments)
> 3. The entire window body adds in horz scrollbar.  This is ugly, and it's not like we can actually scroll the page to see the tooltip, since, as soon as you mouse away  the toolttip is gone.
> 
> Almost like we need a universal tooltip manager to mange using tooltips in all cases.

@CynZhang what do you think?

src/components/Table/Table.test.jsx
```diff
@@ -306,18 +305,18 @@ describe('Table', () => {
     expect(renderRowCountLabel2).toHaveLength(0);
   });
 
-  it('enter key should trigger onDownload', () => {
+  it('click should trigger onDownload', () => {
```
davidicus, line 308: do we not still want to test the keyboard functionality?

scottdickerson: we can't @davidicus because the simulate doesn't really simulate button presses, it actually triggers the onKeyDown handler which doesn't exist anymore.

package.json
scottdickerson: no luck with the `killall node -9` even after reboot, I couldn't do the `yarn start` without upping the heapsize!  not sure if something is strange with the new carbon.

scottdickerson:
ok @stuckless resubmitted after talking with @CynZhang and we gonna go with the native browser tooltips for now until Carbon gives us more options to make the default tooltips less intrusive.

scottdickerson:
@CynZhang in the TableCard the icon size no longer matches once it's next to the CardIcons, should we update the CardIcon size as well?

image

CynZhang:
@scottdickerson yes, those icons in the toolbar should all be 16px. Can you also check if the icon colors are the same?

yarn.lock
```diff
@@ -19357,6 +19365,13 @@ use-resize-observer@^5.0.0:
   dependencies:
     resize-observer-polyfill "^1.5.1"
 
+use-resize-observer@^6.0.0:
+  version "6.1.0"
+  resolved "https://registry.yarnpkg.com/use-resize-observer/-/use-resize-observer-6.1.0.tgz#d4d267a940dbf9c326da6042f8a4bb8c89d29729"
+  integrity sha512-SiPcWHiIQ1CnHmb6PxbYtygqiZXR0U9dNkkbpX9VYnlstUwF8+QqpUTrzh13pjPwcjMVGR+QIC+nvF5ujfFNng==
+  dependencies:
+    resize-observer-polyfill "^1.5.1"
+
```
tay1orjones, line 19374: This is coming from Carbon. We use an older version. It would be very beneficial if we updated our version of this dep to also use v6.1.

I've created a separate issue https://github.com/IBM/carbon-addons-iot-react/issues/1250 but we can include here if you'd prefer @scottdickerson

scottdickerson:
@tay1orjones thanks for the fix, I put in the PR with a dependency on yours, and it should be ready to go now

yarn.lock
```diff
@@ -6085,12 +6092,12 @@ capture-stack-trace@^1.0.0:
...
-  version "7.10.3"
-  resolved "https://registry.yarnpkg.com/carbon-components-react/-/carbon-components-react-7.10.3.tgz#667c5e1bce597123ccd6574204ab2b3dc721e4fc"
-  integrity sha512-L583T2qN5tFiYXHBv/iumw5vAE9T5OJ2l7LdEjY8/WUjzUdgA1iJD5tpYKL4U1ciO12zfWITlJBK8g39IupI/Q==
+carbon-components-react@7.12.0:
+  version "7.12.0"
+  resolved "https://registry.yarnpkg.com/carbon-components-react/-/carbon-components-react-7.12.0.tgz#d41125e774fa40e5c5fbc1d536c1923111273f6b"
+  integrity sha512-zsVD9/Dh8Rgl9LOO7vltS8AXj14eCbmSddQrDMkHeq5JSpQWtIqUAI7A7sT2uAhmowi1hk8d0BAroyGzhLHzAg==
   dependencies:
-    "@carbon/icons-react" "^10.9.3"
+    "@carbon/icons-react" "^10.11.0"
```
tay1orjones, line 6100: Looks like we should've upgraded icons-react too. I've opened a new PR: https://github.com/IBM/carbon-addons-iot-react/pull/1263

PR #1254, opened 2020-06-01
feat(datetime picker): new func prop for tooltip value

### LINKED ISSUE #1251: [DateTimePicker] new function prop to set the tooltip value
### Summary

Add a new function prop to set the tooltip value. The default implementation assumes that the preset values are always defined as offset from current time. This may not always be true if the user has a custom definition of preset e.g. in Predict Asset Timeline widget preset of 1 year is defined as 8 months in past and 4 months in future.

Needed for Maximo Predict July release.

### Available extra resources

Will work on its implementation.

navjeets:
Closes #1251

**Summary**

- Added new function prop "getTooltipValue" for providing the tooltip value for current selection.

- https://github.com/IBM/carbon-addons-iot-react/commit/7a98a949e5011e17835719722db6eb18d29d6477

- Set "getTooltipValue" prop and test the displayed tooltip for a preset selection.


src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -589,6 +593,9 @@ const DateTimePicker = ({
    * @returns {string} an interval string, starting point in time to now
    */
   const getIntervalValue = () => {
+    if (getTooltipValue) {
+      return getTooltipValue();
+    }
```
tay1orjones, line 598: `getIntervalValue` is called in more than one place, so we shouldn't call `getTooltipValue()` here.

```suggestion
```

Could we modify `tooltipText` below to use the callback, and fallback to getIntervalValue if it's not truthy? I believe we should also pass the `currentValue` to the callback.


```diff
           <TooltipDefinition
              align="start"
              direction="bottom"
-              tooltipText={getIntervalValue()}
+              tooltipText={getTooltipValue(currentValue) || getIntervalValue()}
              triggerClassName=""
            >
```

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -124,6 +124,8 @@ const propTypes = {
   disabled: PropTypes.bool,
   /** show the relative custom range picker */
   showRelativeOption: PropTypes.bool,
+  /** get the tooltip value for range selections */
+  getTooltipValue: PropTypes.func,
```
tay1orjones, line 128: I think it makes more sense if we rename this to `renderPresetTooltipText` to mirror the prop that it ultimately gets used for, and follow the same naming convention that we see for props that render things (function or otherwise), like `renderIcon`.  The comment/documentation can specify the purpose, parameter, and that it should return a string.

```suggestion
  /**
   * Function hook used to provide the appropriate tooltip content for the preset time
   * picker. This function takes in the currentValue and should return a string message.
   */
  renderPresetTooltipText: PropTypes.func,
```

src/components/DateTimePicker/DateTimePicker.jsx
navjeets: Thanks. I'll make this change.

src/components/DateTimePicker/DateTimePicker.jsx
navjeets: I had named it initially as only for preset tooltip but thought a more generic prop name will leave the option open to set user defined tooltips for other types as well. However, you are correct. The user may expect the custom tooltip to show for other selection types which the component is not doing thereby resulting in a defect.

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -678,11 +685,11 @@ const DateTimePicker = ({
           <div className={`${iotPrefix}--date-time-picker__menu-scroll`}>
             {!isCustomRange ? (
               <OrderedList nested={false}>
-                {getIntervalValue() ? (
+                {renderPresetTooltipText(currentValue) || getIntervalValue() ? (
                   <ListItem
                     className={`${iotPrefix}--date-time-picker__listitem ${iotPrefix}--date-time-picker__listitem--current`}
                   >
-                    {getIntervalValue()}
+                    {renderPresetTooltipText(currentValue) || getIntervalValue()}
```
tay1orjones, line 692: I think these can stay as they are actually. I don't think we need the tooltipText here since this is the actual list and not the tooltip?

navjeets: That's for displaying the internal value at the top of the expanded div. We do want to show the same text as the tooltip.

navjeets:
> The deploy preview is showing an error.
> 
> image

@davidicus Issue resolved after latest changes.

src/components/DateTimePicker/DateTimePicker.jsx
navjeets: @tay1orjones Can I mark this conversation resolved?

PR #1261, opened 2020-06-02
fix(dashboardactions): show tooltips over dashboard actions and use button

scottdickerson:
The Dashboard Header version of #1238  it prereqs that PR being integrated

**Summary**

- Similar to the Table Toolbar Actions, need to add tooltips over the dashboard actions as well

- fix(DashboardHeader): use the official carbon button for the tooltip

- Verify that in the Dashboard Header actions stories, the sample actions render correctly and response to keyboard and mouse presses


src/components/Dashboard/DashboardHeader.jsx
```diff
@@ -84,31 +85,22 @@ const DashboardHeader = ({
...
-                )}
-              </div>
+                kind="ghost"
+                title={action.labelText}
+                renderIcon={
+                  typeof action.icon === 'string' // legacy support for naming the icon by string
+                    ? icons[action.icon]
+                    : React.isValidElement(action.icon)
+                    ? props => React.cloneElement(action.icon, props)
+                    : action.icon // alternatively you can pass the
```
tay1orjones, line 100: unfinished comment here looks like

PR #1269, opened 2020-06-03
fix(card): ux show tooltips over the card actions and use carbon buttons

scottdickerson:
**Summary**

- These are requested UX changes to improve the usability of the Card toolbar buttons

Fixing all of the toolbar buttons to have mouseover tooltips, and be keyboard navigable inside the CardRangePicker menu.  Adding the selected Time Range to the Card title area and indicating that the selected Time Range is active by coloring the Overflow Menu

image

@CynZhang 
Internal defect here: https://github.ibm.com/wiotp/monitoring-dashboard/issues/1004

- fix(Card): add selectTimeRangeLabel prop
- fix(CardRangePicker): show Default in the label section when selected, set focus and color the active timerange
- fix(CardToolbar): Use the official Carbon buttons for the toolbar actions

- Verify the story/watson-iot-card--basic story card actions work correctly and have mouse over text if you leave your mouse over them.


CynZhang:
@scottdickerson we also need to double check the default color for these icons. I suspect the two icons here are in different colors

scottdickerson:
thanks @CynZhang you're right they're not matching, specifically I'll fix the Expand Collapse to match the other icon button colors in the table

scottdickerson:
icon color is fixed now @CynZhang

src/components/Card/CardToolbar.jsx
```diff
@@ -2,35 +2,29 @@ import React from 'react';
...
-    <button
-      type="button"
-      onClick={onClick}
+    <Button
+      kind="ghost"
       className={classNames(
         `${iotPrefix}--card--toolbar-action`,
-        `${iotPrefix}--card--toolbar-svg-wrapper`
+        `${iotPrefix}--card--toolbar-svg-wrapper`,
+        `${prefix}--btn--icon-only` // can't actually use the hasIconOnly prop because we don't want the tooltip
```
tay1orjones, line 22: Is there some context on why the tooltip isn't desired? It's an a11y affordance that I think we should give for these icon buttons - [some context from Carbon](https://github.com/carbon-design-system/carbon/issues/5069#issuecomment-576375526)

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -5285,9 +5285,9 @@ Map {
         "rollingPeriodLabel": "Rolling period",
         "searchPlaceholder": "Search",
         "selectAllAria": "Select all items",
-        "selectDateRangeLabel": "Select time range",
```
tay1orjones, line 5288: To avoid a breaking change we need to leave this in and deprecate it with a warning and map it to the new `selectTimeRangeLabel ` prop.

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -5572,15 +5572,15 @@ Map {
             "selectAllAria": Object {
               "type": "string",
             },
-            "selectDateRangeLabel": Object {
-              "type": "string",
-            },
```
tay1orjones, line 5577: To avoid a breaking change we need to leave this in and deprecate it with a warning and map it to the new `selectTimeRangeLabel ` prop.

src/components/Card/CardToolbar.jsx
scottdickerson: it's how greedy the iconOnly tooltip is @tay1orjones , we've requested a delay be added so we can keep from popping the tooltip too often for those that already understand what this Icon is for: https://github.com/carbon-design-system/carbon/issues/6178.  We are adding the browser tooltip though for accessibility so if you hover you'll see the in-browser version

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
scottdickerson: on this one we agreed that since this prop only existed for a build or two, we were okay with the rename

src/components/Card/CardToolbar.jsx
scottdickerson: the focus stealing is actually a carbon issue, I've opened this to track: https://github.com/carbon-design-system/carbon/issues/6227

PR #1280, opened 2020-06-03
feat(BarChartCard): variables, units, isExpanded, isEditable

### LINKED ISSUE #1248: [BarChartCard] support variables, units, tooltip, and expanded state
### Summary

BarChartCard should support dynamic variables, expanded table state, and units in the same way as TimeSeriesCard.

Additionally, tooltips should look the same as line charts do. The current behavior is to only show the value without a time or dataset reference

### Available extra resources

Will complete myself

sstone2423:
Closes #1248 

**Summary**

Adds full support that TimeSeriesCard currently has:
  - Dynamic cardVariables support
  - Units to be shown in the legend and tooltip
  -  Tooltip that shows the dataset name, value, unit, color of dataset, and time
  - `isExpanded` state with StatefulTable displaying the chart data
    - For time-based charts, the data will be sorted by timestamp
    - For group charts, the data will be sorted by the grouped category
    - For simple charts, it will be a single row with the defined data
  - `isEditable` state with sample / preview data in grayscale colors (similar to `TimeSeriesCard`)
  - Adds explicit default colors if the user doesn't define them (the same ones carbon charts used)

- Moved all bar chart utility functions to `barChartUtils.js`
- Added and updated tests / stories to reflect new features
- Moved `valueFormatter` from `timeSeriesUtils` to common `cardUtils`
- Added a check for null values before formatting the data as it will cause the graph to blow up
- Added stricter card sizes as anything below MEDIUM is unreadable
- Added BarChartCard to CardRenderer and Dashboard
- Added custom propType warnings for unique cases that will cause unexpected side effects
- Added default colors to lines if not explicitly defined
- Added color band to tooltip similar to `TimeSeriesCard`

1. Go to all BarChartCard stories and see that they still function normally
  - Try the `isEditable` and `isExpandable` knobs on each story to see them working
2. Go to Dashboard stories and see that BarChartCard is functioning and displaying correctly

**Screenshots**

`isEditable`

Screen Shot 2020-06-05 at 1 59 00 PM

`isExpanded`

Screen Shot 2020-06-05 at 1 59 57 PM

Regular tooltip

Screen Shot 2020-06-05 at 2 04 39 PM

Time series tooltip

Screen Shot 2020-06-05 at 2 04 51 PM


sstone2423:
@scottdickerson I went ahead and used the same grayscale colors that were being used in TimeSeriesCard

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -3,212 +3,101 @@ import SimpleBarChart from '@carbon/charts-react/bar-chart-simple';
...
+    series.forEach(datasource => {
+      if (datasource.label) {
+        shouldNotGivePropError = false;
+      }
+    });
+    warning(
+      shouldNotGivePropError,
+      `\`BarChartCard\` of type \`STACKED\` with \`categoryDataSourceId\` AND \`timeDataSourceId\` cannot use \`label\` within series. The labels will be created from the \`categoryDataSourceId\`.`
+    );
+  }
```
tay1orjones, line 86: Could these also be custom proptype validation functions, like the additions you made to the CardPropTypes? We wouldn't want to pop these warnings in production.

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -4075,8 +4079,12 @@ Map {
                   "STACKED",
                 ],
               ],
+              "isRequired": true,
```
tay1orjones, line 4082: Is this fixing a defect? Moving a prop to be required is considered a breaking change. If it should have been this way to begin with though, we can publish it under a minor/patch version change.

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -8594,9 +8597,6 @@ Map {
                               "isRequired": true,
                               "type": "string",
                             },
-                            "unit": Object {
-                              "type": "string",
-                            },
```
tay1orjones, line 8599: We'll need to leave this in, mark it deprecated, and map it to the new prop (`prop={oldProp || newProp}` or something) to avoid a breaking change.

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -8563,9 +8569,6 @@ Map {
                           "isRequired": true,
                           "type": "string",
                         },
-                        "unit": Object {
-                          "type": "string",
-                        },
```
tay1orjones, line 8568: We'll need to leave this in, mark it deprecated, and map it to the new prop (`prop={oldProp || newProp}` or something) to avoid a breaking change.

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -4207,8 +4215,6 @@ Map {
       "size": Object {
         "args": Array [
           Array [
-            "SMALL",
-            "SMALLWIDE",
```
tay1orjones, line 4211: To avoid a breaking change, we'll need to leave these in. If someone has a small barchart card configured today and upgrades, what would be the best thing to happen? render `null`, render 1 size up (medium, mediumthin)? I'm not sure, thoughts? In addition I think we can mark these sizes deprecated to be removed in the next major.

src/components/BarChartCard/BarChartCard.jsx
sstone2423: Yep, thought there was some limitations, but there's actually not. Fixing them

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
sstone2423: Swapped it back and set default props to type `SIMPLE`

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
sstone2423: This prop was not being used at all. The mapping was actually using `props.content.unit` instead of `props.content.series[i].unit`, but this was silently happening. If you look at https://carbon-addons-iot-react.com/?path=/story/watson-iot-timeseriescard--large-units , you'll see that even our story wasn't using this position

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
sstone2423: This prop was not being used at all. The mapping was actually using `props.content.unit` instead of `props.content.series[i].unit`, but this was silently happening. If you look at https://carbon-addons-iot-react.com/?path=/story/watson-iot-timeseriescard--large-units , you'll see that even our story wasn't using this position

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
sstone2423: I think the least invasive thing to do is to render 1 size up. `SMALL` -> `MEDIUM`, `SMALLWIDE` -> `MEDIUMWIDE`.

I've added the deprecation notice, but made it custom so that I could add the additional rules

jest.config.js
```diff
@@ -9,12 +9,15 @@ module.exports = {
       functions: 80,
       lines: 80,
     },
-    './src/components/**/!(ColumnHeaderSelect|FilterHeaderRow|TableToolbar|RowActionsCell|RowActionsError|StatefulTable|StatefulTableDetailWizard|CatalogContent|FileDrop|HeaderMenu|Dashboard|CardRenderer|Attribute|UnitRenderer|ImageHotspots|ImageControls|PageHero|PageTitle|EditPage|AsyncTable|ImageCard|WizardHeader|TableHead|ColumnResize|DateTimePicker|TimeSeriesCard).jsx': {
+    './src/components/**/!(ColumnHeaderSelect|FilterHeaderRow|TableToolbar|RowActionsCell|RowActionsError|StatefulTable|StatefulTableDetailWizard|CatalogContent|FileDrop|HeaderMenu|Dashboard|CardRenderer|Attribute|UnitRenderer|ImageHotspots|ImageControls|PageHero|PageTitle|EditPage|AsyncTable|ImageCard|WizardHeader|TableHead|ColumnResize|DateTimePicker|TimeSeriesCard|BarChartCard).jsx': {
```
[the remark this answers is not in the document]
sstone2423: The last 2 functions that haven't been tested are lacking because charts don't properly render. I've added the issue number and a TODO for the future when carbon charts fixes the rendering in jest issue

PR #1281, opened 2020-06-04
feat(card): allow each card to have its own defaultLabel

scottdickerson:
Related to internal issue https://github.ibm.com/wiotp/monitoring-dashboard/issues/1004 Item `#1`

**Summary**

- Simply allows each card to use it's own i18n labels, but defaults to the dashboard version of the i18n label

- fix(Dashboard): move the memoization of i18n down to the CardRenderer layer since each card might be different
- fix(CardRenderer): allow card specific i18n to override the setting at the Dashboard level

- Verify the dashboard stories


scottdickerson:
need to do this a different way

PR #1289, opened 2020-06-05
feat(timeseries): allow timeseries to show times in GMT

### LINKED ISSUE #1287: [TimeSeriesCard] When passed no time information use GMT time
We are using the TimeSeriesCard to show count-of-activity per day, so there is no time info in our data.  Maybe we should be using a different component (DateSeriesCard??), but we would like an option to tell the graph that the data has no time info so use GMT time for all timespan data.

scottdickerson:
Closes #1287

**Summary**

- Adds a new prop to TimeSeriesCard that will prevent the automatic conversion to the browsers local timezone (the default) and instead allow them to show it to users in GMT

- refactor(CardPropTypes): move the TimeSeriesCard propTypes over to the TimeSeriesCard file 
- feat(TimeSeriesCard): add  the showTimeInGMT prop to prevent conversion of the timestamp into the local timezone when formatting
- fix(TimeSeriesCard): prevent double conversion of the date
- test(TimeSeriesCard.story): simplify the stories so they're more understandable from an outsider perspective

- verify in the TimeSeriesCard stories if you change the showInGMT bit it should flip the displayed timestamp to GMT


src/components/TimeSeriesCard/TimeSeriesCard.story.jsx
```diff
@@ -2,21 +2,23 @@ import React from 'react';
...
-import memoize from 'lodash/memoize';
 
 import { COLORS, CARD_SIZES } from '../../constants/LayoutConstants';
 import { getCardMinSize } from '../../utils/componentUtilityFunctions';
-import { getIntervalChartData as getFakeData, chartData } from '../../utils/sample';
+import { getIntervalChartData, chartData } from '../../utils/sample';
 
 import TimeSeriesCard from './TimeSeriesCard';
 
-const getIntervalChartData = memoize(getFakeData);
```
tay1orjones, line 13: I think this still needs to be memoized so that the data isn't regenerated every time a knob changes. Currently flipping the gmt knob gets new random data every time (see the y axis changing)

scottdickerson: unfortunately @tay1orjones if I leave the memoize in, then when you switch between chart stories, the data never changes.  I figured the knob behavior was less bad then the story switching behavior

tay1orjones: Yeah, static slices of functionality/data for each story is the primary intent of storybook though. 

If we want to show different data for different stories, we should probably regen the data at the individual story level instead of globally for the entire chapter/section.

Issue #1210, opened 2020-05-21
[Table]  Create & Save views


bjornalm:
[Internal issue 167](https://github.ibm.com/Watson-IoT/pal-tracking/issues/167)

**Is this an extension to an existing component, pattern, or a brand new asset?**
This is an EDS item.

**What is the rationale for this contribution.**
In certain situations, a user may need to make use of the same resource table for a number of different purposes, which would benefit from different configurations of things like sorting, columns and filters.

This contribution covers the ability to save a particular configuration as a 'view' and later change between saved views.

**Detail the intended behavior**
- The default table gains a dropdown to the left of the toolbar.
- This dropdown is by default 'default view', when a user makes changes to the tables configuration, this changes to 'unsaved view'

- Once a custom view is created through changing table settings. The user can open the settings (disregard incorrect icon usage - should be from settings gear) overflow and choose to 'save view'
Image: 
![image](https://media.github.ibm.com/user/684/files/3dff6480-200e-11ea-894a-59b647fe53b5)

- This will prompt a small dialog to provide a name and description to the custom view and if it should be the view selected by default when the table first loads.
Image: 
![image](https://media.github.ibm.com/user/684/files/46579f80-200e-11ea-9e8c-c528b64e5006)

- Completing this form will save the view as something selectable in the aforementioned dropdown. 
Image: 
![image](https://media.github.ibm.com/user/684/files/4eafda80-200e-11ea-8249-eca38c109700)

- Also in the dropdown of view selection: default view, any saved views and a 'manage views' action.
image:
![image](https://media.github.ibm.com/user/684/files/553e5200-200e-11ea-91fd-4cabeca5dc04)

- The manage views action will pop a modal with a simple table of views that can be edited (name/description/default?) or deleted.
Image: Not provided.

lukefirth:
Adding some clarifications on behavior after discussion with @bjornalm today:

> 1. When should the “Save new view” should be available?

- When a users current configuration of the table is different from the last selection made in the view picker. Eg: making any configuration changes from the 'default' setup of the table in a fresh instance, and after they select 'view X/Y/Z' and then make a change. In other words, the only time it isn't available is when they haven't changed any of the things that form the view (filters, sorts etc) from what we loaded in for them.
[...]

PR #1296, opened 2020-06-10
1282 table manage views dropdown

### LINKED ISSUE #1282: [table] manage views dropdown
### Summary
As part of #1210 we need to create a custom dropdown component.

**Additional context**
image
image

1. It should accept view items to show in the list
2. Show when the current view has been edited
3. Provide additional menu items for "saving" and for "managing" views
4. Provide callbacks for view change, view save and manage views

bjornalm:
Closes #1282

**Summary**
As part of #1210 we need to create a custom dropdown component that can manage the table views. Once the different components of #1210 are finished a full table story example will be created.

- Instead of adding additional complexity to the TableComponent itself a ViewDropdown component was implemented as an external table sub component that the application must pass in via a render prop (customToolbarContent).

- Created a new folder called TableSubcomponents for this purpose. The ViewDropdown component got its own folder with code, styles, tests. 

- Unlike the underlying Dropdown the ViewDropdown component is stateless. The rationale is that the Application will most likely want to be able to change the selected item dynamically. E.g. when a user saves a new view, that view should probably be selected automatically. 

- A new story for TableSubcomponents was created showing how to use the ViewDropdown in a an application that handles the state. 

- The ViewDropdown was implemented to be highly configurable by using a lighter version of the emerging "Overrides Pattern" used by Uber. https://medium.com/@dschnr/better-reusable-react-components-with-the-overrides-pattern-9eca2339f646

- A ViewDropdownItem component was created to handle the special rendering requirements. We are forced to use the itemToString function to show the selected item they way it was designed, that is the cause of some of the complexity of the ViewDropdownItem. See closed PR https://github.com/carbon-design-system/carbon/pull/5578

- Added unit tests

1. Go to story https://deploy-preview-1296--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-tablesubcomponents--viewdropdown
2. Check that all items are rendered correctly
3. Check that there is a "View all" default
4. Check that "Save as new view" and "Manage views" are available
5. Select "Save as new view" and make sure the corresponding action is called but that the selected item is not changed.
6. Select "Manage views" and make sure the corresponding action is called but that the selected item is not changed.
7. Verify that the selected view has postfix "- edited"
8. Uncheck the knob activeViewEdited and verify that the postfix "- edited" is removed.
9. Select the option "My saved 3 with a very long name.." and make sure it is rendered correctly
10. Use the knob to change the wrapper width to '100px' and verify that the icons of selected item and the Manage Views is removed. Verify that the "View" prefix is removed for the selected item.


src/components/TableSubComponents/ViewDropdown/ViewItemPropTypes.js
```diff
@@ -0,0 +1,10 @@
+import PropTypes from 'prop-types';
+
+const ViewItemPropType = PropTypes.shape({
+  id: PropTypes.string.isRequired,
+  text: PropTypes.string.isRequired,
+  public: PropTypes.bool,
```
stuckless, line 6: what is this being used for?

src/components/TableSubComponents/ViewDropdown/ViewItemPropTypes.js
```diff
@@ -0,0 +1,10 @@
...
+  customAction: PropTypes.func,
```
stuckless, line 7: it would be good to document what this function does and the args.  ie, customAction, if used will override the default changeView action and changeView is never called if you use a customAction.

bjornalm:
> Looks good! A few thoughts and questions for clarity before we merge:
> 
> I'm not sure that we should have a table subcomponents folder and section of the storybook. If a component is exported, it's worth having it's own individual entry with 1 to n stories. If a component is exported and intended to be used with another parent/sibling component (loose coupling of sorts), we've named/prefixed them similarly (Table, TableBodyRow, this could be TableViewDropdown). 

There is however a crucial difference between the ViewDropdown and the TableBodyRow, TableToolbar etc. The Table component is in charge of importing and creating the TableBodyRow so the implementation details are hidden from the app developer. He/she does not have to know about the TableBodyRow component. However, all the actions and and configurations will have to go via the Table props, so I would say this coupling is very strong. 

ViewDropdown on the other hand requires the app developer to import it, write the JSX to use it and serve it to the table using a node prop like the `customToolbarContent`. The table does not have to know about the ViewDropdown, so here I would say the coupling is low.

> If a component is intended to only be used within another component and is _not exported_, it should probably be placed in a story variation of that parent/coupled component.

Tricky. This component should be exported, but only makes sense within the table context. The rationale for creating a new stories section was that the table is extremely crowded as is which is becoming a problem. For the the full "manage views" #1210  delivery I am planning on adding a table story that ties it all together, but the individual pieces I thought would better be exemplified outside the table story.  

> 
> What if these two components lived in their own `TableViewDropdown` folder at the `components` level and we renamed to `TableViewDropdown(Item)`?

That works well, but then there will also be a `TableSaveViewModal` and a `TableManageViewsModal` as part of the "manage views" feature. And if we continue on the more decoupled approach when adding features to our table then the component root folder will contain even more of them. That doesn't have to be a problem, but that was the reason that I choose a separate folder to put them all in.

> 
> The PR doesn't modify `index.js` to export `ViewDropdown` or `ViewDropdownItem`, though I'm not sure if they're intended to be exported? In theory we wouldn't have to export it since it's only intended to be used within the table? I'm not sure - let me know what you think.

Good point. Since the app developer will have to create the `ViewDropdown` (at least for a non stateful table) it needs to be exported in `index.js`. I will fix this.

> 
> I don't mind the overrides pattern, but I question if it's needed for this component. Is there a requirement to support alternate layouts beyond the one we're providing by default? 

The spec does not require it. Currently. But that is also part of the point with the pattern as I envision what our usage could be. Most of the times the specs are not 100% complete (or the programmer misses something), and additional requirements/bug fixes are being added to our existing components. The overrides pattern can work as an escape hatch for critical modifications that we did not expect in advance which will allow our clients to move faster. They could still create an issue to have the modified functionality added to our component, and once we add that, they can upgrade and close the escape hatch. In theory this sounds bad, since it could involve double work, but in practice I think it could save scope and deadlines. From what I understand Uber does not see it as an escape hatch but relies on it as the standard way of allowing modifications. 

> I am concerned that the pattern could increase the surface of our API making it more difficult to iterate on component architecture without a major version bump. 

I think your concern is valid. That is why I wanted you and David and the other seniors discussing this. 

>It seems in this case if we didn't have to rely on the `itemToString` "hack" we could invert control and just provide all these as individual composable pieces?

Not sure I understand what you mean here. I think the ViewDropDownItem would have to be a render function prop of the ViewDropdown in that case and how would we expose that as composable piece? The app developer could place a ViewDropDownItem that we offer in a render function that he/she writes, but that can also be done with the itemToString...

> 
> Sorry for the long brain dump 😅
> Happy to hop on a webex or slack call to chat through - I am not opposed to merging this as-is!

I love the brain dump! This is the kind of discussion I was looking for. Like I told @davidicus  regarding the override pattern; if we think that using the pattern the way described above would make sense as a general rule for our components then this could be a start. **But** if we think that we probably aren't going to adopt this pattern, then I should remove it from this PR as well. Only having it on 1-2 components would probably just add confusion.

tay1orjones:
I think we’re on the same page. That’s a good distinction between the ViewDropdown and TableBodyRow and I agree it’s more loosely coupled to Table.

To recap this:

- The component is exported, and will not be baked into the Table. 
- The intended use here however is only to be used with/provided to the Table.

 I think this further confirms for me that we should prefix the names of these components with `Table`, like `TableViewDropdown`. Also put them in their own folder, or in `./src/components/table`, either is fine. 

For the stories - we can add another layer of nesting within the Table section by using the hierarchy separator: `storiesOf('Watson IoT/Table/TableViewDropdown’, module)`  

I know the `Table` section is already huge, but I think nesting stories into logical component groups or breaking subpieces out into their own section is how we solve the problem of being overloaded with confusing and unhelpful stories in sections.

I'm interested in exploring this pattern in selective places. Table and it's subcomponents is a great place to start. 👍

bjornalm:
> I think we’re on the same page. That’s a good distinction between the ViewDropdown and TableBodyRow and I agree it’s more loosely coupled to Table.
> 
> To recap this:
> 
> * The component is exported, and will not be baked into the Table.
> * The intended use here however is only to be used with/provided to the Table.
> 
> I think this further confirms for me that we should prefix the names of these components with `Table`, like `TableViewDropdown`. Also put them in their own folder, or in `./src/components/table`, either is fine.
> 
> For the stories - we can add another layer of nesting within the Table section by using the hierarchy separator: `storiesOf('Watson IoT/Table/TableViewDropdown’, module)`  
> 
> I know the `Table` section is already huge, but I think nesting stories into logical component groups or breaking subpieces out into their own section is how we solve the problem of being overloaded with confusing and unhelpful stories in sections.
> 
> I'm interested in exploring this pattern in selective places. Table and it's subcomponents is a great place to start. 👍
> 
> Edit - also re-reading [the spec and internal convo](https://github.ibm.com/Watson-IoT/pal-tracking/issues/167#issuecomment-20450051) for the full manage views flow - it sounds like the `TableSaveViewModal` and `TableManageViewsModal` should be provided at the application level. We might have example ones in the storybook, but I believe the intent there is to not add those to our exports.

Very good, I will 

1. remove the TableSubComponents and move the ViewDropdown folder to the table folder
2. rename the ViewDropdown to TableViewDropdown
3. make the story a nested story of the table
4. add some smaller changes based on new info from Luke https://github.com/IBM/carbon-addons-iot-react/issues/1210#issuecomment-645534391

PR #1305, opened 2020-06-12
Limit width on long ListItem values with rowActions

### LINKED ISSUE #1304: [ListItem] overflow text pushes rowActions
### Describe the bug

If the value of a ListItem with rowActions is long enough to truncate, it will push the rowActions component out of the ListItem.

### To Reproduce

Create a ListItem that has rowActions, and give it a long value. See that the rowAction is not usable anymore.

### Expected behavior

The value should truncate sooner.

Screen Shot 2020-06-12 at 12 44 00 PM

JoelArmendariz:
Closes #1304

**Summary**

In the case of a ListItem with rowActions that has a long value, we need to specify a width, so that it doesn't push the action component off of the ListItem.

Add a `__with-actions` suffix for ListItem classnames to limit the 100% width.

In ListItem, extend the value that is rendered to be long enough that it needs to truncate. See that the rowAction is still accessible.


JoelArmendariz:
@scottdickerson I had added a knob in the two rowAction stories for ListItem, as they were the ones that this affected.
In value and secondaryValue, add a long string and see the truncation.
I can add new stories with the long text, but thought just adding the text knobs accomplised this.

tay1orjones:
Technically ready to merge as this is experimental and exempt from the 90% test coverage threshold, but would be a very appreciated bonus if you could cover this piece in tests.

JoelArmendariz:
> Technically ready to merge as this is experimental and exempt from the 90% test coverage threshold, but would be a very appreciated bonus if you could cover this piece in tests.

Thanks @tay1orjones , I have updated the PR with a visibility test.

PR #1306, opened 2020-06-12
feat(flyoutmenu): added flyout menu

### LINKED ISSUE #1266: [FlyoutMenu] aka carbon modeless dialog
### Summary

This is a general purpose menu dialog, typically attached to an icon.
A modal dialog was not used because the user needs to be able to see the data in the table.

**Additional context**
[Internal Watson-IoT/pal-tracking#44](https://github.ibm.com/Watson-IoT/pal-tracking/issues/44)

@lukefirth could you provide the spec documents for this? Also is it pertinent to identify if this is in fact modeless? Is `FlyoutMenu` an appropriate title for this?

Will be similar to [what's available in Graphite](https://pages.github.ibm.com/maximo-app-framework/graphite/react/storybook/?path=/story/components-%E2%98%A2-flyout--flyoutcontroller-relative)

### LINKED ISSUE #1267: [Table] table toolbar flyout menu for simple filters
### Summary

The table should support a FlyoutMenu #1266 that can hold content including but not limited to a collection of filters.

**Additional context**
From the 
[Internal Watson-IoT/pal-tracking#7](https://github.ibm.com/Watson-IoT/pal-tracking/issues/7)

> SImple filter pattern design: Complete/approved. Visible in design site for tables here: https://pages.github.ibm.com/Watson-IoT/IoT-Design-Site/components/datatable/tablefilter
> 
> Will be added to an overall 'filter' pattern when created on the site to cover the page level version that APM uses. This may be completed very soon, but is not technically part of the development workload for the MAS MVP release.

The development implementation outcome of this should be an example of putting the FlyoutMenu from #1266 into the table.

Simple filters will not be standardized, we'll simply provide the ability to place custom content within the flyout in the table toolbar.

marcJV:
Closes #1266
Closes #1267

**Summary**

Added fly out component

Added FlyoutMenu.jsx, FlyoutMenu.story.jsx, tests

Verify updating directions cause the modal to change which side it is rendered in relation to the button. Test button functionality.


bjornalm:
I think the "closes link" is wrong. It is the internal issue number, but we should probably use the external one, which I assume is #1267

bjornalm:
It looks like content can leak outside the flyout
image

bjornalm:
You can check how the component look in Right To Left mode using the story knob RTL
image

bjornalm:
Is the design correctly implemented? Ignoring the content and purely looking at the design specs linked from the issue #1267 it doesn't look exactly the same:

image

vs

image

bjornalm:
Looks like there are some visual differences between the spec and implementation (ignoring the actual content) with colors, shadow & spacing

image
 
vs 

image

src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -0,0 +1,117 @@
...
+import { Tooltip, Button } from 'carbon-components-react';
+
+import { settings } from '../../constants/Settings';
+
+const { iotPrefix } = settings;
+
+const FlyoutMenu = props => {
+  const { children, onApply, onCancel } = props;
+  return (
+    <Tooltip {...props} renderIcon={SettingsAdjust}>
```
bjornalm, line 15: Not sure how to work this into the current tooltip (you might have to define your own tooltip trigger button) but according to the spec the triggering button should stay in "focus mode" while the flyout is open.
https://github.ibm.com/Watson-IoT/pal-tracking/issues/44#issuecomment-16176683

src/components/FlyoutMenu/FlyoutMenu.test.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { mount, shallow } from 'enzyme';
+
+import FlyoutMenu from './FlyoutMenu';
+
+describe('FlyoutMenu', () => {
```
bjornalm, line 6: I would also test the callbacks, and you can also target the testIDs instead of the classes.

src/components/FlyoutMenu/_flyout-menu.scss
```diff
@@ -0,0 +1,26 @@
+@import '../../globals/vars';
+@import '~carbon-components/scss/globals/scss/layout';
+
+.#{$iot-prefix}-flyout-menu {
+  color: initial;
+  background: $ui-background;
+  min-width: 300px;
+  .bx--tooltip__caret {
```
bjornalm, line 8: replace 'bx' with #{prefix} since that is the carbon prefix

src/components/FlyoutMenu/_flyout-menu.scss
```diff
@@ -0,0 +1,26 @@
...
+    padding-top: $spacing-03;
+    margin-top: $spacing-03;
+  }
+
+  &__cancel {
+    margin-right: $spacing-03;
+  }
+}
+
+.#{$iot-prefix}-flyout-menu.bx--tooltip {
```
bjornalm, line 24: #{prefix}

src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -0,0 +1,117 @@
+import './_flyout-menu.scss';
```
tay1orjones, line 1: This should be moved to the central sass entrypoint `./src/styles.scss`. We don't import styles directly in the component. The compiled sass is provided in the exported bundle and also provided to the storybook environment.

src/components/FlyoutMenu/FlyoutMenu.test.jsx
tay1orjones: Yes - also using `testing-library/react` and `testing-library/dom` instead of `enzyme` would be preferred.

src/components/FlyoutMenu/_flyout-menu.scss
```diff
@@ -0,0 +1,26 @@
+@import '../../globals/vars';
+@import '~carbon-components/scss/globals/scss/layout';
+
+.#{$iot-prefix}-flyout-menu {
+  color: initial;
+  background: $ui-background;
+  min-width: 300px;
```
tay1orjones, line 7: I believe there are some use cases floating around from the design team that call for a flyout smaller than 300px width.

```suggestion
```

src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -0,0 +1,117 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { SettingsAdjust16 as SettingsAdjust } from '@carbon/icons-react';
+import { Tooltip, Button } from 'carbon-components-react';
```
tay1orjones, line 6: We can import these relative to the project instead of directly from `carbon-components-react` so existing and future updates/overrides we make are available to this component.

```suggestion
import Tooltip from '../Tooltip';
import Button from '../Button';
```

src/components/FlyoutMenu/FlyoutMenu.story.jsx
```diff
@@ -0,0 +1,86 @@
+import React, { useState } from 'react';
+import { storiesOf } from '@storybook/react';
+import { select } from '@storybook/addon-knobs';
+import { action } from '@storybook/addon-actions';
+import { Button, Tabs, Tab, Dropdown } from 'carbon-components-react';
```
tay1orjones, line 5: Same thing here - we can import the components relative to the project instead of directly from `carbon-components-react`

```suggestion
import Button from '../Button';
import Tab from '../Tab';
import Tabs from '../Tabs';
import Dropdown from '../Dropdown';
```

tay1orjones:
@lukefirth could you do a visual review for us on this? should the top of the flyout be "attached" and directly against the bottom/edge of the trigger button? Could you share an updated spec if things are off a bit?

* [Isolated flyout stories here](https://deploy-preview-1306--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-flyout-menu--basic)
* [Flyout in table toolbar story here](https://deploy-preview-1306--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-maxpages-and-column-resize)

src/components/FlyoutMenu/FlyoutMenu.test.jsx
bjornalm: @tay1orjones Oh, I didn't know that. How come, are we moving away from enzyme?

lukefirth:
Apologies that its taken me so long to come here, but I did not wish to come empty-handed, and I hadn't quite realised the depths of missing logic/definition in the spec we received...

There's a lot going on in what I have to offer around design logic etc for this, so I have updated https://github.com/IBM/carbon-addons-iot-react/issues/1266 to include it all. Please totally feel free to reach out and start a conversation there, here or on slack etc!

tay1orjones:
Thanks @lukefirth! 

@marcJV let's update this PR to match the visual spec luke linked in the original issue:
spec

The positioning logic is outside of the scope of this PR and for the time being will need to be handled at the application level, just like regular tooltips are today.

src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -0,0 +1,223 @@
...
+
+  light: PropTypes.bool,
+};
+
+const defaultProps = {
+  buttonProps: {
+    className: '',
+    renderIcon: SettingsAdjust,
+    size: 'default',
+  },
```
tay1orjones, line 198: I think these should just be at the base/root level and not nested in an object. The button classname probably isn't needed as an option, consumers would target the base `className`, affix `button` and/or `#{$prefix}--btn` to their selector like `.my-flyout button#{$prefix}--btn`

src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -0,0 +1,223 @@
...
+
+  /**
+   * The CSS class names of the flyout menu.
+   */
+  className: PropTypes.string,
+
+  /**
+   * The CSS class names of the trigger UI.
+   */
+  triggerClassName: PropTypes.string,
```
tay1orjones, line 137: It doesn't look like this is used anywhere, can be removed
```suggestion
```

src/components/FlyoutMenu/FlyoutMenu.jsx
tay1orjones: Also would be awesome if `size` was a knob in one of the stories so we could see and validate the visual for it.

src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -0,0 +1,223 @@
...
+  onCancel: PropTypes.func.isRequired,
+
+  /**
+   * On Apply button callback
+   */
+  onApply: PropTypes.func.isRequired,
+
+  i18n: PropTypes.shape({
+    cancelButton: PropTypes.string,
+    applyButton: PropTypes.string,
```
tay1orjones, line 176: We can fully qualify these as `text`. In other places we qualify input `label`s as such, aria-labels as `aria` etc.
```suggestion
    cancelButtonText: PropTypes.string,
    applyButtonText: PropTypes.string,
```

src/components/FlyoutMenu/FlyoutMenu.test.jsx
```diff
@@ -0,0 +1,153 @@
...
+  it('Cancel Action', async () => {
+    const cancelAction = jest.fn();
+
+    const { queryByText } = render(
+      <FlyoutMenu open triggerId="flyout-test" transactional onCancel={cancelAction} />
+    );
+
+    const cancelButton = await waitForElement(() => queryByText(cancelButtonText));
+
+    cancelButton.click();
```
tay1orjones, line 53: What `click()` is being used here? Usually I see [`fireEvent.click(cancelButton)`](https://testing-library.com/docs/react-testing-library/cheatsheet#events) - is this just a proxy for the same thing?

bjornalm:
Not sure if the netlify deployment is showing the latest visual changes but if it does then we should address this little issue with choosing different flyout directions than the default.

image

bjornalm:
Looks like the alignment is off for RTL mode. I think this is an issue for all directions.

image

src/components/FlyoutMenu/FlyoutMenu.test.jsx
```diff
@@ -0,0 +1,251 @@
...
+
+import FlyoutMenu, { FlyoutMenuDirection } from './FlyoutMenu';
+
+const applyButtonText = 'Apply';
+const cancelButtonText = 'Cancel';
+
+describe('FlyoutMenu', () => {
+  it('Renders an open transactional menu', async () => {
+    const { queryByText } = render(
+      <FlyoutMenu defaultOpen transactional triggerId="flyout-test" />
```
bjornalm, line 13: We gonna get console warnings about missing required `iconDescription` for these tests

src/components/FlyoutMenu/FlyoutMenu.test.jsx
```diff
@@ -0,0 +1,251 @@
...
+import '@testing-library/jest-dom';
+import { fireEvent, render, waitFor, waitForElement } from '@testing-library/react';
+
+import FlyoutMenu, { FlyoutMenuDirection } from './FlyoutMenu';
+
+const applyButtonText = 'Apply';
+const cancelButtonText = 'Cancel';
+
+describe('FlyoutMenu', () => {
+  it('Renders an open transactional menu', async () => {
```
bjornalm, line 11: Why are the tests using `async` and   `await waitForElement`? Is it really needed?

src/components/FlyoutMenu/FlyoutMenu.test.jsx
```diff
@@ -0,0 +1,251 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { fireEvent, render, waitFor, waitForElement } from '@testing-library/react';
+
+import FlyoutMenu, { FlyoutMenuDirection } from './FlyoutMenu';
+
+const applyButtonText = 'Apply';
```
bjornalm, line 7: I think we can make the test less brittle by using the actual default values, e.g. `const applyButtonText = FlyoutMenu.defaultProps.i18n.applyButtonText`

src/components/FlyoutMenu/FlyoutMenu.test.jsx
```diff
@@ -0,0 +1,251 @@
...
+
+    const applyButtonTest = await waitForElement(() => queryByText(applyButtonText));
+
+    const cancelButtonTest = await waitForElement(() => queryByText(cancelButtonText));
+
+    expect(applyButtonTest).toBeTruthy();
+    expect(cancelButtonTest).toBeTruthy();
+  });
+
+  it('Renders an open menu', async () => {
```
bjornalm, line 24: This test will easily give us a false positive. I.e. we can remove the `defaultOpen` and the test would still pass... which is proven by the test below. I think it would be better if you added some actual test content to the FlyoutMenu and tested that it was showing/not showing in these open/closed tests.

src/components/FlyoutMenu/FlyoutMenu.test.jsx
```diff
@@ -0,0 +1,251 @@
...
+
+    const applyButtonVisible = queryByText(applyButtonText);
+
+    expect(applyButtonVisible).toBeNull();
+  });
+
+  it('Click to open and close menu', async () => {
+    const { container, queryByText } = render(<FlyoutMenu transactional triggerId="flyout-test" />);
+
+    const button = await container.firstChild.firstChild;
```
bjornalm, line 117: I think you should query using the label (aria-label) instead of this container.firstChild.firstChild in order to better mimic what a human tester would do

src/components/FlyoutMenu/FlyoutMenu.test.jsx
```diff
@@ -0,0 +1,251 @@
...
+  });
+
+  it('Flyout Anchor Points Rendered bottom for bottom start', async () => {
+    const { container } = render(
+      <FlyoutMenu
+        defaultOpen
+        triggerId="flyout-test"
+        transactional
+        direction={FlyoutMenuDirection.BottomStart}
+        onApply={jest.fn()}
```
bjornalm, line 139: Are the props `onApply` and `onCancel` needed for the this and the tests below?

src/components/FlyoutMenu/_flyout-menu.scss
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  &--tooltip-anchor {
+    position: relative;
+    height: 0px;
+    width: 0px;
+  }
+
+  &--open {
+    box-shadow: 0 2px 6px rgba(0, 0, 0, 0.2) !important;
```
bjornalm, line 99: can we find and use the box-shadow in Carbon?

src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -0,0 +1,335 @@
...
+  /**
+   * Optional prop to specify the test id of the flyout
+   */
+  testId: PropTypes.string,
+
+  /**
+   * If true renders the default submit / cancel footer
+   * Can pass in a custom footer
+   */
+  transactional: PropTypes.oneOfType([PropTypes.bool, PropTypes.node]),
```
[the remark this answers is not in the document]
marcJV: This was a suggestion from Taylor, the idea is that there would not be a need for the custom footer if the flyout is not transactional. @tay1orjones should we keep this prop merged or separated?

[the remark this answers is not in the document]
tay1orjones: Lets avoid the `modal` nomenclature

```diff
/**
* Specify whether the flyout should be button-less
*/
-passiveModal: PropTypes.bool,
+isPassiveFlyout: PropTypes.bool,
```

PR #1311, opened 2020-06-16
fix(list): replaced filterdeep with somedeep

### LINKED ISSUE #1106: [List] replace usage of filterDeep with someDeep
Hi there, I was just curious how people use deepdash, 
so may I suggest to use [someDeep](https://deepdash.io/#somedeep) instead of filterDeep here:

https://github.com/IBM/carbon-addons-iot-react/blob/0b93e5012d651a97a48ff85df1a6dd2f7e822de4/src/components/List/List.story.jsx#L352-L366

It may look like:
```js
    const checkSelectedChildren = items => someDeep(items, 
        (value, key) => selectedIds.some(id => key === id));
```
this code is not only shorter but also faster.
filterDeep is overkill here because it's like partial cloning of source object depending on given criteria. It goes from top to bottom over 100% of given data (in your use case)
And someDeep does not create anything - it just iterates and checks and breaks the process on the first match.

And the second point, I didn't test and I am not familiar with this code, but this criteria looks wrong:
https://github.com/IBM/carbon-addons-iot-react/blob/0b93e5012d651a97a48ff85df1a6dd2f7e822de4/src/components/List/List.story.jsx#L355
`key` is field **name** so if you want to check if some deep id listed in array, then you need:
```js
const checkSelectedChildren = items => someDeep(items, 
        (value, key) => key==='id' && selectedIds.some(id => value === id));
```

rawa-resul:
Closes #1106

**Summary**

- Replaced filterDeep with someDeep for added performance in List

**Update**
- Fixed a bug whereby multiple children with the same id were treated as one object. Added `${team}-${player}` to child elements and for their evaluation (pointed out by @tay1orjones )

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
     "cross-env": "^6.0.3",
     "css-loader": "^2.1.0",
     "d3": ">=5.0.0 <=5.14.2",
-    "deepdash": "^4.4.2",
+    "deepdash": "^5.1.2",
     "enzyme": "^3.8.0",
     "enzyme-adapter-react-16": "^1.14.0",
     "enzyme-to-json": "^3.3.5",
--- a/src/components/List/List.story.jsx
+++ b/src/components/List/List.story.jsx
@@ -312,7 +312,7 @@ storiesOf('Watson IoT Experimental/List', module)
           if (item.id === value) {
             filteredItems.push(item.id);
             if (item.children) {
-              const children = searchNestedItems(item.children, value, true);
+              const children = searchNestedItems(item.children, item.id, true);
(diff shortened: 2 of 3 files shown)


rawa-resul:
> Looks like there's a minor defect with selection - selecting certain items are changing selected state for unrelated items in other groups.
> 
> Jun-16-2020 13-15-49
> 
> Edit: looks like the defect is currently present in production. Would be awesome if you could fix it here, but we can open a separate issue if you would prefer.

Thanks for pointing it out. I rather push a functional PR. I will spend more time on it to fix it.

PR #1315, opened 2020-06-16
fix(rowactioncell): change action color

### LINKED ISSUE #1314: [RowActionCell] Colors don't adhere to design
### Describe the bug

Colors were incorrect according the component design

See internal issue: https://github.ibm.com/wiotp/monitoring-dashboard/issues/1043

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: v2.90

sstone2423:
Closes #1314 

**Summary**

- Color was blue by default which did not adhere to the original design

- Swap color to black by default and blue on hover

- Check the row actions story to see that the colors are correct

Screen Shot 2020-06-16 at 12 48 37 PM

### DIFF
--- a/src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
+++ b/src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
@@ -8,14 +8,6 @@
   > * {
     margin-left: $spacing-04;
   }
-  .#{$iot-prefix}--row-actions-cell-btn--icononly {
-    margin-left: 0;
-    color: black; // black by default
-
-    .#{$prefix}--btn__icon {
-      margin-left: 0;
-    }
-  }
 }
 
 .#{$iot-prefix}--row-actions-container__background {
--- a/src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
+++ b/src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
@@ -10,11 +10,16 @@
   }
   .#{$iot-prefix}--row-actions-cell-btn--icononly {
     margin-left: 0;
+    color: unset; // black by default
 
     .#{$prefix}--btn__icon {
       margin-left: 0;
     }
   }
+}
+
+.#{$iot-prefix}--row-actions-cell-btn--icononly:hover {
+  color: $interactive-01; // blue on hover
 }
 
 .#{$iot-prefix}--row-actions-container__background {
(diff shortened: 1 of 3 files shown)


cynthiaz:
@sstone2423 
I may have been misguided as well regarding the color for the hover state. I checked again Carbon's guideline. To clarify, here's the 3 kinds of row action buttons and the color for each state

image

To correct myself, the icon button should be black in both default and hover state, see carbon's example: https://react.carbondesignsystem.com/?path=/story/datatable--with-overflow-menu

CynZhang:
@sstone2423 
I may have been misguided as well regarding the color for the hover state. I checked again Carbon's guideline. To clarify, here's the 3 kinds of row action buttons and the color for each state

image

To correct myself, the icon button should be black in both default and hover state, see carbon's example: https://react.carbondesignsystem.com/?path=/story/datatable--with-overflow-menu

src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
```diff
@@ -11,6 +11,7 @@
   }
   .#{$iot-prefix}--row-actions-cell-btn--icononly {
     margin-left: 0;
+    color: black; // black by default
```
tay1orjones, line 14: Looks like it's Gray 100, not straight black

```suggestion
    color: $ui-05;
```

image

tay1orjones: Actually with the `color` property it should probably be `$text-01`

tay1orjones:
@sstone2423 @CynZhang could you point me to a story in our storybook that shows this defect? This story shows a black overflow menu item and the other action buttons aren't iconOnly, so they're blue as intended 
https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--with-row-expansion-and-actions

CynZhang:
@tay1orjones I believe the one you posted there is the one, and the requested change was to update the icon's hover state color to $text-01 instead of blue. 
The other one I found is https://carbon-addons-iot-react.com/?path=/story/watson-iot-tablebodyrow--row-actions-with-overflow  which also uses Blue color for icon buttons.

tay1orjones:
Thanks @CynZhang - that's a great example I overlooked.

@sstone2423 I believe we should be using `hasIconOnly` for these buttons so we provide the optimal accessibility affordance via the tooltip in addition to the color consistency.

sstone2423:
Good call. Didn't realize there was an iconOnly button already! That makes it way simpler. Looks like its naturally taking the gray100 now

src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss
sstone2423: This style is no longer needed. Using `hasIconOnly` naturally gave the icon the correct color again

PR #1327, opened 2020-06-18
feat(story): edited the title for the table sticky header story

### LINKED ISSUE #1242: [Table] Add experimental tag to sticky header story
There was story that was added a while ago that included the sticky header feature. We need to document the experimental nature of this feature as it is leading people to assume they can use it when not currently supported.

soniadeveloper:
Closes #1242 

**Summary**

Made a note that the sticky header feature for the table is experimental.

- Changed the title for the table sticky header story so it notes that the feature is experimental

- When the storybook is launched (yarn start), scroll down to Table and view the stories within the component. The sticky header story should have (experimental) in the title to indicate.
Screen Shot 2020-06-18 at 12 25 24 PM

### DIFF
--- a/src/components/Table/Table.story.jsx
+++ b/src/components/Table/Table.story.jsx
@@ -2006,7 +2006,7 @@ storiesOf('Watson IoT/Table', module)
     }
   )
   .add(
-    'with sticky header and cell tooltip calculation',
+    'with sticky header (experimental) and cell tooltip calculation',
     () => {
       const renderDataFunction = ({ value }) => (
         <div style={{ position: 'relative' }} data-floating-menu-container>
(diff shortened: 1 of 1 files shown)


tay1orjones:
@soniadeveloper it looks like the status checks in Travis CI are failing because of an out of date snapshot.

As a way to detect potential regressions in cross-component usage (Table uses Button for instance) we use [StoryShots](https://storybook.js.org/docs/testing/structural-testing/) to generate [Jest snapshots](https://jestjs.io/docs/en/snapshot-testing) of our stories in storybook. 

When the stories change, tests will fail. Take a look at the failing snapshot - it should only be the one you modified. If the change is intentional and expected, you can update the snapshot by running `yarn test:update` and then commit/push the updated snapshot file.

PR #1333, opened 2020-06-19
deprecate Dashboard,  CardRenderer

### LINKED ISSUE #1332: [Dashboard / CardRenderer] deprecate
### Summary

Dashboard and CardRenderer are only being used by the Monitor team. Not only that, if another project wanted to use them, they're very complicated and heavily biased to Monitor use-cases. Deprecating both of these will take 2 pretty heavy components off of the addons library and allow Monitor to more efficiently update them and not worry as much about trying to keep them general for other projects

upcoming release?

Would like this to be removed in the next major release

### Available extra resources

Will do myself

sstone2423:
Closes #1332 

**Summary**

- Deprecates Dashboard and CardRenderer for removal in next major version
- Exports DashboardHeader

- Added deprecation notices to both components and Dashboard stories

- Check that the stories show the deprecation notice, and that both console warnings are shown in development


tay1orjones:
@stuckless are these being used by your team?

stuckless:
@tay1orjones if I understand correctly, we are using ValueCard (and other cards) with DashboardLayout and I don't think we are affected by this.   thx.

src/components/Dashboard/DashboardHeader.jsx
```diff
@@ -69,6 +70,13 @@ const DashboardHeader = ({
   actions,
   onDashboardAction,
 }) => {
+  if (__DEV__) {
+    warning(
+      false,
+      'Dashboard component has been deprecated and will be removed in the next release of `carbon-addons-iot-react`.'
+    );
+  }
+
```
tay1orjones, line 79: Lets not deprecate this one as it's a reusable piece we'll want to export/make available and make future improvements to.

sstone2423: Removed the deprecation. Thanks!

PR #1339, opened 2020-06-23
(BarChartCard / TimeSeriesCard) domain range functionality

### LINKED ISSUE #1338: [BarChartCard / TimeSeriesCard] domain range functionality
### Summary

Internal issue: https://github.ibm.com/wiotp/monitoring-dashboard/issues/996

Adding a domain range to the current time series cards will allow us to always the show same date range regardless of the actual data being passed to the card props.

For example, if I have data that is from January 1-10, but I only want to display data from January 3-10, the domain range allows that.

### Available extra resources

Will work on myself

sstone2423:
Closes #1338 

**Summary**

- Adds domain ranges to BarChartCard and TimeSeriesCard
- Fixes a BarChartCard bug where the chart was never able to be 'SIMPLE'

- Updated to latest Carbon Charts version to pick up this bug fix (https://github.com/carbon-design-system/carbon-charts/pull/666)
- Changed the conditional logic for the different types of charts to support 'SIMPLE'
- Added new `domainRange` prop that accepts numbers (timestamps) or date instances

- See https://deploy-preview-1339--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-timeseriescard--medium-single-line-domain-range
- See https://deploy-preview-1339--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-barchartcard--simple-bar-time-series

To see the domains actually change, the storybook will need to be run locally. I tried using knobs, but it wasn't forcing a rerender of the chart


yarn.lock
```diff
@@ -2300,17 +2300,28 @@
...
+    "@carbon/colors" "10.4.0"
+    "@carbon/utils-position" "1.1.1"
+    date-fns "2.8.1"
+    lodash-es "4.17.15"
+    resize-observer-polyfill "1.5.0"
+
+"@carbon/charts@^0.32.7":
+  version "0.32.8"
+  resolved "https://registry.yarnpkg.com/@carbon/charts/-/charts-0.32.8.tgz#f191c8b5136c15774b64e6e80cf9cf1d336a4c4a"
+  integrity sha512-KQDmf+4aD4Mm4++gbm6lYkQYqCOO6GkqexeIcAFtyh8C/geq25f2xJZ0kOlrrilskcDUD0GbEOwyO7XB3oLJFQ==
```
tay1orjones, line 2324: I believe this duplicate resolution is because `@carbon/charts-react` has `@carbon/charts` as a hard dep. Do we need to specify `@carbon/charts` as a dep at all in our package.json since they already list is as such?

sstone2423: Oh this is my mistake. I believe we can remove it

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -41,6 +41,8 @@ const BarChartCard = ({
   isLoading,
   interval,
   className,
+  domainRange,
+  endDate,
```
scottdickerson, line 45: is endDate actually being used anymore?

sstone2423: Nope, thought I caught all of those. Removing it now

yarn.lock
sstone2423: Removed and verified that the charts are still working correctly

sstone2423: But unfortunately this is blowing up on `yarn build` now

sstone2423: Changed this to use ^v0.32.11 as its the actual dependency of @carbon/charts-react

PR #1341, opened 2020-06-23
chore(testing): eslint rules to improve consistency in testing-library

### LINKED ISSUE #1342: [testing] use screen to query DOM elements
### Summary

In #1341 the `eslint-plugin-testing-library` and the `prefer-screen-queries` rule was added. There's quite a few spots where we're not using `screen` that need to be refactored. Currently these are marked as warnings, and print to the console:

```
warning  Use screen to query DOM elements, `screen.queryAllByText`     testing-library/prefer-screen-queries
```

Items needed to mark this complete:
- [x] refactor the tests with warnings to use `screen`
- [x] update the rule to be `error` level to enforce usage of `screen` in `.eslintrc`

```diff
- "testing-library/prefer-screen-queries": "warn",
+ "testing-library/prefer-screen-queries": "error",
```

tay1orjones:
Closes #1342  

**Summary**

- Small update to add some eslint rules to help make our `testing-library` tests more consistent and follow best practices.

- add rule, fix some errors
- ~going to defer `screen` warnings for now to #1342~ @soniadeveloper added a fix for this to this PR, thanks Sonia!

- status checks should pass
- you can locally run `yarn lint` to see the output


soniadeveloper:
#1342: Screen warnings have been updated to an error and it is now being used in testing files as needed.

src/components/Accordion/AccordionItemDefer.test.jsx
```diff
@@ -1,37 +1,47 @@
...
   it('renders content when expanded', () => {
-    const { container, getByTestId } = render(
+    const { container } = render(
       <Accordion>
         <AccordionItemDefer id="a" title="Title one">
           <p>This content</p>
         </AccordionItemDefer>
       </Accordion>
     );
-    expect(getByTestId('accordion-item-deferred').lastElementChild.childElementCount).toEqual(0);
```
davidicus, line 17: This data-testid is on a React component. When would be an appropriate use for the way we had it over the screen query method if not this example?

tay1orjones: For cases where you can't match by role or text or it doesn't make sense (e.g. the text is dynamic). We should instead try to use queries that reflect the experience of visual/mouse users as well as those that use assistive technology - `getByRole`, `getByLabelText`, etc.
Testing library docs have a good overview of query preference - https://testing-library.com/docs/guide-which-query

Also, while the destructure assignments `{ container, getByTestId }` are available and present in the documentation, the emerging best practice is to query using `screen`. [The eslint rule has further info](https://github.com/testing-library/eslint-plugin-testing-library/blob/master/docs/rules/prefer-screen-queries.md) on why screen is considered best practice:

> _DOM Testing Library (and other Testing Library frameworks built on top of it) exports a screen object which has every query (and a debug method). This works better with autocomplete and makes each test a little simpler to write and maintain._

tay1orjones:
@davidicus I've updated this PR with some updated documentation for testing in our `contributing.md`

sstone2423:
Looks like Travis builds are failing here

tay1orjones:
Looks like a few tests were added in `master` in the wrong format. I'll update

Issue #963, opened 2020-02-26
[Storybook] upgrade to v6 - use addon-docs instead of addon-info


tay1orjones:
### Summary

as of storybook v6 [`addon-docs`](https://github.com/storybookjs/storybook/tree/next/addons/docs) replaces the now-deprecated `addon-info`. With it comes a much more robust way to document components.

There's some info in the migration guide on how to update ([1](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#docs-mode-docgen), [2](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#unified-docs-preset), [3](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#description-doc-block)) and also some pitfalls (notably as of now (v5.2-5.3) there's only support for one prop table per component).

There is new documentation for the migration to v6 in the storybook repo's migration guide.

PR #1348, opened 2020-06-25
feat(table): Save view modal

### LINKED ISSUE #1299: [table] save view modal
### Summary

As part of #1210 we need to create a modal that lets the user save a view.

**Additional context**
image

bjornalm:
Closes #1299

**Summary**
A highly customisable modal built for saving a table view. The component is built outside the actual table but only make sense in the context of saving a table view. A full example with this component and the table views will be completed as part of #1210. 

- Created a TableSaveViewModal component that should be exported
- Created a TableSaveViewForm component as a subcomponent of the TableSaveViewModal
- Created 4 stories as a sub section of the table stories that demonstrates and with info explains the usage and how to customise the modal
- Wrote unit tests

- Go to https://deploy-preview-1348--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table-tablesaveviewmodal--default-with-knobs
  - verify that all the knobs and actions works as expected
- Go to /https://deploy-preview-1348--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table-tablesaveviewmodal--prefilled 
  - make sure that the form is prefilled
- Go to https://deploy-preview-1348--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table-tablesaveviewmodal--custom-description-and-public-checkbox-using-overrides 
  - verify that the public checkbox is disabled and that the description is replaced by a working Accordion and that the story info makes sense
- Go to https://deploy-preview-1348--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table-tablesaveviewmodal--custom-form-using-overrides 
  - verify that the overridden form works and that the story info makes sense


src/components/Table/TableSaveViewModal/TableSaveViewModal.jsx
```diff
@@ -0,0 +1,171 @@
...
+   * onChange : Called when view title input value is changed
+   */
+  actions: PropTypes.shape({
+    onSave: PropTypes.func,
+    onClose: PropTypes.func,
+    onClearError: PropTypes.func,
+    onChange: PropTypes.func,
+  }).isRequired,
+  /** Shows this string as a general modal error when present */
+  error: PropTypes.string,
```
tay1orjones, line 27: This might be better suited to live in the `i18n` object since it will always be some sort of internationalized/translated string.

src/components/Table/TableSaveViewModal/TableSaveViewModal.test.jsx
```diff
@@ -0,0 +1,260 @@
...
+    expect(actions.onClearError).not.toHaveBeenCalled();
+
+    fireEvent.click(getByRole('button', { name: 'closes notification' }));
+    expect(queryByText('my test error')).not.toBeInTheDocument();
+    expect(actions.onClearError).toHaveBeenCalled();
+  });
+
+  it('replaces the strings with i18n strings', () => {
+    const i18nTest = {
+      modalTitle: 'modal-title-test',
```
tay1orjones, line 94: I think like this idea. We've not done this before, but makes sense to ensure that i18n strings are appropriately used when configured to prevent silent regressions moving forward. In theory prop types and default props should ensure these are always used, but the variable could be used without actually being rendered to the DOM. 

In the future we might be able to achieve this easier or automatically via snapshots of some kind.

src/components/Table/TableSaveViewModal/TableSaveViewModal.test.jsx
```diff
@@ -0,0 +1,260 @@
...
+    };
+
+    const {
+      getByText,
+      getByLabelText,
+      queryByText,
+      queryByLabelText,
+      getByTitle,
+      queryByTitle,
+    } = render(<TableSaveViewModal actions={actions} open testID="my-modal" i18n={i18nTest} />);
```
tay1orjones, line 111: you can avoid this block by using `screen` instead. It's a best practice to aim for moving forward.

import `screen` from `@testing-library/react` above and then:

```suggestion
render(<TableSaveViewModal actions={actions} open testID="my-modal" i18n={i18nTest} />);
```

and then in your queries it's just `expect(screen.getByWhatever(thing)).toBeWhatever();`

tay1orjones:
@lukefirth could you do a visual review on this for us via the deploy preview links in the main comment? 

It all looks good from my perspective - you might have an idea on the double scrollbar issue above?

.storybook/story-styles.scss
```diff
@@ -2,3 +2,7 @@
   padding: 3rem;
   height: 100vh;
 }
+
+.storybook-container .info__overlay p {
+  margin: 12px 0;
+}
```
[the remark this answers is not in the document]
bjornalm: It is for making texts in the info overlay appears as actual paragraphs when we are using multiple line breaks in the text it self. I noticed that the different text sections were wrapped in <p> tags but the p margins are reset to 0 by default. I also noticed that some story info texts was trying to work around this by inserting a <br> which doesn't work.

image

src/components/Table/TableSaveViewModal/TableSaveViewForm.jsx
```diff
@@ -0,0 +1,135 @@
...
+import PropTypes from 'prop-types';
+
+import { TextInput } from '../../TextInput';
+import { Checkbox } from '../../Checkbox';
+import { settings } from '../../../constants/Settings';
+import { OverridePropTypes } from '../../../constants/SharedPropTypes';
+
+const { prefix } = settings;
+
+const { iotPrefix } = settings;
```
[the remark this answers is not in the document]
bjornalm: good catch

src/components/Table/TableSaveViewModal/TableSaveViewModal.jsx
bjornalm: Yes, I considered that, but other components, like PageWizard and StatefulPageWizard that have both and error prop and an i18n object prop, keep them separated. I think that is because the error prop is also slightly different than normal i18n props since the error prop is also the trigger to make the error appear.

src/components/Table/TableSaveViewModal/TableSaveViewModal.test.jsx
bjornalm: Cool, having this automated would be great.

src/components/Table/TableSaveViewModal/TableSaveViewModal.test.jsx
bjornalm: Awesome, thanks! I will change it.

bjornalm:
> Looks solid! A few minor points in the comments below but nothing to hold up the merge.
> 
> Requesting changes though for these double scrollbars that are on the default story with an error. If we can fix them easily would be a solid improvement.
> 
> image

This seems to be related specifically to the scenario where the error message is showing, and I think it has to do with the margins. If we conditionally remove the bottom margin of the `bx--modal-content` (or at least reduce it from 3rem to 1rem) when the error is showing it looks like this problem goes away. But I think that should be done in a separate issue, so I will create new PR for that and verify the solution.

.storybook/story-styles.scss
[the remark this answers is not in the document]
tay1orjones: Ah interesting, sounds good. Yet another thing that #963 will help us with!

src/components/Table/TableSaveViewModal/table-save-view-modal.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../../globals/vars';
+
+.#{$iot-prefix}--save-view-modal__body-text {
+  margin-bottom: 1.5rem;
```
davidicus, line 4: Should be a carbon token

davidicus, line 4:
```suggestion
  margin-bottom: $spacing-06;
```

PR #1357, opened 2020-06-30
[Table] Fix RTL issues with row action overflow menu

### LINKED ISSUE #1197: [table] Overflow menu opens wrong way in RTL
[Internal Watson-IoT/pal-tracking#426](https://github.ibm.com/Watson-IoT/pal-tracking/issues/426)

### Describe the bug

Surfaced in #1112, the overflow actions dropdown opens the wrong way in RTL

 rtl wrong way dropdown
### To Reproduce

1. Go to a table story with overflow actions
2. Trigger RTL in storybook

### Expected behavior

The overflow opens the opposite way in RTL.

Side note - would this be solved if we used something like RTLCSS? That's what the Carbon team is recommending for most of their RTL issues like this with direction-aware declarations (margin-left, etc.)

davidicus:
Closes #1197 

**Summary**

Fixed bug where RowActions overflow dropdown was not opening the correct way based off the document language direction. 

- Pass a langDir prop down from Table to RowActionsCell.jsx which renders the overflow menu. 

- Go to story that has row actions
- Click on overflow row action button
- Menu should open inside the table
- Click on RTL knob to switch language direction
- Click on row action button
- Menu should be flipped and still open inside the table.


sstone2423:
Looks like the netlify failed here. Could you restart the build?

tay1orjones:
Looks like we ran into their build timeout limit - I just restarted

davidicus:
@tay1orjones there is nothing in the code that should change the placement of the dropdown. This PR is only manipulating the `flipped` prop on the `OverflowMenu`. Good call on the initial render though. It only happens the very first time you start up the storybook and not when you refresh. Weird behavior. But I provided a default in the hook in case there is not a lang attribute on the HTML tag.

tay1orjones:
Yeah, I think [`getFloatingPosition()` logic in FloatingMenu from Carbon](https://github.com/carbon-design-system/carbon/blob/4ccebf48f4aecb4d44fda9987a57a546c2dff821/packages/react/src/internal/FloatingMenu.js#L87-L140) is flawed and producing values that cause the misalignment in RTL. Seems that we could:

1. Shadow/copy that component (not ideal as it's technically an internal)
1. Implement our own positioning logic to override the values given by the Carbon code
1. Fix the code up in carbon - this is something that RTLCSS isn't going to fix

I prefer option 3. We can merge this as-is, it is an improvement on the root bug you're aiming to fix. We can open a new issue to track the work to improve the positioning logic in Carbon. They might even already have an issue open on it, I haven't looked.

src/components/Table/Table.test.jsx
```diff
@@ -499,6 +516,23 @@ describe('Table', () => {
...
+    expect(document.body.childNodes[2].className.includes('bx--overflow-menu--flip')).toBeTruthy();
+
+    document.documentElement.setAttribute('dir', 'rtl');
+
+    rerender(<Table columns={tableColumns} data={[tableData[1]]} options={options} />);
+    await fireEvent.click(getByTestId('Table-row-1-row-actions-cell-overflow'));
+    expect(document.body.childNodes[2].className.includes('bx--overflow-menu--flip')).toBeFalsy();
+
+    unmount();
+  });
```
tay1orjones, line 534:
```suggestion
    const { rerender } = render(
      <Table columns={tableColumns} data={[tableData[0]]} options={options} />
    );
    await fireEvent.click(screen.getByTestId('Table-row-0-row-actions-cell-overflow'));
    expect(screen.childNodes[2].className.includes('bx--overflow-menu--flip')).toBeTruthy();

    document.documentElement.setAttribute('dir', 'rtl');

    rerender(<Table columns={tableColumns} data={[tableData[1]]} options={options} />);
    await fireEvent.click(screen.getByTestId('Table-row-1-row-actions-cell-overflow'));
    expect(screen..childNodes[2].className.includes('bx--overflow-menu--flip')).toBeFalsy();
  });
```

davidicus: I think Kent Dodds says when you are operating on the root element like I do here for the RTL that you should unmount after the test. to avoid leaking those changes into other test. I am trying to find the issue comment

davidicus:
> Checks are failing from the new eslint testing rules.

@tay1orjones changed!

davidicus:
I am unable to reproduce that offset issue

PR #1367, opened 2020-07-06
TileCatalog and Pagination RTL fixes

### LINKED ISSUE #451: RTL: TileCatalog: Various
- Page arrows point wrong way
- Not sure if this is a problem or not, but the multi-select channel stays on the right side of the card instead of shifting to the left side

image

- Search icon moves to left align in the element.  doesn't move in LTR
image
image

Jessepy:
Closes #451

**Summary**

- Fixed arrow direction and padding in RTL. 
- Fixed tile padding and checkmark location in RTL
- The search icon moving also appears to be a non-issue, but a change can be added if wanted

- added dir="ltr" to arrow components in pagination so in RTL they're flipped
- added padding in pagination so the text is the right distance away in RTL
- added styles that override RadioTile defaults in RTL
- snapshots updated

- Verify both in Pagination and TileCatalog the bottom bar has the correct padding and arrow direction in RTL
- Verify that when a tile is selected in TileCatalog, the checkmark switches sides in RTL and the padding is correct


davidicus:
There are still some padding issues in RTL mode around the plus icon and the checkmark does not move to the opposite side.

Jessepy:
Fixed checkmark location and padding in RTL

PR #1394, opened 2020-07-13
[Combobox]: add default props for onChange and onInputChange

### LINKED ISSUE #1392: [Combobox] Storybook is blowing up when you type into input
### Describe the bug

Combobox story is crashing when you try to interact with the component

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-experimental-combobox--default
2. Try to interact with component

davidicus:
Closes #1392 

**Summary**

Not sure how this was introduced but without a default `onInputChange` our storybook was blowing up. This would have been true for onChange as well except we pass onChange handler into story. This PR just uses our default function for the default for both methods. 

- added default function to Combobox.jsx

- Go to Combobox story and choose an item. The component should not throw an error


src/components/ComboBox/ComboBox.jsx
```diff
@@ -37,6 +38,8 @@ const defaultProps = {
   editOptionText: '-Create',
   hasMultiValue: false,
   items: [],
+  onInputChange: defaultFunction('onInputChange'),
```
tay1orjones, line 41: Is this required? It's not defined in proptypes above. 

We haven't used this helper outside of the table props before, but I'm not opposed to using it here for now. You might consider null or a noop though, see https://github.com/IBM/carbon-addons-iot-react/issues/1401

src/components/ComboBox/ComboBox.jsx
```diff
@@ -37,6 +38,8 @@ const defaultProps = {
...
+  onChange: defaultFunction('onChange'),
```
tay1orjones, line 42: Do we need to define defaults for props that are required?

davidicus:
> It sounds like the fix to this should be to give the function prop in the stories. I'm not sure I understand how giving a function that console logs as a default prop helps the consumer

We are defaulting to null and doing a check for the function before calling it

SECURITY.md
```diff
@@ -11,8 +11,8 @@
...
 
-As this is a UI library, it's highly uncommon to see a security vulnerability directly within this codebase, but it is possible. 
+As this is a UI library, it's highly uncommon to see a security vulnerability directly within this codebase, but it is possible.
 
-If a reported vulnerability is within the codebase, the issue will be added to the current sprint and someone will begin to investigate immediately. Some components/modules we export are a direct proxy of a module from the [Carbon Design System](https://github.com/carbon-design-system/carbon). If the vulnerability is there, a maintainer will contact a member of the Carbon team and we'll work with them to investigate. 
+If a reported vulnerability is within the codebase, the issue will be added to the current sprint and someone will begin to investigate immediately. Some components/modules we export are a direct proxy of a module from the [Carbon Design System](https://github.com/carbon-design-system/carbon). If the vulnerability is there, a maintainer will contact a member of the Carbon team and we'll work with them to investigate.
```
tay1orjones, line 16: This is fine - I wrote this doc in the github interface in the security getting started wizard thing. So your git hook ran prettier which corrected some whitespace

Issue #1382, opened 2020-07-08
[Repo] Remove console noise in our test output


davidicus:
When running test the log can be quite busy. Its hard to read important information in the logs amongst all the noise. 

We should refactor components to get rid of console warnings and deprecation notices.

bjornalm:
Yes great. Perhaps this deserves another issue but there is still a console warning from the `TileGallery` when running the Storybook so it would be nice to get rid of that one as well.

davidicus:
> Yes great. Perhaps this deserves another issue but there is still a console warning from the `TileGallery` when running the Storybook so it would be nice to get rid of that one as well.

Yes. I think this issue should fix that console warning in storybook.

davidicus:
Will complete when we have a lull in priorities.

PR #1395, opened 2020-07-13
[Repo] 1382 test log cleanup

davidicus:
related to #1382 

**Summary**

PR to fix several test log issues. 

- Swapped out a couple icon names that were deprecated from Carbon for their new versions.
- Changed date format for our default Absolute date in DateTimePicker. 
- Changed test for ImageCard to test for implementation instead of looking for the console warning. Then added ignore to test since we will never trigger that if statement in our testing suite
- Added `truncateCellText` to TableCellRenderer

- Go to each test file and run them. You should not see any console statements or deprecation warnings.

### DIFF
--- a/src/components/DateTimePicker/DateTimePicker.story.jsx
+++ b/src/components/DateTimePicker/DateTimePicker.story.jsx
@@ -28,9 +28,9 @@ export const defaultRelativeValue = {
 export const defaultAbsoluteValue = {
   timeRangeKind: PICKER_KINDS.ABSOLUTE,
   timeRangeValue: {
-    startDate: '04/01/2020',
+    startDate: '2020-04-01',
     startTime: '12:34',
-    endDate: '04/06/2020',
+    endDate: '2020-04-06',
     endTime: '10:49',
   },
 };
--- a/src/components/ImageCard/CardIcon.jsx
+++ b/src/components/ImageCard/CardIcon.jsx
@@ -16,6 +16,8 @@ const propTypes = {
 };
 
 /** This component calls out to our renderIconByName callback function with the icon name OR fails over to our legacy card icons from our legacy icon bundle */
+// We test the implementation and do not want to trigger this console in our test logs
+/* istanbul ignore next */
 const CardIcon = ({ icon, renderIconByName, title, color, width, height, className }) => {
   if (__DEV__ && !renderIconByName && !icons[icon]) {
     warning(
--- a/src/components/ImageCard/CardIcon.jsx
+++ b/src/components/ImageCard/CardIcon.jsx
@@ -16,7 +16,7 @@ const propTypes = {
 };
 
 /** This component calls out to our renderIconByName callback function with the icon name OR fails over to our legacy card icons from our legacy icon bundle */
-// We test the implementation and do not want to trigger this console in our test logs
+/* We test the implementation and do not want to trigger this console in our test logs */
 /* istanbul ignore next */
 const CardIcon = ({ icon, renderIconByName, title, color, width, height, className }) => {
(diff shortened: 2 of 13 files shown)


src/components/ImageCard/CardIcon.test.jsx
```diff
@@ -1,22 +1,22 @@
...
-  afterAll(() => {
-    spy.console.mockRestore();
-    // globally this is false, but we need it true so the warning pops
-    global.__DEV__ = false;
-  });
   it('validate default', () => {
-    render(<CardIcon icon="bogus" title="title" color="#FFFFFF" />);
-    expect(spy.console).toHaveBeenCalled();
+    const { rerender } = render(<CardIcon icon="help" title="title" color="#FFFFFF" />);
+    // Help icon is the defaul used when Icon does not match list. Grabbing the path value
```
sstone2423, line 9: Small typo on `default`

tay1orjones:
@sstone2423 We brought in `user-event` to be able to more concretely test the new combobox - specifically the [`type` event](https://github.com/testing-library/user-event#typeelement-text-options) was very helpful.

We haven't really had a convo around where we should be using it vs `fireEvent`, but it in theory sounds like a good idea to use it in most places as it "tries to simulate the real events that would happen in the browser as the user interacts with it." Explore using it and let us know. They technically consider the repo a WIP but it's already a very popular package with a solid maintenance team, including Kent himself.

src/components/DateTimePicker/DateTimePicker.story.jsx
```diff
@@ -28,9 +28,9 @@ export const defaultRelativeValue = {
 export const defaultAbsoluteValue = {
   timeRangeKind: PICKER_KINDS.ABSOLUTE,
   timeRangeValue: {
-    startDate: '04/01/2020',
+    startDate: '2020-04-01',
     startTime: '12:34',
-    endDate: '04/06/2020',
+    endDate: '2020-04-06',
```
tay1orjones, line 33: @davidicus did you figure out that warning with the date formats, or is this okay to merge?

src/components/ImageCard/CardIcon.jsx
```diff
@@ -16,6 +16,8 @@ const propTypes = {
 };
 
 /** This component calls out to our renderIconByName callback function with the icon name OR fails over to our legacy card icons from our legacy icon bundle */
+/* We test the implementation and do not want to trigger this console in our test logs */
+/* istanbul ignore next */
```
tay1orjones, line 20: Alternatively we can mock the console and expect the console to be called with a message that .includes(theWarningMessage)

One time we had the condition wrong for one of these warnings and we would've caught it if it was actually wrapped in a test.

tay1orjones: I see now we previously had mocked the console as I mention in the above comment and you refactored to remove and test the implementation. 

I'm fine not having a test around the `if (__DEV__ && !renderIconByName && !icons[icon]) {` condition but it's something we _can_ break and would potentially erroneously show the error. Might be worth a new issue for a broader discussion around if/when we test `warning()` usage (deprecations, errors like this, misconfigurations like in Breadcrumb with ResizeObserver polyfill, etc)

src/components/Table/TableBody/RowActionsCell/RowActionsCell.jsx
```diff
@@ -186,6 +186,7 @@ class RowActionsCell extends React.Component {
           className={`${iotPrefix}--row-actions-container`}
         >
           <div
+            data-testid="row-action-container-background"
```
tay1orjones, line 189: Curious, what does this fix?

PR #1397, opened 2020-07-14
Add tests for i18n strings to ensure they appear on the DOM

### LINKED ISSUE #1351: [testing] ensure i18n strings are covered in tests
### Summary

Right now we don't explicitly ensure that i18n strings passed into components typically through an `i18n` prop object are covered in tests. It's possible right now for a component to be modified to no longer render an i18n string to the DOM and tests would still pass.

**Additional context**
In https://github.com/IBM/carbon-addons-iot-react/pull/1348#discussion_r445844626 @bjornalm added a test to cover the i18n strings which spawned this whole idea.

My initial thought is to attempt to automate this as much as possible and use snapshots. Any component with an i18n prop should have a snapshot attached of each i18n string being set to something, ensure they're rendered to the DOM, and then additionally track changes to what's rendered to the DOM via the snapshot.

This is a nice-to-have improvement with low priority

Jessepy:
Closes #1351

**Summary**

- Many components use i18n strings, but it isn't being tested if they are used. This fixes that.

Added tests to:
- BarChartCard
- DateTimePicker
- ImageHotspots
- ListHeader
- ListItem
- PageTitleBar
- PageWizard
- Table
- TableSaveViewModal
- TableViewDropdown
- TileCatalogNew

- Look at tests, make sure everything looks okay and tests pass.


src/components/PageWizard/PageWizard.test.jsx
```diff
@@ -125,4 +126,38 @@ describe('PageWizard', () => {
...
+    expect(screen.getByLabelText(i18nTest.close)).toBeInTheDocument();
+    expect(screen.queryByText(i18nDefault.next)).not.toBeInTheDocument();
+    expect(screen.queryByText(i18nDefault.cancel)).not.toBeInTheDocument();
+    expect(screen.queryByLabelText(i18nDefault.close)).not.toBeInTheDocument();
+
+    render(
+      <PageWizard currentStepId="step3" i18n={i18nTest}>
+        {content}
+      </PageWizard>
+    );
```
tay1orjones, line 157: I think this is going to render a second instance into the DOM instead of replacing the old one. You might want to either use `rerender` or change the currentStep by clicking/interacting with the component.

src/components/Table/Table.test.jsx
```diff
@@ -558,4 +560,156 @@ describe('Table', () => {
...
+    expect(screen.getAllByLabelText(i18nTest.editButtonAria)[0]).toBeInTheDocument();
+    expect(screen.getAllByText(i18nTest.searchLabel)[0]).toBeInTheDocument();
+    expect(screen.getAllByPlaceholderText(i18nTest.searchPlaceholder)[0]).toBeInTheDocument();
+    expect(screen.getAllByTitle(i18nTest.clearFilterAria)[0]).toBeInTheDocument();
+    expect(screen.getAllByLabelText(i18nTest.filterAria)[0]).toBeInTheDocument();
+    expect(screen.getAllByLabelText(i18nTest.openMenuAria)[0]).toBeInTheDocument();
+    expect(screen.getAllByText(i18nTest.batchCancel)[0]).toBeInTheDocument();
+    expect(
+      screen.getAllByText((content, element) => {
+        if (RegExp(`.*\\s${i18nTest.itemSelected}.*`).test(content)) {
```
tay1orjones, line 638: Same thing here on the TextMatch that I commented on the other above.

src/components/Table/Table.test.jsx
```diff
@@ -558,4 +560,156 @@ describe('Table', () => {
...
+        i18n={i18nTest}
+        options={{
+          hasFilter: true,
+        }}
+      />
+    );
+    expect(screen.getAllByText(i18nTest.emptyMessageWithFilters)[0]).toBeInTheDocument();
+    expect(screen.queryByText(i18nDefault.emptyMessageWithFilters)).not.toBeInTheDocument();
+
+    render(
```
tay1orjones, line 698: Here also with the multiple `renders` concern.

src/components/ImageCard/ImageHotspots.test.jsx
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { act, isDOMComponent } from 'react-dom/test-utils';
+import { render } from '@testing-library/react';
+import { screen } from '@testing-library/dom';
```
davidicus, line 5: Is there a reason we brought in screen through the DOM package here instead of the React package like your other test?

src/components/ImageCard/ImageHotspots.test.jsx
```diff
@@ -1,6 +1,9 @@
...
+import '@testing-library/jest-dom/extend-expect';
```
davidicus, line 6: is this necessary to bring in for the expect assertions we have here?

src/components/Table/Table.test.jsx
```diff
@@ -558,4 +560,179 @@ describe('Table', () => {
...
+          },
+        }}
+      />
+    );
+    expect(screen.getAllByText(i18nTest.columnSelectionConfig)[0]).toBeInTheDocument();
+    expect(screen.queryByText(i18nDefault.columnSelectionConfig)).not.toBeInTheDocument();
+  });
+
+  it('i18n string test 2', () => {
+    const i18nTest = {
```
davidicus, line 676: Not sure if these i18nTest objects are different but if not we can define them in the top of the file and reuse them in each test.

src/components/TileCatalogNew/TileCatalogNew.test.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { render, fireEvent, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
```
davidicus, line 3: let's be sure to double check whether we need to pull this in for each of these test.

tay1orjones: https://github.com/IBM/carbon-addons-iot-react/issues/1400

src/components/ImageCard/ImageHotspots.test.jsx
tay1orjones: https://github.com/IBM/carbon-addons-iot-react/issues/1400

src/components/ImageCard/ImageHotspots.test.jsx
Jessepy: No, just lazy coding on my part. Fixed.

PR #1402, opened 2020-07-14
fix(dashboardgrid): upgrade react-grid-layout, fix card proptypes

### LINKED ISSUE #1175: [react-grid-layout] update dep version, reduce console noise
[Internal Watson-IoT/pal-tracking#408](https://github.ibm.com/Watson-IoT/pal-tracking/issues/408)

### Summary

We have quite a few console warnings when the test suite is ran:
```
Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.
      
      * Move data fetching code or side effects to componentDidUpdate.
      * If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
      * Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.
      
      Please update the following components: ReactGridLayout, ResponsiveReactGridLayout
```

`react-grid-layout` has been updated to use an optimized version of `componentShouldUpdate`, so if we update, we'll get rid of some of these. 

Also sounds like we might see a perf boost if we memoize the `children` in the grid - ([docs link](https://github.com/STRML/react-grid-layout/blob/master/README.md#performance))

scottdickerson:
Closes #1175

**Summary**

- upgrade the react-grid-layout and react-resizable components to remove the console errors and get a new feature that react-grid-layout has

- chore(package.json): upgrade the package, remove the direct dependency on react-resizable
- fix(dashboard): new version of react-grid-layout is rendering the resizable handle in the wrong place, so we need to override it's styles to hide. 
- fix(CardPropTypes): allow i18n.defaultLabel to be a node, not just strings.

- Compare the old DashboardGrid stories to the new one, the only difference is that the outer padding margin is gone (which is actually good!, we had to have padding workaround in our code)


src/components/Dashboard/_dashboard.scss
```diff
@@ -60,3 +60,7 @@
     }
   }
 }
+// workaround to hide the resize handles in react-grid-layout
+.react-resizable-hide .react-resizable-handle {
```
tay1orjones, line 64: Could we scope this under the root dashboard style? It's possible this is used elsewhere outside of dashboard (now or in the future) and I don't think we'd want to globally override this.

sstone2423: This might need to go under `DashboardGrid`'s style sheet since this component is being deprecated

tay1orjones: Ah yeah good call - let's move this there and scope it under the root DashboardGrid class.

PR #1365, opened 2020-07-06
feat(yarn): changed yarn install to no longer build

Jessepy:
Closes #1328

**Summary**

- Yarn install no longer automatically calls yarn build

**Change List (commits, features, bugs, etc)**

- removed yarn build from "prepare"
- added "- yarn build" to Travis, as we still need to build after we install

**Acceptance Test (how to verify the PR)**

- run yarn install, make sure it doesn't build. make sure that Travis still works.

PR #1350, opened 2020-06-25
fix(repo): Updated hard coded values to use Carbon tokens

soniadeveloper:
Closes #303 

**Summary**

Replaced hardcoded margin/padding and color values with tokens used in the @carbon/layout and @carbon/colors libraries.

**Change List (commits, features, bugs, etc)**

- Hardcoded values now use carbon tokens

**Acceptance Test (how to verify the PR)**

- Running yarn start on the command line should load the storybook as normal
- Going through component and SCSS files should show that tokens are being used appropriately.

PR #1406, opened 2020-07-16
fix(tabletoolbar): search remains focused as you type

### LINKED ISSUE #1386: [Table] Regression: Can no longer change the search via props on table
### Describe the bug

When a table is created if you later set new props on the table with a search the table ignores it.

### To Reproduce

```
  test('toolbar search should render with default value', () => {
    const wrapper = mount(
      <Table
        columns={tableColumns}
        data={tableData}
        actions={mockActions}
        options={{
          hasSearch: true,
        }}
        view={{
          toolbar: {
            search: {
              defaultValue: '',
            },
          },
        }}
      />
    );

    expect(wrapper.find('.bx--search-input')).toHaveLength(1);
    expect(wrapper.find('.bx--search-input').prop('value')).toEqual('');
    expect(wrapper.find('.bx--search-input').html()).toContain(`aria-hidden="true"`);

    wrapper.setProps({ view: { toolbar: { search: { defaultValue: 'ferrari' } } } });
    wrapper.update();

    expect(wrapper.find('.bx--search-input').prop('value')).toEqual('ferrari');
    expect(wrapper.find('.bx--search-input').html()).toContain(`aria-hidden="false"`);

    wrapper.setProps({ view: { toolbar: { search: { defaultValue: '' } } } });
    wrapper.update();

    expect(wrapper.find('.bx--search-input').prop('value')).toEqual('');
    expect(wrapper.find('.bx--search-input').html()).toContain(`aria-hidden="true"`);
  });
```

### Expected behavior

Updating the props should update the table.

### Additional Notes

The test I pasted in there is the ORIGINAL TEST.  I guess when you upgraded Carbon this test broke, and the fix, apparently was to rewrite the test logic and not test the original behaviour that this test was testing for.  So in essence the fix was to remove the failing test.  If you upgrade carbon and a test requires that you need to refactor the logic of your test, that should send alarm bells that Carbon just broke you... and you are going to break someone else, if you continue.  Rewriting the test logic to test for something completely different, is an effective way to get tests to pass, but it invalidates the purpose to have testing in the first place :(

https://github.com/IBM/carbon-addons-iot-react/commit/5289496c89a49b3d99348fe06d97c9793b02969d#diff-f47894e32d4dbf684d2ac3f89df148c9

tay1orjones:
Closes #1386

**Summary**

- Improve logic around when we create a new instance of the TableToolbarSearch when defaultValue is set by tracking changes coming from typing (internal onChange from the Carbon component) vs prop changes coming from outside the component. 

- re-add change to remove yarn prepare from build step - we somehow lost this in merging #1365 
- fix statefultable and table reducer to also respect the changes to toolbar props
- improve story of pre-filled search with buttons to demonstrate programatically clearing search or updating search to a new value

- http://localhost:3000/?path=/story/watson-iot-table--with-pre-filled-search updates search and clears search when buttons are clicked, focus will remain in the input while typing.


src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -143,115 +143,132 @@ const TableToolbar = ({
...
+              // not match the currentSearchValue, we can infer that a new defaultValue has been
+              // set externally and we should render a new instance of the component.
+              // https://github.com/IBM/carbon-addons-iot-react/issues/1386
+              key={
+                ((search.defaultValue || search.defaultValue === '') &&
+                  search.defaultValue !== currentSearchValue) ||
+                ((search.value || search.value === '') && search.value !== currentSearchValue)
+                  ? `${search.defaultValue}${search.value}`
+                  : `${tableId}-toolbar-search`
+              }
```
tay1orjones, line 203: Unfortunately the diff for this component is awful because I had to remove the implicit return to use useState, but this is what actually fixes the problem.

scottdickerson: thanks for the comment this will help us understand why in the future, I believe the !isEmpty(search.value) from lodash will check the search.value || search.value === '' in a easy to read way

scottdickerson:
@tay1orjones what's the latest on this one, it's holding off on a merge to beta for us?

yarn.lock
```diff
@@ -2088,7 +2088,7 @@
     levenary "^1.1.1"
     semver "^5.5.0"
 
-"@babel/preset-env@^7.4.5", "@babel/preset-env@^7.8.7":
+"@babel/preset-env@^7.4.5":
```
tay1orjones, line 2091: This was erroneously added in #1350 - package.json was updated but yarn.lock wasn't regened.

tay1orjones:
@stuckless Thanks for catching that - I've revamped this to a more comprehensive fix with the help of @scottdickerson 

Ready for a re-review

tay1orjones:
Something was lost in the cleanup I did from the lint errors - going to revisit

tay1orjones:
@scottdickerson @davidicus ready for review! deploy preview is now showing the fix under the [pre-filled search story](https://deploy-preview-1406--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--with-pre-filled-search)

Issue #1413, opened 2020-07-20
Experimental components are not using snapshot testing


sstone2423:
### Describe the bug

Experimental components are not using snapshot testing, meaning there is a loss of actual test coverage when there is indeed stories that could be used for snapshots.

### Expected behavior

All components should have stories and they should all be using snapshot testing

### Specific timeline issues / requests

Do you want this work within a specific time period? Is it related to an
upcoming release?

Working on myself

tay1orjones:
@sstone2423 what is remaining for this one?

sstone2423:
Meant to tag the other issue before. Here's the remaining components that have not been enabled because of errors: https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1412

tay1orjones:
Going to close this as it's primarily complete - we'll track the re-enabling of the ones with errors in #1412

Issue #1412, opened 2020-07-20
[SimpleList / ListItem / TileCatalogNew] fix snapshot unique key warnings


sstone2423:
### Describe the bug

Multiple unique key warnings are being thrown in SimpleList, ListItem, and TileCatalogNew. I've added a snapshot exception  in https://github.com/IBM/carbon-addons-iot-react/pull/1414 for all 3 components as they're causing the snapshots to fail 

### To Reproduce

1. Re-enable the snapshots by removing the regex exception in `StorybookSnapshots.test.js`
2. Run `yarn test:update`
3. See unique key warnings

### Expected behavior

Snapshots should pass with no warnings or errors

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: v2.100.2

### Specific timeline issues / requests

Do you want this work within a specific time period? Is it related to an
upcoming release?

No specific timeline, but non-unique keys can cause unexpected behavior in production

sstone2423:
Looks like I already had an issue open for List as well. Can probably close https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1412 in favor of this one

tay1orjones:
I'm delivering a fix for this alongside https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1664

PR #1414, opened 2020-07-20
fix(experimental): enabled snapshots for all experimental components

sstone2423:
Part of #1413

**Summary**

- Re-enables snapshot testing for experimental components
- Disables snapshot testing for TileCatalogNew, SimpleList, and ListItem as they have unique key warnings that require more work to fix. Tracking with #1412 

- Removed Experimental regex from `StorybookSnapshots.test.js`
- Added regex to pass testing on TileCatalogNew, SimpleList, and ListItem
- Added `labelText` prop to List stories to remove warning
- Added keys to SimpleList story action components to remove warning
- Fixed proptype in TileCatalogNew to not require the prop `tiles` to remove warning
- Added warning exception for ComboBox experimental prop to remove warning
- Added exception for HierarchyList's ref and mocked `scrollIntoView` as its not implemented in jsdom
- Fixed TileGallery propType to remove warning
- Fixed TileCatalogNew tiles function by removing 2nd argument

- Run `yarn test:update` successfully

### DIFF
--- a/src/components/List/ListHeader/ListHeader.jsx
+++ b/src/components/List/ListHeader/ListHeader.jsx
@@ -45,6 +45,7 @@ const ListHeader = ({ title, buttons, search, i18n, isLoading }) => {
       {search && !isLoading ? (
         <div className={`${iotPrefix}--list-header--search`}>
           <Search
+            id={`${iotPrefix}--list-header--search`}
             placeHolderText={i18n.searchPlaceHolderText}
             onChange={search.onChange}
             size="sm"
--- a/src/components/TileCatalogNew/TileCatalogNew.jsx
+++ b/src/components/TileCatalogNew/TileCatalogNew.jsx
@@ -20,7 +20,7 @@ const propTypes = {
   /** Number of rows to be rendered per page */
   numRows: PropTypes.number,
   /** A collection of tiles to to rendered  */
-  tiles: PropTypes.arrayOf(PropTypes.node).isRequired,
+  tiles: PropTypes.arrayOf(PropTypes.node),
   /** Set to true if a search is needed */
   hasSearch: PropTypes.bool,
   /** Call back function of search */
@@ -66,6 +66,7 @@ const defaultProps = {
     searchPlaceHolderText: 'Enter a value',
     error: 'An error has occurred. Please make sure your catalog has content.',
   },
+  tiles: null,
 };
 
 const TileCatalogNew = ({
--- a/src/components/TileCatalogNew/TileCatalogNew.jsx
+++ b/src/components/TileCatalogNew/TileCatalogNew.jsx
@@ -66,7 +66,7 @@ const defaultProps = {
     searchPlaceHolderText: 'Enter a value',
     error: 'An error has occurred. Please make sure your catalog has content.',
   },
-  tiles: null,
+  tiles: [],
 };
 
 const TileCatalogNew = ({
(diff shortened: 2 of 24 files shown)


src/components/List/ListHeader/ListHeader.jsx
```diff
@@ -45,6 +45,7 @@ const ListHeader = ({ title, buttons, search, i18n, isLoading }) => {
       {search && !isLoading ? (
         <div className={`${iotPrefix}--list-header--search`}>
           <Search
+            id={`${iotPrefix}--list-header--search`}
```
tay1orjones, line 48: What's this one for?

src/components/TileCatalogNew/TileCatalogNew.jsx
```diff
@@ -66,6 +66,7 @@ const defaultProps = {
     searchPlaceHolderText: 'Enter a value',
     error: 'An error has occurred. Please make sure your catalog has content.',
   },
+  tiles: null,
```
tay1orjones, line 69: Would it be more appropriate for this to be an empty array to match the prop types?

```suggestion
  tiles: [],
```

src/components/TileGallery/TileGallery.jsx
```diff
@@ -1,12 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import TileGallerySection from './TileGallerySection';
-
 const propTypes = {
   /** Component children's to be rendered */
-  children: PropTypes.arrayOf(PropTypes.shape({ type: PropTypes.oneOf([TileGallerySection]) }))
-    .isRequired,
+  children: PropTypes.node.isRequired,
```
tay1orjones, line 6: Memory is fuzzy on the details here, but I think we were trying to explictly require that a TileGallerySection be used as children here instead of allowing anything. Aiming for total adherence to the design spec.

What was the warning?

src/components/List/ListHeader/ListHeader.jsx
sstone2423: If the id isn't set, a random id is produced on every render, causing the snapshot to never be the same

src/components/TileCatalogNew/TileCatalogNew.jsx
sstone2423: To match the current implementation, it would need to be null. It would probably better to fix the conditional as it makes more sense to have an empty array rather it be null
```
{tiles && tiles.length && !error ? (
            <div className={`${iotPrefix}--tile-catalog--tile-canvas--content`}>{renderGrid()}</div>
          ) : (
            <Tile className={`${iotPrefix}--tile-catalog--empty-tile`}>
              <>
                <Bee32 />
                <p>{error || i18n.error}</p>
              </>
            </Tile>
          )}
```

src/components/TileGallery/TileGallery.jsx
sstone2423: First, the `children` prop here is actually returning an object, rather than an array. After removing the `arrayOf`, there is another issue in which there is a surrounding React fragment meaning the children would never equal `TileGallerySection` unless it is a single instance of it with no parent

src/components/TileCatalogNew/TileCatalogNew.jsx
tay1orjones: We're not bound by semver on this component, we can make that update if it makes sense. Good point on the conditional

src/components/TileGallery/TileGallery.jsx
tay1orjones: Ah yeah just took a look at the story. Even further - I'm not sure why this is a component if it's only job is to wrap children in a class. Beyond the scope here though. Looks good 👍

src/components/TileCatalogNew/TileCatalogNew.jsx
sstone2423: Went ahead and swapped this to your recommended change

PR #1417, opened 2020-07-20
feat(popups): Popup Positioning

### LINKED ISSUE #972: [Component Proposal] usePopoverPosition - hook to improve contextual awareness rendering for tooltip, flyout, etc
This is a proposal for a hook or wrapper component that can be used to detect proximity to window edge in order to handle replacement of "pop in" elements like tooltip or overflow menus. 

This could calculate children size and its placement on window. If children/component ref go off window we need to feed back the information to the children/component ref that will allow them to reposition themselves using the props available from Carbon. 

Is complete when:
- [ ] Hook or wrapper component that can calculate placement logic
- [ ] Update "pop in" components to use hook/wrapper internally behind a prop to handle positioning for downstream consumers:
  - [ ] Tooltip
  - [ ] Overflow
  - [ ] Flyout

Jessepy:
resolves #972 

**Summary**
- In several instances, the components that create popups like OverflowMenu, Tooltip, and FlyoutMenu, go offscreen.
- To resolve this, a utility function was made in PopupPositioning.js. This function will reposition the popups if they're offscreen.
- To use this functionality, you pass the utility function into the menuOffset prop of the component. In the specific case of OverflowMenu, when flipped is true, it needs to be passed into menuOffsetFlip
- For OverflowMenu and Tooltip, if a direction is specified, it will default to that direction unless that direction is offscreen.
- For FlyoutMenu, direction is completely ignored and behaves as described in [#1266](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1266#issuecomment-645654330)

1. Created PopupPositioning.js, which contains the utility function
2. Created PopupPositioning.test.jsx which tests that the function is working as intended with the components
3. Created PopupPositiong.story.jsx which can be used to see the functionality and how to use it
4. Edited FlyoutMenu.jsx to support the menuOffset prop being a function. Also added a max width to the Flyout as described in [#1266](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1266#issuecomment-645654330)
5. Changed flyout-menu.scss to allow variable after distance

Go to the PopupPositioning story, verify that when Auto Positioning is checked the popups appear within the viewport.


src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -220,7 +234,7 @@ const FlyoutMenu = ({
             triggerId={triggerId}
             tabIndex={tabIndex}
           >
-            <div ref={tooltipContentRef}>
+            <div ref={tooltipContentRef} style={{ maxWidth: '960px' }}>
```
tay1orjones, line 237: Why is the maxWidth required? We should remove it if possible

src/components/FlyoutMenu/_flyout-menu.scss
```diff
@@ -59,12 +59,12 @@ $shadow: 0.4rem;
       content: '';
       position: absolute;
       top: -$shadow-plus-padding;
-      left: -1rem;
+      left: var(--after-offset);
```
tay1orjones, line 62: This needs a default set for when the component isn't using the helper function and/or the conditional in flyout menu fails https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1417/files#diff-f98c4644941be7c8613bb9b77020bc7aR100

src/utils/PopupPositioning.js
```diff
@@ -0,0 +1,225 @@
...
+    return tooltipOffset(menuBody, menuDirection, menuElement);
+  }
+
+  if (menuBody.classList.contains(`${prefix}--overflow-menu-options`)) {
+    return overflowMenuOffset(menuBody, menuElement, flipped);
+  }
+
+  if (menuElement.classList.contains(`${iotPrefix}--flyout-menu`)) {
+    return flyoutMenuOffset(menuBody.parentElement.parentElement, menuDirection, menuElement);
+  }
```
tay1orjones, line 221: This is sort of brittle - Carbon doesn't consider their classnames as part of the API that adheres to semver, so they can change them and would break this logic. Maybe it's easier to just have separate modules for flyoutPositioning, tooltipPositioning, etc?

tay1orjones:
After some review @davidicus and I have found that the core logic around positioning values is a solid starting point, though there are places where component props/attributes/classes are accessed and modified imperatively outside of the react render cycle. Functionally this mostly works, but there are a couple edge cases related to browser resize that cause some odd rendering effects.

We need to retool this a bit to remain in sync with when the underlying components are (re)rendered. We're going to close this unmerged as it's becoming stale and unfortunately isn't a high priority at the moment for the core team to take on and push it over the finish line. We'll begin work on a new approach with this PR as a starting/inspiration point hopefully sometime soon. 

Thanks for the hard work @Jessepy and getting this off to a solid start! 🎉

PR #1420, opened 2020-07-22
fix(table): column resize failing for initially hidden column with un…

### LINKED ISSUE #1419: [table] column resize failing for initially hidden column with undefined width
### Describe the bug

Table resizing doesn't work when the user toggles the visibility of an initially hidden column that did not have a width property in the columns prop.

### To Reproduce

1. Go to a story that has column resize and column management with initially hidden columns without width, e.g. story https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-expansion-maxpages-and-column-resize
2. Click on "Column selection" button
3. Click on "Secret information field" to toggle the visibility
4. See how the columns width looks strange and how the resize doesn't work

### Expected behavior

The hidden column should be given a width, possibly based on an average of existing column widths, and then shown. The remaining columns should adjust properly and the column resizing should continue to work.

### Environment/versions:
- OS: all
- Browser: all

### Additional context

Noticed as part of this conversation #1358

bjornalm:
Closes #1419 

**Summary**
Table resizing doesn't work when the user toggles the visibility of an initially hidden column that did not have a width property in the columns prop.

- The fix was to create a new width (based on a column average) in the case a toggled column that should be shown did not have an existing width property.
- Refactored the `calculateWidthOnShow` and its sub functions for improved readability
- added a unit test to prevents future regression 

1. go to https://deploy-preview-1420--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-maxpages-and-column-resize
2. Click on "Column selection" button
3. Click on "Secret information field" to toggle the visibility
4. Verify that the "Secret information" column becomes visible with a width that is close to the average column widths
5. Verify that the column resize is still working properly


src/components/Table/TableHead/TableHead.test.jsx
```diff
@@ -285,6 +285,43 @@ describe('TableHead', () => {
...
+      };
+      myProps.columns = [
+        { id: 'col1', name: 'Column 1', width: '200px' },
+        { id: 'col2', name: 'Column 2', width: '200px' },
+        { id: 'col3', name: 'Column 3' },
+      ];
+
+      mockGetBoundingClientRect.mockImplementation(() => ({ width: 200 }));
+
+      const wrapper = mount(<TableHead {...myProps} />);
```
davidicus, line 305: Are we able to use the testing library in this case?

bjornalm: Yes probably, this test is almost identical to the previous test so I figured that I'd take the quick route for this fix and do a copy paste. 

Do you want me to rewrite it using testing library?

tay1orjones: Preferably, but I agree with taking a speedier copy/paste route for bugs where the tests are already consistently written with enzyme.

PR #1425, opened 2020-07-23
Add priority label to issue templates

tay1orjones:
New label so it's easier to filter issues that need to be prioritized during backlog grooming call.


tay1orjones:
@sstone2423 Sort of. `needs priority` is more specifically for items that need a `val` identified by OM in the weekly prioritization meeting. `needs triage` is a more general call for the PAL team to review or address. Bugs that we need to determine severity/validity, questions that haven't been addressed, enhancements/features that need technical/arch guidance, etc.

PR #1449, opened 2020-07-29
fix(table): pass filteredData to onDownloadCsv

### LINKED ISSUE #1448: [Table] Table download to csv should only send the filtered/searched data
### Describe the bug

When clicking the download to csv button after filtering/searching the data, the full data is downloaded from the table, instead of the filtered data

This occurs in all Card instances that use Table/StatefulTable and csv download utility

### To Reproduce

1. Go to a Table instance that has filtering or searching
2. Filter out some values
3. Click download csv button
4. See that ALL tableData has been exported rather than just the filtered data that is visible

### Expected behavior

Should return the filtered data only

### Environment/versions:

- OS: MacOS,
- Browser:chrome
- `carbon-addons-iot-react` version: `v2.100.2`

sstone2423:
Closes #1448 

**Summary**

- Passed the filteredData back to the `onDownloadCsv` callback so the CSV will be generated with the active filtered data

- Pass `data` prop to TableToolbar
- Added `data` propType to TableToolbar
- Updated the publicAPI snapshot to reflect the new prop
- Changed `onClick` prop of TableToolbarSVGButton to call the `onDownloadCsv` callback with the filtered `data` prop
- Changed all cards that use `onDownloadCsv` to use the new callback `data` argument
- I didn't do anything but `yarn install` from the master branch so I'm not sure where the yarn.lock change came from

1. Go to any story that implements `onDownloadCsv` (I was using TableCard)
2. Click the CSV download toolbar button
3. See that the output is correct
4. Filter the table with any of the filters or by searching
5. Click the CSV download button again and see that it only shows data that has been filtered

### DIFF
--- a/src/components/BarChartCard/BarChartCard.jsx
+++ b/src/components/BarChartCard/BarChartCard.jsx
@@ -192,7 +192,7 @@ const BarChartCard = ({
               }}
               actions={{
                 toolbar: {
-                  onDownloadCSV: () => csvDownloadHandler(tableData, title),
+                  onDownloadCSV: filteredData => csvDownloadHandler(filteredData, title),
                 },
               }}
               view={{
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -446,6 +446,7 @@ const Table = props => {
               'rowEditBarButtons'
             ),
           }}
+          data={data}
         />
       ) : null}
       <div className="addons-iot-table-container">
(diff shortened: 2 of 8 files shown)


tay1orjones:
`yarn.lock` change comes from #1350 - the `package.json` was modified and then rolled back but the yarn.lock wasn't regened. So all good 👍

tay1orjones:
CSV is different based on my search
image
image

Issue #1344, opened 2020-06-24
[List] ability to reorder rows


tay1orjones:
From 
[Internal Watson-IoT/pal-tracking#264](https://github.ibm.com/Watson-IoT/pal-tracking/issues/264)

> **Is this an extension to an existing component, pattern, or a brand new asset?**
> This is an extension to the list component to be able to support a mode where the rows inside can be reordered.
> 
> **What is the rationale for this contribution.**
> This was an EDS item
> ...
> 
> 
> **Detail the intended behavior**
> Currently only designed here (scroll to reorder mode). https://ibm.invisionapp.com/share/GFNZALEUJ5H#/screens/319100292
> ![image](https://media.github.ibm.com/user/684/files/07360c80-5252-11ea-9ba3-e3530a382347)
> 
> 
> Highly suspect that detail designs will need to be finalised with development before much work can continue
> .

tay1orjones:
`List` does not use the Table under the hood, so there isn't a dependency here. We can deliver this independent of the table reordering of rows.

PR #1451, opened 2020-07-30
feat(list): added simple reordering

marcJV:
Closes #

#1344 

Adds the ability for drag and drop for the list component

Single drag and drop for Simple List

Multiple and nested drag and drop for Hierarchy List

Added bulk reorder for Hierarchy List

React DND for List, with only simple callbacks, Simple List handles single drag and drop, hierarchy list handles nesting and multiple dnd as well

Simple list basic with reorder, you can drag and drop items into two targets, above and below. When hovering over a list item, a line will show which drop target is highlighted.

In Hierarchy List with reorder it defaults to single drag and drop. In this case there will be three drop targets. Above, below and nesting. The lines for above and below should be the same as the simple drag and drop, but nested will have the whole container highlighted with an outline. Dropping in the nested location will place the dropped item as the first child.

Enable multiple nesting in the editing style here. You can then click the check boxes and drag and drop multiple items.

An option in the header will also appear to handle bulk reorder. When clicked a modal will appear and you will then be able to click a row to view its children, or click the radio button to drop your selection as the first children of that list item. If an item does not contain children, if you click the row, the radio button will select by default.

Added an 'Allow Item Movement' variable to storybook to show what happens when the component does not allow the drag and drop to complete

### DIFF
--- a/src/components/List/HierarchyList/_hierarchy-list.scss
+++ b/src/components/List/HierarchyList/_hierarchy-list.scss
@@ -1,0 +1,101 @@
+@import '../../../globals/vars';
+
+.#{$iot-prefix}--hierarchy-list-bulk {
+  &-header {
+    width: 100%;
+    background-color: $focus;
+
+    display: flex;
+
+    &--title {
+      padding-left: $spacing-05;
+      margin: auto;
+
+      flex-grow: 1;
+      color: $ui-background;
+    }
+
+    &--button-container {
+      display: flex;
+      flex-direction: row;
+    }
+
+    &--divider {
+      height: $spacing-04;
+      padding-right: $spacing-02;
+      margin: auto 0 auto $spacing-02;
+      border-left: 1px solid $ui-background;
+    }
+
+    &--button {
+      padding: 0 $spacing-08 0 $spacing-05 !important;
+
+      &-no-icon {
+        padding: 0 $spacing-05;
+      }
+    }
+  }
+
+  &-modal {
+    &--breadcrumb {
+      overflow-y: scroll;
+
+      > button {
+        padding: 0 !important;
+      }
+    }
+
+    .#{prefix}--modal-content {
+      padding-right: 0 !important;
+    }
+
+    &--list {
+      display: flex;
+
+      width: 100%;
+
+      align-items: flex-start;
+      margin-top: rem(6px);
+      flex-direction: column;
+
+      &-item {
+        display: flex;
+        position: relative;
+        width: 100%;
+        background-color: $ui-background;
+
+        &:hover {
+          background-color: $hover-ui;
+        }
+
+        &::after {
+          height: 1px;
+
+          content: '';
+          background: $ui-03;
+          position: absolute;
+          bottom: 0px;
+          right: $spacing-08;
+          left: 0;
+        }
+
+        &-button {
+          display: flex;
+          align-content: flex-start;
+          padding: auto 0;
+          padding-right: $spacing-04;
+          flex-grow: 1;
+        }
+
+        &-value {
+          margin: auto 0;
+        }
+      }
+    }
+
+    &--radio {
+      padding: $spacing-05;
+      margin: 0 !important;
+    }
+  }
+}
--- a/src/components/List/HierarchyList/_hierarchy-list.scss
+++ b/src/components/List/HierarchyList/_hierarchy-list.scss
@@ -28,7 +28,7 @@
     }
 
     &--button {
-      padding: 0 $spacing-08 0 $spacing-05 !important;
+      padding: 0 $spacing-08 0 $spacing-05;
 
       &-no-icon {
         padding: 0 $spacing-05;
@@ -41,12 +41,8 @@
       overflow-y: scroll;
 
       > button {
-        padding: 0 !important;
+        padding: 0;
       }
-    }
-
-    .#{prefix}--modal-content {
-      padding-right: 0 !important;
     }
 
     &--list {
@@ -95,7 +91,7 @@
 
     &--radio {
       padding: $spacing-05;
-      margin: 0 !important;
+      margin-right: 0 !important;
     }
   }
 }
--- a/src/components/List/List.story.jsx
+++ b/src/components/List/List.story.jsx
@@ -69,7 +69,7 @@ export const sampleHierarchy = {
       'New York Mets': {
         'Jeff McNeil': '3B',
         'Amed Rosario': 'SS',
-        'Michael ConfortoMichael ConfortoMichael ConfortoMichael Conforto': 'RF',
+        'Michael Conforto': 'RF',
         'Pete Alonso': '1B',
         'Wilson Ramos': 'C',
         'Robinson Cano': '2B',
@@ -487,7 +487,7 @@ storiesOf('Watson IoT Experimental/List', module)
         isLoading={boolean('isLoading', false)}
       />
     </div>
-  ));
+  ))
   .add('basic (single column) with reorder', () => {
     const SingleColumnReorder = () => {
       const startData = Object.entries(
--- a/src/components/List/List.story.jsx
+++ b/src/components/List/List.story.jsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 import { text, boolean } from '@storybook/addon-knobs';
-import { Add16, Edit16, Star16, Close16, Checkmark16, NumberSmall_132 } from '@carbon/icons-react';
+import { Add16, Edit16, Star16, Close16, Checkmark16 } from '@carbon/icons-react';
 import cloneDeep from 'lodash/cloneDeep';
 import someDeep from 'deepdash/someDeep';
 
@@ -496,15 +496,9 @@ storiesOf('Watson IoT Experimental/List', module)
       const editing = boolean('isEditing,', true);
 
       const onItemMoved = (dragProps, hoverProps) => {
-        console.log(`raw drag: ${dragProps} raw hover: ${hoverProps}`);
-
         const dragCard = listItems[dragProps.index];
         listItems.splice(dragProps.index, 1);
         listItems.splice(hoverProps.index, 0, dragCard);
-
-        console.log(
-          `drag: ${dragProps.index} hover: ${hoverProps.index} items: ${JSON.stringify(listItems)}`
-        );
 
         setListItems([...listItems]);
       };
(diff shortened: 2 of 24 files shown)


tay1orjones:
It's a bit buried, but @amymillett mentioned [in the original internal issue](https://github.ibm.com/Watson-IoT/pal-tracking/issues/264#issuecomment-19878131) that list row reordering should follow the same pattern as table row reordering. This would include support for nested/hierarchy items, as well as dragging multiple rows at a time.

You can convert this to a draft and add those functions, or alternatively I'm okay with a phased delivery too. This PR can close the linked issue and we can create a new issue for enhanced list row reordering - heirarchy, multiple rows. Either way works 👍

tay1orjones:
Recap of our convo with @lukefirth (so I don't forget 😅 ) - this will follow the same visual style/flow specified for table reordering.

Remaining items to include here:

- [ ] [Multi select reorder](https://media.github.ibm.com/user/174580/files/25e6bb00-906b-11ea-9678-8c7792c38ed7)
- [ ] Hierarchy - [add items into a category aka "move underneath"](https://media.github.ibm.com/user/174580/files/136e8080-906f-11ea-9bae-a45abc3ee529) (three "targets" per row: above, below, and middle. middle adds to the category)
- [ ] Modal flow ([spec link](https://github.ibm.com/ai-applications/pal/issues/160#issuecomment-19878086) --- [image link](https://media.github.ibm.com/user/174580/files/993ffb00-9072-11ea-9e7e-14fb85b454a6)), because drag and drop isn't ideal for a11y. Ideally both methods (dnd and modal flow) would be available to the user at the same time.

Also discussed that the actual reordering function/logic can be abstracted into an exported helper function so this is as complete as possible for someone to pick up and use without much config

davidicus:
There is some weirdness that I am seeing where the rows are duplicated on HierarchyList.
image
 
and also where Titles are being duplicated. 
image
This stems from trying to drag an item into a parent multiple times.

marcJV:
> There is some weirdness that I am seeing where the rows are duplicated on HierarchyList.
> image
> 
> and also where Titles are being duplicated.
> This stems from trying to drag an item into a parent multiple times.

This has been fixed. The duplicate name was in the data, and that has been fixed as well.

src/components/List/HierarchyList/_hierarchy-list.scss
```diff
@@ -0,0 +1,101 @@
...
+
+    &--divider {
+      height: $spacing-04;
+      padding-right: $spacing-02;
+      margin: auto 0 auto $spacing-02;
+      border-left: 1px solid $ui-background;
+    }
+
+    &--button {
+      padding: 0 $spacing-08 0 $spacing-05 !important;
```
sstone2423, line 31: We should try to avoid `!important`. Can you use more specificity instead?

src/components/List/HierarchyList/_hierarchy-list.scss
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+  }
+
+  &-modal {
+    &--breadcrumb {
+      overflow-y: scroll;
+
+      > button {
+        padding: 0 !important;
```
sstone2423, line 44: We should try to avoid `!important`. Can you use more specificity instead?

src/components/List/HierarchyList/_hierarchy-list.scss
```diff
@@ -0,0 +1,101 @@
...
+      }
+    }
+
+    .#{prefix}--modal-content {
+      padding-right: 0 !important;
```
sstone2423, line 49: We should try to avoid `!important`. Can you use more specificity instead?

src/components/List/HierarchyList/_hierarchy-list.scss
```diff
@@ -0,0 +1,101 @@
...
+
+        &-value {
+          margin: auto 0;
+        }
+      }
+    }
+
+    &--radio {
+      padding: $spacing-05;
+      margin: 0 !important;
```
sstone2423, line 98: We should try to avoid `!important`. Can you use more specificity instead?

src/components/List/List.story.jsx
```diff
@@ -64,7 +69,7 @@ export const sampleHierarchy = {
       'New York Mets': {
         'Jeff McNeil': '3B',
         'Amed Rosario': 'SS',
-        'Michael ConfortoMichael ConfortoMichael ConfortoMichael Conforto': 'RF',
```
sstone2423, line 67: I think this was originally here to make sure long list items render correctly. Could you leave it so we can quickly check list items in the future? It does look like a typo though so maybe leave a comment on it?

src/components/List/ListItem/ListItem.jsx
```diff
@@ -1,25 +1,45 @@
...
+import ListTarget, { TargetSize } from './ListTarget';
+
 const { iotPrefix } = settings;
 
-// internal component
-const ListItemWrapper = ({ id, isSelectable, onSelect, selected, isLargeRow, children }) =>
-  isSelectable ? (
+export const ItemType = 'ListItem';
+
+const ListItemWrapper = ({
```
sstone2423, line 17: With this added code, I think its time to make a separate component file for `ListItemWrapper`

src/components/List/ListItem/ListTarget.jsx
```diff
@@ -0,0 +1,82 @@
...
+  Half: 'half',
+  Full: 'full',
+};
+
+const ListTargetPropTypes = {
+  connectDropTarget: PropTypes.func.isRequired,
+  isOver: PropTypes.bool,
+  targetPosition: PropTypes.string.isRequired,
+  targetSize: PropTypes.oneOf([TargetSize.Third, TargetSize.Half, TargetSize.Full]),
+  // eslint-disable-next-line react/no-unused-prop-types
```
sstone2423, line 21: Whats the purpose of this prop if its unused?

src/components/List/ListItem/ListTarget.jsx
```diff
@@ -0,0 +1,82 @@
...
+      ref={instance => {
+        if (connectDropTarget) {
+          connectDropTarget(instance);
+        }
+      }}
+    />
+  );
+};
+
+/* istanbul ignore next */
```
sstone2423, line 58: I don't think this should be ignored. Is it dropping you below the threshold?

src/components/List/ListItem/ListTarget.jsx
```diff
@@ -0,0 +1,82 @@
...
+    const target = hoverProps.targetOverride ?? hoverProps.targetPosition;
+
+    // Check if drop is allowed
+    if (hoverProps.itemWillMove(dragId, hoverId, target)) {
+      hoverProps.onItemMoved(dragId, hoverId, target);
+    }
+  },
+};
+
+/* istanbul ignore next */
```
sstone2423, line 72: I don't think this should be ignored. Is it dropping you below the threshold?

src/components/List/ListItem/_list-item.scss
```diff
@@ -4,16 +4,94 @@
...
+  }
+
+  &--drop-targets {
+    width: 100%;
+    height: rem(40px);
+    display: flex;
+    flex-direction: column;
+    position: absolute;
+    align-items: stretch;
+    // pointer-events: none;
```
sstone2423, line 22: Is this needed?

src/components/List/SimpleList/SimpleList.jsx
```diff
@@ -111,6 +132,23 @@ const SimpleList = ({
     pageOfPagesText: page => i18n.pageOfPagesText(page),
   };
 
+  /* istanbul ignore next */
```
sstone2423, line 135: This shouldn't be ignored as its code that will get hit multiple times

src/components/List/SimpleList/SimpleList.story.jsx
```diff
@@ -338,4 +340,36 @@ storiesOf('Watson IoT Experimental/SimpleList', module)
       </div>
     )),
     { info: { text: `` } }
-  );
+  )
+  .add('basic - SimpleList with reorder', () => {
```
sstone2423, line 344: Could this just be named `with reorder` ? It looks like some of the stories are using `list with` as a convention so either of those would be easier to identify

src/utils/DragAndDropUtils.jsx
```diff
@@ -0,0 +1,122 @@
...
+};
+
+export const EditingStyle = {
+  Single: 'single',
+  Multiple: 'multiple',
+  SingleNesting: 'single-nesting',
+  MultipleNesting: 'multiple-nesting',
+};
+
+export const editingStyleIsMultiple = editingStyle => {
```
sstone2423, line 16: Can you add jsdoc comments for all of the functions in this file?

src/components/List/List.story.jsx
davidicus: Lets rename to " A long list title to showcase what happens when titles is too long for the container...yada yada" or whatever. This way it doesn't look accidental.

src/components/List/HierarchyList/_hierarchy-list.scss
marcJV: In this case bx--radio-button-wrapper is overriding the margin here. I will need the !important here, but i removed the other ones.

src/components/List/List.story.jsx
marcJV: Will put this back, it looked like a visual bug at first.

src/components/List/ListItem/ListTarget.jsx
davidicus: Yes. we should remove these.

davidicus:
There is an issue with the items being able to be dropped into themselves and then disappearing from list.

davidicus:
> For the `allow item movement` knob, its working great, but its not obvious to the user why they can't drag. Maybe you could remove the select icon or add some kind of disabled icon?

I agree with this sentiment. If the user can not re-order we should probably not render the icon or allow the outlines when dragging.

src/components/List/_list.scss
```diff
@@ -10,7 +10,6 @@
   }
   &--content {
     flex: 1 1 0 0;
-    background: $ui-01;
```
davidicus, line 13: I am adding this back as this leaves a blank space in list that are not large enough 
image

src/components/List/ListItem/ListTarget.jsx
sstone2423: Could we still address this?

davidicus:
I will be closing this in Favor of #1564. @sstone2423 if you could check out this other PR? there are some changes from this PR. 

@marcJV can you address @sstone2423 comment above about the unused prop over in the new issue?

PR #1454, opened 2020-07-30
feat(TimeSeriesCard/BarChartCard): add zoomBar functionality

sstone2423:
Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1456

**Summary**

- Adds alpha zoomBar functionality to TimeSeriesCard and BarChartCard using the new `zoomBar` prop
- Restricts the zoomBar to cards of size MEDIUMWIDE, LARGE, LARGEWIDE as the other sizes are too small to render both the chart and the zoomBar
- Note: This feature is experimental in carbon charts
```
zoomBar: {
  axes: string,   // defaults to top as thats the only axis supported at the moment
  enabled: boolean,
  initialZoomDomain: array<string|number>    // same as the current `domainRange` prop
}
```

- Added 2 stories to show the zoomBar feature
- Updated to latest carbon charts to pick up this feature

- Go to BarChartCard and TimeSeriesCard `with zoomBar` stories and test out the zoom bar

### DIFF
--- a/src/components/TimeSeriesCard/TimeSeriesCard.jsx
+++ b/src/components/TimeSeriesCard/TimeSeriesCard.jsx
@@ -76,7 +76,8 @@ const TimeSeriesCardPropTypes = {
     /** Optionally addes a zoom bar to the chart */
     zoomBar: PropTypes.shape({
       /** Determines which axis to put the zoomBar */
-      axes: PropTypes.string,
+      axes: PropTypes.oneOf(['top']), // top is the only axes supported right now
+      // axes: PropTypes.oneOf(['top', 'bottom', 'left', 'right']), // TODO: When the other axes are supported, swap to this proptype
       /** Determines whether the zoomBar is enabled */
       enabled: PropTypes.bool,
       /** Optional domain to zoom to by default. Can be a timestamp or date string */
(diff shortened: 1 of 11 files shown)


src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -68,6 +73,17 @@ const TimeSeriesCardPropTypes = {
     ),
     /** optional units to put in the legend */
     unit: PropTypes.string,
+    /** Optionally addes a zoom bar to the chart */
+    zoomBar: PropTypes.shape({
+      /** Determines which axis to put the zoomBar */
+      axes: PropTypes.string,
```
tay1orjones, line 79: I'm not familiar with the prop model that much. Is there a specific set of axes? What format should the string be in? Wondering if we should make it an enum for clarity
```
axes: PropTypes.oneOf(['x-axis', 'y-axis']),
```
or
```
axes: PropTypes.oneOf(['top', 'bottom', 'left']),
```

tay1orjones: I think we also would need to update this in `CardPropTypes`

sstone2423: Oh good call. I'll update it to be `axes: PropTypes.oneOf(['top', 'bottom', 'left']),` , where I could actually enforce the `'top'` until it the other 3 are supported

Issue #1358, opened 2020-07-01
Table columns resizing is not working correctly


agupta2-ibm-maximo:
Table columns resizing is not working correctly

Team,

We are observing an issue on table where column resizing is not working, As per our observations, Table is not able to fetch width of columns correctly and need to reload page to make it work sometimes even that did not work, Due to this column resizing feature is not working.
Please refer attached screenshots ("currentColumnWidths" object not able to populate width showing "undefined")
![WidthData](https://user-images.githubusercontent.com/56728062/86273793-64719f80-bbee-11ea-809c-12562bca3cba.JPG)
![NoWidthData](https://user-images.githubusercontent.com/56728062/86273797-663b6300-bbee-11ea-9981-f5de79dd9098.JPG)

JordanWSmith15:
FYI, this environment shows it breaking: https://pages.github.ibm.com/maximo-app-framework/graphite/reference-app/#/tables

davidicus:
There seems to be several props that are undefined. Do you know why these are not being set?

![image](https://user-images.githubusercontent.com/1590966/86277109-28b1e800-bba4-11ea-90de-a9c8f3bd3b2d.png)

davidicus:
Thanks for reporting this bug. Please provide some more info on the options that are set for this Table.

hsharma1-ibm-maximo:
Thanks for looking into the issue, Please refer attached screenshots of options we are setting, First screenshot (resize not working) and second screenshot (resize working). They both looks identical. Also we observe even if resizing work after page load clicking on one of sortable column also breaks resizing feature. Please suggest.

![notWorking](https://user-images.githubusercontent.com/67732023/86331406-afc79480-bc66-11ea-8203-0ea5a457d933.JPG)
![working](https://user-images.githubusercontent.com/67732023/86331414-b1915800-bc66-11ea-8460-aab92be1f6a3.JPG)

hsharma1-ibm-maximo:
Steps to replicate - 
Storybook URL - https://pages.github.ibm.com/maximo-app-framework/graphite/react/storybook/

Scenario 1 -
1. Navigate to storybook using above URL.
2. Check "Can Resize" knob.
3. Perform resizing it should work.
4. Uncheck and re-check the "Can Resize" knob.
5. Perform resizing it did not work.

Scenario 2 -
1. Navigate to storybook.
2. Check "Can Resize" knob.
3. Perform resizing it should work.
[...]

PR #1455, opened 2020-07-30
fix(table): column resize fails when adding columns

### LINKED ISSUE #1428: [table] adding new column definitions after initial rendering fails when hasResize is true
### Describe the bug
Adding new column definitions after initial rendering fails when hasResize is true. This issue was identified in the discussion
https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1358

### To Reproduce

1. Manually add a new column after the initial rendering. I don't think we currently have a story for this
but story code below can be used.
2. Click on 'Add column'
3. Notice how the new column header is missing

### Expected behavior

The new column header should be showing and resizing should work. It should also work with adding two new or more new column at a time

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context

Story example code:

```
  .add(
    'with resize, onColumnResize callback and no initial column width',
    () => {
      return React.createElement(() => {
        const modifiedColumns = [
          ...tableColumns,
          {
            id: 'newField',
            name: 'Brand new field',
          },
        ];

        const modifiedOrdering = [
          ...defaultOrdering,
          {
            columnId: 'newField',
            isHidden: false,
          },
        ];

        const [myColumns, setMyColumns] = useState(tableColumns);
        const [myOrdering, setMyOrdering] = useState(defaultOrdering);

        const modify = () => {
          setMyColumns(modifiedColumns);
          setMyOrdering(modifiedOrdering);
        };

        const onColumnResize = cols => setMyColumns(cols);
        return (
          <>
            <button onClick={modify}>Add column</button>
            <Table
              options={{
                hasResize: true,
                wrapCellText: select('wrapCellText', selectTextWrapping, 'always'),
              }}
              columns={myColumns}
              view={{
                filters: [],
                table: {
                  ordering: myOrdering,
                },
              }}
              data={tableData}
              actions={{ ...actions, table: { ...actions.table, onColumnResize } }}
            />
          </>
        );
      });
    }
  )
```

upcoming release?

date, the better!_

bjornalm:
Closes #1428

**Summary**
Adding new column definitions (by adding them to the columns and ordering props) after initial rendering fails when hasResize is true. The same problem appears when removing columns. This issue was identified in the discussion #1358

After the initial rendering the resizable column widths are now updated in the case of
a. The columns and order props are modified, either with new columns or by having columns removed.
b. The columns prop is modified and all the visible columns in that new prop state have a width.

1. Updated a table story to be able to reproduce and test the scenario where columns are added and removed by an application. This is different from toggling the visibility of already defined columns.
image

2. Fixed the cause of the issue in the useDeepCompareEffect of the TableHead.jsx. Now that hook will also check if any columns are added or removed and deal with that.
3. Added additional functions to columnWidthUtilityFunctions.js and refactored some existing ones to be able to handle multiple column changes (i.e. add/show or remove/hide) at the same time.
4. Added a small improvement to prevent columns from shrinking below the minimum size when new columns are added or shown. 
5. Added unit tests to cover the new functionality
6. Cleaned up some left over comments in TableSaveViewModal.test.jsx
7. Introduced a warning for supplying incorrect width in the column prop. A column width should be a string containing the width and the pixel unit  e.g. `100px` or have the value `undefined`. For backwards compatibility the resizing can also handle an empty string value which will have the same effect as not defining a width.

1. Go to the modified story [/watson-iot-table--with-resize-oncolumnresize-callback-no-initial-column-width-and-column-management](https://deploy-preview-1455--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--with-resize-oncolumnresize-callback-no-initial-column-width-and-column-management)
2. Use the add and remove buttons and the input fields to add/remove columns

### DIFF
--- a/src/components/Table/Table.story.jsx
+++ b/src/components/Table/Table.story.jsx
@@ -9,6 +9,8 @@ import Add from '@carbon/icons-react/lib/add/16';
 import Edit from '@carbon/icons-react/lib/edit/16';
 import { spacing03 } from '@carbon/layout';
 import { Add20, TrashCan16, SettingsAdjust16 as SettingsAdjust } from '@carbon/icons-react';
+import cloneDeep from 'lodash/cloneDeep';
+
 import {
   Tooltip,
   TextInput,
@@ -17,9 +19,7 @@ import {
   Button,
   FormGroup,
   Form,
-} from 'carbon-components-react';
-import cloneDeep from 'lodash/cloneDeep';
-
+} from '../../index';
 import { getSortedData, csvDownloadHandler } from '../../utils/componentUtilityFunctions';
 import FullWidthWrapper from '../../internal/FullWidthWrapper';
 import FlyoutMenu, { FlyoutMenuDirection } from '../FlyoutMenu/FlyoutMenu';
@@ -1694,7 +1694,7 @@ storiesOf('Watson IoT/Table', module)
                 <Checkbox
                   labelText="add as hidden column(s)"
                   id="isHiddenCheckbox"
-                  value={isHidden}
+                  defaultChecked={isHidden}
                   onChange={() => setIsHidden(!isHidden)}
                 />
               </FormGroup>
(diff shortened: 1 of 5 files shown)


src/components/Table/Table.story.jsx
```diff
@@ -9,7 +9,15 @@ import Add from '@carbon/icons-react/lib/add/16';
...
-import { Tooltip, TextInput, Checkbox, ToastNotification, Button } from 'carbon-components-react';
+import {
+  Tooltip,
+  TextInput,
+  Checkbox,
+  ToastNotification,
+  Button,
+  FormGroup,
+  Form,
+} from 'carbon-components-react';
```
tay1orjones, line 20: I know it's just a story file, but would prefer to use as few direct imports from `carbon-components-react` as possible and instead use our entrypoint.

```suggestion
} from '../../index';
```

bjornalm: I agree, will change this.

PR #1458, opened 2020-07-31
fix(table): allow empty row actions for some rows of a table

### LINKED ISSUE #1457: [Table] if one row has actions and another doesn't in the same table the rows actions column is missing instead of empty
### Describe the bug

To recreate, create a Table where one row has action and another doesn't in the same table.  To recreate, set rowActions to `[]` for that one row. The picture shows where the rows without actions show a strange empty column

image

scottdickerson:
Closes #1457 

**Summary**

- Allow an empty array to be passed for 'row actions' and have it render the row actions cell with empty actions for that row so the table isn't missing a column for that row

- test(Table.story): have one of the stories skip row actions for a row every four rows
- fix(TableBodyRow): allow empty row actions to actually render the cell
- fix(RowActionsCell): allow empty row actions to actually render the cell so the table looks good

- Test the one updated story: `watson-iot-table--with-row-expansion-and-actions`

### DIFF
--- a/src/components/Table/Table.story.jsx
+++ b/src/components/Table/Table.story.jsx
@@ -1154,32 +1154,33 @@ storiesOf('Watson IoT/Table', module)
         columns={tableColumns}
         data={tableData.map((i, idx) => ({
           ...i,
-          rowActions: [
-            idx % 4 === 0
-              ? {
-                  id: 'drilldown',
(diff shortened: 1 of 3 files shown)


scottdickerson:
@tay1orjones @davidicus can you take a look?

Issue #1086, opened 2020-04-13
[Table] columns rendering incorrectly


sstone2423:
### Describe the bug

When rendering different tables based on state conditions, incorrect columns from previous renders intermittently in the newly rendered table.

For example, in Monitor we render a different table based on the current data item that is selected. This bug only occurs on a second re-render of the page as the first render does not have previous data to clash with.

![Screen Shot 2020-04-13 at 1 55 14 PM](https://user-images.githubusercontent.com/25177649/79151033-5cdcf000-7d8f-11ea-9048-166c9a6131d7.png)

### To Reproduce

1. Go to a component using Table that renders different data based off conditional state
2. Force a re-render
3. See duplicate key warnings and *possibly* see an extra empty column being rendered from the previous table's state

### Expected behavior

No warnings should appear and the component should only render the current table state's columns, not the previous

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: 2.63.0

### Additional context

Internal issue: https://github.ibm.com/wiotp/monitoring-dashboard/issues/934

### Specific timeline issues / requests

Do you want this work within a specific time period? Is it related to an
upcoming release?

No specific timeline

_NB: The core contributors will try to work with your timeline, but it's not
guaranteed. The earlier you make a request in advance of a desired delivery
date, the better!_

tay1orjones:
Are you able to work around this by regenerating the key of the table itself when the data changes?

Are you using `Table` or `StatefulTable` - if StatefulTable, this could be a bug in the reducer.

sstone2423:
I'll try that out. Yes this is StatefulTable

tay1orjones:
I think this will be fixed by #1470

tay1orjones:
@sstone2423 this should be fixed in v2.111.4 as we're now using unique id's by default for table components

PR #1470, opened 2020-08-06
fix(table): with multiple tables, select all button doesn't work

### LINKED ISSUE #1084: [StatefulTable] cannot select all in StatefulTable if that table is in a modal
### Describe the bug

I add a StatefulTable in a modal, table data is populated, each line can be selected, but "Select all" checkbox cannot be selected, when click "Select all" checkbox, no change in the table, only the modal "X" close icon is focused.

### To Reproduce

1. Add StatefulTable in a modal, show in page
2. Click on 'select all' checkbox in the modal
3. See error, no line is selected, "X" close icon in the modal is focused

image

> https://codesandbox.io/s/codesandbox-yxbkg

### Environment/versions:

- OS: [e.g. MacOS, Windows]
- Browser: [e.g. firefox, chrome]
- `carbon-addons-iot-react` version: [e.g. `v2.63.0`]

upcoming release?

date, the better!_

Jessepy:
Closes #1084

**Summary**

- When there is a table in a modal with another table outside the modal, you can't select all. Fixed by adding unique id to select all button.

- As tables are created, the select all button has an incrementing id
- Updated snapshots that had select all buttons
- Added test to ensure show it works

- make sure that test runs without error
- if you want to go the extra mile, verify visually by adding 2 tables, one in a modal, in a story then ensure select all buttons work

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -498,7 +498,7 @@ const Table = props => {
             clearSelectionText={i18n.clearSelectionAria}
             openMenuText={i18n.openMenuAria}
             closeMenuText={i18n.closeMenuAria}
-            tableId={`${id || tableId}-head`}
+            tableId={id || tableId}
             tableState={{
               isDisabled: rowEditMode || singleRowEditMode,
               activeBar: view.toolbar.activeBar,
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -86,7 +86,12 @@ const propTypes = {
     filters: PropTypes.arrayOf(
       PropTypes.shape({
         columnId: PropTypes.string.isRequired,
-        value: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.bool]).isRequired,
+        value: PropTypes.oneOfType([
+          PropTypes.string,
+          PropTypes.number,
+          PropTypes.bool,
+          PropTypes.arrayOf(PropTypes.string),
+        ]).isRequired,
       })
     ),
     toolbar: PropTypes.shape({
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -4,6 +4,7 @@ import merge from 'lodash/merge';
 import pick from 'lodash/pick';
 import { Table as CarbonTable, TableContainer } from 'carbon-components-react';
 import isNil from 'lodash/isNil';
+import uniqueId from 'lodash/uniqueId';
 import classnames from 'classnames';
 import { useLangDirection } from 'use-lang-direction';
 
@@ -28,8 +29,6 @@ import TableBody from './TableBody/TableBody';
 import Pagination from './Pagination';
 
 const { iotPrefix } = settings;
-
-let tableIdNumber = 0;
 
 const propTypes = {
   /** DOM ID for component */
@@ -180,7 +179,7 @@ const propTypes = {
 };
 
 export const defaultProps = baseProps => ({
-  id: `Table_${tableIdNumber}`,
+  id: null,
   useZebraStyles: false,
   lightweight: false,
   title: null,
@@ -317,6 +316,7 @@ const Table = props => {
   const { maxPages, ...paginationProps } = view.pagination;
   const langDir = useLangDirection();
 
+  const [tableId] = useState(() => uniqueId('table-'));
   const [, forceUpdateCellTextWidth] = useState(0);
 
   const useCellTextTruncate = useMemo(
@@ -374,7 +374,7 @@ const Table = props => {
 
   const rowEditMode = view.toolbar.activeBar === 'rowEdit';
   const singleRowEditMode = !!view.table.rowActions.find(action => action.isEditMode);
-  tableIdNumber += 1;
+
   return (
     <TableContainer
       style={style}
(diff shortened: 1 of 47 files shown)


src/components/Table/TableHead/TableHead.jsx
```diff
@@ -279,7 +282,7 @@ const TableHead = ({
                     https://github.com/IBM/carbon-components-react/issues/1088 */}
             <Checkbox
               disabled={isDisabled}
-              id="select-all"
+              id={`select-all${selectAllId}`}
```
tay1orjones, line 285: What if we instead modified this component to take the tableId prop and pass it down from `<Table>`?

It would then be unique to the table but also would be configurable

Jessepy: Sounds like a good solution, I'll get that done!

src/components/Table/Table.jsx
```diff
@@ -178,7 +180,7 @@ const propTypes = {
 };
 
 export const defaultProps = baseProps => ({
-  id: 'Table',
+  id: `Table_${tableIdNumber}`,
```
tay1orjones, line 183: This incrementing counter feels odd. I don't think it's necessary as consumers (and our stories) should configure a tableID, although it's not considered a required prop. 

Should we just use `Date.now()`?

sstone2423: > I think something is up with the merge from resolving the conflicts - there are now lots of unrelated changes in here @davidicus @sstone2423
> 
> RowActionsCell, Table 'onColumnResize callback' story, ComposedModal, TableHead, columnWidthUtilityFunctions, etc

It looks like that charting changes are correct here. The diff is the same as master for the obvious ones that I know in BarChartCard. Could we close and reopen this, or reset the base and hopefully the diff will re-correct

sstone2423:
> I think something is up with the merge from resolving the conflicts - there are now lots of unrelated changes in here @davidicus @sstone2423
> 
> RowActionsCell, Table 'onColumnResize callback' story, ComposedModal, TableHead, columnWidthUtilityFunctions, etc

@tay1orjones 
It looks like that charting changes are correct here. The diff is the same as master for the obvious ones that I know in BarChartCard. Could we close and reopen this, or reset the base and hopefully the diff will re-correct?

sstone2423:
@tay1orjones the diff is normal after resetting the base 👍

src/components/Table/Table.jsx
bjornalm: Looks like a lot of people are adding counters to combat these kind of problems, but I  like @tay1orjones suggests another option is to generate a unique ID. With this solution the table is getting a new ID on each re-render, which doesn't seem optimal.

Perhaps we isolate the fix to the TableHead.jsx file and reduce some table complexity while keeping the ID fixed while the component is alive? We could use lodash and the react state do something like `const [ checkboxID ] = useState(() => uniqueId('select-all-'))` in TableHead.jsx.

I did notice that the checkbox itself has the following comment

```
/* TODO: Replace checkbox with TableSelectAll component when onChange bug is fixed
                    https://github.com/IBM/carbon-components-react/issues/1088 */
```

The URL is dead now, but we should probably look into that before adding more workarounds with the current checkbox.

tay1orjones: Even further, could probably avoid state and just useMemo?
```js
const tableID = useMemo(() => `table-${Date.now()}-${tableId}`, [tableId]); // could use counter instead of Date.now
```

IIRC That comment is pretty old, from the initial implementation of the Table. I've opened https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1520 to address separately

src/components/Table/TableHead/TableHead.jsx
```diff
@@ -40,6 +40,7 @@ const { iotPrefix } = settings;
 const { TableHead: CarbonTableHead, TableRow, TableExpandHeader } = DataTable;
 
 const propTypes = {
+  id: PropTypes.string.isRequired,
```
tay1orjones, line 43: I think we usually follow the convention that internal table components take a `tableId` instead of `id`
```suggestion
  tableId: PropTypes.string.isRequired,
```

src/components/Table/Table.jsx
bjornalm: Yea, useMemo could work but it is not guaranteed that the ID will not change, according to the docs:
"In You may rely on useMemo as a performance optimization, not as a semantic guarantee"

src/components/Table/Table.jsx
```diff
@@ -395,7 +397,7 @@ const Table = props => {
       view.toolbar.customToolbarContent ||
       tooltip ? (
         <TableToolbar
-          tableId={id}
+          tableId={id || tableId}
```
stuckless, line 400: trying to understand what this is doing... basically if we are passed an ID then we use it, otherwise we use the tableId which comes from the state that is some randomly generated id?  Why is tableId something that is stateful?

tay1orjones: I think github auto collapsed the outdated conversation we had about it in from my review: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1470#discussion_r471747991

src/components/Table/Table.jsx
stuckless: Seems to me that IF the ID is required for the table to operate correctly, then it should be required.  Hacking the id to avoid using a require prop seems wrong.

tay1orjones:
This may also fix https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1086

src/components/Table/Table.jsx
davidicus: @stuckless we are  using the useState hook to get a value that is not updated on re-renders. Since the default is only evaluated once and we do not provide an update method it will be unique for each table instance.

PR #1474, opened 2020-08-06
fix(BarChartCard/TimeSeriesCard): enable zoomBar when isExpanded is true

### LINKED ISSUE #1473: [BarChartCard/TimeSeriesCard] zoomBar doesn't show when card is expanded, but at incompatible size
### Describe the bug

If the card initially has an incompatible size with zoomBar, the zoomBar fails to render when `isExpanded` is true. Since `isExpanded` forces the card to increase its size through style, but doesn't update its actual `size`

### To Reproduce

1. Go to implementation of TimeSeriesCard or BarChartCard with zoomBar and isExpandable properties set and a size of MEDIUM
2. Click on expand button
3. See that the zoomBar does not display in the expanded state

### Expected behavior

zoomBar should always display in expanded state if the zoombar is enabled

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: `v2.106.2`

sstone2423:
Closes #1473 

**Summary**

- The `size` prop doesn't actually get updated when `isExpanded` is true. It just changes the container size through style, meaning the `zoomBar` conditional on the compatible sizes never gets triggered.

- Added an additional conditional to check for `isExpanded` when adding the `zoomBar` option
- Added a conditional in BarChartCard to check if the card is time-based as thats the only time we should zoom in
- Updated `isExpanded` stories to have the zoomBar enabled
- Move TimeSeriesCard legend to bottom

- Go to `isExpanded` stories of both TimeSeriesCard and BarChartCard and see that zoomBar is visible

### DIFF
--- a/.storybook/__snapshots__/Welcome.story.storyshot
+++ b/.storybook/__snapshots__/Welcome.story.storyshot
@@ -3313,6 +3313,18 @@ exports[`Storybook Snapshot tests and console checks Storyshots 0/Getting Starte
                   <div
                     className="bx--structured-list-td"
                   >
+                    FlyoutMenu
+                  </div>
+                </div>
+                <div
+                  className="bx--structured-list-row"
+                >
+                  <div
+                    className="bx--structured-list-td"
+                  />
+                  <div
+                    className="bx--structured-list-td"
+                  >
                     determineCardRange
                   </div>
                 </div>
--- a/package.json
+++ b/package.json
@@ -113,8 +113,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.10.2",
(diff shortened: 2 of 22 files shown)


sstone2423:
@tay1orjones Thats by design. It looks a little bit odd because the right side of the card doesn't have the same amount of padding but the zoom bar is only the width of the actual chart, not the axes labels. If that first, left side handle was all the way to the left, it would give the appearance that you're actually viewing data thats off the screen if that makes sense

tay1orjones:
Yeah, that makes sense. Sounds like the goal in the future will be the legend and zoom bar on the bottom of the chart. Legend on the bottom of the chart makes sense for now 👍

PR #1486, opened 2020-08-11
feat(CardRangePicker): allow for custom time range options

### LINKED ISSUE #1485: [CardRangePicker] allow for custom time ranges
### Summary

It would be useful to be able to give custom time ranges in the CardRangePicker. For an example use case in Monitor, we need different time ranges for TimeSeriesCards when raw metric values are used so that we can have a smaller like Last 8 hours, Last 4 hours, Last 2 hours, Last hour

upcoming release?

September release

### Available extra resources

Will add myself

sstone2423:
Closes #1485 

**Summary**

- Allows the CardRangePicker to receive custom `timeRangeOptions`

- Added a new prop `timeRangeOptions` that is identical to the current implementation of the default options to Card, CardToolbar, and CardRangePicker (so that it can be passed all the way through)
- Moved the current default options to CardToolbar so they could get their default i18n labels and so that other components using CardToolbar can receive the same default options
- Updated tests to use the new prop
- Added a test to ensure the custom option does display
- Added a story for custom `timeRangeOptions`

- Go to Card / custom range selector story and see that the options are customizable with the knob

### DIFF
--- a/src/components/Card/CardToolbar.jsx
+++ b/src/components/Card/CardToolbar.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import omit from 'lodash/omit';
 import { Close16, Popup16 } from '@carbon/icons-react';
@@ -6,6 +6,7 @@ import { OverflowMenu, OverflowMenuItem, Button } from 'carbon-components-react'
 import classnames from 'classnames';
 
 import { settings } from '../../constants/Settings';
+import { TimeRangeOptionsPropTypes } from '../../constants/CardPropTypes';
 import { CARD_ACTIONS } from '../../constants/LayoutConstants';
 
 import CardRangePicker, { CardRangePickerPropTypes } from './CardRangePicker';
@@ -35,15 +36,19 @@ const propTypes = {
   isExpanded: PropTypes.bool,
   className: PropTypes.string,
   ...omit(CardRangePickerPropTypes, 'onClose'),
-  /** Generates the available time range selection options */
-  timeRangeOptions: PropTypes.objectOf(PropTypes.string),
+  /** Generates the available time range selection options. Each option should include 'this' or 'last'.
+   * i.e. { thisWeek: 'This week', lastWeek: 'Last week'}
+   */
+  timeRangeOptions: TimeRangeOptionsPropTypes, // eslint-disable-line react/require-default-props
 };
+
 const defaultProps = {
   isEditable: false,
   isExpanded: false,
   className: null,
   timeRangeOptions: null,
 };
+
 const CardToolbar = ({
   i18n,
   width,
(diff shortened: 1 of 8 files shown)


src/constants/CardPropTypes.js
```diff
@@ -395,19 +395,10 @@ export const CardPropTypes = {
...
-    'thisWeek',
-    'thisMonth',
-    'thisQuarter',
-    'thisYear',
-    '',
-  ]),
+  /** Optional selected range to pass at the card level */
+  timeRange: PropTypes.string,
+  /** Generates the available time range selection options */
+  timeRangeOptions: PropTypes.objectOf(PropTypes.string),
```
tay1orjones, line 401: How does someone know what the intended format of these strings is?

src/components/Card/CardToolbar.jsx
```diff
@@ -48,9 +51,24 @@ const CardToolbar = ({
...
+    last24Hours: i18n.last24HoursLabel,
+    last7Days: i18n.last7DaysLabel,
+    lastMonth: i18n.lastMonthLabel,
+    lastQuarter: i18n.lastQuarterLabel,
+    lastYear: i18n.lastYearLabel,
+    thisWeek: i18n.thisWeekLabel,
+    thisMonth: i18n.thisMonthLabel,
+    thisQuarter: i18n.thisQuarterLabel,
+    thisYear: i18n.thisYearLabel,
+  };
```
tay1orjones, line 70: Should this be in a useEffect so it's updated if the i18n props change?

src/components/Card/CardToolbar.jsx
```diff
@@ -35,11 +35,14 @@ const propTypes = {
   isExpanded: PropTypes.bool,
   className: PropTypes.string,
   ...omit(CardRangePickerPropTypes, 'onClose'),
+  /** Generates the available time range selection options */
+  timeRangeOptions: PropTypes.objectOf(PropTypes.string),
```
tay1orjones, line 39: Should this be imported from the CardPropTypes so we use that proptypes definition abstraction everywhere?

src/components/Card/CardToolbar.jsx
sstone2423: Since its not in a hook, this should get re-called with any re-render, including a prop change. I could see the benefit of memoizing it maybe?

src/components/Card/CardToolbar.jsx
sstone2423: The problem here is that its required in CardRangePicker, but not required here or in Card

src/components/Card/CardToolbar.jsx
sstone2423: Or I should set these as the default i18n in CardRangePicker

src/components/Card/CardToolbar.jsx
tay1orjones: Ah 👍

src/components/Card/CardToolbar.jsx
sstone2423: Went ahead and memoized the defaults based off the i18n and timeRangeOptions prop

src/constants/CardPropTypes.js
sstone2423: Added more documentation and examples to the prop type definition

sstone2423:
Lets hold this. Need to test another scenario first

PR #1490, opened 2020-08-12
feat(zoombar): enable simple slider view

### LINKED ISSUE #1489: [TimeSeriesCard / BarChartCard] enable simple slider view
### Summary

Enable the simple slider view for zoomBar's

**Additional context**

https://github.com/carbon-design-system/carbon-charts/issues/698

upcoming release?

September release

### Available extra resources

I will add

sstone2423:
Closes #1489 

**Summary**

- Enabled the simple slider view for the zoomBar
- See https://github.com/carbon-design-system/carbon-charts/issues/698 for more information about the design specs

- Abstracted `ZoomBarPropTypes` as they are the same prop types in both components
- Updated `zoomBar` prop model to take in a new `view` prop that can be `slider_view` or `graph_view`
- Updated carbon charts to pick up this new feature
- Updated TimeSeriesCard stories to have a slider_view and graph_view story
- Updated BarChartCard to have a slider_view story, and not graph_view because it looks a bit odd with it
- Default both charts to use the `slider_view`

- Go to `with zoomBar` stories for TimeSeriesCard and BarChartCard and see that both the simple_view and graph_view are still working correctly

### DIFF
--- a/src/components/BarChartCard/BarChartCard.jsx
+++ b/src/components/BarChartCard/BarChartCard.jsx
@@ -192,7 +192,7 @@ const BarChartCard = ({
                       top: {
                         enabled: zoomBar.enabled,
                         initialZoomDomain: zoomBar.initialZoomDomain,
-                        type: zoomBar.type || 'slider_view', // default to slider view
+                        type: zoomBar.view || 'slider_view', // default to slider view
                       },
                     },
                   }
--- a/src/components/BarChartCard/BarChartCard.jsx
+++ b/src/components/BarChartCard/BarChartCard.jsx
@@ -192,7 +192,7 @@ const BarChartCard = ({
                       top: {
                         enabled: zoomBar.enabled,
                         initialZoomDomain: zoomBar.initialZoomDomain,
-                        type: zoomBar.type ? zoomBar.type : 'slider_view', // default to slider view
+                        type: zoomBar.type || 'slider_view', // default to slider view
                       },
                     },
                   }
--- a/src/components/BarChartCard/BarChartCard.jsx
+++ b/src/components/BarChartCard/BarChartCard.jsx
@@ -192,6 +192,7 @@ const BarChartCard = ({
                       top: {
                         enabled: zoomBar.enabled,
                         initialZoomDomain: zoomBar.initialZoomDomain,
+                        type: zoomBar.type ? zoomBar.type : 'slider_view', // default to slider view
                       },
                     },
                   }
--- a/src/components/TimeSeriesCard/TimeSeriesCard.jsx
+++ b/src/components/TimeSeriesCard/TimeSeriesCard.jsx
@@ -544,7 +544,7 @@ const TimeSeriesCard = ({
                         top: {
                           enabled: zoomBar.enabled,
                           initialZoomDomain: zoomBar.initialZoomDomain,
-                          type: zoomBar.type || 'slider_view', // default to slider view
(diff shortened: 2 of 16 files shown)


src/components/TimeSeriesCard/TimeSeriesCard.story.jsx
```diff
@@ -683,7 +683,7 @@ storiesOf('Watson IoT/TimeSeriesCard', module)
             zoomBar: {
               enabled: true,
               axes: 'top',
-              // initialZoomDomain: []
+              view: select('view', ['slider_view', 'graph_view'], 'slider_view'),
```
tay1orjones, line 686:
```suggestion
              view: select('zoomBar.view', ['slider_view', 'graph_view'], 'slider_view'),
```

src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -554,6 +544,7 @@ const TimeSeriesCard = ({
                         top: {
                           enabled: zoomBar.enabled,
                           initialZoomDomain: zoomBar.initialZoomDomain,
+                          type: zoomBar.type ? zoomBar.type : 'slider_view', // default to slider view
```
tay1orjones, line 547:
```suggestion
                          type: zoomBar.type || 'slider_view', // default to slider view
```

src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -192,6 +192,7 @@ const BarChartCard = ({
                       top: {
                         enabled: zoomBar.enabled,
                         initialZoomDomain: zoomBar.initialZoomDomain,
+                        type: zoomBar.type ? zoomBar.type : 'slider_view', // default to slider view
```
tay1orjones, line 195:
```suggestion
                        type: zoomBar.type || 'slider_view', // default to slider view
```

src/components/TimeSeriesCard/TimeSeriesCard.story.jsx
sstone2423: I think this may be happening because of all of the memoization in the card. I went ahead and made 2 separate stories instead

src/components/BarChartCard/BarChartCard.jsx
sstone2423: I think this may be happening because of all of the memoization in the card. I went ahead and made 1 slider_view story as the graph_view doesn't seem useful for BarChartCard

src/components/TimeSeriesCard/TimeSeriesCard.jsx
sstone2423: Updated to use ||

src/components/TimeSeriesCard/TimeSeriesCard.story.jsx
sstone2423: This is actually a bug. Looking into it now

sstone2423: Great find. I was actually calling the improper prop. I also realized that there's already an `content` object knob where the view can be changed

tay1orjones: Honestly I liked the old knob setup because it showed the options instead of being a freeform text input with no context as to what the options are, but it does flip the type of zoombar now 👍

Issue #1495, opened 2020-08-13
[HierarchyList] OverflowMenu on the list items don't scroll


JoelArmendariz:
### Describe the bug

If there is an `OverflowMenu` added to the `ListItems` in an `overflow: scroll` enabled `HierarchyList`, the menu does not stay with the menu icon.

### To Reproduce

Add a HierarchyList story with OverflowMenu and `overflow: scroll` to see the menu positioning doesn't follow.

### Expected behavior

The menu should stay stuck to the menu icon.

![d8b2a000-b569-11ea-97cd-a403868b84a1](https://user-images.githubusercontent.com/53092833/90145824-f6b8a780-dd45-11ea-86c0-17aa18ee1ca8.png)

PR #1496, opened 2020-08-13
fix(hierarchy-list): add prop for hierarchy menus

JoelArmendariz:
Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1495

**Summary**

When an `OverflowMenu` is used in the `HierarchyList`, the menu is positioned incorrectly because it it implemented as a portal. As a result it is positioned absolute, relative to the body instead of the container we would expect.

Relevant issue in Monitor: https://github.ibm.com/wiotp/monitoring-dashboard/issues/1125

Relevant issue in Carbon: https://github.com/carbon-design-system/carbon/issues/4755

Adding `data-floating-menu-container` and `style={{ position: 'relative' }}` to a wrapper div forces the menu to be placed back into the DOM where we would expect it, and to be positioned relative to the `HierarchyList` container itself, rather than the body

Go to the `With OverflowMenu` story  and verify that the position stays with the menu icon through the `scrollIntoView` and as you scroll.

### DIFF
--- a/src/components/List/HierarchyList/HierarchyList.jsx
+++ b/src/components/List/HierarchyList/HierarchyList.jsx
@@ -264,39 +264,41 @@ const HierarchyList = ({
   const delayedSearch = useCallback(debounce(textInput => handleSearch(textInput), 150), [items]);
 
   return (
-    <List
-      title={title}
-      buttons={buttons}
-      search={
-        hasSearch
-          ? {
-              value: searchValue,
-              onChange: evt => {
-                setSearchValue(evt.target.value);
-                delayedSearch(evt.target.value);
-              },
-            }
-          : null
-      }
-      items={pageSize != null ? itemsToShow : filteredItems}
-      expandedIds={expandedIds}
-      toggleExpansion={id => {
-        if (expandedIds.filter(rowId => rowId === id).length > 0) {
-          // remove id from array
-          setExpandedIds(expandedIds.filter(rowId => rowId !== id));
-        } else {
-          setExpandedIds(expandedIds.concat([id]));
+    <div data-floating-menu-container style={{ position: 'relative' }}>
+      <List
+        title={title}
+        buttons={buttons}
+        search={
+          hasSearch
+            ? {
+                value: searchValue,
+                onChange: evt => {
+                  setSearchValue(evt.target.value);
+                  delayedSearch(evt.target.value);
+                },
+              }
+            : null
         }
-      }}
-      i18n={i18n}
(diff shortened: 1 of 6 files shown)


src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -264,39 +264,41 @@ const HierarchyList = ({
...
-      }
-      items={pageSize != null ? itemsToShow : filteredItems}
-      expandedIds={expandedIds}
-      toggleExpansion={id => {
-        if (expandedIds.filter(rowId => rowId === id).length > 0) {
-          // remove id from array
-          setExpandedIds(expandedIds.filter(rowId => rowId !== id));
-        } else {
-          setExpandedIds(expandedIds.concat([id]));
+    <div data-floating-menu-container style={{ position: 'relative' }}>
```
tay1orjones, line 267: This style should be in the .scss file and applied via a class. Do we need the data attribute for something?

sstone2423, line 267: Make sure you add the comment and carbon issue link here so others know why its here

JoelArmendariz: I have the addons issue linked in the PR description already. I will add the carbon issue to the description.

JoelArmendariz: I will extract to scss.

The data attribute forces the Menu to be in the DOM as a child of what is rendering it. Currently the implementation of this in Carbon has it as a portal (https://reactjs.org/docs/portals.html) which places it outside of our tree of components. I have updated the description of this PR to have the relevant Carbon issue that had reference to this fix.

JoelArmendariz: This doesn't seem to be an issue for `SimpleList`. https://carbon-addons-iot-react.com/?path=/story/watson-iot-experimental-simplelist--list-with-overflow-menu

sstone2423: At the top of the PR description, set this PR to close that addons issue.

I meant to add a code comment above that `data-floating-menu-container` so developers will understand why its there

JoelArmendariz: Yes. I added it. Sorry, I miss-read.

JoelArmendariz:
@tay1orjones I have moved the data attribute and position fixed to live on the list item, which fixes the scrolling issue. Thanks for the suggestion!

Issue #1494, opened 2020-08-13
[DateTime Picker] Cancel doesn't work as expected after setting the date range programmatically


navjeets:
### Describe the bug

On setting the date range programmatically, the value displayed in text box changes as expected. However, clicking open the expanded view and clicking "Cancel" button, doesn't revert the selection to the programmatically set date range. Instead it gets set to the default value. It appears that the programmatically set value is not being applied properly (incorrect internal state update).

### To Reproduce

1. Pass a preset default value.
2. Programmatically set an absolute date range.
3. Click the calendar icon to show expanded view.
4. Click on back button and then on cancel button.
5. See that the current selection is not reverted to programmatically set value but the default value.

### Expected behavior

On clicking cancel, the current selection should revert to the programmatically set value (which is the previous value).

### Environment/versions:

- OS: all
- Browser: all

### Additional context

N/A

### Specific timeline issues / requests

N/A

navjeets:
Fixed by changes under PR 1497

PR #1497, opened 2020-08-13
fix(datetime picker): externally set range not selected in calendar

### LINKED ISSUE #1482: [DateTime Picker] Externally set date range is not selected in calendar
### Describe the bug

Externally set date range is not properly selected in calendar. E.g. If the range value of 2020-04-01 12:34 to 2020-04-06 10:49 is set, the calendar shows July 31 2025 as the active selection.
https://carbon-addons-iot-react.com/?path=/story/watson-iot-experimental-datetime-picker--selected-absolute

DateTimePicker_Issue

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-experimental-datetime-picker--selected-absolute
2. Click on the calendar icon
3. See the issue

### Expected behavior

The active selection in calendar should match the set date range.

### Environment/versions:

- OS: All
- Browser: All

### Additional context
N/A

N/A

navjeets:
Closes #1482 #1494

**Summary**

- The external value is passed in as "defaultValue". Updated code to ensure that the external value is properly set and assigned to "lastAppliedValue" as well.
- The absolute value was not being passed to Carbon DatePicker correctly. Current code passes the absolute value startDate and endDate to DatePickerInput as "value" prop. DatePickerInput doesn't have any "value" prop. Passing the start and end date as an array to DatePicker's "value" prop.
- DatePicker dateFormat has been set as "m/d/Y". However, the format of the date that is passed to the "value" prop depends on the "dateTimeMask". So when we set the "dateTimeMask" to anything but "MM/DD/YYYY", the current selection is not set in calendar. Changed code to format the absolute value dates to "MM/DD/YYYY" before passing to DatePicker so that its independent of the format specified in "dateTimeMask".

- https://github.com/carbon-design-system/carbon-addons-iot-react/commit/49187d9a2065cfb7316100ab39d75bd853adeb60

- Tested the changes in Predict Asset Timeline widget.

### DIFF
--- a/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/src/components/DateTimePicker/DateTimePicker.jsx
@@ -830,8 +830,8 @@ const DateTimePicker = ({
                         value={
                           absoluteValue
                             ? [
-                                moment(absoluteValue.startDate).format('MM/DD/YYYY'),
-                                moment(absoluteValue.endDate).format('MM/DD/YYYY'),
+                                moment(absoluteValue.startDate, dateTimeMask).format('MM/DD/YYYY'),
+                                moment(absoluteValue.endDate, dateTimeMask).format('MM/DD/YYYY'),
                               ]
                             : ''
                         }
--- a/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/src/components/DateTimePicker/DateTimePicker.jsx
@@ -488,8 +488,7 @@ const DateTimePicker = ({
     });
   };
 
-  const parseDefaultValue = (lastValue = null) => {
-    const parsableValue = lastValue || defaultValue;
+  const parseDefaultValue = parsableValue => {
     const currentCustomRangeKind = showRelativeOption
       ? PICKER_KINDS.RELATIVE
       : PICKER_KINDS.ABSOLUTE;
@@ -551,7 +550,9 @@ const DateTimePicker = ({
   useEffect(
     () => {
       if (defaultValue || humanValue === null) {
-        parseDefaultValue();
+        parseDefaultValue(defaultValue);
+        setLastAppliedValue(defaultValue);
+        setIsExpanded(false);
       }
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
@@ -570,6 +571,7 @@ const DateTimePicker = ({
   const onApplyClick = () => {
     setIsExpanded(false);
     const value = renderValue();
+    setLastAppliedValue(value);
     const returnValue = {
       timeRangeKind: value.kind,
       timeRangeValue: null,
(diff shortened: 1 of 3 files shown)


navjeets:
> @navjeets could we add a unit test to cover this functionality?

Sure. I'll add a new unit test for checking internal state after updating the default value.

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -551,7 +550,9 @@ const DateTimePicker = ({
   useEffect(
     () => {
       if (defaultValue || humanValue === null) {
-        parseDefaultValue();
+        parseDefaultValue(defaultValue);
+        setLastAppliedValue(defaultValue);
+        setIsExpanded(false);
```
tay1orjones, line 555: Why should the menu close when the defaultValue changes? What if the `expanded` prop is passed in as true?

src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -828,19 +827,17 @@ const DateTimePicker = ({
                         ref={datePickerRef}
                         onChange={onDatePickerChange}
                         onClose={onDatePickerClose}
+                        value={
+                          absoluteValue
+                            ? [
+                                moment(absoluteValue.startDate, dateTimeMask).format('MM/DD/YYYY'),
+                                moment(absoluteValue.endDate, dateTimeMask).format('MM/DD/YYYY'),
```
tay1orjones, line 834: I don't think it's necessary to format on every render. We also do this same format in `onDatePickerChange`, so on non-initial renders I think the format isn't necessary. 

Could we unify this and improve the performance by placing the format inside a useEffect that only runs when the absoluteValue changes? Or maybe a useEffect with an empty dependency array so it's only ran on mount and unmount?

src/components/DateTimePicker/DateTimePicker.jsx
navjeets: Removed setIsExpanded(false);

navjeets:
> > @navjeets could we add a unit test to cover this functionality?
> 
> Sure. I'll add a new unit test for checking internal state after updating the default value.

Added new test case 'should render with programmatically set absolute range'

src/components/DateTimePicker/DateTimePicker.jsx
navjeets: Moved the date formatting code to parseDefaultValue() function so that it executes only when the default value is set.

src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -348,6 +348,61 @@ describe('DateTimePicker', () => {
     ).toHaveLength(1);
   });
 
+  it('should render with programmatically set absolute range', () => {
+    const wrapper = mount(<DateTimePicker {...dateTimePickerProps} />);
```
davidicus, line 352: Since this is a sev 1 bug fix I am not going to hold up the PR but in the future we want to utilize the react testing library for any new test.

davidicus:
@navjeets Thank you for the contribution and getting this bug fixed!

PR #1519, opened 2020-08-18
feat(listitem): add tags

### LINKED ISSUE #1518: [ListItem] Add tags to item
### Summary

As part of https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1308 we need to be able to show tags (pills) in the list items.

**Additional context**
image

bjornalm:
Closes #1518 

**Summary**
Adds the possibility to add one or more tags to an ListItem in a List.
image
image

1. Added a new tag prop to the ListItem.jsx and a new ListTagsPropTypes PropType in SharedPropTypes.js since the prop type is shared between List and ListItem
2. Updated ListItem tests and stories
3. Modified the content prop in List.jsx to include an optional tags prop
4. Updated List tests and stories

1. Go to /story/watson-iot-experimental-listitem--basic-w-knobs
2. Enable tags using the knobs, verify that it looks ok
3. Enable all the other features and make sure it all plays well together
4. Verify that the rest of the ListItem stories look good
5. Go to /story/watson-iot-experimental-list--with-tags
6. Make sure that the tag is rendered properly for each item
7. Verify that the rest of the List stories look good

### DIFF
--- a/src/components/List/ListItem/ListItem.story.jsx
+++ b/src/components/List/ListItem/ListItem.story.jsx
@@ -16,6 +16,8 @@ storiesOf('Watson IoT Experimental/ListItem', module)
     const iconComponent =
       iconName === 'Star16' ? Star16 : iconName === 'StarFilled16' ? StarFilled16 : null;
     const rowActionSet = select('row action example', ['none', 'single', 'multi'], 'none');
+    const tagsConfig = select('tags example', ['none', 'single', 'multi'], 'none');
+
     const rowActionComponent =
       rowActionSet === 'single'
         ? [
@@ -39,6 +41,12 @@ storiesOf('Watson IoT Experimental/ListItem', module)
             </OverflowMenu>,
           ]
         : [];
+    const tagsData =
+      tagsConfig === 'single'
+        ? [{ type: 'blue', content: 'default' }]
+        : tagsConfig === 'multi'
+        ? [{ type: 'blue', content: 'default' }, { type: 'red', content: 'warning' }]
+        : undefined;
     return (
       <div style={{ width: 400 }}>
         <ListItem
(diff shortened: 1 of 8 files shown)


src/constants/SharedPropTypes.js
```diff
@@ -4,3 +4,12 @@ export const OverridePropTypes = PropTypes.shape({
   props: PropTypes.object,
   component: PropTypes.elementType,
 });
+
+export const ListTagsPropTypes = PropTypes.arrayOf(
+  PropTypes.shape({
+    /** The content of the tag */
+    content: PropTypes.string,
+    /** The type (normally color, see Tag component) */
+    type: PropTypes.string,
```
tay1orjones, line 13: I would be okay with this being `PropTypes.arrayOf(PropTypes.node)` instead of an object structure so we won't have to track and update the structure to match `Tag` prop names when/if they change or are added to. This would also allow folks to use the Tag skeleton if the tag content is async/loading, as well as the additional `disabled` and `title` props available on Tag.

The example in storybook and the documentation of the `tags` prop specify that the intention is for an array of `Tag`s to be passed in.

src/components/List/ListItem/ListItem.story.jsx
```diff
@@ -185,4 +194,13 @@ storiesOf('Watson IoT Experimental/ListItem', module)
         ]}
       />
     </div>
+  ))
+  .add('with Tags', () => (
+    <div style={{ width: 400 }}>
+      <ListItem
+        id="list-item"
+        value={text('value', 'List Item')}
+        tags={[{ type: 'blue', content: 'default' }, { type: 'red', content: 'warning' }]}
```
tay1orjones, line 203: If we modify the proptype to be an array of nodes, this could be something like:

```suggestion
        tags={[<Tag type='blue' title='descriptor'>default</Tag>, <Tag type='red' disabled>disabled tag</Tag>]}
```

src/constants/SharedPropTypes.js
bjornalm: Yes, good idea.

PR #1521, opened 2020-08-19
Filter tags group initial implementation

### LINKED ISSUE #676: [Filters] Design for displaying filter tags outside a table
### Summary
https://github.ibm.com/Watson-IoT/pal-tracking/issues/29
https://jsw.ibm.com/browse/GRAPHITE-4806
### Description
When filters are set outside the context of a table, there's no clear pattern for invoking them, representing them, and dismissing them. In APM, we have filters that are applied to all visualizations on a page, including tables, maps, and charts.

We need an indicator that represents active filters and can be dismissed. When it's dismissed, the filters are cleared. Ideally, the tag would clearly indicate what the filter is, and if space is limited, it will have to truncate or remove content responsively.

### Name of Pattern or Asset
Add the name and a link to the pattern you are referencing in IoT or Carbon
https://ibm.invisionapp.com/share/U2NYW8D7DFV#/screens/319285777

### User outcome
APM user needs to see that there are active filters on her page and understand what they are, and she needs to be able to quickly clear them.

If the tag content is truncated, user should be able to tap / hover on the tag to see the details. She should be able to tap on the "x" to clear filters.

Proposal
https://ibm.invisionapp.com/d/main#/console/15350624/319220018/preview

Needed by APM

davidicus:
Closes #676 

**Summary**

Wrapper to add the functionality of collapsing filter tags that do not fit in their containers into an overflow menu. 

- Added `FilterTags.jsx`, `FilterTags.story.jsx`, `_filter-tags.scss`, and `FilterTags.test.jsx `
- Added import of styles to `styles.scss`
- Added export of this component to `index.js`

- Checkout story that was added for this component.

### DIFF
--- a/src/components/FilterTags/FilterTags.jsx
+++ b/src/components/FilterTags/FilterTags.jsx
@@ -1,24 +1,24 @@
 import React, { useRef, useEffect, useState, Children } from 'react';
 import PropTypes from 'prop-types';
 import useResizeObserver from 'use-resize-observer';
-import { OverflowMenuItem, OverflowMenu, Tag } from 'carbon-components-react';
+import { OverflowMenuItem, OverflowMenu } from 'carbon-components-react';
 import { Close16 } from '@carbon/icons-react';
-
+/* eslint-disable-next-line react/prop-types */
 const DefaultWrapper = React.forwardRef(({ children, ...props }, ref) => {
   return (
-    <span {...props} ref={ref}>
+    <div {...props} ref={ref}>
       {children}
-    </span>
+    </div>
   );
 });
-
+/* eslint-disable-next-line react/prop-types */
 const OverflowTag = ({ children }) => (
   <span>
     {children} <Close16 />
   </span>
 );
 
-const FilterTags = ({ children, hasOverflow, startingIndex }) => {
+const FilterTags = ({ children, hasOverflow, id }) => {
   const overFlowContainerRef = useRef(null);
   useResizeObserver({
     ref: overFlowContainerRef,
@@ -30,9 +30,10 @@ const FilterTags = ({ children, hasOverflow, startingIndex }) => {
 
   useEffect(
     () => {
+      /* istanbul ignore else */
       if (hasOverflow && overFlowContainerRef) {
         setOverflowItems([]);
-        setVisibleItems(childrenItems.slice(startingIndex));
+        setVisibleItems(childrenItems.slice(0));
       }
     }, // eslint-disable-next-line react-hooks/exhaustive-deps
     [children]
@@ -40,6 +41,7 @@ const FilterTags = ({ children, hasOverflow, startingIndex }) => {
 
   useEffect(
     () => {
+      /* istanbul ignore else */
       if (overFlowContainerRef.current) {
         const clientWidth = overFlowContainerRef.current?.clientWidth;
         const scrollWidth = overFlowContainerRef.current?.scrollWidth;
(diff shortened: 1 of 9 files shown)


src/components/FilterTags/FilterTags.jsx
```diff
@@ -0,0 +1,150 @@
+import React, { useRef, useEffect, useState, Children } from 'react';
+import PropTypes from 'prop-types';
+import useResizeObserver from 'use-resize-observer';
```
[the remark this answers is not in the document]
davidicus: This reminds me. The useResize that Marc added is still breaking the rules of Hooks in conditionals. We need to check back with the useResizeObserver repo to see if they were willing to put a check for support.

src/components/FilterTags/FilterTags.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+  useEffect(
+    () => {
+      /* istanbul ignore else */
+      if (hasOverflow && overFlowContainerRef) {
+        setOverflowItems([]);
+        setVisibleItems(childrenItems.slice(0));
+      }
+    }, // eslint-disable-next-line react-hooks/exhaustive-deps
+    [children]
```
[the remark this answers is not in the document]
davidicus: adding `hasOverflow` to key prop to just reset component if this prop changes.

src/components/FilterTags/FilterTags.jsx
```diff
@@ -0,0 +1,149 @@
...
+  const overFlowContainerRef = useRef(null);
+  useResize(overFlowContainerRef);
+  const childrenItems = Children.map(children, child => child);
+  const breakingWidth = useRef([]);
+  const [overflowItems, setOverflowItems] = useState([]);
+  const [visibleItems, setVisibleItems] = useState(childrenItems);
+
+  useEffect(
+    () => {
+      /* istanbul ignore else */
```
sstone2423, line 37: Is there a reason to not get coverage here? Its a viable use-case that Overflow is not set

src/components/FilterTags/FilterTags.jsx
```diff
@@ -0,0 +1,149 @@
...
+        setOverflowItems([]);
+        setVisibleItems(childrenItems.slice(0));
+      }
+    }, // eslint-disable-next-line react-hooks/exhaustive-deps
+    [children]
+  );
+
+  useLayoutEffect(
+    () => {
+      /* istanbul ignore else */
```
sstone2423, line 48: Is there a reason to not get coverage here? Its a viable use-case that Overflow is not set

src/components/FilterTags/FilterTags.jsx
sstone2423: Just realized you replied to similar comments from @tay1orjones . The `else` branch is here that this useEffect doesn't do anything so I think it should be tested for (which it looks it will be covered by your tests)

davidicus: There is no else logic which is what the coverage is complaining about. So it falsely reports a lack of coverage..

src/components/FilterTags/FilterTags.jsx
davidicus: same as above. I am only "ignoring" the lack of an else statement not the code present.

davidicus:
> It looks like the cursor only switches to 'pointer' on the 'More' overflow tag when you're hovering over the blue outlined box. Its not immediately intuitive to me that I can click it if I hover over too quickly and miss the cursor change

Good call on the pointer styles. I will add.

PR #1523, opened 2020-08-20
Multiselect table filter

JoelArmendariz:
**Summary**

Relevant ticket item: https://github.ibm.com/wiotp/monitoring-dashboard/issues/728

Adds the ability to filter a table based on a multiselect dropdown.

Add `isMultiselect` option on the `filter` prop for `Table`
Render `Multiselect` for filtering in the `FilterHeaderRow`
Update `TableReducer` logic to handle multiselect filters.

Go to the `Table example with multiselect filtering` story in storybook and confirm that the multiselect properly filters the table.

Screen Shot 2020-08-20 at 12 56 52 PM

### DIFF
--- a/src/components/Table/TableHead/FilterHeaderRow/_filter-header-row.scss
+++ b/src/components/Table/TableHead/FilterHeaderRow/_filter-header-row.scss
@@ -47,6 +47,15 @@
     padding-right: $spacing-03;
   }
 
+  .#{$iot-prefix}--filterheader-multiselect {
+    .bx--list-box__selection--multi {
+      background-color: #393939;
+    }
+    .bx--list-box__selection--multi > svg {
+      fill: #fff;
+    }
+  }
+
   .#{$prefix}--tag--filter {
     background-color: transparent;
 
--- a/src/components/Table/TablePropTypes.js
+++ b/src/components/Table/TablePropTypes.js
@@ -112,6 +112,7 @@ export const TableColumnsPropTypes = PropTypes.arrayOf(
     filter: PropTypes.shape({
       /** I18N text for the filter */
       placeholderText: PropTypes.string,
+      isMultiselect: PropTypes.bool,
       options: PropTypes.arrayOf(
         PropTypes.shape({
           id: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.bool]).isRequired,
--- a/src/components/Table/TableViewDropdown/__snapshots__/TableViewDropdown.story.storyshot
+++ b/src/components/Table/TableViewDropdown/__snapshots__/TableViewDropdown.story.storyshot
@@ -47,7 +47,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Table
           >
             <span
               className="bx--list-box__label"
-              htmlFor="downshift-10-input"
+              htmlFor="downshift-11-input"
               id="dropdown-field-label-iot--view-dropdown"
               title={
                 <TableViewDropdownItem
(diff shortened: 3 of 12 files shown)


src/components/Table/TableHead/FilterHeaderRow/_filter-header-row.scss
```diff
@@ -47,6 +47,15 @@
     padding-right: $spacing-03;
   }
 
+  .#{$iot-prefix}--filterheader-multiselect {
```
sstone2423, line 50: Could you add comments why this is needed?

src/components/Table/TablePropTypes.js
```diff
@@ -112,6 +112,7 @@ export const TableColumnsPropTypes = PropTypes.arrayOf(
     filter: PropTypes.shape({
       /** I18N text for the filter */
       placeholderText: PropTypes.string,
+      isMultiselect: PropTypes.bool,
```
sstone2423, line 115: Could you add the same prop comments as above?
```suggestion
/** if isMultiselect and isFilterable are true, the table is filtered based on a multiselect */
      isMultiselect: PropTypes.bool,
```

src/components/Table/TableViewDropdown/__snapshots__/TableViewDropdown.story.storyshot
```diff
@@ -47,7 +47,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Table
           >
             <span
               className="bx--list-box__label"
-              htmlFor="downshift-10-input"
```
sstone2423, line 50: Any idea why this changed?

src/components/Table/__snapshots__/Table.story.storyshot
```diff
@@ -2653,8 +2653,8 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Table
           >
             <label
               className="bx--pagination__text"
-              htmlFor="bx-pagination-select-5"
```
sstone2423, line 2656: Any idea why these changed?

src/components/Table/TableViewDropdown/__snapshots__/TableViewDropdown.story.storyshot
JoelArmendariz: Not sure why these changed.

src/components/Table/TableHead/FilterHeaderRow/_filter-header-row.scss
JoelArmendariz: Added

src/components/Table/TablePropTypes.js
JoelArmendariz: Added

JoelArmendariz:
Thanks @tay1orjones . That does sound very beneficial to hopefully catch some bugs with this as versions change. I will add some RTL test cases.

src/components/Table/TableViewDropdown/__snapshots__/TableViewDropdown.story.storyshot
tay1orjones: This is generated by downshift - the id counter is added to on every new instance of downshift (when snaps are generated) if an id isn't passed in. We are passing the ID into the carbon component, so I'm not sure why it's using the generated one instead of the one we pass in.

https://github.com/downshift-js/downshift/blob/a71005bd879d05d7dcb892d1e0dc5c6ca74e9d39/src/utils.js#L113-L119

I don't think there's any way for us to prevent it other than resetting the counter before snaps are ran. Either way, can ignore for now

JoelArmendariz:
@tay1orjones I have added some RTL test cases to cover preselected filters, non-preselected filters, adding filters, removing filters, and clearing the multiselect box.

PR #1529, opened 2020-08-22
fix(TimeSeriesCard): don't filter null values

### LINKED ISSUE #1528: [TimeSeriesCards] null values aren't displayed
### Describe the bug

When TimeSeriesCard is passed null values, it skips over them without any indication of the missing data.

### To Reproduce

1. Go to any TimeSeriesCard story
2. Pass it null values (along with valid values for the bigger picture)
3. See that the chart only displays the non-null values as if the null values never happened

### Expected behavior

TimeSeriesCard's should show all valid data, including null values. Its important to know that a device had missing or null data and there should be a gap between the lines to show this

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: 2.111.1

### Additional context

Related internal issue: https://github.ibm.com/wiotp/monitoring-dashboard/issues/523

upcoming release?

None

sstone2423:
Closes #1528 

**Summary**

- When null values are passed to TimeSeriesCard, we should display a gap in the line charts where the data is missing to give the user an indication of the null values

Related Carbon Charts PR: https://github.com/carbon-design-system/carbon-charts/pull/464/files

- Remove null filtering of data values
- Add additional conditional to timestamp to make sure the `undefined === undefined` race condition does not occur

- Go to any TimeSeriesCard story and pass it a mix of null and valid values
- You should see a gap in the line where the null values were passed

### DIFF
--- a/src/components/TimeSeriesCard/TimeSeriesCard.jsx
+++ b/src/components/TimeSeriesCard/TimeSeriesCard.jsx
@@ -156,10 +156,10 @@ export const formatChartData = (timeDataSourceId = 'timestamp', series, values) 
       // First filter based on on the dataFilter
       const filteredData = filter(values, dataFilter);
       if (!isEmpty(filteredData)) {
-        // have to filter out null values from the dataset, as it causes Carbon Charts to break
         filteredData
           .filter(dataItem => {
-            return !isNil(dataItem[dataSourceId]) && dataItem[timeDataSourceId] === timestamp;
+            // only allow valid timestamp matches
+            return !isNil(dataItem[timeDataSourceId]) && dataItem[timeDataSourceId] === timestamp;
           })
           .forEach(dataItem =>
             data.push({


src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -156,10 +156,10 @@ export const formatChartData = (timeDataSourceId = 'timestamp', series, values)
       // First filter based on on the dataFilter
       const filteredData = filter(values, dataFilter);
       if (!isEmpty(filteredData)) {
-        // have to filter out null values from the dataset, as it causes Carbon Charts to break
         filteredData
           .filter(dataItem => {
-            return !isNil(dataItem[dataSourceId]) && dataItem[timeDataSourceId] === timestamp;
+            // only allow valid timestamp matches
+            return !isNil(dataItem[timeDataSourceId]) && dataItem[timeDataSourceId] === timestamp;
```
scottdickerson, line 162: Isn't this still filtering out the null values here?

sstone2423: This is filtering the null or undefined timestamps. Without this, we wouldn't know when the null value came in. The swap here was to stop filtering out the actual null data values. The timestamp should still be valid though

bjornalm (approved):

PR #1537, opened 2020-08-25
feat(table): Manage view modal

### LINKED ISSUE #1308: [table] manage existing views modal
### Summary
As part of #1210 we need to create a modal that lets the user manage existing views.

**Additional context**
The modal would look like this:
image

- search name of existing views
- filter on public true/false
- list views displaying (name, description, private/public, default/not default
- trigger "view edit mode"
- trigger view deletion

**Questions**
@davidicus @lukefirth @amymillett
1. Do we need to support server search/filtering, i.e. some sort of async update that might take time?
2. How does a long list look like? e.g. use scrolling? load from server while scrolling? pagination?
3. Editing a view will open up the same modal as the "Save view" but pre filled with the name etc . So I assume that the "Manage Views Modal" will close, but what happens after the view was saved. Does the "Manage Views Modal" reopen or stay closed?
4. What happens when the user clicks on delete, should there not be a warning before a destructive action? Or should an undo option be presented in a toast or something?
5. Are the colors correct? The default modal background is white and the default search field is gray, in this design it is the other way around. 
6. Should all views in the list always be editable and deletable or should that be a property of the list item itself?

bjornalm:
Closes #1308

**Summary**
A highly customisable modal built for managing table views. The component is built as component outside the actual table but only make sense in the context of table view management.

Default:
image

Heavily customised:
image

1. Created component TableManageViewsModal that lets a user list, search, filter and take actions on views. The state needed for search, pagination etc is handled by the application via callbacks.
2. Created a TableManageViewsList component as a subcomponent of the TableManageViewsModal that in turn uses the List compoinent.
3. Created 2 stories as a sub section of the table stories. One that demonstrates and with info explains the full usage and the other how to customise the modal. 
4. Wrote unit tests

* Go to [/story/watson-iot-table-tablemanageviewsmodal--with-callbacks-implemented](https://deploy-preview-1537--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table-tablemanageviewsmodal--with-callbacks-implemented)
  * Test the knobs 
  * Test that all the functionality is working according to spec
  * Verify the info text
* Go to [story/watson-iot-table-tablemanageviewsmodal--with-custom-row-actions-custom-rendering-and-no-pagination](https://deploy-preview-1537--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table-tablemanageviewsmodal--with-custom-row-actions-custom-rendering-and-no-pagination)
  * Test that it looks ok and that the actions are working as storybook actions
  * Verify the info text

### DIFF
--- a/.storybook/__snapshots__/Welcome.story.storyshot
+++ b/.storybook/__snapshots__/Welcome.story.storyshot
@@ -2569,6 +2569,18 @@ exports[`Storybook Snapshot tests and console checks Storyshots 0/Getting Starte
                   <div
                     className="bx--structured-list-td"
                   >
+                    TableManageViewsModal
+                  </div>
+                </div>
+                <div
+                  className="bx--structured-list-row"
+                >
+                  <div
+                    className="bx--structured-list-td"
+                  />
+                  <div
+                    className="bx--structured-list-td"
+                  >
                     TableSkeletonWithHeaders
                   </div>
                 </div>
--- a/src/components/List/List.jsx
+++ b/src/components/List/List.jsx
@@ -24,6 +24,8 @@ export const itemPropTypes = {
 };
 
 const propTypes = {
+  /** Specify an optional className to be applied to the container */
+  className: PropTypes.string,
   /** list title */
   title: PropTypes.string,
   /** search bar call back function and search value */
@@ -64,6 +66,7 @@ const propTypes = {
 };
 
 const defaultProps = {
+  className: null,
   title: null,
   search: null,
   buttons: [],
@@ -87,6 +90,7 @@ const defaultProps = {
 const List = forwardRef((props, ref) => {
   // Destructuring this way is needed to retain the propTypes and defaultProps
   const {
+    className,
     title,
     search,
     buttons,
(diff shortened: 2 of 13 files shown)


davidicus:
There are some minor visual defects in RTL mode.

bjornalm:
> There are some minor visual defects in RTL mode.

Good catch, I've fixed a missing spacing between the checkbox and the search field and removed quoting ("") around the view name in the delete warning message that looked strange in RTL for some reason. The modal header also looked a bit strange but it is apparently how the standard modal header looks in RTL. I also notice the tooltip of the row actions, but that should be fixed in the ListItem component, so I created an issue for that [here](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1540).  Are there any other RTL issues that you noticed?

bjornalm:
> Not sure why it's got the component listed as `F` in the info section of the story. It looks good otherwise.
> 
> image

That is strange. I don't see that locally in dev mode, so it might be related to the build. @tay1orjones, any ideas? We are moving away from the Storybook addon-info right?

PR #1553, opened 2020-08-28
feat(hierarchylist) allow list items to be expanded by default #1545

### LINKED ISSUE #1545: [HierarchyList] allow list items to be expanded by default
There is no way to expand list items by default in the HierarchyList component. 

The HierarchyList currently initializes the internal expandedIds state to an empty array. If instead the state is initialized by a new defaultExpandedIds HierarchyList property, users will be able to expand list items by default.

tom-fitts:
Closes #1545

**Summary**

- This enhances the HierarchyList component to all list items to be expanded by default. The property array defaultExpandedIds has been introduced, which will take an array of list item ids that should be expanded by default.

- Added new HierarchyList property called defaultExpandedIds.
- Add a new storybook use case.

- Verify that item ids specified in the defaultExpandedIds array are expended when the HierarchyList renders.


sstone2423:
Looks like theres some snapshots failing after the merge with master @tom-fitts

src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -36,6 +36,8 @@ const propTypes = {
   pageSize: PropTypes.string,
   /** Item id to be pre-selected */
   defaultSelectedId: PropTypes.string,
+  /** Item ids to be pre-expanded */
+  defaultExpandedIds: PropTypes.array,
```
tay1orjones, line 40: Checks are failing because linting is failing due to rule [`react/forbid-prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md). 

This needs to specify what type is in the array. I think they're strings, right?
```suggestion
  defaultExpandedIds: PropTypes.arrayOf(PropTypes.string),
```

tom-fitts: @tay1orjones , I made the suggested to the code and added a unit test.

tay1orjones (approved):

Issue #1527, opened 2020-08-22
[cardUtilityFunctions] card variable functions should not iterate over the entire card object


sstone2423:
### Describe the bug

Currently when finding and replacing card variables in `card` objects, our logic attempts to iterate and search the entire card object, which includes functions and React nodes. We should really only iterate over the places that will actually have variables such as `title`, `description`, and `content`. This would save us computation time and complexity

Related functions: `getCardVariables`, `replaceCardVariables`

### Expected behavior

Only iterate over the card properties that could have variables

### Additional context

Related issue and PR:
https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1524
https://github.ibm.com/wiotp/maximo-asset-monitor-ui

### Specific timeline issues / requests

No specific timeline

PR #1567, opened 2020-09-02
fix(cardUtilityFunctions): do not loop over nullish values

sstone2423:
Related to https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1527

**Summary**

- If a card's props had a nested null value, `getCardVariables` `Object.values(card)` will fail, as its unable to convert null to an object

- Do not return null values to the accumulator in the reduce function

- Added a test case for null and React symbol values (the 2 main edge cases)


scottdickerson (approved):
LGTM this is due to a crazy javascript error

tay1orjones (approved):

PR #1570, opened 2020-09-03
feat(table): added column overflow menu

### LINKED ISSUE #1375: [Table] Column overflow menu
### Summary

From [Internal Watson-IoT/pal-tracking#161](https://github.ibm.com/Watson-IoT/pal-tracking/issues/161)

> This is an EDS item
> 
> Certain table actions (to come) affect only a specific column of data, this contribution covers a mechanism for how to access/apply these actions using an overflow menu.
> 
> Include mockups of any fidelity, the more spec the better as this will enable higher quality feedback and evaluation.
> 
> Image:
> image
> 

marcJV:
Closes #1375, and #1597 

**Summary**

- Adds support for an overflow menu in the the column header

- Added column header and callback for both table and stateful table

- Check out the `Simple Stateful Example with column overflow menu` story and click the overflow item + options. Observe the storybook actions tab for callbacks.

- Check column size adjust and column resort to make sure those are working correctly


bjornalm:
Did you verify that this works with the different column resize scenarios?

marcJV:
> Did you verify that this works with the different column resize scenarios?

Yes, they currently work on multiple resize scenarios. We may have to revisit the min-width when the overflow is shown because clipping can occur at the current min-width when the overflow button is shown

src/components/Table/TableHead/TableHead.jsx
```diff
@@ -350,14 +354,44 @@ const TableHead = ({
...
               </TableCellRenderer>
+
+              {matchingColumnMeta.options !== undefined && matchingColumnMeta.options !== null ? (
+                <OverflowMenu
+                  className={`${iotPrefix}--table-head--overflow`}
+                  direction="bottom"
+                  data-testid="table-head--overflow"
+                  flipped={columnIndex !== 0}
+                  onClick={e => {
+                    e.stopPropagation();
```
davidicus, line 376: Can we remove this since we are stoping propagation in the `OverflowMenuItem`?

src/components/Table/TableHead/TableHead.jsx
```diff
@@ -350,14 +354,44 @@ const TableHead = ({
...
+                  }}
+                >
+                  {matchingColumnMeta.options.map((option, index) => (
+                    <OverflowMenuItem
+                      itemText={option.text}
+                      primaryFocus={index === 0}
+                      key={`${columnIndex}--overflow-item-${index}`}
+                      onClick={e => {
+                        e.stopPropagation();
```
davidicus, line 385: let's define this onClick function outside of the mapping function and pass the event in
```
onClick={e => handleOnClick(e)}
```

src/components/Table/TableHead/_table-head.scss
```diff
@@ -91,15 +105,22 @@
...
   }
 
-  vertical-align: middle;
+  &--table-header {
+    .#{$prefix}--table-header-label {
+      height: 100%;
+      min-width: var(--table-header-width);
+      max-width: var(--table-header-width);
+      white-space: nowrap;
+      overflow-x: hidden;
```
davidicus, line 119: let's combine these into the shorthand property. 
```
      overflow: hidden;
```

davidicus (dismissed):

src/components/Table/TableHead/TableHead.jsx
marcJV: if the header is clickable (if column sorting is available) the overflow menu item will still trigger the sort.

bjornalm:
There seems to be a misalignment between the first column's overflow button and the rest
image

bjornalm:
The actual menu is out of place for the first column
image

src/components/Table/TableHead/_table-head.scss
```diff
@@ -91,15 +105,21 @@
...
-    text-overflow: ellipsis;
+.#{$iot-prefix}--table-head {
+  &--overflow {
+    flex-shrink: 0;
   }
 
-  vertical-align: middle;
+  &--table-header {
+    .#{$prefix}--table-header-label {
+      height: 100%;
```
bjornalm, line 115: Why was this height added? It seems to be the cause of this misalignment:
image

bjornalm (changes requested):
> > Did you verify that this works with the different column resize scenarios?
> 
> Yes, they currently work on multiple resize scenarios. We may have to revisit the min-width when the overflow is shown because clipping can occur at the current min-width when the overflow button is shown

I think that if the labels are being clipped we should fix that before we finish this. Perhaps we should ask @lukefirth to comment on this. Perhaps it is best to dynamically set the column min-width depending on the presence of the "overflow menu", and "column sort" since they both use up the label space. And we should make sure we are not wasting any space by excessive paddings/margins like shown below between the overflow menu button and the sort button.
image

src/components/Table/TableHead/_table-head.scss
marcJV: a fair amount of these issues popped up when I merged in master. I will resolve the issues that were fixed from that.

src/components/Table/TableHead/_table-head.scss
```diff
@@ -9,12 +9,26 @@
     height: 3rem;
     border-top: none;
     border-bottom: none;
+    position: relative;
```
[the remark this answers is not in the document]
marcJV: i have not seen any other side effects from this

bjornalm (approved):
Nice work!

marcJV:
> > > Did you verify that this works with the different column resize scenarios?
> > 
> > 
> > Yes, they currently work on multiple resize scenarios. We may have to revisit the min-width when the overflow is shown because clipping can occur at the current min-width when the overflow button is shown
> 
> I think that if the labels are being clipped we should fix that before we finish this. Perhaps we should ask @lukefirth to comment on this. Perhaps it is best to dynamically set the column min-width depending on the presence of the "overflow menu", and "column sort" since they both use up the label space. And we should make sure we are not wasting any space by excessive paddings/margins like shown below between the overflow menu button and the sort button.
> image

This has been fixed: adds more padding if the overflow item is present and even more when overflow and sort is present. Enough so to keep the label without clipping

tay1orjones (approved):

PR #1583, opened 2020-09-08
feat(table): create save view - statefulTable

### LINKED ISSUE #1210: [Table]  Create & Save views

bjornalm:
Closes #1210

**Summary**
This PR ties together all the components needed to let the user Create, Save, List, Select and manage views in a new story.

- Created a new story for Create, Save and Manage views for a StatefulTable. The story simulates a small app
with its own state and includes plenty of explanatory comments.
- Created a new story for Create, Save and Manage views for a normal Table. The story simulates a small app
with its own state and includes plenty of explanatory comments, but not all view altering callbacks have been implemented (because that would be building a new stateful table)
- Added a column resize reducer so that the StatefulTable could manage the state of the current columns width. This was needed to export the current widths.
- Added an option hasUserViewManagement to the Table/StatefulTable that enables some new functionality needed for this feature. The option allows backwards compatibility and reduces the risk of introducing regressions. 
- Added a onUserViewModified callback to Table/StatefulTable that is called each time the view prop, column prop or search value is changed. Only active if hasUserViewManagement=true.
- Added a ref to store the current searchValue in Table. This is needed since the current search value is not managed by the StatefulTable via the props like normal, but in the Carbon ToolbarSearch where we can't access it.
- Modified the tableRegister reducer to read ordering and default value from the state (when hasUserViewManagement is active). 
- Modified TableSaveViewForm to prevent form submission which was redirecting the page on enter. 
- Modified TableSaveViewModal to merge the i18n prop with the default values.
- removed clipping from the table toolbar that was preventing the TableViewDropdown from showing correctly
- Added a wrapper with a style prop to the TableViewDropdown so that we could set some demo styles
- Carefully Updated tests where needed

1. Verify the new story '[watson-iot-table--stateful-example-with-create-save-views](https://deploy-preview-1583--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-create-save-views)' by exploring all the functionality compared to the specs and common sense.
2. Verify that the existing Table stories still work as expected.

_A note on rendering_
When a user first modifies a filter (or change sort, types in search etc), the Table/StatefulTable render function is called. The table triggers callback `onUserViewModified` which in the story updates state `selectedViewEdited` and the table render function is called. The `selectedViewEdited` is passed as a prop to the TableViewDropdown which is updated and the table render function is called again. So in total there could be 3 rendering calls, but this only happens if the change modifies the `selectedViewEdited`. There is nothing specific in `TableViewDropdown` that causes this additional render except the fact that it depends on the `selectedViewEdited` and is passed in as a node to the table via the `customToolbarContent`.


davidicus (approved):
I think this is good to go. Looks good!

filipecorrea (approved):

PR #1586, opened 2020-09-09
chore(card): added gauge and value card expand stories

### LINKED ISSUE #1461: [Dashboard cards] Need "pop out" expand icon in card header
### Summary

Need "expand" option in dashboard card header.

image

**Additional context**
Health & Predict are following the pattern that Monitor set for expanding cards to show a larger view of the data, but it looks like the expand button isn't available in the IoT cards.  We need to expand a value card first, but also need this option on chart cards.

Click on the expand icon in the upper right to trigger an action.
image

_______
FYI: In Health case, this trigger would open a dialog that has additional info, but we do not need that capability built into the trigger.
image

upcoming release?

Health needs this for our October 2020 release.  
Lack of this button blocks creation of the dialog that is invoked when the user clicks the button.

### Available extra resources

The Predict team hard-coded this into their chart cards, copying the Monitor team, and should hopefully be able to contribute it back.

marcJV:
Closes #1461

**Summary**

- Added stories to show cards can have expanded states
- Added ability to change the icon for the expand button

- Added custom icon story to card
- Added expand story in gauge card
- Added expand story in value card

Verify the above stories


src/components/GaugeCard/GaugeCard.story.jsx
```diff
@@ -69,6 +69,35 @@ storiesOf('Watson IoT Experimental/GaugeCard', module)
       </div>
     );
   })
+  .add('basic with expand', () => {
```
sstone2423, line 72: It might be helpful to add something into the expanded content as an example as to something that would be there. It looks very weird being so empty, and the arrow seems to go off center.

Screen Shot 2020-09-12 at 10 03 22 AM

src/components/ValueCard/ValueCard.story.jsx
```diff
@@ -38,6 +38,37 @@ storiesOf('Watson IoT/ValueCard', module)
...
+              },
+            ]),
+          }}
+          availableActions={{
+            expand: true,
+          }}
+          breakpoint="lg"
+          size={size}
+          values={{ occupancy: number('occupancy', 88) }}
+          onCardAction={() => action('Value Card Expand Action')}
```
bjornalm, line 67: Clicking the expand button causes an error
image

src/constants/CardPropTypes.js
```diff
@@ -413,6 +413,11 @@ export const CardPropTypes = {
   isEditable: PropTypes.bool,
   /** goes full screen if expanded */
   isExpanded: PropTypes.bool,
+  /**
+   * Define the icon render to be rendered.
+   * Can be a React component class
```
bjornalm, line 418: This comment is a bit confusing, does it have to be a react **class**?

bjornalm (changes requested):
Nice work, just a couple of fixes needed.

src/constants/CardPropTypes.js
marcJV: This is the same comment that appears in Button.jsx in the carbon library, keeping it the same to avoid confusion for passing in renderIcons

src/components/ValueCard/ValueCard.story.jsx
marcJV: Fixed

src/components/GaugeCard/GaugeCard.story.jsx
marcJV: Fixed this by removing the expand update, this is to keep it in line with the other expand stories

bjornalm (approved):

tay1orjones (approved):

Issue #1563, opened 2020-09-01
[CardCodeEditor] enable editing of dashboard card config json


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

In support of the dashboard builder/editor #1372, provide a way to edit the card configration json directly through an editor placed in a modal. 
- Per the spec, editor should support fullscreen and truncation through "show more" button. We should be able to utilize the [multi-line CodeSnippet component](https://carbon-addons-iot-react.com/?path=/story/codesnippet--multi-line) for this and use the Monaco editor within.
- On save, the edited JSON should be passed via a callback.

Consuming applications are already using the [Monaco editor](https://github.com/microsoft/monaco-editor), and I believe the [Monaco JSON plugin](https://github.com/microsoft/monaco-json).
- Add Monaco editor as a peer dependency
- Add Monaco JSON plugin as a peer dependency
- Update installation documentation (and the story for this component) to highlight that in order for this to work they must install this dependency. We might also consider editing the default installation instructions on the readme to highlight this. In theory if someone isn't using this component they shouldn't have to install it, but they'll get peer dep warnings in the console.

**Additional context**
![image](https://user-images.githubusercontent.com/3360588/91899194-6bef0c80-ec62-11ea-8dff-b474d2f523af.png)

Issue #1562, opened 2020-09-01
[CardEditForm] configurable form for editing dashboard card parameters


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

In support of the dashboard builder/editor #1372

Build a component that maps to the JSON configuration schema of a dashboard card, mapping the configrable keys to form inputs.

Most form inputs would be standard - some could be enhanced such as the "Card type" which could potentially use Dropdown with icon selection buttons, #1259 (this could be a stretch goal though - perhaps we use the stock dropdown in the mvp)

This will probably also need to support some sort of 'slot' pattern where components could be passed in and the value map to the json key, such as the 'upload image' section in the spec below. Also will want to keep in mind that the consuming applications will likely be using Formik, so our API design for this should be informed by that.

@bryancboyd I think you have a solid idea in your mind of how this would work beyond these broad strokes I've outlined. it sounds like you'll take this on, so will defer additional implementation details/conversation to your eventual PR 🙂 

**Additional context**
![image](https://user-images.githubusercontent.com/3360588/91898138-e74fbe80-ec60-11ea-9802-f258937aa583.png)

davidicus:
has been de-prioritized for Monitor. Removing from Q3

davidicus:
@sstone2423 Is there anything left for this issue to accomplish or can we close this one out?

sstone2423:
@davidicus the base idea of this implemented, but we are still missing card-specific form items for all card types except TimeSeriesCard.

There is also no way of adding custom card types at the moment beyond the basic 'Custom'. I think this could be a separate issue though.

There is also no way of adding additional custom form items for specific use-cases like the varying ways consumers use `dataSource`

There is also virtually no real validation for cards by default. We simply do a basic valid JSON check so we would need to add in some kind of proper schema for the basic necessities like card size that can cause the editor to blow up (although there is some error boundaries to prevent  a complete crash

I'd say these could all be created into separate issues though if you'd like to close this one out

tay1orjones:
Closing - follow up issues created to continue this work:

- [CardEditor] support editing for TableCard #1739
- [CardEditor] support editing for ValueCard #1740
- [CardEditor] support editing for ImageCard #1741

If there are more we should create, please do so or let me know 👍

Issue #1561, opened 2020-09-01
[CardGalleryHeader] dashboard builder/editor pane header


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

In support of the dashboard builder/editor #1372

The Gallery/Editing pane should have a standardized header. From the spec it appears to have two states/configurations
1. Gallery
1. Return to Gallery

This would be a very small component but something that we could publish for ease of design consistency and standardization. Would need to accept props: 
- i18n object prop for the strings
- a `bool` prop `isGallery` which would toggle the two different states. 
- callback for onGalleryClose (the arrow)
- callback for onReturnToGallery (the "open gallery" link and the icon)

**Additional context**

![image](https://user-images.githubusercontent.com/3360588/91897471-f41fe280-ec5f-11ea-81b2-0527c10e47bc.png)
![image](https://user-images.githubusercontent.com/3360588/91897622-2a5d6200-ec60-11ea-8b53-20cf39cb2aeb.png)

bryancboyd:
How about we treat this component as the parent and layout manager of both the CardGalleryList and CardEditForm components (maybe, `<CardEditor>`), and include the header rendering?

Example usage:

```
<CardEditor
  isGallery
  supportedTypes={[CARD_TYPES.LINE, CARD_TYPES.VALUE]}
  i18n={{
    listItemTimeSeries: 'Time series line',
    listItemSimpleBar: 'Simple bar',
    etc...
  }}
  onClose={/* triggered when gallery closes */}
  onAddCard={(cardType) => {}}
  onCardChange={(newCardJSON) => {}}
/>
```

Which would render:

```
<div className="iot--card-editor">
  <div className="iot--card-editor--header">
    { isGallery ? (Gallery -->) : <Button>Open gallery</Button> }
  </div>
  { isGallery ? <CardGalleryList /> : <CardEditForm /> }
</div>
```

tay1orjones:
@bryancboyd what's left for this one since your PR?

sstone2423:
This is managed by CardEditor if the gallery is not showing and CardGalleryList if the gallery is showing. The reasoning is that CardGalleryList can use SimpleList's header

davidicus:
> This is managed by CardEditor if the gallery is not showing and CardGalleryList if the gallery is showing. The reasoning is that CardGalleryList can use SimpleList's header. It would be a good idea to still create a proper CardEditorHeader to keep it consistent though

[...]

Issue #1560, opened 2020-09-01
[CardGalleryList] List variant with large items, pictograms


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Create a variation of the List component in support of the dashboard builder/editor #1372

- CardGalleryList component would accept array of list items with card type/title that would map to predefined list. These list items would not support hierarchy (`children`). The `icon` prop could be repurposed to allow pictograms from `@carbon/pictograms-react` 
- It would be nice if we could enforce certain types of cards to have certain pictograms by default. Maybe the predefined list of card types would map item `value` strings to an associated default pictogram from `@carbon/pictograms-react` 
- Additionally a 'custom' list item could be configured and passed in an optional pictogram. If no pictogram is provided, we could choose a default, something generic like `<Analyze />` maybe?
![image](https://user-images.githubusercontent.com/3360588/91896123-f5e8a680-ec5d-11ea-80df-e7778d5dac41.png)
- per the spec below this will need to support:
  - Search 
  - A hover/selected/highlighted state for each item - @CynZhang could you give more specific details on the states of these?
- An 'add' callback via a hasIconOnly button on the right of the item. Text should probably be truncated into a tooltip if it doesn't fit

**Additional context**
![image](https://user-images.githubusercontent.com/3360588/91896731-f170bd80-ec5e-11ea-95d3-90d4016f0527.png)

tay1orjones:
This requests highlights the limitations of a non-composable set of components. The List is not composable. 

If the List _was_ composable, we could simply create a new type of ListItem with the visual/functional updates.

davidicus:
@bryancboyd can you comment on this issue what we discussed about this issue. I believe we are de-prioritizing this right now, correct?

sstone2423:
What still needs to be done:
- [ ] add custom card icon
- [ ] add ability for consumer to add their own card types and icons

PR #1596, opened 2020-09-12
feat(cardeditor): initial drop of experimental component

bryancboyd:
**Summary**

This PR provides an initial implementation that covers work for the following issues:

#1560 
#1561 
#1562 
#1563 

We'll provide further improvements and additions to this component over the next 2 weeks, but the component delivered in this PR is functional enough to start being imported and used by applications

- Added new Experimental component `<CardEditor>`

- Jest tests pass
- Click around the storybook `<CardEditor>`interactive story.  The JSON editor allows you to fully modify the JSON definition of the card, and the card edit form will provide inputs to modify card size and title.  If no card definition is passed to CardEditor, the component opens in "gallery mode" where the user can select a card type from a list.

The interactive story also renders the card definition JSON on the left side of the screen, updated in response to the `onChange` event defined on CardEditor.

image


src/components/CardEditor/CardEditForm/_card-edit-form.scss
```diff
@@ -0,0 +1,24 @@
...
+    padding-bottom: $spacing-05;
+  }
+  &--footer {
+    & > button {
+      width: 100%;
+    }
+  }
+  .temp-json-editor {
+    textarea {
+      height: 30rem;
```
sstone2423, line 21: This can be added to the actual TextArea component with the `columns` prop

sstone2423 (changes requested):
Looks good, but is there any kind of state to go back to the existing card that is being editted? If you start in the default story, then click `view gallery`, theres no way of going back to the existing card that is being editted

bryancboyd:
Yeah, in fact I was thinking about removing the showGallery state from the component completely and replacing with a callback (onShowGallery) so that the implementer can manage the "mode" of the editor component (add (gallery) or edit (form)). I'll do that for now.

In proposed "Dashboard Editor" patterns I've seen, the Card Editor would default to the "gallery" state when no cards are selected on the preview dashboard. Then, when either A) a user selects an existing card, or B) a user selects a card type from the gallery, the editor would switch over to the "form" state. With this behavior, selecting "View gallery" would essentially deselect the currently-editing card.

(I wouldn't be surprised if the "view gallery" button gets removed by the time we are finished)

sstone2423 (approved):

filipecorrea (dismissed):
`Table` and `TableViewDropdown` snapshots needs to be updated.

src/components/CardEditor/CardEditor.test.jsx
```diff
@@ -0,0 +1,99 @@
...
+      <CardEditor
+        value={defaultCard}
+        onShowGallery={actions.onShowGallery}
+        onChange={actions.onChange}
+        onAddCard={actions.onAddCard}
+      />
+    );
+    userEvent.click(
+      screen.getByRole('button', {
+        name: CardEditor.defaultProps.i18n.openGalleryButton,
```
tay1orjones, line 72: love this, great idea to use the default props

tay1orjones (approved):
looks great!

PR #1605, opened 2020-09-14
feat(list-item): add disabled prop and functionality

### LINKED ISSUE #1603: [ListItem] add a disabled state for a ListItem
### What package is this for?

- [x] React
- [ ] Angular

### Summary

This is for an enhancement where a `List` component can be populated by 'disabled' `ListItems`.

**Additional context**
image

September release

### Available extra resources

Myself

JoelArmendariz:
Closes #1603

**Summary**

Adds a `disabled` prop on `ListItem` that disables the item.

 - Add `disabled` prop
 - Lower opacity for disabled items

Go to the `with disabled` story in `ListItem` and alter the knobs to verify the styling and functionality.

Screen Shot 2020-09-14 at 3 26 06 PM

### DIFF
--- a/src/components/List/List.jsx
+++ b/src/components/List/List.jsx
@@ -117,6 +117,7 @@ const List = forwardRef((props, ref) => {
       content: { value, secondaryValue, icon, rowActions, tags },
       isSelectable,
       isCategory,
+      disabled,
     } = item;
 
     return [
@@ -132,6 +133,7 @@ const List = forwardRef((props, ref) => {
           nestingLevel={level}
           value={value}
           icon={icon}
+          disabled={disabled}
           iconPosition={iconPosition}
           secondaryValue={secondaryValue}
           rowActions={rowActions}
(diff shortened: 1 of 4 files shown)


sstone2423 (approved):

bjornalm (approved):
Nice work. Since we already had an option to disable the actions only it makes sense to introduce a new prop in order not to break old functionality, but is there ever a use case for disabling the item while having the actions enabled? If not I think we should let this new disabled prop also disable the action.

JoelArmendariz:
> Nice work. Since we already had an option to disable the actions only it makes sense to introduce a new prop in order not to break old functionality, but is there ever a use case for disabling the item while having the actions enabled? If not I think we should let this new disabled prop also disable the action.

@bjornalm Interestingly, the design in monitor that calls for this prop is a use case for keeping the action enabled.
We are using the action to deactivate the `ListItem`, but keeping an action on the item to be able to undo the deactivation.
It seemed best to leave the action to the user, being that the `ListItem` supports a custom component that can set its own state.

image

PR #1609, opened 2020-09-16
feat(table): enable single nested hierarchy styling

### LINKED ISSUE #1600: [Table] parent row selection doesn't select all children nested rows
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

If parent rows are selected, all children nested rows should be selected as well. If any children rows are selected, the parent row should be indeterminate

### To Reproduce

1. Go to https://deploy-preview-1596--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-row-nesting-and-fixed-columns
2. Click on any parent row
3. Expand the children rows
4. See that no children are selected

### Expected behavior

When selecting a parent, all children should be selected as well

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: v2.114.0

### Additional context

See HierarchyList for an example of how this is implemented

upcoming release?

No specific timeline

### LINKED ISSUE #1602: [Table] add hierarchy functionality / styling
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Table currently has nested rows, but doesn't have the idea of hierarchy like the List component. For single-nested child rows, we should not add a gray spacer and we should align the checkboxes. For multi-nested rows, the gray spacer should be present

- [ ] checkboxes should be aligned
- [ ] there should be no gray spacer

**Additional context**
internal issue: https://github.ibm.com/wiotp/monitoring-dashboard/issues/1363

Screenshot:

image

upcoming release?

September release

### Available extra resources

Myself

sstone2423:
Closes #1602
Closes #1600 

**Summary**

- The current Table only supports multiple hierarchy nested styling, but the release of List has given a new use-case where single nested hierarchies can exist, and should have similar styling to List
- This PR adds a new prop `hasSingleNestedHierarchy` as an option that will disable the gray offset if enabled

Related internal Monitor design: https://ibm.invisionapp.com/share/KWO0FZ5BHQ9#/screens/319695909

- added `hasSingleNestedHierarchy` prop
- removed padding when select all checkbox was enabled as it was creating an odd offset where it was not aligned with the first level parents (this also fixed #1600 )
- Updated snapshots, but not sure why the pagination id increased
- Added new single nested hierarchy story

- Go to /?path=/story/watson-iot-table--stateful-example-with-single-nested-hierarchy in the netlify preview
  - ensure that all first level checkboxes are aligned
  - expand one of the rows
    - see that there is no gray offset

**Regression Test (how to make sure I didn't break things)**
- Go to ?path=/story/watson-iot-table--stateful-example-with-row-nesting-and-fixed-columns
  - ensure that the checkboxes are still aligned
  - expand one of the rows
    - see that there IS a gray offset still (as this is not a single hierarchy)

- Go to table resize AND fixed column stories to ensure that the padding still looks correct in the select all checkbox in the table head AND the offset still works correctly with those (if they include nested rows)

Screen Shot 2020-09-16 at 12 40 17 PM

Screen Shot 2020-09-16 at 12 40 13 PM


src/components/Table/Table.jsx
```diff
@@ -52,6 +52,8 @@ const propTypes = {
     hasRowSelection: PropTypes.oneOf(['multi', 'single', false]),
     hasRowExpansion: PropTypes.bool,
     hasRowNesting: PropTypes.bool,
+    /** If the hierarchy only has 1 nested level of children */
+    hasSingleNestedHierarchy: PropTypes.bool,
```
bjornalm, line 56: This is just an idea, but perhaps a name like "hasSingleLevelRowNesting" would be easier to understand. Or maybe it is even better to expand the existing `hasRowNesting` the same way `hasFilter` has been expanded to be either a boolean or some additional configuration data. This extra config data can then be more specific, e.g. {addOffset: false}. With the later option we don't have to introduce a new option prop just for this and if we in the future want to add more rowNesting configurations we are on the right track.

src/components/Table/Table.story.jsx
```diff
@@ -813,13 +813,89 @@ storiesOf('Watson IoT/Table', module)
 
         <br />
 
-        You must also set hasRowExpansion to true in your table options
+        You must also set hasRowExpansion and hasRowNested to true in your table options
```
bjornalm, line 816: hasRowNested -> hasRowNesting

src/components/Table/Table.story.jsx
```diff
@@ -813,13 +813,89 @@ storiesOf('Watson IoT/Table', module)
...
+                }
+              }
+            ]
+          }
+        ]
+        ~~~
+
+        <br />
+
+        You must also set hasRowExpansion and hasRowNested to true in your table options
```
bjornalm, line 891: hasRowNested -> hasRowNesting

bjornalm (approved):
Nice work! I give an approval right away, but please consider the comment regarding the new prop and its naming

src/components/Table/Table.jsx
sstone2423: Great idea. I confuse myself with the syntax around nesting, hierarchy, and expanding as well so this would help. I'll expand upon the `hasRowNesting`

src/components/Table/Table.story.jsx
sstone2423: oops good catch!

sstone2423: Fixed

src/components/Table/Table.jsx
sstone2423: Went with `hasRowNesting: false` OR `hasRowNesting: { hasSingleLevelRowNesting: true }`

JoelArmendariz (approved):
Looks good to me

tay1orjones (approved):

PR #1611, opened 2020-09-17
fix(wizardmodal): progessindicator not changing content

### LINKED ISSUE #1585: [WizardModal] Clicking ProgressIndicator does not change content
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Clicking steps from `ProgressIndicator` inside `WizardModal` won't show the content of the related step, although `ProgressIndicator` shows the clicked step as active. However, when clicking on the "Next" button, the content of the next step will be shown and the `ProgressIndicator` jumps to the related step.

### To Reproduce

1. Open the WizardModal with multiple steps
2. Click on "Next" or "Back". This will show the content of the related step.
3. Click on a different step in the `ProgresssIndicator`.
4. The content will stay the same, although it should have changed.

The bug can be found on the documentation site: https://carbon-addons-iot-react.com/?path=/story/watson-iot-wizardmodal--basic-wizard-modal and under the `WizardModal` Tab in this Stackblitz: https://stackblitz.com/edit/carbon-table-example?file=index.js

### Expected behavior

When clicking on any step of the `ProgressIndicator` inside `WizardModal`, the related content for the given step should be displayed.

### Environment/versions:

- OS: Windows
- Browser: Chrome
- `carbon-addons-iot-react` version: `2.114.8`

### Additional context

I would assume that this is probably caused by the fact that `setStep` is passed to `ProgessIndicator` instead of `onClickItem` [as seen here](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/master/src/components/ProgressIndicator/ProgressIndicator.jsx#L344), and since `ProgressIndicator` only has `onClickItem` as prop [as seen here](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/master/src/components/ProgressIndicator/ProgressIndicator.jsx#L344), the component won't pass the clicked step back to `WizardModal`.

I'd appreciate if you can let me know whether this can be fixed soon. Other than that I can also work with `PageWizard` for now, so it's not too pressuring.

jonguenther:
Closes #1585 

**Summary**

- Fixes not being able to jump between steps when clicking on steps of ProgressIndicator

- Fix WizardModal to use correct prop `onClickItem`

- Clicking on steps of ProgressIndicator should change the content of the modal to the correct step


sstone2423 (commented):
Nice! Could you add a react-testing-library interaction test for this so we can prevent this in the future?

tay1orjones:
Thanks for the PR @jonguenther! 🎉  

Let me know if you'd like me to help out with the unit test. I don't believe I can directly push to this pull request since you initiated it from your fork.

jonguenther:
Sorry for the late reply. I was looking into getting the testing to work, but it took a bit longer than expected.

The test I added checks whether clicking on a progress step in the `WizardModal` will show the content of the next step, and also if progressing is blocked if `onValidate` returns false.

tay1orjones (approved):
No worries - looks great, thanks for adding the test!

marcJV (approved):
👍

PR #1612, opened 2020-09-17
feat(suiteheader): add survey notification

bryancboyd:
**Summary**

- Adds support for displaying a toast notification with a link to a product survey.
- The `useSuiteHeaderData` hook contains logic to determine whether or not a survey should be shown, based on an API `/users/{userId}/surveys/{surveyId}` that returns the initial user interaction time, the time since the last survey was displayed, and some interval values.

image

- Jest tests should pass
- A new storybook story "Header with survey notification" renders and contains a survey link.


jest.config.js
```diff
@@ -1,5 +1,9 @@
 module.exports = {
-  collectCoverageFrom: ['src/components/**/*.js?(x)', '!src/**/*.story.js?(x)'],
+  collectCoverageFrom: [
+    'src/components/**/*.js?(x)',
+    '!src/**/*.story.js?(x)',
+    '!src/**/hooks/*.js',
```
bryancboyd, line 5: Note: The hook (merged in a previous PR) wasn't being included in the per-file coverage report, but was being included in the global count.  This hook is a reference implementation (closer to a story, or AsyncTable, than an actual component) so I recommend we defer coverage until we have a larger strategy for how the PAL will maintain and export data hooks.

src/components/SuiteHeader/SuiteHeader.story.jsx
```diff
@@ -92,6 +92,40 @@ const sideNavLinks = [
   },
 ];
 
+/* Sample of SuiteHeader usage with data hook
```
bryancboyd, line 95: This story failed jest with the "state change needs to be wrapped in act" error.  The code has value as a reference sample, so I left it in (commented out).  I'm happy to remove it if needed.

src/components/SuiteHeader/i18n/index.js
```diff
@@ -22,510 +22,8 @@ const SuiteHeaderI18N = {
...
-    switcherLearnMoreLink: '進一步瞭解',
-    switcherRequestAccess: 'Contact your administrator to request application access',
-    whatsNew: '新增功能',
-    documentation: '說明文件',
-    requestEnhancement: '要求加強功能',
-    about: '關於',
-    support: '支援',
-    gettingStarted: '開始使用',
+    surveyTitle: solutionName => `Enjoying ${solutionName}?`,
+    surveyText: 'Click here to help us improve the product',
```
blechner, line 26: I've just got an update from id and the string is "To help us improve the product, **click here**.", with only the "click here" part being clickable. But I am ok with the way it is now.

blechner (approved):
LGTM

jest.config.js
tay1orjones: Yeah, fair. The related component is under experimental anyway. I've opened https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1615 to track work towards guidance on unit testing hooks we export and reimplementing the testing thresholds.

tay1orjones: Actually just realized the related component isn't under experimental, but we should be fine as this is directly for your team to consume.

tay1orjones (approved):

Issue #1597, opened 2020-09-13
[Table] Select all checkbox is misaligned


sstone2423:
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The select all checkbox is not being centered

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--simple-stateful-example
2. See misaligned select all checkbox

### Expected behavior

Should be centered

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: v2.114.6

### Additional context

![Screen Shot 2020-09-13 at 1 16 03 PM](https://user-images.githubusercontent.com/25177649/93025431-7d6dd800-f5c3-11ea-9558-a0e5429fb142.png)

### Specific timeline issues / requests

Do you want this work within a specific time period? Is it related to an
upcoming release?

September release

marcJV:
This was fixed in most cases, but is still misaligned in 'Stateful Example with expansion, maxPages, and column resize' and a few others like that story
![image](https://user-images.githubusercontent.com/1293444/93629924-b5ec2800-f9ae-11ea-8f27-22d5bc352120.png)

PR #1614, opened 2020-09-18
feat(table): toolbar is not shown when only row actions are present

### LINKED ISSUE #1423: [Table] Row actions cause empty toolbar
### Describe the bug

If RowActions are being used in a Table, the Table renders an empty toolbar, even if there is no toolbar content defined.

### To Reproduce

Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--with-row-expansion-and-actions

Notice the blank toolbar, due to the Table having row actions.

### Expected behavior

There should be no toolbar if there is no toolbar content defined.

Screen Shot 2020-07-22 at 2 07 45 PM

marcJV:
Closes #1423

**Summary**

Row actions no longer render a toolbar, also added if multiselect is present and there is a selection, the toolbar is shown.

- Removes the check for row actions when rendering the toolbar, but adds in check for multiselect and selection

- Verify that Table 'with row expansion and actions' no longer supplies an empty toolbar
- Verify that there are no other row actions that require toolbar support


stuckless:
Is this a regressions 
image

I'm noticing stories with row details are providing a misalignment of the table header with selections

marcJV:
> Is this a regressions
> image
> 
> I'm noticing stories with row details are providing a misalignment of the table header with selections

This is also showing up in master from another PR, this may require some triage though is already part of: #1597. The misalignment may need to be a separate issue.

src/components/Table/Table.story.jsx
```diff
@@ -2247,7 +2247,7 @@ storiesOf('Watson IoT/Table', module)
   .add(
     'with row expansion and actions',
     () => (
-      <Table
+      <StatefulTable
```
tay1orjones, line 2250: Why switch this one from Table to StatefulTable?

tay1orjones (approved):
Looks good pending my one comment - I think we can leave the story as is

davidicus (approved):

PR #1617, opened 2020-09-21
fix(list): adjust hierarchy indentations to match spec

### LINKED ISSUE #1568: [HierarchyList] invalid indentation padding and arrow direction
Currently Items with no children show up without any indent and are not aligned with items at the same level that do have children, since those are indented for the expand/collapse carat. This implies incorrectly a hierarchical relationship between these items.

Expected behavior is that all items  at the same level will be horizontally aligned. If an item does not have child items it should be indented to match the ones that are indented for carats.

Screenshot shown here:  
(In this situation, Home and My Reports are at the same level as the items below. They should be indented.)

image (2)

tay1orjones:
Closes #1568

**Summary**

- HierarchyList indentation is improved for when there's a mix of category and non-category items.

image

- Root level items with no children align with items that have children
- Children items have slightly more indentation than their parent
- Flipped the caret direction to be consistent with sidenav and accordion. When the section is closed, the arrow points down.

- View and validate the new [`HierarchyList \ with mixed hierarchies`](https://deploy-preview-1617--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-experimental-hierarchylist--with-mixed-hierarchies) story
- Check the other HierarchyList stories to ensure no visual regressions

### DIFF
--- a/src/components/CardEditor/__snapshots__/CardEditor.story.storyshot
+++ b/src/components/CardEditor/__snapshots__/CardEditor.story.storyshot
@@ -775,6 +775,14 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT Exper
                   className="iot--list-item"
                 >
                   <div
+                    className="iot--list-item--nesting-offset"
+                    style={
+                      Object {
+                        "width": "30px",
+                      }
+                    }
+                  />
+                  <div
                     className="iot--list-item--content"
                   >
                     <div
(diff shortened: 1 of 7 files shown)


tay1orjones:
@JessicaGentry @jm986 I went ahead and knocked this out today - let me know if this doesn't meet all your needs.

sstone2423 (commented):
Looks good, but beware I recently made this change to single nested hierarchy tables recently as a UX change.

We may need to update tables to also reflect this design change

See this story for an example: https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-single-nested-hierarchy

I didn't think to add a row that had no parent category

sstone2423 (approved):

marcJV (approved):
Looks good to me

PR #1620, opened 2020-09-23
feat(cardeditor): implementation of the card code editor

### LINKED ISSUE #1563: [CardCodeEditor] enable editing of dashboard card config json

davidicus:
Closes #1563

**Summary**

A code editor modal that supports code highlighting and language specific intellisense. 

- Added `monaco-editor/react` npm package.
- Created `CardCodeEditor` component
- Created `_card-code-editor.scss` and added it to `styles.scss`
- Created two new stories for the new component
- Exported `CardCodeEditor` from `index.js`
- Added test for `CardCodeEditor`

- Go to the new CardCodeEditor stories and play around.


stuckless (commented):
@davidicus this looks like a good addition, although not sure why it is in a dialog vs just an editor component.

I was looking at pulling in Ace Editor, but, since you are pulling in this one, I might look at using the same.

Couple questions...
1. If you exposed an "Editor" component that wrapped much of Manaco, then I wouldn't need add a dependency for it explicitly and that way we could be sure we were using the same copy/version instead of potentially having 2 versions pulled in.
2. Does monaco support gutters where you can attach errors to the gutters?  This was one reason I was looking at Ace, although to be fair, I didn't really look at Monaco, since I've used Ace in the past.

How much of an overhead is monaco-editor going to add to the bundle sizes?  I'm wondering if components like this that tend to bring in a large set of other dependencies should be put into a separate component repo?

JordanWSmith15:
@davidicus Agree that it should be a component by itself, rather than embedded as part of a dialog. I have a use case for this as a 80% section of a page. It seems like it also needs to adopt the height of it's parent (it is blowing out vertically on a small dialog)

bryancboyd:
@davidicus I'm not sure this component needs its own export, especially if we are framing it as a "Card" code editor.

Perhaps there is utility in providing a simple Monaco wrapper, as Jordan suggested, and calling it a CodeEditor?  And CardCodeEditor wraps in a modal, provides callbacks for our custom dashboard card json validation, etc?

I suggest we remove the CardCodeEditor package export and only scope this as being used by the CodeEditor component.

bryancboyd:
This perfectly fits the needs of the Dashboard editor though 👏  looking good

davidicus:
@stuckless @JordanWSmith15 
 
> 1. If you exposed an "Editor" component that wrapped much of Manaco, then I wouldn't need add a dependency for it explicitly and that way we could be sure we were using the same copy/version instead of potentially having 2 versions pulled in.

It is in a dialog because that is what the spec I was building for required. That being said, I can definitely expose an editor component on it's own that I consume here. 
> 2. Does monaco support gutters where you can attach errors to the gutters?  This was one reason I was looking at Ace, although to be fair, I didn't really look at Monaco, since I've used Ace in the past.

Monaco supports a lot of functionality and I believe most of the features that are available with VSCode. I have not see the errors in the gutters along the line numbers but Monaco has it on the scroller or minimap.

image


> How much of an overhead is monaco-editor going to add to the bundle sizes? I'm wondering if components like this that tend to bring in a large set of other dependencies should be put into a separate component repo?

The package itself is about 5.5kB minified + gzipped
https://bundlephobia.com/result?p=@monaco-editor/react@3.6.3

What product is the editor requirement for? I will open up another issue to refactor this into its own component. I would like to add the product tags so we have the context in our priority call. 

@bryancboyd I can remove the export

bryancboyd (approved):

tay1orjones (approved):

PR #1622, opened 2020-09-23
feat(suiteheader): Implementation of i18n support and fixing of a redirect issue

blechner:
**Summary**

- Adding another API call to `useSuiteHeaderData` hook to load the translated strings that will be passed to `SuiteHeader` component.
- Fixing an issue where the settings button in `SuiteHeader` would redirect to the login page instead of the last accessed page when `isAdminView` is set to `true`.
- Please don't mind the changes in `SuiteHeader.story.jsx`, the changes are in a story that is commented out because it is using the useSuiteHeaderData hook (and we had decided to comment out this story because of "missing required prop" errors in tests, caused by async data loading, which causes the component to be loaded without some required props in the first render).

- fixed settings button redirect issue
- added `GET /i18n/header/<lang>` API call
- included the `lang` to a `useCallback`'s watch list 
- fixing an "import order" lint error

- No new tests have been added (as it has been decided that data hooks don't need to comply with minimum coverage policy at this moment).

### DIFF
--- a/src/components/SuiteHeader/SuiteHeader.jsx
+++ b/src/components/SuiteHeader/SuiteHeader.jsx
@@ -185,7 +185,7 @@ const SuiteHeader = ({
                       ),
                       onClick: () => {
                         window.location.href = isAdminView
-                          ? document.referrer !== ''
+                          ? document.referrer !== '' && document.referrer.indexOf('auth') < 0
                             ? document.referrer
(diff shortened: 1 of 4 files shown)


bryancboyd (approved):
well done 👍

tay1orjones (approved):

PR #1625, opened 2020-09-24
feat(wizardmodal): add onBack callback

### LINKED ISSUE #1624: [WizardModal] Add onBack callback
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Add a callback prop `onBack` that notifies the parent component when the user goes to a previous step.

**Additional context**
Currently, the wizard modal can only go "forward" by verifying steps. When a user goes back and content needs to be adjusted for that in the parent component, a callback should be available to notify the parent which previous step has been selected.

jonguenther:
Closes #1624 

**Summary**

- Add `onBack` callback to let parent component know if user selects previous step and pass key of step

- added `onBack` prop
- added `changedBack()` to check whether user went back and if `onBack` should be called 

- added test that will check whether callback is called when previous step of `ProgressIndicator` is clicked
- added test that will check whether callback is called when "previous" button is called

**Note:** Has changes that are included in #1611


src/components/WizardModal/WizardModal.jsx
```diff
@@ -80,18 +84,32 @@ class WizardModal extends Component {
   };
 
   handlePrevious = () => {
-    this.setState(state => ({ step: state.step - 1 }));
+    const stepBack = this.state.step - 1;
+    this.changedBack(stepBack);
+    this.setState(() => ({ step: stepBack }));
```
tay1orjones, line 89: Do we need to validate here if we've gone back? `handlePrevious`'s purpose is to decrement the `step`. I think we can just call the callback directly.

Also since we're using existing state to update the state and [there is no guarantee when setState is called](https://reactjs.org/docs/react-component.html#setstate), _I think_ we need to use the callback/updater function still.

```suggestion
    this.setState(
      state => ({ step: state.step - 1 }),
      () => {
        if (this.props.onBack) {
          this.props.onBack(this.state.step);
        }
      }
    );
```

tay1orjones (changes requested):
Looks good! Just one comment below to address

davidicus (approved):
Looks good, once Taylor's comment is addressed.

tay1orjones (approved):

tay1orjones:
Thanks @jonguenther - sorry for the delay!

PR #1630, opened 2020-09-24
fix(table): fixed late loading

### LINKED ISSUE #1191: [StatefulTable] Can't set totalItems when late loading
### To Reproduce

We are using late loading data in StatefulTable, so at first time, the tabledata sent to StatefulTable is just a part of data, not complete data. 
we want to use totalItems field to specify the real total data count, instead of table data count. However i found totalItems  doesn’t work. 
I found this  https://github.com/IBM/carbon-addons-iot-react/pull/372  seems someone has fixed it but actually still not.
here is my version of carbon-addons-iot-react: 2.62.0

### Expected behavior
When using late loading, we expect to set correct totalItems to do pagination.

### Environment/versions:

- OS: [ MacOS,]
- `carbon-addons-iot-react` version: [ `2.62.0`]

upcoming release?
Our product will be released in September, so it will be good if can be fix before that timeline.

marcJV:
Closes #1191 

**Summary**

Fixes the stateful table from ignoring totalItems prop

Added story to show the ability to have a totalItem different than the current dataset size

Stateful Table with Late load data, click the button and observe


src/components/Table/StatefulTable.jsx
```diff
@@ -256,7 +261,8 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
         },
         pagination: {
           ...view.pagination,
-          totalItems: filteredData.length,
+          totalItems:
+            (state.data?.length || 0) === filteredData.length ? totalItems : filteredData.length,
```
bjornalm, line 265: I find this line combined with the totalItems line above really difficult to follow. Is it possible to use some well named constants and restructure this logic to better explain what is going on?

Something along the lines below (I haven't tried it but you get the gist)

```
const filteredCount = filteredData.length;
const paginationCount = state.view.pagination?.totalItems;
const currentlyLoadedDataCount = state.data?.length || 0;

const totalItems = filteredCount && filteredCount !== currentlyLoadedDataCount ? 
  filteredCount  : 
  (paginationCount || currentlyLoadedDataCount);
```

Perhaps this can be simplified further

bjornalm (changes requested):
Nice work. Just a couple of things. Is it possible to create story where the pagination works and still demonstrate the point of this fix?

marcJV:
> Nice work. Just a couple of things. Is it possible to create story where the pagination works and still demonstrate the point of this fix?

This is fixed in the latest, let me know if that is sufficient.

bjornalm (approved):
Nice work

davidicus (approved):

PR #1631, opened 2020-09-24
1181 piechart dashboard card

### LINKED ISSUE #1181: [PieChart] Card shows data using the carbon-charts Pie Chart
### Summary

Create a Dashboard Card type that uses Pie Charts
image
https://carbon-design-system.github.io/carbon-charts/?path=/story/pie--pie

It should have the same abilities as every Dashboard Card, to change the Range, etc

**Additional context**
https://github.ibm.com/wiotp/monitoring-dashboard/issues/221

We're trying to ship this in June in Maximo Asset Monitor

bjornalm:
Closes #1181 

**Summary**
Added a new PieChartCard
image

- Added a new card called PieChartCard that contains a PieChart and a Table
- Using Table instead of StatefulTable like the other chart cards since there is no need for a state for this simple table
- There is no series prop, just a values prop that contains slice objects with both the label and the value since this is what the underlying PieChart api is expecting. The actual value must be called "value" but the label prop can be called anything and be controlled via 'groupDataSourceId'
- Added the properties for the expected customisations but also added the overrides pattern since due to the lack of actual specs
- Added a small addition to the overrides pattern so that a prop now also can be a function that returns an object, instead of just being an object. Added a small helper function in componentUtilityFunctions.js to check this.
- Created a card utility function called increaseSmallCardSize that is shared between the BarChartCard and this new new PieChartCard
- Added unit tests

1. Go to '?path=/story/watson-iot-piechartcard--custom-tooltip' 
2. Verify that the different stories work and make sense


marcJV (changes requested):
image

The thin card sizes clips the categories when it breaks to two or more rows

marcJV:
image

When isLoading is true, should we hide the graph itself?

src/utils/cardUtilityFunctions.js
```diff
@@ -354,3 +354,20 @@ export const chartValueFormatter = (value, size, unit, locale) => {
...
+
+/**
+ * Charts render incorrectly if size is too small, so change their size to MEDIUM
+ * @param {string} size card size
+ */
+export const increaseSmallCardSize = (size, cardName) => {
+  const warnMsg = `${cardName} does not support card size ${size}`;
+  let returnSize;
+  if (size === CARD_SIZES.SMALL) {
+    console.warn(warnMsg); // eslint-disable-line no-console
```
marcJV, line 366: Should we have a test expecting a warning in the console when small or small-wide is selected?

marcJV (changes requested):

src/utils/cardUtilityFunctions.js
tay1orjones: The warning should at the very least be limited to development environments via the `warning` package/fuction that is already installed as a dep.

bjornalm:
> image
> 
> When isLoading is true, should we hide the graph itself? When both fields are present, the graph draws over the loading animation. Should the legend be available when loading?

Can you show me the exact configuration that produces this, I can't reproduce it. What browser are you using?

davidicus:
>@davidicus, it seems that this is fixed with a patch upgrade to 0.36.4 of @carbon/charts & carbon/charts-react, I assume that is ok?

@bjornalm that should be fine

src/constants/CardPropTypes.js
```diff
@@ -295,6 +296,47 @@ export const BarChartCardPropTypes = {
...
+    groupDataSourceId: PropTypes.string,
+    /**
+     * Function to format the labels. Input to the function is a wrapped data object containing additional
+     * chart label info such as x & y positions etc */
+    labelsFormatter: PropTypes.func,
+    /** The position of the legend in relation to the chart, can be 'bottom' or 'top'. */
+    legendPosition: PropTypes.string,
+  }),
+  /** Used to overide the internal components and props for advanced customisation */
+  overrides: PropTypes.shape({
```
sstone2423, line 317: Is this the new standard for our components? Seems like it opens up the gates to too much freedom

sstone2423 (approved):
Looks good!

src/constants/CardPropTypes.js
bjornalm: Probably not a new standard for every component, but perhaps an escape hatch to be used/provided if needed. For the PieChartCard we did not have much of a spec so I thought it would make sense to use this pattern here. We have already used it for a couple of components.

marcJV (approved):
Great job!

davidicus (approved):

PR #1632, opened 2020-09-24
chore: initial monorepo setup

cal-smith:
So, this generated quite a bit of diff to go through, most of which is just files moving around.

Changes:
- added lerna, `lerna.json`, and yarn workspaces
- moved all the `carbon-addons-iot-react` sources into `packages/react`
- netlify should now deploy from `packages/react/storybook-static` ... We'll want to check what carbon/carbon-charts are doing for monorepo storybook deploys
- added some scripts in the root to handle building/linting/testing everything in the repo ... most package specific scripts can stay isolated to each package.

We may want to make use of github actions to give us a bit more control over the builds - travis only gives us one job per event (PR, push, etc), while actions will let us run multiple jobs (i.e. one per project).

Issues:
- the build doesn't work at the moment. I'm going to try and chase down _where_, but it seems to be rollup related
- deleting the `yarn.lock` and re-generating causes all sorts of strange issues. It may be worth pinning the versions of everything in `package.json`

TODO:
- [x] fix the build
- [x] add versioning. We'll have to figure out [which mode we want to use ](https://github.com/lerna/lerna#how-it-works), but I imagine "fixed" will be the easiest to manage (at the cost of potentially excessive versioning if several packages are updating ... a non issue with a staged/time-based release schedule).
- [x] update developer/contributor guidance
- [x] automate as much as possible, document the rest


tay1orjones (commented):
🎉  Thanks for starting this! 🙏 Lets work together on the build issues and versioning. 

Deleting/regen yarn.lock has caused issues in the past with storybook. Is a delete/regen necessary?

Would be awesome if we could use independent mode to decouple the workstreams and avoid the "empty bump" versions where some packages don't have any changes but are incremented because a sibling package has an update. I need to read up on lerna to see if replaces all of our current semantic release workflow, or just part. Determining major|minor|patch version bump from commit message format, adding/publishing tags, adding releases to GitHub, automatically compiling release notes into the GitHub releases, etc. I assume we can probably automate all this with Travis. 

A couple additional items on my mind:
- We should update developer guidance to instruct how to run the project, etc with the new setup. Even if it's just "`cd` into folder x, then run the command `yarn`" etc
- If the release/versioning process requires any manual steps besides a merge of pull request, we should document it so that we can empower a small group of core maintainers to own the process. Just want to avoid you or I being the only ones able to do releases. Ideally it'll all be automated though like it is now and we don't have to think about it.

package.json
```diff
@@ -1,107 +1,46 @@
 {
-  "name": "carbon-addons-iot-react",
+  "name": "ai-apps-pal",
```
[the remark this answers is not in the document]
cal-smith: Gotcha 👍 just needed something to place hold

cal-smith:
> 🎉 Thanks for starting this! 🙏 Lets work together on the build issues and versioning.
> 
> Deleting/regen yarn.lock has caused issues in the past with storybook. Is a delete/regen necessary?

Not necessarily, it's just an easy trap to fall into, definitely not something we need to tackle as part of this though.

> Would be awesome if we could use independent mode to decouple the workstreams and avoid the "empty bump" versions where some packages don't have any changes but are incremented because a sibling package has an update. I need to read up on lerna to see if replaces all of our current semantic release workflow, or just part. Determining major|minor|patch version bump from commit message format, adding/publishing tags, adding releases to GitHub, automatically compiling release notes into the GitHub releases, etc. I assume we can probably automate all this with Travis.

In my experience lerna completely replaces semantic release. I'll have to check carbon/carbon-charts to see how they handle the "empty bump" situation ...

Given that we'll naturally end up with a few projects in the repo, I would be inclined to follow Carbons lead and run weekly/monthly releases ... We don't need to be quite as particular (patches weekly, minors monthly), but it should help avoid a situation where we're bumping versions _too_ frequently for consumers to track - We've mostly managed to avoid that with `carbon-components-angular` but some weeks it definitely gets a bit silly.

> A couple additional items on my mind:
> 
>     * We should update developer guidance to instruct how to run the project, etc with the new setup. Even if it's just "`cd` into folder x, then run the command `yarn`" etc
> 
>     * If the release/versioning process requires any manual steps besides a merge of pull request, we should document it so that we can empower a small group of core maintainers to own the process. Just want to avoid you or I being the only ones able to do releases. Ideally it'll all be automated though like it is now and we don't have to think about it.

- Absolutely! I'll add that to the TODOs
- Definitly ... Strong preference towards fully automating the process 👍

cal-smith:
Build fixed! Turns out it was a fairly simple change in the rollup config to handle symlinks better - `'node_modules/**` -> `/node_modules/`

package.json
```diff
@@ -23,7 +23,8 @@
     "lint": "lerna run --stream lint",
     "test": "lerna run --stream test",
     "start": "lerna run --stream start",
-    "clean": "lerna run --stream clean"
+    "clean": "lerna run --stream clean",
+    "publish-npm": "lerna publish --conventional-commits --create-release github --yes"
```
cal-smith, line 27: This is equivalent to running `semantic-release`. `--conventional-commits` generates the changelog, and uses the conventional spec to determine which bump to apply (the usual `feat`/`fix`/`BREAKING CHANGE` stuff). `--create-release github` publishes a GitHub release with the changelog. `--yes` just skips the confirmation prompts.

Under the hood, lerna uses [`lerna changed`](https://github.com/lerna/lerna/tree/master/commands/changed#readme) to figure out which packages to publish versions for, it then uses [`lerna version`](https://github.com/lerna/lerna/tree/master/commands/version#readme) to figure out which version to publish. My understanding is that packages that don't update for a given release will stay on an older version, while packages that have had an update will all jump to the same version.

Conceptually this makes sense to me, since we'll want to be able to point consumers to a unified version of each package.

Options and alternatives:

We can use [GitHub actions `schedule` feature](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#schedule) to preform weekly/daily/whatever releases as a means of adding predictability to the version bump schedule.

Another option would be adding a `release` branch that we merge into when we're ready to make a release.

We could also take the approach Carbon has with `lerna publish from-package` and simply merge a PR with updated `package.json` versions whenever a release is needed. The automation would take over from that point and generate the appropriate changelogs and releases. There are some risks with this approach (additional manual work) but it would give each package better control over it's version.

We can also combine a few of these approaches - Weekly automated conventional-commits releases with intermediate from-package releases as needed.

davidicus (commented):
This looks pretty straight forward so far. Thanks for getting this in here. I saw the 669 files changed and died a bit inside but I see they are mostly file changes!

packages/react/package.json
```diff
@@ -0,0 +1,277 @@
...
+    "rollup-plugin-filesize": "^6.0.0",
+    "rollup-plugin-json": "^4.0.0",
+    "rollup-plugin-node-builtins": "^2.1.2",
+    "rollup-plugin-node-globals": "^1.4.0",
+    "rollup-plugin-node-resolve": "^4.0.0",
+    "rollup-plugin-postcss": "^2.0.3",
+    "rollup-plugin-replace": "^2.1.0",
+    "rollup-plugin-uglify": "^6.0.4",
+    "sass-loader": "^7.1.0",
+    "semantic-release": "^16.0.0-beta.39",
```
tay1orjones, line 261:
```suggestion
```

package.json
tay1orjones: Awesome, that makes sense and is very helpful. I guess the question then just becomes, "when and how do we run `npm-publish`"?

> Another option would be adding a release branch that we merge into when we're ready to make a release.

I think this makes the most sense to me. So far we've followed the _"what's in `master` is what's in production (`latest` version)"_ mantra by relying on Travis CI to run the build/publish step when commits are merged into master. This also democratizes releases by allowing contributors to merge their own (approved) PRs and the CI contain the the publish process and github/npm access tokens as secure env variables.

What if the release branch was `master`? With semantic-release we were going to achieve this and follow a time based release schedule by redirecting all new work to a new `next` branch, which would publish on every PR to the `next` channel. On timed intervals (2 weeks, 1 month, whatever), we would PR `next` into `master`, which would trigger a new release on `latest`. Caveat there being that while the `next -> master` PR is open, no new work can go into `next` - all PRs would be temporarily halted until it's merged. Consumers would have the option of following the `next` releases (the same experience today) or follow the time based `latest`. [Semantic release's docs do a better job of explaining](https://github.com/semantic-release/semantic-release/blob/311c46580f57fbe809735d61fea498789d32a61e/docs/recipes/distribution-channels.md#publishing-on-distribution-channels) 😅 

Additionally, if we needed to patch an old version we could open a new `2.100.x` branch that would publish commits on that branch only as patches, `2.100.1`, `2.100.2`, etc

By no means is this how I think we should do it moving forward, more of just an assessment of what @davidicus and I have been discussing for a few months and how we've relied on semantic-release. This boils down to a few open questions:

* How do we support rapid iteration with product teams via (ideally automated) `next` and `latest` distribution channels?
* How would we do maintenance releases to patch old, non-current minor versions without running the publish manually? (rare problem, but could happen)
* Could we augment the publish steps to include posting a comment to PRs with what `next` and `latest` version it was published in? (this has been super helpful in tracking down the source of bugs)
* Would it be better or easier to contain all this automation in Travis, Circle CI (standardize on what Carbon's using) or GH Actions?

tay1orjones: I've opened PR https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1694 to begin to address some of this. I've added some base workflows for CI/test actions and one to test doing our publish step from an action. These are so much faster than Travis CI.

I've answered a lot of my own questions:

> - How do we support rapid iteration with product teams via (ideally automated) next and latest distribution channels?

Actions! It's clearer to me now how we can tie specific actions to specific branches and run scripts (semantic-release, lerna, etc) with params to set up automated publishing for `next`/`latest`.

> - How would we do maintenance releases to patch old, non-current minor versions without running the publish manually? (rare problem, but could happen)

Set up an action for a branch that is created from the tag we want to start from. Create 2.100.x branch at tag 2.100.0 - new PRs to this branch should only increment patch. Have the action auto generate these new tags as appropriate and publish to npm. 

Actions can also be ran automatically on push/pull request/etc, but also can be triggered manually. Carbon did this recently with their storybook deploy - https://github.com/carbon-design-system/carbon/pull/7061

> - Could we augment the publish steps to include posting a comment to PRs with what next and latest version it was published in? (this has been super helpful in tracking down the source of bugs)

Yes! We can upload artifacts from different github action jobs and download them into new jobs. The publish command could output the version to be released, and on successful release post a comment to the issue with the issue number via [actions/github-script](https://github.com/actions/github-script#comment-on-an-issue)

> - Would it be better or easier to contain all this automation in Travis, Circle CI (standardize on what Carbon's using) or GH Actions?

I think GH Actions is a great place to start.

cal-smith:
I'm going to have to fiddle with this some more to get the `@next` and `@latest` behaviour correct 👍

package.json
cal-smith: Oh wow somehow I missed all this! Totally agree with all the above though!

cal-smith:
👀 I'll take a look at what's breaking the actions

cal-smith:
_Should_ be good to go on the release logic ... I expect we'll need to spend some time debugging the versioning logic when we merge (it's a big change from semantic release, and there's a few more moving parts), but otherwise, I think this is pretty well set up.

cal-smith:
I think that, barring issues that will come up post-merge (I'm sure there will be some ... either packages not publishing, or versions not working quite right), this is ready to go!

cal-smith:
Hopefully we can get this in before I have to resolve too many more merge conflicts 😂

.travis.yml
```diff
@@ -25,5 +25,4 @@ install:
 script:
   - yarn build
 
-after_success:
-  - yarn publish-npm
+after_success: bash ./scripts/relese.sh
```
tay1orjones, line 28:
```suggestion
after_success: bash ./scripts/release.sh
```

scripts/release.sh
```diff
@@ -0,0 +1,27 @@
...
+  exit 1;
+fi
+
+if [[ $TRAVIS_BRANCH == "master" ]]; then
+  # graduate the relase with --conventional-graduate
+  lerna version --conventional-commits --conventional-graduate --create-release github --yes
+  # publish the packages that were just versioned
+  lerna publish from-package --dist-tag next --yes
+  # push changes back to next so we can start work on the next release
+  git push origin next
```
tay1orjones, line 16: Will this have sufficient permissions?

cal-smith: It'll be using the same git credentials travis has when it clones the repo to build it, so it _should_. Looking at Carbon Charts, they explicitly add a token - https://github.com/carbon-design-system/carbon-charts/blob/master/scripts/deploy.sh#L9-L11 - But I'm not 100% sure if that's necessary.

theiliad: I think I remember having to specifically add those lines bc travis wouldn't publish

cal-smith: Cool, I'll add something similar

tay1orjones: We have both `next` and `master` configured as protected branches, so I don't think this will work. We just turned off travis in favor of a github action. It can be swapped out to run this script instead, but [actions are limited in the same way](https://github.community/t/allowing-github-actions-bot-to-push-to-protected-branch/16536/2).

We can do this manually for now and/or find a way to do it via the rest API as mentioned in that thread.

cal-smith: Interesting 🤔 may be best if we do it manually for now and investigate our options after the switch

tay1orjones (approved):

davidicus (approved):
Looks good to merge after we resolve the conflicts. Let's hope this is the last time

cal-smith:
conflicts resolved locally, just running tests now ...

cal-smith:
Done! Good to go!

theiliad:
> Done! Good to go!

🎉

theiliad (approved):
obv don't rely on my approval but looks good

PR #1636, opened 2020-09-27
Some fixes and enhancements to `SuiteHeader` component

blechner:
**Summary**

- Survey link opens in a new window
- A custom logout function can be passed to ` SuiteHeader`
- Adding an extra filter to only include applications that have sync state `SUCCESS` in `SuiteHeaderAppSwitcher`

- Logout test is now `async` to account for the custom function that is executed during logout process.


src/components/SuiteHeader/SuiteHeader.jsx
```diff
@@ -106,6 +109,7 @@ const SuiteHeader = ({
   applications,
   sideNavProps,
   surveyLink,
+  onRequestLogout,
```
stuckless, line 112: I feel that the header itself should not be managing "links" but rather just call onLogout passing the current link that it knows about.  People can then override the onLogout function to do as they please.  This is true for changing applications as well.  ie, the navigation should call a function that can be overridden.  In graphite both logout and change application, show help, etc, all go through application-level APIs rather than hitting URL directly.  ie, there really should be a single standard uniform way in which the application header will leave an application for whatever reason.

stuckless (commented):
@bryancboyd I know you've most of the work with the suite header.  Next week, I'd like to set up some time for you and I to work through some of the alignment goals of making this header consumable to other teams.

src/components/SuiteHeader/SuiteHeader.jsx
bryancboyd: Sounds good @stuckless -- let's work on an interface that is nice for graphite to consume.

One possibility (to simplify the interface) would be to pass the header a single function prop `onRouteChange` that would be called (if defined) on every header link click event, with the type of action (logout, help, switch application, etc) and the URL. If the function returns true, the link redirect occurs in the header-- if false, it does not. This would give applications full control on how to handle (or fail) the route change.

blechner: @bryancboyd @stuckless The `onRouteChange` function intercepts all route changes now. It is a `Promise<boolean>` that, if resolved with `true`, will allow the redirect to happen. If it resolves with `false`, it won't redirect. `onRouteChange` will be called with `(routeType, url, params)`. `routeType` is a string that can identify, for example, if the redirect is to a logout page, or to a documentation page. `url` is the actual URL to redirect to. The 3rd argument is only included when `routeType` is `APPLICATION`, so `params` will be `{appId: someapplicationid}`, for example

src/components/SuiteHeader/SuiteHeader.test.jsx
```diff
@@ -72,27 +73,80 @@ describe('SuiteHeader', () => {
...
     delete window.location;
     window.location = { href: '' };
     render(<SuiteHeader {...commonProps} />);
-    userEvent.click(screen.getAllByRole('button', { name: 'Log out' })[0]);
+    await userEvent.click(screen.getAllByRole('button', { name: 'Log out' })[0]);
     expect(window.location.href).toBe(commonProps.routes.logout);
   });
-  it('admin link from admin view takes you to navigator route', () => {
+  it('clicks logout link (but no redirect)', async () => {
+    delete window.location;
```
davidicus, line 84: Do we want to store the original value for window.location and restore after test?

blechner: @davidicus I've added `beforeEach()` and `afterEach()` to all tests that delete `window.location` so that it is restored after each test.

bryancboyd:
The latest changes look good 👍

bryancboyd (approved):

tay1orjones (approved):

PR #1641, opened 2020-09-28
feat(valuecard): added ability to override custom formatting

### LINKED ISSUE #844: [Value Card] Need a way to provide custom formatting of the value
### Describe the bug

The value card today just renders a number and doesn't allow for custom formatting of that number.

eg, You wan to show a Value Card that is Money?

$12K USD

It would be helpful if in the rendering of the value, IF there is a custom render/formatting function then that is used to do the formatting of the number instead of the default behaviour.  Perhap the custom formatter could get passed a reference to the default formatter so that the customer formatter can call it, and then add in any additional formatting.

marcJV:
Closes #844 

**Summary**

Value card lacked a custom formatting tool, this PR adds in a function that takes in the original value, and the pending formatted value. This function expects a return with however the developer wants the field formatted.

- features: added a customFormatter hook.

- Unit testing tests this. There is also a story at: `?path=/story/watson-iot-valuecard--small-with-custom-formatter`


src/constants/CardPropTypes.js
```diff
@@ -74,6 +74,7 @@ export const ValueCardPropTypes = {
   cardVariables: PropTypes.objectOf(
     PropTypes.oneOfType([PropTypes.string, PropTypes.func, PropTypes.number, PropTypes.bool])
   ),
+  customFormatter: PropTypes.func,
```
bjornalm, line 77: Perhaps we should add a comment that explains this function and the parameters

bjornalm (approved):
Great work!

src/components/WizardModal/WizardModal.test.jsx
```diff
@@ -156,32 +156,36 @@ describe('WizardModal', () => {
...
-      screen.getByTestId('iot--progress-step-button-main-step3').click();
-      
-      expect(screen.getByTitle('step2').children[0]).toEqual(beforeClick2);
-      expect(screen.getByText('page 2')).toBeDefined();
-
-      // clicking on page 1 should go back to step 1
-      screen.getByTestId('iot--progress-step-button-main-step1').click();
-      expect(screen.getByTitle('step1').children[0]).toEqual(beforeClick1);
-      expect(screen.getByText('page 1')).toBeDefined();
+    render(
```
bjornalm, line 159: I also got this but excluded it from my PR.  It looks like auto-formatting from Prettier and perhaps we should just commit this and get rid of it. It is a bit strange that is is formatted since we don't touch the file...

tay1orjones: Yes, please leave it in. This was introduced in https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1611

Status checks must've been bypassed locally by the author. I'll look into adding a prettier validation to our Travis builds to prevent in the future https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1659

src/constants/CardPropTypes.js
marcJV: added a comment for this

tay1orjones (approved):

PR #1645, opened 2020-10-01
feat(table): added option alwaysTruncate

### LINKED ISSUE #1196: [table] add option to force table cell texts to truncate
# Summary
The wrapCellText prop can currently have the values:
• ‘auto’ we apply wrapping only for tables with normal columns. For tables with fixed or resizable columns we use truncation.
• ‘always’ we wrap in all scenarios including tables with fixed or resizable columns (i.e. current situation).
• ‘never’ we never wrap and as a consequence the tables with normal columns* get wide cells and more often horizontal scrolling. For tables with fixed or resizable columns we use truncation.

Feedback and questions shows to some confusion where app developers want to force text truncation for non resizable/fixed width tables. So we should add the option `alwaysTruncate` to support this.

**Additional context**
The reason that we only truncate for tables with fixed or resizable columns, and this is important but difficult to communicate, is that if we would add truncation for tables with dynamic widths the column widths would be based on the table headers rather than the actual content of the tables. Hence for “normal” tables we would go from a smart column width distribution like this

image

To a truncated but not very optimized layout like this
image

But at least the app developer would have the choice to accept such column width distribution in order to get cell text truncation.

bjornalm:
After discussion we will close this PR and use #1651 instead. The code refactoring will go in a separate issue for the next major release where we also remove the export of TableHead and TableBody.

**Summary**
The wrapCellText prop could have the values:
• ‘auto’ we apply wrapping only for tables with normal columns. For tables with fixed or resizable columns we use truncation.
• ‘always’ we wrap in all scenarios including tables with fixed or resizable columns (i.e. current situation).
• ‘never’ we never wrap and as a consequence the tables with normal columns* get wide cells and more often horizontal scrolling. For tables with fixed or resizable columns we use truncation.

Feedback and questions shows to some confusion where app developers want to force text truncation for non resizable/fixed width tables. So we added the option `alwaysTruncate` to support this.

See #1196 for more context

- added the option 'alwaysTruncate' to the table prop `options.wrapCellText`.
- refactored the unit tests to use React Testing Library and made them more robust
- added additional unit tests
- merged the internal props wrapText && truncateCellText into a new prop called cellTextOverflow since they now were overlapping. Removed unused values for the new cellTextOverflow prop.
- improved and refactored the memoized function that calculates the value of the `cellTextOverflow`. Hopefully this code is easier to understand and more aligned with the unit tests.

Nota bene! You might have to decrease the width of your browser window in order to trigger truncation of the table cell texts.

1. In general, verify that the table stories that have a `wrapCellText` knob will use truncation if the new 'alwaysTruncate' option us selected.

2. Especially, verify that stories that don't use `hasResize` and don't have fixed column widths still will truncate when the new 'alwaysTruncate' option us selected.
E.g. /story/watson-iot-table--stateful-example-with-multiselect-filtering

3. If you are ambitious (which you of course are), verify that the old stories work the same way for the old `wrapCellText` knob alternatives.

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -77,6 +77,12 @@ const propTypes = {
     hasUserViewManagement: PropTypes.bool,
     /** If true removes the "table-layout: fixed" for resizable tables  */
     useAutoTableLayoutForResize: PropTypes.bool,
+    /**
+     * auto - Wrap for tables with dynamic columns widths and truncate for tables with fixed or resizable columns
+     * always - Wrap if needed for all table column configurations
+     * never - Tables with dynamic columns widths grow larger and tables with fixed or resizable columns truncate.
+     * alwaysTruncate - Alwyas truncate if needed for all table column configurations
+     */
     wrapCellText: PropTypes.oneOf(['always', 'never', 'auto', 'alwaysTruncate']),
   }),
 
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -77,7 +77,7 @@ const propTypes = {
     hasUserViewManagement: PropTypes.bool,
     /** If true removes the "table-layout: fixed" for resizable tables  */
     useAutoTableLayoutForResize: PropTypes.bool,
-    wrapCellText: PropTypes.oneOf(['always', 'never', 'auto']),
+    wrapCellText: PropTypes.oneOf(['always', 'never', 'auto', 'alwaysTruncate']),
   }),
 
   /** Initial state of the table, should be updated via a local state wrapper component implementation or via a central store/redux see StatefulTable component for an example */
@@ -369,13 +369,26 @@ const Table = props => {
   const [tableId] = useState(() => uniqueId('table-'));
   const [, forceUpdateCellTextWidth] = useState(0);
 
-  const useCellTextTruncate = useMemo(
-    () =>
-      options
-        ? options.wrapCellText !== 'always' &&
-          ((options.hasResize && !options.useAutoTableLayoutForResize) ||
-            columns.some(col => col.hasOwnProperty('width')))
-        : undefined,
+  const cellTextOverflow = useMemo(
+    () => {
+      const fixedTableLayout = !options?.useAutoTableLayoutForResize;
+      const hasInitalColumnWidths = columns.some(col => col.hasOwnProperty('width'));
+      const hasCalculatedColumnWidths = options.hasResize && fixedTableLayout;
+      const dynamicCellWidths = !hasInitalColumnWidths && !hasCalculatedColumnWidths;
+      const TEXT = { TRUNCATE: 'truncate', PREVENT_WRAP: 'prevent-wrap', WRAP: undefined };
+
+      switch (options?.wrapCellText) {
+        case 'alwaysTruncate':
+          return TEXT.TRUNCATE;
+        case 'never':
+          return dynamicCellWidths ? TEXT.PREVENT_WRAP : TEXT.TRUNCATE;
+        case 'auto':
+          return dynamicCellWidths ? TEXT.WRAP : TEXT.TRUNCATE;
+        case 'always':
+        default:
+          return TEXT.WRAP;
+      }
+    },
     [options, columns]
   );
 
@@ -533,8 +546,7 @@ const Table = props => {
                 'useAutoTableLayoutForResize',
                 'hasSingleRowEdit'
               ),
-              wrapCellText: options.wrapCellText,
-              truncateCellText: useCellTextTruncate,
+              cellTextOverflow,
             }}
             columns={columns}
             filters={view.filters}
@@ -608,8 +620,7 @@ const Table = props => {
                 'shouldExpandOnRowClick',
                 'shouldLazyRender'
               )}
-              wrapCellText={options.wrapCellText}
-              truncateCellText={useCellTextTruncate}
+              cellTextOverflow={cellTextOverflow}
               ordering={view.table.ordering}
               rowEditMode={rowEditMode}
               actions={pick(
--- a/src/components/Table/Table.story.jsx
+++ b/src/components/Table/Table.story.jsx
@@ -55,7 +55,7 @@ const STATUS = {
   BROKEN: 'BROKEN',
 };
 
-const selectTextWrapping = ['always', 'never', 'auto'];
+const selectTextWrapping = ['always', 'never', 'auto', 'alwaysTruncate'];
 
 const renderStatusIcon = ({ value: status }) => {
   switch (status) {
@@ -700,7 +700,7 @@ storiesOf('Watson IoT/Table', module)
             ...initialState.options,
             hasResize: true,
             hasFilter: select('hasFilter', ['onKeyPress', 'onEnterAndBlur'], 'onKeyPress'),
-            wrapCellText: select('wrapCellText', selectTextWrapping, 'always'),
+            wrapCellText: select('wrapCellText', selectTextWrapping, 'alwaysTruncate'),
             hasSingleRowEdit: true,
           }}
         />
--- a/src/components/Table/Table.test.jsx
+++ b/src/components/Table/Table.test.jsx
@@ -12,8 +12,6 @@ import { mockActions } from './Table.test.helpers';
 import Table, { defaultProps } from './Table';
 import TableToolbar from './TableToolbar/TableToolbar';
 import EmptyTable from './EmptyTable/EmptyTable';
-import TableBodyRow from './TableBody/TableBodyRow/TableBodyRow';
-import TableHead from './TableHead/TableHead';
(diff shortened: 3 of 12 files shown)


src/components/Table/Table.jsx
```diff
@@ -77,7 +77,7 @@ const propTypes = {
     hasUserViewManagement: PropTypes.bool,
     /** If true removes the "table-layout: fixed" for resizable tables  */
     useAutoTableLayoutForResize: PropTypes.bool,
-    wrapCellText: PropTypes.oneOf(['always', 'never', 'auto']),
+    wrapCellText: PropTypes.oneOf(['always', 'never', 'auto', 'alwaysTruncate']),
```
stuckless, line 80: I think it would help if the comment for this prop described what the intention of each prop means.

src/components/Table/Table.story.jsx
```diff
@@ -700,7 +700,7 @@ storiesOf('Watson IoT/Table', module)
             ...initialState.options,
             hasResize: true,
             hasFilter: select('hasFilter', ['onKeyPress', 'onEnterAndBlur'], 'onKeyPress'),
-            wrapCellText: select('wrapCellText', selectTextWrapping, 'always'),
+            wrapCellText: select('wrapCellText', selectTextWrapping, 'alwaysTruncate'),
```
stuckless, line 703: wouldn't we want all options in this list so that we can see how always, never, auto and alwaysTruncate behaves?

src/components/Table/Table.test.jsx
```diff
@@ -464,79 +462,219 @@ describe('Table', () => {
     expect(wrapper.find('.bx--search-input').html()).toContain(`aria-hidden="true"`);
   });
 
-  it('cells should always wrap by default', () => {
```
stuckless, line 467: It's scary to me that tests are being removed.  Why not just keep the existing tests in place to catch regressions, and then add in the new tests?

src/components/Table/Table.test.jsx
```diff
@@ -567,54 +705,6 @@ describe('Table', () => {
     unmount();
   });
 
-  it('cells should wrap (not truncate) for wrapCellText:auto + resize + table-layout:auto', () => {
```
stuckless, line 570: Why not just leave in all the previous tests, especially since they are verifying using an alternate methodology.

src/components/Table/TableBody/TableBody.jsx
```diff
@@ -51,8 +52,7 @@ const propTypes = {
     }),
   ]),
   hasRowActions: PropTypes.bool,
-  wrapCellText: PropTypes.oneOf(['always', 'never', 'auto']).isRequired,
-  truncateCellText: PropTypes.bool.isRequired,
```
stuckless, line 55: shouldn't these just be deprecated?

src/components/Table/TableBody/TableBody.jsx
```diff
@@ -130,8 +131,7 @@ const TableBody = ({
   shouldExpandOnRowClick,
   shouldLazyRender,
   ordering,
-  wrapCellText,
-  truncateCellText,
```
stuckless, line 134: so we are removing 2 props and replacing with a single prop, isn't that a breaking change?

src/components/Table/TableCellRenderer/TableCellRenderer.test.jsx
```diff
@@ -34,104 +38,73 @@ describe('TableCellRenderer', () => {
...
-      <TableCellRenderer wrapText="never" truncateCellText={false}>
-        {cellText}
-      </TableCellRenderer>
-    );
-    expect(wrapper2.find(`.${iotPrefix}--table__cell-text--truncate`)).toHaveLength(0);
+    rerender(<TableCellRenderer>{cellText}</TableCellRenderer>);
+    expect(screen.getByText(cellText)).not.toHaveClass(textTruncateClassName);
   });
 
-  it('does not truncat when wrapText={always}', () => {
```
stuckless, line 53: Again, very scared with the need to remove tests :(

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -1711,21 +1721,6 @@ Map {
             "hasSingleRowEdit": Object {
               "type": "bool",
             },
-            "truncateCellText": Object {
```
stuckless, line 1714: So with the removal of these props, is this forcing a major version change?

stuckless (changes requested):
Overall I think the feature concept is good (and I think it's really what people expect when they used 'always' to begin with.  I'm a little scared of the downstream impact of these changes.

src/components/Table/Table.story.jsx
bjornalm: Yes indeed, and you have all options defined in the `selectTextWrapping` variable. The `alwaysTruncate` is just the selected default for this story.

src/components/Table/Table.test.jsx
bjornalm: We are moving our tests away from enzyme and this test still exists but under another a slightly different name and refactored for the React Testing Library:

https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1645/files/073c843872db029503f1157850b83b5c63acbb3b#diff-f47894e32d4dbf684d2ac3f89df148c9R495-R500

src/components/Table/Table.test.jsx
bjornalm: I wanted to make the tests more resilient and not rely on checking for props, especially for the internal components. Since the props wrapText && truncateCellText would be overlapping and confusing after this PR I wanted to merge them and reduce the table code complexity, which meant that the tests relying on props and not HTML outcome would have to be refactored.

stuckless: I get the logic, but, the existing unit tests are in place to protect against regressions.  If you having to remove tests and completely rewrite them as a result of a change, then that tells me you are hiding a regression.  Leaving the existing tests in place and then create a separate story LATER to remove what you considered redundant tests makes more sense, since, those existing tests will help you catch regressions, but they can't help if you create new tests that you think are 1-1 tests of the original, and erase the ones.  I'm a little sensitive here because I've seen a number of cases in table/dashboard where tests were erased as part of a change and then I'm left dealing with finding those when I'm taking an update.

src/components/Table/TableBody/TableBody.jsx
bjornalm: Thanks for bringing this to my attention. I was under the impression that TableBody and TableHead were internal components (like the TableBodyRow and the TableCellRenderer), but are they in fact exported. The TableBody and TableHead were not actually using these props, only passing them down to the children, that no longer will use them.

Let me discuss with David how to best handle this. They should indeed be deprecated and perhaps I can add some kind of deprecation mapper to the new prop for the unlikely case where someone is directly using these props without the Table component. That way we don't have to introduce this as a breaking change.

src/components/Table/TableBody/TableBody.jsx
bjornalm: See deprecation comment above

src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
bjornalm: See deprecation comment above

src/components/Table/Table.test.jsx
bjornalm: So in this particular case it is basically a trade off between a) introducing more complexity to the table when there is the option to reduce it and b) accepting the risk and rewriting the existing suboptimal tests to allow us the possibility to refactor the table and improve our future test situation.

stuckless: But you are not accepting any risk, are you creating it :)  You are asking me to accept your risk, and given that I'm now closing out 3 weeks effort to move to the latest Carbon/IoT and I'll be pushing into 4 weeks of that effort before it is done... I'm not sure I like it when you guys create "acceptable" risk :(

src/components/Table/TableBody/TableBody.jsx
stuckless: If there are not exported, then I'm ok with that... and to be clear, I'm not using these components directly, anyways, so, I'm ok if there are considered internal and you remove them.

src/components/Table/Table.test.jsx
bjornalm: Language confusion, I meant "accepting the risk with rewriting the existing... ". I believe we all share the risk since our goal is always delivering changes without introducing regressions. But to ease your worry I have created a new PR that does not contain the code refactoring, hence keeping the old unit tests. I will close this PR and we will add the refactoring in a separate Issue for a major release where we also remove the export of TableHead and TableBody.

New PR is available here
https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1651

PR #1651, opened 2020-10-05
feat(table): added option alwaysTruncate (no refactor)

### LINKED ISSUE #1196: [table] add option to force table cell texts to truncate

bjornalm:
Closes #1196

**Summary**
The wrapCellText prop could have the values:
• ‘auto’ we apply wrapping only for tables with normal columns. For tables with fixed or resizable columns we use truncation.
• ‘always’ we wrap in all scenarios including tables with fixed or resizable columns (i.e. current situation).
• ‘never’ we never wrap and as a consequence the tables with normal columns* get wide cells and more often horizontal scrolling. For tables with fixed or resizable columns we use truncation.

Feedback and questions shows to some confusion where app developers want to force text truncation for non resizable/fixed width tables. So we added the option `alwaysTruncate` to support this.

See #1196 for more context

- Added option alwaysTruncate
- Added more unit tests

Nota bene! You might have to decrease the width of your browser window in order to trigger truncation of the table cell texts.

1. In general, verify that the table stories that have a wrapCellText knob will use truncation if the new 'alwaysTruncate' option us selected.

2. Especially, verify that stories that don't use hasResize and don't have fixed column widths still will truncate when the new 'alwaysTruncate' option us selected.
E.g. /story/watson-iot-table--stateful-example-with-multiselect-filtering


stuckless (approved):

src/components/Table/Table.jsx
```diff
@@ -77,7 +77,13 @@ const propTypes = {
     hasUserViewManagement: PropTypes.bool,
     /** If true removes the "table-layout: fixed" for resizable tables  */
     useAutoTableLayoutForResize: PropTypes.bool,
-    wrapCellText: PropTypes.oneOf(['always', 'never', 'auto']),
+    /**
+     * auto - Wrap for tables with dynamic columns widths and truncate for tables with fixed or resizable columns
+     * always - Wrap if needed for all table column configurations
+     * never - Tables with dynamic columns widths grow larger and tables with fixed or resizable columns truncate.
+     * alwaysTruncate - Alwyas truncate if needed for all table column configurations
```
tay1orjones, line 84:
```suggestion
     * alwaysTruncate - Always truncate if needed for all table column configurations
```

tay1orjones (changes requested):
One minor spelling correction 😄

tay1orjones (approved):

PR #1662, opened 2020-10-06
chore(prettier): update prettier, related deps, add check to CI

### LINKED ISSUE #1659: [repo] update to prettier v2, add formatting validation check to CI
### What package is this for?

- [x] React
- [ ] Angular

### Summary

It's currently possible to bypass status checks (precommit hook) locally (not running prettier), open a PR, status checks pass, and merge in code with invalid formatting.

Then, in subsequent PR's that have run the precommit hook, unrelated files will be formatted causing an odd diff. We saw this recently introduced in https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1611 and noticed in the diff of https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1641#discussion_r499563618

- [x] Add prettier diff to Travis CI status checks to fail build if there is invalid formatting
- [x] Update dep to [Prettier v2](https://prettier.io/blog/2020/03/21/2.0.0.html)
- [x] Add Carbon's prettier config to standardize on the same intra-project formatting conventions
- [x] Format the codebase (will be a large diff 😬 )

tay1orjones:
Closes #1659 

**Summary**

- Update the formatting deps and settings for the repository
- Reformat codebase

- Add prettier diff to Travis CI status checks to fail build if there is invalid formatting
- Update dep to [Prettier v2](https://prettier.io/blog/2020/03/21/2.0.0.html)
- Add Carbon's prettier config to standardize on the same intra-project formatting conventions
- Format the codebase (will be a large diff 😬 , though you'll see I have broken the changes into three primary initial commits for easier review/understanding):
  1. https://github.com/carbon-design-system/carbon-addons-iot-react/commit/495e54ee39751d1e9fd4cc7c4f927aea2b6efb0b - updating of deps, config, and CI checks
  1. https://github.com/carbon-design-system/carbon-addons-iot-react/commit/1644c8407ea05e7cdcb7900c91f847ba663a3904 - reformat codebase (this is the large unwieldy diff)
  1. https://github.com/carbon-design-system/carbon-addons-iot-react/commit/be9259e8075a97ff1558370a8fa4a35b824ccab7 - regen snapshots
  1. ... additional commits will follow due to updating with latest from master and reformatting/fixing lint errors.

- Status checks should pass
- Take a look over the stories of the modified component stories. No logic has been directly modified, so there should be no functional change here - this is simply a reformat.


config/stylelint/sassMsgFormatter.js
```diff
@@ -27,12 +27,12 @@ function filterForErrors(result) {
...
   let filteredArray = [];
   let isUnique = true;
-  results.forEach(result => {
+  results.forEach((result) => {
     if (filteredArray.length == 0) {
       // if the filtered array is empty then it's automatically unique
       filteredArray.push(result);
     } else {
-      filteredArray.forEach(uniqueResult => {
+      filteredArray.forEach((uniqueResult) => {
```
davidicus, line 35: Do we want the added parens in our code base? Not opposed I suppose.

jest.config.js
```diff
@@ -78,17 +86,21 @@ module.exports = {
     './src/components/Table/AsyncTable/AsyncTable.jsx': {
       statements: 56,
       branches: 50,
-      lines: 60,
+      lines: 57,
```
davidicus, line 89: is there a reason this coverage exception went down?

davidicus (approved):
LGTM. Just a couple minor comments

jest.config.js
tay1orjones: It's a misnomer due to the formatting. Some code that was previously on one line is now spread across multiple lines and it's catching that the per-line coverage is now lower. There's no actual change in what we're testing.

config/stylelint/sassMsgFormatter.js
tay1orjones: [The blog announcement for Prettier v2](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev) has a good explanation that rings true for me:

> ## Change default value for arrowParens to always
> Since version 1.9, Prettier has had an option to always wrap arrow function arguments with parentheses. In version 2.0, this behavior has become the default.
> ```
> // Input
> const fn = (x) => y => x + y;
> 
> // Prettier 1.19
> const fn = x => y => x + y;
> 
> // Prettier 2.0
> const fn = (x) => (y) => x + y;
> ```
> At first glance, avoiding parentheses in the isolated example above may look like a better choice because it results in less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments, default values, or a variety of other things. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the change.

bjornalm (approved):
LGTM

Issue #124, opened 2019-04-01
[Table]: selection box support indeterminate state based on the selection of their children state


scottdickerson:
Pulled from comments:
> This issue clarifies the general selection state across parent and child row.
> The IoT Table component should follow the core behavior of the Carbon table component, and secondarily any common hierarchical table component row selection industry practices.
> 
> In summary
> 
> * Selection states are none (empty), all (filled), partial (dash). There are selection icons for each.
> * Selecting the parent will select all children
> * Unselecting the parent will unselect all children
> * Changing the selection of a child will change the state of the parent to partial
> * Selecting a partial parent will make the parent and children selected
> 
> Optional
> 
> * A selection state change may be revoked with a ctrl-z

matsgothe:
This issue clarifies the general selection state across parent and child row.
The IoT Table component should follow the core behavior of the Carbon table component, and secondarily any  common hierarchical table component row selection industry practices.

In summary
- Selection states are none (empty), all (filled), partial (dash). There are selection icons for each.
- Selecting the parent will select all children
- Unselecting the parent will unselect all children
- Changing the selection of a child will change the state of the parent to partial
- Selecting a partial parent will make the parent and children selected

Optional
- A selection state change may be revoked with a ctrl-z

matsgothe:
@scottdickerson please review

davidicus:
This issue is stale. Closing. Please re-open if needed.

matsgothe:
This is a base table feature. If we do row selection in a hierarchy we need this parent state visualization. Its the way hierarchies works in any tool. 
@amymillett

bjornalm:
> @bjornalm In #1671 you mention that you would open an issue to track this work. If there is a duplicate issue can we close one of them?

That comment was about deprecating `view.table.isSelectAllSelected` and `view.table.isSelectAllIndeterminate` props and the issue created was this one: https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1673

I think the original PR https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1671 solved the problems described here, but please verify that (or I can take a deeper look if you want).

bjornalm:
Correction #1671 was only concerned with the table header checkbox, not the parent child relationships of each row. This is a new enhancement.

PR #1671, opened 2020-10-08
fix(table): auto select the select all checkbox

### LINKED ISSUE #1464: [Table] When all the rows are selected through prop, the 'AllSelected' checkbox isn't checked
### Describe the bug
When all the rows of the table are selected through view.table.selectedIds prop, the 'AllSelected' checkbox isn't automatically checked 

Here's a table where I pass the view.table.selectedIds prop as an array that matches the length the entire table data, but the top checkbox shows unselected

image

### To Reproduce

Simply instantiate a table with props where all the selectedIds that are selectable are selected

```
<CommonTable
        {...this.props}
        lightweight
        columns={tableColumns}
        data={tableData}
        options={{
          hasRowSelection: 'multi',
        }}
        view={{
          table: { selectedIds: selectedDimensions },
        }}
        actions={{
          table: { onRowSelected: onSelect, onSelectAll },
        }}
      />
```

bjornalm:
Closes #1464

**Summary**
This fix automatically checks the checkbox if all rows are selected. If some rows are selected, the checkbox gets the Indeterminate state. 

If the `view.table.isSelectAllSelected` is present as a prop then that value will be used instead and if `view.table.isSelectAllIndeterminate` is present that value will be used. See the unit tests for the exact functionality.

Like I commented in the issue itself:
I question if this really is a bug to be fixed? Since the app actually controls the view.table.isSelectAllSelected like the suggested workaround shows I'm not sure it makes sense to handle this value automatically. With this fix implemented, what is the use case of the view.table.isSelectAllSelected (and possibly view.table.isSelectAllIndeterminate) if those values are automatically calculated? So I think we should go with either

a) Don't add this fix
b) Add this fix and if we don't have a use case for overriding the automatically calculated props view.table.isSelectAllSelected and view.table.isSelectAllIndeterminate then we should probably deprecate them.
c) **CURRENT PR -** We combine view.table.isSelectAllSelected and view.table.isSelectAllIndeterminate with a default automatic behaviour like this issue describes.

1. Removed the default value "false" for the view.table.isSelectAllSelected 
2. Added automatic calculation for isSelectAllSelected and isSelectAllIndeterminate
3. Added unit tests

1. Go to story ?path=/story/watson-iot-table--with-multi-select-and-batch-actions
2. Select "Table" from the Type of Table knob
3. Add `row-0,row-1,row-2,row-3,row-4,row-5,row-6,row-7,row-8,row-9` to the selectedIds knob
4. Make sure the "Select all is selected"
5. Play with the combination of `isSelectAllSelected` and `isSelectAllIndeterminate` knobs to make sure they override the automatic behaviour

### DIFF
--- a/src/components/Table/Table.jsx
+++ b/src/components/Table/Table.jsx
@@ -241,7 +241,7 @@ export const defaultProps = baseProps => ({
     },
     table: {
       expandedIds: [],
-      isSelectAllSelected: false,
+      isSelectAllSelected: undefined,
       selectedIds: [],
       rowActions: [],
       sort: {},
@@ -432,6 +432,18 @@ const Table = props => {
   const rowEditMode = view.toolbar.activeBar === 'rowEdit';
   const singleRowEditMode = !!view.table.rowActions.find(action => action.isEditMode);
 
+  const allRowsAreSelected = view.table.selectedIds.length === visibleData.length;
+  const someRowsAreSelected = view.table.selectedIds.length > 0 && !allRowsAreSelected;
+
+  const noSelectAllProp = view.table.isSelectAllSelected === undefined;
+  const isSelectAllSelected = noSelectAllProp ? allRowsAreSelected : view.table.isSelectAllSelected;
+
+  const noIndeterminateProp = view.table.isSelectAllIndeterminate === undefined;
+  const isSelectAllIndeterminate =
+    noIndeterminateProp && noSelectAllProp
+      ? someRowsAreSelected
+      : view.table.isSelectAllIndeterminate;
+
   return (
     <TableContainer
       style={style}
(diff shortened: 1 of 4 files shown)


sstone2423:
This is an interesting problem. I was recently implementing a NestedTable and was really surprised that addons doesn't provide the selection logic. I think its beneficial to allow the user to implement their own selection logic, but I'd like if addons had a default. As for a regular, non-nested table, I can't think of a reason why a consumer would want to implement their own selectAll logic so I'd vote to deprecate it. It would also help bring down the very high number of prop options we have

tay1orjones:
Something to keep in mind with this is that we have a longstanding request https://github.com/carbon-design-system/carbon-addons-iot-react/issues/124 for rows with child/nested rows to support indeterminate state on the parent when a child row is selected.

It doesn't look like this really blocks that future functionality in any way, but will be an addition to this logic at some point (unless you want to look at it now in conjunction with this)

sstone2423 (approved):

tay1orjones (approved):

bjornalm:
> Something to keep in mind with this is that we have a longstanding request #124 for rows with child/nested rows to support indeterminate state on the parent when a child row is selected.
> 
> It doesn't look like this really blocks that future functionality in any way, but will be an modification to this logic at some point (unless you want to look at it now in conjunction with this)

I'll rather keep them separated since I'm already "complicating" this PR with different scenarios on how to handle the fix

bjornalm:
I will create a new issue to deprecate the props `isSelectAllSelected` and `isSelectAllIndeterminate`

PR #1674, opened 2020-10-08
feat(combobox): allow listening to onBlur event to trigger custom logic

### LINKED ISSUE #1672: [ComboBox]: if I blur the combobox with a new entry, it doesn't create it
The combobox forces me to push 'Enter' while in it to actually create a new item.  Instead if I leave the combobox by clicking away from it, I'd like it to create the new item automatically onBlur

scottdickerson:
Closes #1672

**Summary**

- Adds onBlur event and provides an example story that adds an item to a list onBlur

- This allows us to get a UX when we want to add the items

- Verify that the combobox doesn't lose its value when you leave it with a new value in the stories: https://deploy-preview-1674--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-experimental-combobox--default


tay1orjones (changes requested):
Should this only be enabled if `addToList` prop is true?
Should handleKeyPress be a prop so you can override it instead?

Blank tags are created in the hasMultiValue story
dont work

scottdickerson:
@tay1orjones I made the change to only act this way if addToList is true

tay1orjones (approved):
@scottdickerson I think you could copy the `adds to list (not tags) when addToList is passed and no hasMultiValue` unit test and make a small modification to click away from the input and we'd have this covered.

I believe status checks are failing from outdated snaps - the added onBlur handler should show up in related snaps.

tay1orjones (changes requested):
After a review with @dianatran18 @davidicus, we think this probably shouldn't be the default behavior. Some additional concerns included when you click (or tab) away from the input and add the current text to the list there's no way to then manage the list after the fact (remove items, etc). Overall makes it much easier to 

Could your ux be achieved instead through `onBlur` and `onKeyDown` being made to be configurable props so you can override this in your application for the specific use cases you've identified?

scottdickerson:
@tay1orjones

sstone2423 (approved):

tay1orjones (approved):

tay1orjones:
Looks great, thanks! I just pulled down and ran `yarn format` to fix the CI - this should be good to go 👍

PR #1676, opened 2020-10-09
fix(table): fixed table stories showing error

### LINKED ISSUE #1666: [Table] table stories showing  error
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The console shows error:
`Warning: React does not recognize the 'hasRowNesting' prop on a DOM element`

### To Reproduce

1. In your local dev go to http://localhost:3000/?path=/story/watson-iot-table--stateful-example-with-expansion-maxpages-and-column-resize
2. See the error in the console log

I think this is due to that we are sending the `hasRowNesting` as a prop to the `StyledTableExpandRowExpanded` which passes it along to the underlying `TableExpandRow`. 

Instead we should probably pass the prop `hasSingleNestedHierarchy` to the `StyledTableExpandRowExpanded` and use spread in the parameters to exclude it from being passed down.

### Expected behavior
No error thrown and shown in the console

### Environment/versions:
- OS: [e.g. MacOS]
- Browser: [e.g. chrome]

### Additional context
image

bjornalm:
Closes #1666

**Summary**
Removes hasRowNesting={false} from the table tr 

- used the props['data-row-nesting'] for StyledComponents instead

1. Open the browser web tool console 
2. Go to https://deploy-preview-1676--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-table--stateful-example-with-expansion-maxpages-and-column-resize
3. Verify that there is no error message 
4. Expand a row
5. Verify that there is no error message

### DIFF
--- a/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
+++ b/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
@@ -181,7 +181,9 @@ const StyledTableExpandRow = styled(({ hasRowSelection, ...props }) => (
       // if single nested hierarchy AND there are children rows (meaning this is a parent),
       // bolden all cells of this row
       props =>
-        props.hasRowNesting?.hasSingleNestedHierarchy && props['data-child-count'] > 0
+        props['data-row-nesting'] &&
+        props['data-row-nesting'].hasSingleNestedHierarchy &&
+        props['data-child-count'] > 0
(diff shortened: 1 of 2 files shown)


tay1orjones (approved):

sstone2423 (commented):
Thanks for this! I missed it

sstone2423 (approved):

PR #1679, opened 2020-10-09
Some minor fixes to the SuiteHeader component

blechner:
- Fixing a value in `ROUTE_TYPES` (copy/paste error)
- Adding a `useEffect` to capture changes in `surveyData` after the component is mounted.
- Adding a `z-index` to the toast notification to make sure the links are always clickable.


tay1orjones (approved):

src/components/SuiteHeader/_suite-header.scss
```diff
@@ -59,6 +59,7 @@
   right: 0;
   max-width: rem(425px);
   width: auto;
+  z-index: 1;
```
tay1orjones, line 62: There's a z-index function that we can use to keep values consistent, it takes variables as params - [here's the source/list](
https://github.com/carbon-design-system/carbon/blob/73d94961ba828afc72fbb82a853d304caffbd023/packages/components/src/globals/scss/_layout.scss#L14-L23)


```suggestion
  z-index: z('overlay');
```

tay1orjones (approved):
One minor comment but overall LGTM

cgirani (approved):

src/components/SuiteHeader/_suite-header.scss
blechner: Thanks for pointing this out @tay1orjones 
I didn't know about this `z-index` function 👍

PR #1682, opened 2020-10-13
fix(dashboardgrid): specify container padding for all breakpoints

### LINKED ISSUE #1680: [DashboardGrid] no containerPadding value for `max` breakpoint
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

DashboardGrid's container padding changes at the `max` breakpoint:
bounce

It appears that [the default configuration from the `react-grid-layout`](https://github.com/STRML/react-grid-layout/blob/0.18.3/lib/ResponsiveReactGridLayout.jsx#L145) does not contain a default `containerPadding` value for the largest breakpoint, `max`.

This results in the container padding going from `[0 ,0]` to `null` when the viewport hits the `max` breakpoint. When containerPadding is null, it defaults back to the margin default which we set to 16px
https://github.com/carbon-design-system/carbon-addons-iot-react/blob/b3a91067d02da3a7211856e6823c7c55c6438bbd/src/components/Dashboard/DashboardGrid.jsx#L147

To fix, we need to provide a valid [0,0] `containerPadding` for all of our breakpoints.

### To Reproduce

View [the story](https://carbon-addons-iot-react.com/?path=/story/watson-iot-dashboard-grid--dashboard-all-card-sizes), resize your browser. In the react dev tools you can see the `containerPadding` prop changing between `[0,0]` and `null`

tay1orjones:
Closes #1680 

**Summary**

- Invalid margin is no longer applied to the grid container at any breakpoint

- View [the default DashboardGrid story](https://deploy-preview-1682--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-dashboard-grid--dashboard-default-layouts), resize and see that margin is always 0 for the container.

bouncy


bjornalm (approved):
Nice work

sstone2423 (approved):

PR #1704, opened 2020-10-19
Next

### LINKED ISSUE #1385: [dep] update semantic release
### Describe the bug

Semantic release is out of beta. Two deps that semantic release rely on have security vulnerabilities. I'd rather update semantic release than update the deps under the hood, so I've closed #1378, #1379 in favor of this issue to update semantic release itself.

[Semantic release v16 changelog](https://github.com/semantic-release/semantic-release/releases/tag/v16.0.0) has a list of breaking changes, we should validate if they'll impact us. We can probably update to --latest, v17.x - the only breaking change there was requiring a slightly higher version of Node 10, but we're using Node 12.

davidicus:
Closes #1385 

**Summary**

Update our semantic release version to latest (17.2.1) and create our `next` branch. This will allow us to now publish after each pr to this `next` branch and move our main channel to timed releases. 

- Update semantic release
- change travis.yml to run semantic release on `master` and `next` branch

- Merge into master
- merge PRs into next and see it publish.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,15 @@
     ]
   },
   "release": {
-    "branches": "['+([0-9])?(.{+([0-9]),x}).x', 'master', {name: 'next', channel: 'next', prerelease: 'next'}, 'next-major', {name: 'beta', prerelease: true}"
+    "branches": [
+      "+([0-9])?(.{+([0-9]),x}).x",
+      "master",
+      {
+        "name": "next",
+        "channel": "next",
+        "prerelease": "next"
+      }
+    ]
   },
   "dependencies": {
     "@babel/runtime": "^7.10.2",
--- a/package.json
+++ b/package.json
@@ -260,7 +260,7 @@
     "rollup-plugin-replace": "^2.1.0",
     "rollup-plugin-uglify": "^6.0.4",
     "sass-loader": "^7.1.0",
-    "semantic-release": "^16.0.0-beta.39",
+    "semantic-release": "^17.2.1",
     "storybook-addon-rtl": "^0.2.2",
     "storybook-readme": "^5.0.8",
     "style-loader": "^0.23.1",
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     ]
   },
   "release": {
-    "branches": "['+([0-9])?(.{+([0-9]),x}).x', 'master', {name: 'next', channel: 'next', prerelease: 'next'}, 'next-major', {name: 'beta', prerelease: true}, {name: 'alpha', prerelease: true}]"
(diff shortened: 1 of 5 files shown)


tay1orjones (approved):

tay1orjones (changes requested):
Something I hadn't thought about is the storybook. As we add things to `next`, do we need a live production-like environment?

Additionally I did some tests on a [sample repo](https://github.com/carbon-design-system/repo-transfer-test) and confirmed that this isn't going to slow down our version number changes. `next` will publish versions as we do today, one per PR. When `next` is merged into master, `latest` is updated with the most recent version on `next`.

Rudimentary git representation here:
```
2.136.0 - latest
  2.137.0 - next
  2.137.1 - next
  2.137.2 - next
  2.138.0 - next
  --- merge next into master ---
2.138.0 - latest
```

A consumer following `latest` would jump from 2.136.0 to 2.138.0

Alternatively, if we follow a prerelease naming structure, 2.137.0-next.0, 2.137.0-next.1, etc, consumers who want to follow all pre-releases on `next` would need to put `"carbon-addons-iot-react": "next"` in their `package.json`. Otherwise normal semver ranges (caret, etc) are [automatically limited](https://docs.npmjs.com/misc/semver#prerelease-tags) to the immediate major.minor.patch version. 

I think this makes the most sense - I can add a config to support this and some documentation as well.

package.json
```diff
@@ -112,6 +112,9 @@
       "@commitlint/config-conventional"
     ]
   },
+  "release": {
+    "branches": "['+([0-9])?(.{+([0-9]),x}).x', 'master', {name: 'next', channel: 'next', prerelease: 'next'}, 'next-major', {name: 'beta', prerelease: true}, {name: 'alpha', prerelease: true}]"
```
davidicus, line 116: @tay1orjones does this `prerelease` for next need to be `true` instead of `next`?

tay1orjones: `prerelease: true` will automatically use the `name` in the version name, but I configured it with the expanded syntax so it's more clear as to what we've configured.

As an example:
```
{name: 'next', channel: 'next', prerelease: 'true'} would result in the format:
2.136.0-next.1

{name: 'next', channel: 'next', prerelease: 'next'} results in the format:
2.136.0-next.1

{name: 'next', channel: 'next', prerelease: 'dragons'} would result in the format:
2.136.0-dragons.1

{name: 'dragons', channel: 'next', prerelease: 'true'} would result in the format:
2.136.0-dragons.1
```

tay1orjones (approved):
This is ready - I've confirmed the new version syntax is publishing and have made a netlify subdomain for next.carbon-addons-iot-react.com

blechner (approved):

PR #1705, opened 2020-10-19
Feat/icon dropdown

### LINKED ISSUE #1259: [Dropdown] with icon selection buttons
### Summary

This contribution enables users to quickly and easily pick between options that are visually recognisable rather than necessarily recognisable by name. Specifically when picking between chart types.

**Additional context**
Design context from the internal issue [pal-tracking#174](https://github.ibm.com/watson-iot/pal-tracking/issues/174):

> https://ibm.invisionapp.com/share/U2NYW8D7DFV#/screens/319194959
> 
> - This dropdown presents itself mostly as a regular dropdown before interaction, with the notable exception of an icon in the value along with text:
> 
> - Once opened, this dropdown supplies a space for custom content. The desired example would be selection icon buttons #153 
> - In this case, the selected button will appear both as an icon and text value in the dropdown, but also in the 'selected state' in the dropdowns suggestion box.
> 
> - When a new value is selected, the icon and text in the main field updates.
> 
> 
> This is used in the EDS 'analysis' UX pattern (no consumers presently): https://ibm.invisionapp.com/share/U2NYW8D7DFV#/screens/319195042
> Image: 
> image

### LINKED ISSUE #1373: [Dropdown] with icons (with and without label)
### Summary

From 
[Internal Watson-IoT/pal-tracking#175](https://github.ibm.com/Watson-IoT/pal-tracking/issues/175)

> This is an EDS item.
> 
> In certain cases, text labels in dropdowns are not able to satisfactorily help users identify what is being referenced in order to make an informed selection, especially when scanning rapidly.
> 
> This contribution aims to support the text values of options in a dropdown with iconography.
> 
> 
> * This follows all the traditional stylings and behaviours of a traditional carbon dropdown component: https://www.carbondesignsystem.com/components/dropdown/code
> * With the notable exception of a 20px black fill icon left aligned inside the field, and next to each option in the suggestion box.
> 
> Found: https://ibm.invisionapp.com/share/U2NYW8D7DFV#/screens/319194959
> 
> Image:
> image
> 
> image

### LINKED ISSUE #1374: [Dropdown] with icons (with and without label) - Inline version
### Summary

From  [Internal Watson-IoT/pal-tracking#176](https://github.ibm.com/Watson-IoT/pal-tracking/issues/176)

> 
> The same as the rationale for the standard variant #1373 but with lower visual weight for use inline in table etc.
> 
> 
> * The same as the carbon inline version of the dropdown component: https://www.carbondesignsystem.com/components/dropdown/code
> * With the exception of a 20px black fill icon left aligned in both the field and the suggested values.
> 
> Found here: https://ibm.invisionapp.com/share/U2NYW8D7DFV#/screens/319194959
> Image:
> image

marcJV:
Closes #1373
Closes #1374 
Closes #1259

**Summary**

- Added icon and icon with label variant for dropdown

- Added Dropdown, Dropdown.story.jsx, Dropdown.test.jsx

- Two stories were created at
  - https://deploy-preview-1705--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-dropdown--default
  - https://deploy-preview-1705--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-dropdown--default-with-icons


tay1orjones (changes requested):
I think it would improve maintainability overall if the `Dropdown` wasn't overridden/replaced and this was a new `IconDropdown` component. Isolating this feature into it's own allows us to upgrade Carbon without having to do integration testing to determine if our features break the base dropdown. Additionally we won't have to copy and maintain/update the stories across new versions.

Both the `.scss` file and the component `.jsx` need to be exported.

It [sounds like](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1373#issuecomment-712893518) this is still in progress. I'm marking it as a draft, feel free to mark it back as ready-to-review if I'm mistaken! 🙂

src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,98 @@
...
+const Dropdown = ({
+  selectedViewId,
+  items,
+  overrides,
+  hasIconsOnly,
+  itemToString,
+  actions: { onChangeView, ...otherActions },
+  ...other
+}) => {
+  const MyDropDown = overrides?.dropdown?.component || CarbonDropDown;
```
davidicus, line 33: Do we need an override here?

src/components/IconDropdown/_icon-dropdown.scss
```diff
@@ -0,0 +1,164 @@
...
+    outline-style: none;
+    border: none;
+    transition: initial;
+    box-shadow: none;
+
+    > .#{$prefix}--list-box__menu-item--active {
+      background: transparent;
+
+      > .#{$prefix}--list-box__menu-item__option {
+        > .#{$iot-prefix}--dropdown__image-button.bx--btn.bx--btn--icon-only.bx--tooltip__trigger:hover::before {
```
tay1orjones, line 36: The specificity of these selectors and =>3 levels deep nesting concerns me. The more specific, the more likely they are to break with future versions of Carbon.

Do the uses of the child combinator `>` in this file require that or could we use the less specific descendant combinator? Could we apply classes to the button's tooltip to prevent the long selector chain?

Completely understand if not - sometimes these things are somewhat unavoidable. Any improvements we can make would be solid!

src/components/IconDropdown/index.js
```diff
@@ -0,0 +1 @@
+export { Dropdown, DropdownSkeleton } from 'carbon-components-react';
```
tay1orjones, line 1: Why move these here? I think this file should be removed. The Dropdown and DropdownSkeleton should already exported in the main index.js. Other components can import it from there (`../../index.js or whatever), or directly from carbon-components-react to prevent circular references.

tay1orjones (changes requested):

src/components/Dropdown/DropdownWithIcon.jsx
```diff
@@ -0,0 +1,65 @@
...
+  selectedViewId,
+  items,
+  itemToString,
+  actions: { onChangeView, ...otherActions },
+  ...other
+}) => {
+  const onSelectionChange = (changes) => {
+    const { selectedItem } = changes;
+
+    onChangeView(selectedItem);
```
bjornalm, line 35: Is this perhaps a leftover from the TableViewDropdown?

src/components/Dropdown/Dropdown.story.jsx
```diff
@@ -1 +1,113 @@
...
+  invalid: boolean('Show form validation UI (invalid)', false),
+  invalidText: text(
+    'Form validation UI content (invalidText)',
+    'A valid value is required'
+  ),
+});
+
+storiesOf('Watson IoT/Dropdown', module).add('with icons and labels', () => {
+  return React.createElement(() => {
+    const [selectedViewId, setSelectedViewId] = useState(items[1].id);
```
bjornalm, line 92: selectedOption instead of selectedView?

src/components/Dropdown/Dropdown.story.jsx
```diff
@@ -1 +1,113 @@
...
+      <div
+        style={{
+          width: select('wrapper width', ['300px', '100px'], '300px'),
+        }}>
+        <DropdownWithIcon
+          {...props()}
+          items={items}
+          selectedViewId={selectedViewId}
+          actions={{
+            onChangeView: (viewItem) => {
```
bjornalm, line 104: This action in not defined in the propTypes for DropdownWithIcon

src/components/Dropdown/DropdownWithIcon.jsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  const renderLabel = (item) => {
+    return (
+      <>
```
bjornalm, line 40: Do you need the fragment when you are wrapping it all in a div anyway?

src/components/IconDropdown/_icon-dropdown.scss
marcJV: This breaking has been a worry for me as well, I will take another dive through this tonight, but I was trying to be as specific as possible to not screw up the other dropdowns in the codebase. Will comment here on my findings

src/components/Dropdown/DropdownWithIcon.jsx
```diff
@@ -0,0 +1,65 @@
...
+    })
+  ).isRequired,
+};
+
+const defaultPropTypes = {
+  itemToString: null,
+};
+
+const DropdownWithIcon = ({
+  selectedViewId,
```
bjornalm, line 26: You are not using this one. It should probably be renamed and then passed down to the Dropdown as the prop `selectedItem`

src/components/Dropdown/DropdownWithIcon.jsx
```diff
@@ -0,0 +1,65 @@
...
+      items={items}
+      actions={otherActions}
+      onChange={onSelectionChange}
+      {...other}
+      itemToString={itemToString !== null ? itemToString : renderLabel}
+    />
+  );
+};
+
+DropdownWithIcon.propTypes = Dropdown.propTypes && propTypes;
```
bjornalm, line 62: I don't understand this. You are just assigning `propTypes` but only if `Dropdown.propTypes` is truthy....

src/components/Dropdown/DropdownWithIcon.jsx
```diff
@@ -0,0 +1,65 @@
...
+DropdownWithIcon.defaultProps = Dropdown.defaultProps && defaultPropTypes;
```
bjornalm, line 63: same here

bjornalm (changes requested):
Partial review, will continue later

src/components/Dropdown/DropdownWithIcon.jsx
marcJV: I need to pull the selected item outside of the changes from here; i moved it down further since I no longer need it in that exact spot.

src/components/Dropdown/Dropdown.story.jsx
marcJV: renamed to selectedItem to keep in line with all the other item names

src/components/Dropdown/DropdownWithIcon.jsx
marcJV: I may need some direction for this, I want to merge the proptypes, but Dropdown's proptypes are not exported, should I copy them over to our index file from carbon and then append the Dropdown as dropdownProps and pass them along to the dropdown directly? Currently changing to an array of the props, it passes the validation as I want, but syntactically it looks wrong.

src/components/Dropdown/Dropdown.story.jsx
marcJV: see the other comment for this. This is part of Dropdown's proptypes that I wish to use

src/components/Dropdown/DropdownWithIcon.jsx
marcJV: Relooked into this, the second approach while passes validation, fails pr snapshots. The above passes validation for propTypes and passes snapshot tests, but I agree it is likely still incorrect. Any help on this would be appreciated.

bjornalm: Since what you actually want is merging the two propTypes we need to either have the Dropdown's proptypes exported or copied. Perhaps @davidicus can weigh in

src/components/Dropdown/Dropdown.story.jsx
bjornalm: Are you sure `onChangeView ` is part of the Dropdown's proptypes? Isn't it called `onChange` there?

src/components/Dropdown/DropdownWithIcon.jsx
```diff
@@ -0,0 +1,140 @@
...
+  direction: 'bottom',
+  ariaLabel: '',
+  invalid: false,
+  invalidText: '',
+  onChange: () => {},
+  warn: false,
+  warnText: '',
+};
+
+const DropdownWithIcon = ({ id, items, itemToString, onChange, ...other }) => {
```
bjornalm, line 108: I don't remember but I think we said that this component would just be an iot extension of the normal dropdown, so it doesn't need a new name.

src/components/Dropdown/DropdownWithIcon.jsx
```diff
@@ -0,0 +1,140 @@
...
+  const onSelectionChange = (changes) => {
+    const { selectedItem } = changes;
```
bjornalm, line 110: Why are we destructuring to only pass the selectedItem to the parent component? If we want compatibility with the carbon Dropdown we should not change the callback parameters.

src/components/IconDropdown/IconDropdown.jsx
```diff
@@ -0,0 +1,260 @@
...
+      heightStyle + defaultHelperPadding * 2
+    }px)`;
+
+    return selectedFooter !== undefined && selectedFooter !== null ? (
+      <div
+        className={`${iotPrefix}--dropdown__footer`}
+        style={{
+          width: widthStyle,
+          transform: direction === 'top' ? topTranslate : bottomTranslate,
+          paddingTop: direction !== 'top' ? `${heightStyle}px` : 0,
```
bjornalm, line 185: we could use `direction === 'bottom'` for better readability

src/components/IconDropdown/IconDropdown.jsx
```diff
@@ -0,0 +1,260 @@
...
+
+            if (change.highlightedIndex !== undefined) {
+              setHighlightedIndex(change.highlightedIndex);
+            }
+          },
+        }}
+        {...other}
+        itemToString={itemToString}
+      />
+      {isOpen && direction !== 'top' && <Footer />}
```
bjornalm, line 252: we could use `direction === 'bottom'` for better readability

bjornalm (changes requested):
Nice work, I think we are almost there.

It looks like a small miss-alignment on the right side here: 
image

bjornalm (approved):

src/components/IconDropdown/_icon-dropdown.scss
```diff
@@ -0,0 +1,152 @@
...
+.#{$iot-prefix}--dropdown__image-button:hover {
+  border-color: transparent;
+
+  ::before {
+    opacity: 0;
+  }
+}
+
+.#{$iot-prefix}--icon-dropdown__footer {
+  @include box-shadow();
```
tay1orjones, line 14: I think the box shadow should be applied uniformly so that the box shadow is present with or without the footer. Currently it sort of flashes on and off. This capture isn't the best but you can see the shadow flashing on and off on hover.
hover shadow

tay1orjones (approved):
~A couple small things that we can defer into a new issue if you would prefer~ I opened #1771 for these small visual fixes.

When the Dropdown is open, the listbox has a blue border. IconDropdown does not have this blue border.
image image

tay1orjones:
I opened https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1771 for the small visual fixes.

PR #1706, opened 2020-10-20
feat(exports): update exports to pass through carbon components

tay1orjones:
**Summary**

- @jonguenther pointed out a few components that are available in Carbon but not available through our package

- update index.js exports

- run a local build, these should be there
- the exports table in [the welcome story](https://deploy-preview-1706--carbon-addons-iot-react.netlify.app/?path=/story/0-getting-started--about-storybook) should be updated to show these as available from our package


sstone2423 (approved):

davidicus (approved):
this is good to go. I think we need to write a script to automate this.

PR #1713, opened 2020-10-20
CardEditor timeseries support

JoelArmendariz:
Summary
Adds support for Line Chart configuration in the larger DashboardEditor component.

Adds:
`CardEditFormContent`
`CardEditFormSettings`
`DataSeriesFormItem`

Go to the `DashboardEditor` `Default` story => add a `Time Series line card` => configure each field in the `CardEditor` and confirm that the JSON payload in the JSON editor modal is updated along with the visible card in the dashboard.
Make sure the JSON updates the editor, and the editor updates the JSON.


tay1orjones (changes requested):
A couple notes of feedback

<details>
  <summary>1. The sidebar background/layout is broken when the browser window size is not tall enough to accommodate the entire sidebar without a scrollbar.</summary>

  image
</details>


<details>
  <summary>2. I know this isn't directly related to your PR, but I noticed that when the header/title wraps, it doesn't have appropriate padding. @sstone2423 was this accounted for in the original spec? I wonder if it should truncate with an ellipsis instead of wrap?</summary>

  image
</details>

JoelArmendariz:
@tay1orjones Can you provide a screenshot for the layout breaking? I'm having trouble recreating it.

scottdickerson (approved):
LGTM

src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
```diff
@@ -0,0 +1,250 @@
...
+          id="title"
+          labelText={mergedI18n.cardTitle}
+          light
+          onChange={(evt) => onChange({ ...cardJson, title: evt.target.value })}
+          value={title}
+        />
+      </div>
+      <div className={`${baseClassName}--input`}>
+        <TextArea
+          id="description"
```
stuckless, line 172: Should the IDs be merged with the parent id?  ideally IDs should be unique in the DOM but I have  feeling that id="description" would be quite common, should this be something like, {`${this.props.id}_cardeditor_description`}

stuckless (commented):
just a comment about the naming of IDs

src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
JoelArmendariz: Good observation, I agree. I will update this.

JoelArmendariz:
@tay1orjones I have addressed the first comment in your feedback.
The second looks like it is an issue with the `<h2>` in the `PageTitleBar` component. I can open a separate issue to address whether to add ellipsis.

tay1orjones (approved):

Issue #1203, opened 2020-05-18
[Repo] provide e2e testing framework


tay1orjones:
[Internal Watson-IoT/pal-tracking#427](https://github.ibm.com/Watson-IoT/pal-tracking/issues/427)

### Summary

Currently we are unable to reliably test certain functions of components via `jest`/`enzyme`/`testing-library/react`.  We also need a way to test for visual regressions. 

- Keyboard focus/interaction in jsdom is not consistent with browser behavior (at the moment, cursor position via `selectionStart` and `selectionEnd` is either beginning of string or end of string, nothing in between) 
- jsdom also [has no support for layout properties](https://github.com/jsdom/jsdom#unimplemented-parts-of-the-web-platform), which means our tests for responsive sizing with sizeMe (Dashboard `Card`s) or ResizeObserver (`Breadcrumb`) are inherently a bit flawed. 
- Additionally, enzyme's simulate [doesn't actually simulate anything](https://github.com/enzymejs/enzyme/issues/1982#issuecomment-455729916)

We need a way to test components in browser, and we need documentation specifying how and when this is done. These tests should be used very sparingly as they're costly/slow.

**Additional context**

[Cypress](https://www.cypress.io/) is a good candidate for this. [testing-library has an adapter](https://testing-library.com/docs/cypress-testing-library/intro), so the syntax can be similar to our unit tests.

New alpha component cypress renderer.
https://docs.cypress.io/guides/component-testing/introduction#What-is-Component-Testing

Visual diffing plugin
https://github.com/meinaart/cypress-plugin-snapshots

tay1orjones:
It is worth mentioning that cypress also has some gaps in their support of native browser events, https://github.com/cypress-io/cypress/issues/311 and they recently pulled future improvements off their roadmap.

I'm unsure how this realistically impacts the cost/benefit ratio of cypress and it's non-selenium based architecture.

Some exploration into the topic would be beneficial to make an informed decision.

PR #1722, opened 2020-10-22
feat(dashboardgrid): allow resizing by dragging the border

### LINKED ISSUE #1598: [DashboardGrid / Card] allow resizing by dragging the border
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Currently, the only way to change the size is to do it programmatically. It could be beneficial to allow the user to drag the border of a card while in a DashboardGrid to resize it and see the changes or previews live.

This would probably only be worth effort if we can truly lock the resizing to each of our predefined sizes. If we have to actually try to adapt our styling to each completely custom size, it would be costly

**Additional context**
React grid layout allows resizing of items and allows min and max widths / heights so I believe this is possible.

Related props: https://github.com/STRML/react-grid-layout#grid-item-props

Related demo: https://strml.github.io/react-grid-layout/examples/0-showcase.html

upcoming release?

No specific timeline

### Available extra resources

bjornalm:
Closes #1598 

**Summary**
Added resize functionality to all cards when present in a DashboardGrid. The feature is based on the underlying resize functionality of the react-grid-layout which passes down a resize handler to the grid items (cards). The DashboardGrid listens to all resize callbacks and makes sure the only valid card sizes are suggested and available as drop while dragging.

The actual card content is showing during resize and I believe performance is acceptable. In the stories added the card also gets transformed between the defined card sizes during the resize drag process.

- Modified DashboardGrid to manage card resizing using strict card sizes. It now exposes onCardSizeChange and onResizeStop. Both callbacks include the current name and size of the card + all additional parameters passed from the react-grid-layout that might be needed to manage/store the new state of the grid. 
- Modified Card to wrap the onStart/onStop callbacks of the resize handle so that the card can know when it is resized (and show resize border)
- Modified all the card components to extract the resize handles from children and explicitly pass them via a special prop to the underlying Card component. This was needed since most cards don't pass down the children to the Card and some pass children as a function. 
- All cards (including the base Card) has a isResizable prop that controls if the card is resizable in the DashboardGrid.
- Added useMemo to PieChart so that the sample data isn't regenerated during resize
- Removed the mock in the mock of componentUtilityFunctions.js in the DashboardGrid.test and tested the result instead.
- Created unit tests
- Created a story that shows a single card that can be resized. 
- Created a story that shows all the card in different sizes where some cards have minimum sizes defined.
- adjusted the CSS to be more inline with the PAL design
- adjusted the coverage threshold 

I tested in Chrome, Safari & Firefox.
Checkout out the two stories and plat around with the resizing and knobs. Verify that performance and functionality is ok.

https://deploy-preview-1722--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-dashboard-grid--dashboard-all-cards-as-resizable

https://deploy-preview-1722--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-dashboard-grid--dashboard-resizable-card

**Note!**
If we upgrade to the latest version (1.1.1) we can use the horizontal (west) & vertical (south) border resize as shown in the PAL design. I've tried it and it works, but since it requires a major upgrade of the react-grid-layout I will create a separate issue for that.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -13,11 +13,16 @@ module.exports = {
       functions: 80,
       lines: 80,
     },
-    './src/components/**/!(ColumnHeaderSelect|FilterHeaderRow|TableToolbar|RowActionsCell|RowActionsError|StatefulTable|StatefulTableDetailWizard|CatalogContent|FileDrop|HeaderMenu|Dashboard|CardRenderer|Attribute|UnitRenderer|ImageHotspots|ImageControls|PageHero|PageTitle|EditPage|AsyncTable|ImageCard|WizardHeader|TableHead|ColumnResize|DateTimePicker|TimeSeriesCard|BarChartCard).jsx': {
+    './src/components/**/!(ColumnHeaderSelect|FilterHeaderRow|TableToolbar|RowActionsCell|RowActionsError|StatefulTable|StatefulTableDetailWizard|CatalogContent|FileDrop|HeaderMenu|Dashboard|CardRenderer|Attribute|UnitRenderer|ImageHotspots|ImageControls|PageHero|PageTitle|EditPage|AsyncTable|ImageCard|WizardHeader|TableHead|ColumnResize|DateTimePicker|TimeSeriesCard|BarChartCard|DashboardGrid).jsx': {
       statements: 80,
       branches: 80,
       functions: 80,
       lines: 80,
+    },
+    './src/components/Dashboard/DashboardGrid.jsx': {
+      statements: 74,
+      branches: 61,
+      lines: 74,
     },
     './src/components/BarChartCard/BarChartCard.jsx': {
       // TODO: Add tests for tooltip interaction and formatting when below issue is solved
--- a/jest.config.js
+++ b/jest.config.js
@@ -20,9 +20,9 @@ module.exports = {
       lines: 80,
     },
     './src/components/Dashboard/DashboardGrid.jsx': {
-      statements: 74,
+      statements: 73,
       branches: 61,
-      lines: 74,
+      lines: 73,
     },
     './src/components/BarChartCard/BarChartCard.jsx': {
       // TODO: Add tests for tooltip interaction and formatting when below issue is solved
--- a/src/components/Dashboard/DashboardGrid.jsx
+++ b/src/components/Dashboard/DashboardGrid.jsx
@@ -13,6 +13,7 @@ import {
   DASHBOARD_BREAKPOINTS,
   DASHBOARD_COLUMNS,
   DASHBOARD_CONTAINER_PADDING,
+  CARD_SIZES,
 } from '../../constants/LayoutConstants';
 import { DashboardLayoutPropTypes } from '../../constants/CardPropTypes';
 
@@ -63,6 +64,10 @@ export const DashboardGridPropTypes = {
   onLayoutChange: PropTypes.func,
   /** Optionally listen to window resize events to update a dashboard template */
   onBreakpointChange: PropTypes.func,
+  /** Callback for when a card has been resized */
+  onResizeStop: PropTypes.func,
+  /** Callback for when a card has been resized by drag */
+  onCardSizeChange: PropTypes.func,
 };
 
 const defaultProps = {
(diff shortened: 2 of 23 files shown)


scottdickerson (approved):
LGTM

jest.config.js
```diff
@@ -13,12 +13,17 @@ module.exports = {
...
       lines: 80,
     },
-    './src/components/**/!(ColumnHeaderSelect|FilterHeaderRow|TableToolbar|RowActionsCell|RowActionsError|StatefulTable|StatefulTableDetailWizard|CatalogContent|FileDrop|HeaderMenu|Dashboard|CardRenderer|Attribute|UnitRenderer|ImageHotspots|ImageControls|PageHero|PageTitle|EditPage|AsyncTable|ImageCard|WizardHeader|TableHead|ColumnResize|DateTimePicker|TimeSeriesCard|BarChartCard).jsx': {
+    './src/components/**/!(ColumnHeaderSelect|FilterHeaderRow|TableToolbar|RowActionsCell|RowActionsError|StatefulTable|StatefulTableDetailWizard|CatalogContent|FileDrop|HeaderMenu|Dashboard|CardRenderer|Attribute|UnitRenderer|ImageHotspots|ImageControls|PageHero|PageTitle|EditPage|AsyncTable|ImageCard|WizardHeader|TableHead|ColumnResize|DateTimePicker|TimeSeriesCard|BarChartCard|DashboardGrid).jsx': {
       statements: 80,
       branches: 80,
       functions: 80,
       lines: 80,
     },
+    './src/components/Dashboard/DashboardGrid.jsx': {
```
davidicus, line 22: Why did we have to make exceptions here?

bjornalm: I found it difficult to write tests with React Testing Library and their strategy (test the way a user would test) in combination with the underlying react-grid-layout library. I was not able to get drag events working in the test environments and exposing additional functions by export required refactoring that made the code worse.

I can a) spend some more time on it to se if I can find a way to proper test it, b) use enzyme, c) leave as is

tay1orjones: Yeah that makes sense - I think it would be better tested via cypress in a real browser #1203

tay1orjones (commented):
I'm not sure if there's any way to improve this, but at times the reflow on resize is a bit jarring and not what I'd expect.

<details>
  <summary>Here I'm just making it smaller, but when I let go it reflows all wonky.</summary>

  auto reflow
</details>






<details>
  <summary>Other times if I get it on a boundary it causes some wild reflow render flashing</summary>

wonky
</details>

bjornalm:
> I'm not sure if there's any way to improve this, but at times the reflow on resize is a bit jarring and not what I'd expect.
> 
> Here I'm just making it smaller, but when I let go it reflows all wonky.
> auto reflow
> 
For this story I haven't implemented a state handler for the reordering of cards, only for the resize, so I guess that you had reordered your cards and when you then resize a card the new card size is loaded with the layout of the original positions. The easiest fix is probably that I disable the 'isEditable' option for this story, alternately implement the state of the reordering as well.  

> Other times if I get it on a boundary it causes some wild reflow render flashing
I might be able to mitigate or fix this. I will investigate.

tay1orjones:
@bjornalm 👍  If there's a way for the consumer to provide a state handler for the reordering of cards, we can defer those pieces into separate issues if you prefer

src/components/Dashboard/DashboardGrid.jsx
```diff
@@ -141,6 +221,8 @@ const DashboardGrid = ({
   layouts,
   onLayoutChange,
   onBreakpointChange,
+  onCardSizeChange,
+  onResizeStop: onResizeStopCallback, // TODO: remove?
```
tay1orjones, line 225:
```suggestion
  onResizeStop: onResizeStopCallback,
```

tay1orjones (approved):

bjornalm:
> @bjornalm 👍 If there's a way for the consumer to provide a state handler for the reordering of cards, we can defer those pieces into separate issues if you prefer

We never created such a story, but I assume that clients that use the `isEditable` of the DashboardGrid has implemented a state handler so that they can actually save the new layout, otherwise it wouldn't be very useful. Should I create an issue for adding such a story **and** story for `isEditable` combined with this new resize functionality?

bjornalm:
@tay1orjones I think I managed to get the crazy reflow fixed. We now calculate for the possibility that a card is jumping around during resize and that the x and y value will change as a consequence. The fix compensates for that so that for example a card that is pushed to a new row because it grew big will have the dropzone stay that new size. You can check it out here
https://deploy-preview-1722--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-dashboard-grid--dashboard-all-cards-as-resizable

tay1orjones:
@bjornalm Wow, yes that is such an improvement!

And yes you're right about `isEditable`, sorry for my confusion. I don't think we need a story showcasing how to wrap a prop in state to toggle on/off. A story for isEditable combined with this new resize functionality could be useful, though not a requirement for this PR.

PR #1734, opened 2020-10-28
fix(tablecard): last column duplicated on resize

### LINKED ISSUE #1701: [TableCard] Last column duplicated on resize
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
Changing the resize prop can duplicate the last column.

### To Reproduce

1. Open the browser debug console and go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-tablecard--table-with-row-expansion
2. Change size to LargeWide
3. Notice how the last column is now duplicated

### Expected behavior
No duplication of last column

### Environment/versions:

- OS: [e.g. MacOS]
- Browser: [e.g. chrome ]

### Additional context

image

bjornalm:
Closes #1701

**Summary**
I've been all over the place with this one trying to figure out where to add the fix. The core problem was that the CardTable was using the StatefulTable and at the same time modifying the number of columns externally (for smaller sizes some columns should not be shown). When a column was removed the StatefulTable could be sending a `columns` prop of length 6 and an `ordering` prop of length 5 to the Table in a render call before the state had synced the two props. The `columns` prop is used in the Table to create the default `ordering` prop of length 6 and since the Table merges default props with normal props a duplication could occur. 

I was unable to reproduce this behaviour anywhere else then in the CardTable scenario, so I finally settled to fix the issue there.

- Updated CardTable to hide columns instead of removing them
- Added unit tests

1. Go to https://deploy-preview-####--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-tablecard--table-with-thresholds
2. Change the size to Large and make sure the Toolbar icon for "Expand to full screen" is showing
3. Change the size to LargeWide
4. Make sure the pressure column is only shown once and make sure the Toolbar icon for "Expand to full screen" is not showing

### DIFF
--- a/src/components/TableCard/TableCard.jsx
+++ b/src/components/TableCard/TableCard.jsx
@@ -566,26 +566,7 @@ const TableCard = ({
             ? i.filter
             : { placeholderText: mergedI18n.defaultFilterStringPlaceholdText }, // if filter not send we send empty object
         }))
-        .concat(hasActionColumn ? actionColumn : [])
-        .map((column) => {
-          const columnPriority = column.priority || 1; // default to 1 if not provided
-          switch (newSize) {
-            case CARD_SIZES.LARGETHIN:
-              return columnPriority === 1 ? column : null;
-
-            case CARD_SIZES.LARGE:
-              return columnPriority === 1 || columnPriority === 2
-                ? column
-                : null;
-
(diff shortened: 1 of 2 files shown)


tay1orjones (approved):
Interesting bug - kudos for figuring it out! 🎉

PR #1749, opened 2020-11-03
1621 card selection

### LINKED ISSUE #1621: [Card] Add isSelected prop to Card (and ValueCard, BarChartCard, etc...)
### What package is this for?

- [x] React
- [ ] Angular

### Summary

In the dashboard editor work, there is a need to render a card as "selected" for editing.

image

If we want to render the border on the outside of the tile content (which might be preferable for usage in a dashboard), the following should work:

```
{
  border: solid $spacing-01 $interactive-04;
  box-sizing: content-box;
  margin: -$spacing-01;
}
```

davidicus:
Closes #1621 

**Summary**

Add `isSelected` prop to base `Card` component that will allow for selection styling.  The component spreads props that will allow for event handlers for click, focus and blur, allowing teams to customize their selection logic. 

- added `isSelected` prop to `Card.jsx`
- added stateful component to story that showcases selection. 

- click and tab through the Card component story and see the styling change.


marcJV (approved):

src/components/Card/Card.jsx
```diff
@@ -54,6 +55,11 @@ const CardWrapper = ({
   ...others
 }) => {
   const validOthers = filterValidAttributes(others);
+  const [isCardSelected, setIsCardSelected] = useState(false);
```
sstone2423, line 58: does this actually need state? It looks like the prop should be able to handle it. See how `isResizable` is being used at line 251. I'd imagine this would work the same way

davidicus: Probably not. I originally thought we would have to pass state setters as arguments to onblur handlers when I first created this. I will double check and remove if unnecessary

davidicus: @sstone2423 removed! thanks

src/components/Card/_card.scss
```diff
@@ -12,6 +12,12 @@ $iot-header-padding: $spacing-05;
   display: flex;
   flex-direction: column;
   overflow: hidden;
+
+  &__selected {
+    border: solid $spacing-01 $interactive-02;
+    box-sizing: content-box;
+    margin: -$spacing-01;
```
sstone2423, line 19: Is this negative margin necessary? I have it currently implemented in DashboardEditor and it bleeds the card into the upper div

Screen Shot 2020-11-04 at 3 21 52 PM

example can be seen here: https://deploy-preview-1751--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-experimental-dashboardeditor--default

davidicus: you get a jump in the card layout otherwise. Can look into it

davidicus: Removing negative margins for this and the resize.

sstone2423 (approved):

PR #1751, opened 2020-11-03
feat(DashboardEditor): add breakpoint switcher

### LINKED ISSUE #1750: [DashboardEditor] There should be a dynamic way of changing layouts / breakpoints
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Currently, the only way of changing layouts or breakpoints is to resize your window, which limits users to their current screen size. This causes issues for users trying to set up their dashboards for screens smaller or larger than their current window because they either can't in the case of larger screen sizes, or they need to manually resize their window and hope to get the right layout. The other issue is that the user doesn't know what breakpoint they are at if they are manually resizing their window

upcoming release?

November release

sstone2423:
Closes #1750 

**Summary**

- Adds a breakpoint switcher consisting of a ContentSwitcher and IconSwitches. 
- breakpoint switcher prop model:
```
breakpointSwitcher: PropTypes.shape({
    enabled: PropTypes.bool,
    initialValue: PropTypes.string,
  }),
```

- A user can select 4 options:
  - Fit to screen (gives regular functionality as the DashboardGrid, meaning it will adjust to the current window size)
    - when this is selected, the breakpoint info text is not visible and the grid will use the window's size to determine between the 3 allowed layout sizes seen below
  - X-large
  - Large
  - Medium
- The dashboard preview has some styling changes as well
  - 1 darker gray background defines the entire container size
  - 1 lighter gray background to define the true DashboardGrid space
  - A dashed border along the DashboardGrid space to give an additional indication
  - Text above the DashboardGrid space to describe the current layout / breakpoint size
- Adds a new `supportedLayouts` prop to DashboardGrid that allows the consumer to define which layouts they want to enable

- add window.resize useEffect to force a resize after layout changes, which then forces react-grid-layout to resize and reorder the cards correctly
- added with breakpoint switcher story
- changed IconSwitch's event handler props to not be required, as regular switches in ContentSwitcher do not require it
- added `isNil` check to ValueCard's dataState to try and stop proptype warnings when no `dataState` is being passed
- added `supportedLayouts` prop and testing to DashboardGrid

- Go to `with breakpoint switcher` story
- The story will start on 'fit to screen', so test that the editor acts as it did before. Try adding your cards, and then resizing your window and see that DashboardGrid is resizing and reordering correctly, and that there is no breakpoint info text visible
- Then swap to each size and do the same checks as above, but now the size of your window should only affect the inner dashboard preview. If the window goes smaller than the breakpoint size, there should be a scroll bar. If the window goes larger than the breakpoint size, you should see a border, and a darker background showing where the true dashboard grid stops. There should be breakpoint info text to describe the currently selected layout


scottdickerson (approved):
LGTM

JoelArmendariz:
Looks like there's some weird padding issues around the edges that cut off the `selected` border.

before:
Screen Shot 2020-11-04 at 3 58 12 PM

this pr:
Screen Shot 2020-11-04 at 3 58 29 PM

JoelArmendariz (approved):
LGTM

davidicus:
image

@sstone2423 Seems like its slightly off.

davidicus (approved):

Issue #1615, opened 2020-09-18
[test] unit test custom hooks


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Custom exported hooks intended for consumption [aren't subject to testing coverage thresholds](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1612/files#r490317885) for a couple of reasons

1. We haven't defined expectations of what/when/how hooks are to be tested.
1. As a result, we don't have any guidance on how best to test hooks. (Linking to [this article](https://kentcdodds.com/blog/how-to-test-custom-react-hooks) might be sufficient though)
1. Some hooks may be more of just a reference implementation, similar in purpose to a story in storybook. These could/should be exempt from testing thresholds.

PR #1757, opened 2020-11-05
feat(suiteheader) Separating data fetching logic in a util function.

blechner:
**Summary**

- Separating the logic that assembles the data to be consumed by `SuiteHeader` from the hook, in an exported function, so that applications can import this function instead of using the hook.

- No component has been changed in this PR, so no new tests have been added.


src/components/SuiteHeader/SuiteHeader.story.jsx
```diff
@@ -227,6 +228,99 @@ export const HeaderWithSurveyNotification = () => {
...
+        surveyData={surveyData}
+      />
+    );
+  }
+  return <StatefulExample />;
+};
+
+HeaderWithHook.story = {
+  name: 'Header with hook',
+};
```
tay1orjones, line 264: Is there something wrong with these stories if they're uncommented and included in storybook?

blechner: @tay1orjones Yes, these stories are testing async data loading, which is not the scope of regular component stories, it is just to make sure that the data hook is working as expected during development (the data hook is excluded from coverage as well). Besides, these stories mount the component without data in the first render, which would make the stories fail as some properties are required. Also, for some reason, stories fail for containing async tasks, with this error:
```
Failed: "Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in %s.%s"
```
If you prefer, I can just delete the stories instead of commenting them out.

jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
     'src/components/**/*.js?(x)',
     '!src/**/*.story.js?(x)',
     '!src/**/hooks/*.js',
+    '!src/**/util/*.js',
```
sstone2423, line 6: Is there a reason that this shouldn't be tested?

tay1orjones: I agree this should probably be covered in tests if others are relying on it. All the API calls would need to be mocked though. @blechner what were your thoughts on exempting this? At the very least this glob should be much more specific, we don't want to ignore all files in `util` folders.
```suggestion
    '! src/components/SuiteHeader/util/suiteHeaderData.js',
```

src/components/SuiteHeader/SuiteHeader.story.jsx
tay1orjones: I'm fine either way. If it's helpful in development I think we could leave them uncommented, but if the stories aren't helpful in development and also fail, then we can remove them.

jest.config.js
sstone2423: Considering this is the header that multiple products will be using, I'd think we would want even more coverage than our usual threshold, especially with data fetching involved

blechner: How about we open an issue for increasing util/ test coverage and prioritize it accordingly?  This specific PR doesn't contain any logic changes (just refactoring), so it would make sense to me to contribute tests in a separate logical unit of work.

blechner: This PR is just a refactor, test coverage could be added later but should be a separate piece of work. I'd welcome some help mocking out the API layer, i'm happy to come along and add the logical tests once that is done.

blechner: btw @tay1orjones I've added your suggested commit, but it had a typo, a space between the `!` and the `src/`, it is fixed now 👍

tay1orjones: @blechner thanks! Sorry for the typo. 

Non-experimental components in the storybook are required to have 80% test coverage. That said, we haven't defined or documented when/how best to test hooks #1615. I have opened https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1810 to specifically track test coverage for SuiteHeader and it's subcomponents/hooks.

tay1orjones (approved):

sstone2423 (approved):

PR #1785, opened 2020-11-11
fix(icondropdown): fixed outline and dropshadow and top direction

### LINKED ISSUE #1771: [IconDropdown] fix listbox box-shadow and add border
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Feedback from https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1705#pullrequestreview-527143316

When the Dropdown is open, the listbox has a blue border. IconDropdown does not have this blue border, but should.
image image

The box shadow should be applied uniformly so that the box shadow is present with or without the footer. Currently it sort of flashes on and off. This capture isn't the best but you can see the shadow flashing on and off on hover.
hover shadow

marcJV:
Closes #1771

**Summary**

- Fixed dropshadow from appearing and disappearing with the footer. The focus outline has also been restored when opening the menu. Furthermore the 'top' direction would break with multiple lines of text, also fixed.

- See above

- Head on over to https://deploy-preview-1785--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-experimental-icondropdown--default 
- Open menu, observe drop shadow exists. Also observe blue outline exists. Mouse over items, footer should show as well with the outline and dropshadow updates. Change orientation to 'top' from 'bottom'. Repeat steps.


src/components/IconDropdown/IconDropdown.jsx
```diff
@@ -156,6 +160,9 @@ const IconDropdown = ({
       ? validationText.clientHeight + defaultHelperPadding
       : 0;
 
+  const labelHeight =
+    dropdownLabel !== undefined ? dropdownLabel.clientHeight : 0;
```
tay1orjones, line 164: I think these measurements should be made [via a `ref` in a `useCallback`](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node)?

marcJV: updated, this is a better way to encapsulate this, thanks!

tay1orjones (approved):

davidicus (approved):

PR #1790, opened 2020-11-11
Add translated strings for timerange labels

JoelArmendariz:
Closes #

**Summary**

Need to translate time range labels

 - Add to and display from the `i18n` default prop
 - Small fix to the dropdown items in the dataSeries where the options were unnecessarily taking the labels as the `text` prop instead of the dataSourceId.

Go to any `DashboardEditor` story, add or select a timeseries card and ensure that the labels for the time range appear correctly.


src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
```diff
@@ -155,8 +173,8 @@ const CardEditFormContent = ({
 
   const validTimeRanges = getValidTimeRanges
     ? getValidTimeRanges(cardConfig, selectedDataItems)
-    : Object.keys(defaultTimeRangeOptions);
-
+    : defaultTimeRangeOptions;
+  console.log('i18n', mergedI18n);
```
tay1orjones, line 177:
```suggestion
```

scottdickerson:
@JoelArmendariz I think you might need to keep the data item labels for those in the Summary dashboard view that you're trying to 'simplify' the names of

JoelArmendariz:
> @JoelArmendariz I think you might need to keep the data item labels for those in the Summary dashboard view that you're trying to 'simplify' the names of

@scottdickerson The labels will still show under the dropdown where the user can edit each dataItem. UX decided that the dropdown would be a "source of truth" and exclusively show the dataSourceIds, while the list under it would show the user created labels, even if the names don't match up between the list and the dropdown.

scottdickerson (approved):
LGTM

bjornalm (approved):

PR #1794, opened 2020-11-12
fix(hierarchylist): need to useDeepCompareEffect to stop infinite loop

scottdickerson:
Closes # internal bug: https://github.ibm.com/wiotp/monitoring-dashboard/issues/1633

**Summary**

- Because of the useEffect comparison, objects that come from redux can cause infinite render loops to happen.

-fix(List): memoize i18n as well.

- Verify searching within HierarchyList still works

### DIFF
--- a/src/components/List/HierarchyList/HierarchyList.jsx
+++ b/src/components/List/HierarchyList/HierarchyList.jsx
@@ -244,7 +244,8 @@ const HierarchyList = ({
     cancelMoveClicked();
   };
 
-  useEffect(
+  useDeepCompareEffect(
+    // have to use deep compare to accurately compare items
     () => {
       // Expand the parent elements of the defaultSelectedId
       if (defaultSelectedId) {
--- a/src/components/List/List.jsx
+++ b/src/components/List/List.jsx
@@ -1,4 +1,4 @@
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { DragDropContext } from 'react-dnd';
@@ -144,7 +144,7 @@ const List = forwardRef((props, ref) => {
     itemWillMove,
     emptyState,
   } = props;
-  const mergedI18n = { ...defaultProps.i18n, i18n };
+  const mergedI18n = useMemo(() => ({ ...defaultProps.i18n, i18n }), [i18n]);
   const selectedItemRef = ref;
   const ListHeader = overrides?.header?.component || DefaultListHeader;
   const renderItemAndChildren = (item, index, parentId, level) => {


sstone2423 (approved):

src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -244,7 +244,8 @@ const HierarchyList = ({
     cancelMoveClicked();
   };
 
-  useEffect(
+  useDeepCompareEffect(
```
tay1orjones, line 247: We're already running a deep compare effect anytime the items change up on line 209. Would be best to encapsulate this logic there together so we're not deep comparing the items twice.

tay1orjones (changes requested):

src/components/List/HierarchyList/HierarchyList.jsx
tay1orjones: Also worth noting that this opens the HierarchyList up to a potential out of memory error if `items` ever contains a React.Element or Node. useDeepCompareEffect can't compare props of these types.

scottdickerson: hmm unfortunately I can't because I need a different trigger for the second logic (i.e. I only want it to trigger if the selectedId changes). @tay1orjones.  The first one should trigger every time the items change

tay1orjones: React [will bail out](https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update) of the upper `setFilteredItems(items);` state call if `items` is the same, so it should be fine to run `setFilteredItems(items);` when either `defaultSelectedId` or `items` changes.

tay1orjones: Ah nevermind, I see what you're saying. This is just really expensive and is very likely to cause perf issues with heavily nested hierarchies. Is there any option to prevent the infinite loop upstream before passing new `items` to the component?

scottdickerson: it is possible to memoize the items so the same object is passed each time, but I think the symptom is soo bad unfortunately it's probably better to protect it for external consumers of the component @tay1orjones

tay1orjones: Yeah, it's an unfortunate consequence of having these monolith prop objects. Let's merge this as-is to unblock you. I do want to follow up with you to see more specifically what's causing the infinite loop in your project.

I've opened https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1797 to look at this in regards to the comparison of nodes. As soon as you upgrade your babel deps to >=7.12.x, you'll get an out of memory error.

tay1orjones (approved):

PR #1795, opened 2020-11-12
Prevent internal downshift ids from changing on every snapshot update

tay1orjones:
**Summary**

- No longer will our snapshots be updated unnecessarily with the `downshift-#` id in aria-roles, htmlFor, etc when snaps are regenerated

- modify jest test setup file to reset the internal downshift ID before each test

- status checks pass, you can see in the snapshot files that the id's have been reset to `0` and should no longer increment.


sstone2423 (approved):

marcJV (approved):

tay1orjones:
@davidicus I'll need your approval here as it modifies the public api snapshot

davidicus (approved):

PR #1800, opened 2020-11-13
Create codeql-analysis.yml

### LINKED ISSUE #1799: Setup CodeQL for vulnerability detection
### Summary

[CodeQL](https://securitylab.github.com/tools/codeql) should be setup as a GitHub workflow to automatically detect common vulnerability and coding errors.

filipecorrea:
Closes #1799 

**Summary**

- Setup CodeQL workflow to automatically detect common vulnerability and coding errors.

- CodeQL scan should pass


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,35 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ next, beta, master ]
+  pull_request:
+    branches: [ next, beta, master ]
+  schedule:
+    - cron: '39 21 * * 6'
```
tay1orjones, line 9: Do we need to run this on a schedule if we're running it on every push and pr?

tay1orjones (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
```
tay1orjones, line 19: I think this part is unnecessary as there is only one job in this file and we can explicitly declare the language below

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
```
tay1orjones, line 29:
```suggestion
        languages: javascript
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
```
tay1orjones, line 32: I think this is only for compiled languages, which wouldn't apply to our package. https://github.com/github/codeql-action#usage

> Autobuild attempts to build any compiled languages (C/C++, C#, or Java).
> If this step fails, then you should remove it and run the build manually (see below).

I'd think we'd want to run a `yarn install && yarn build` so the analyze happens against compiled as well as the non-compiled source code. Thoughts? I'm not super familiar with this action setup, just looking through the docs linked above.

tay1orjones (changes requested):
Putting a "request changes" on here just based on the new feedback, but let me know if any of my comments are off base and invalid.

.github/workflows/codeql-analysis.yml
filipecorrea: You're right, this shouldn't apply to our package. I'll remove this step.

tay1orjones (approved):

davidicus (approved):

Issue #1797, opened 2020-11-13
[repo] useDeepCompareEffect, dequal - RangeError: Maximum call stack size exceeded


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

`useDeepCompareEffect` [does not support](https://github.com/kentcdodds/use-deep-compare-effect/issues/7) comparing props (or prop objects/arrays that contain props) of type Element or Node, `useDeepCompareEffect` is unable to compare them due to these including the entire (very large) React tree.

This presents as an error of the underlying comparison library within `useDeepCompareEffect`, `dequal` in both browser and/or tests:

```
RangeError: Maximum call stack size exceeded

      at dequal (node_modules/dequal/dist/index.js:9:16)
      at dequal (node_modules/dequal/dist/index.js:77:28)
      at dequal (node_modules/dequal/dist/index.js:77:28)
      at ...
```

One way to partially fix this is to stop blindly comparing huge prop objects, like `table.view` for instance, and instead compare the specific props within that are not of type Element or Node. I've submitted a PR for this for StatefulTable/Table https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1782 but we have additional components that we need to investigate

- [x] [`CardRenderer.jsx`](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/d63c51871be9006f698354af6821d5cbdb637294/src/components/Dashboard/CardRenderer.jsx#L31)
- [x] [`ImageHotspots.jsx`](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/1644c8407ea05e7cdcb7900c91f847ba663a3904/src/components/ImageCard/ImageHotspots.jsx#L365)
- [x] [`HierarchyList.jsx`](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/581e8e247530a6979354de727c6694285d0fd0b4/src/components/List/HierarchyList/HierarchyList.jsx#L209) (this is confirmed to be comparing Nodes within the `items` prop and is prone to this error)
- [x] [`TableHead.jsx`](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/2aac0dc996a2f361cc05c0abcea87292a0c50b69/src/components/Table/TableHead/TableHead.jsx#L285)
[...]

PR #1804, opened 2020-11-16
patch to master -- fix(listitem): need to support render function

scottdickerson:
Closes # internal https://github.ibm.com/wiotp/monitoring-dashboard/issues/1633

**Summary**

- fix(ListItem): if you take a full React element as the rowAction prop, the clone can go into infinite recursion, this is because the React Element can point to it's parent, to fix allow you to pass a render prop so if you have complex rowActions you can use the renderProp
- fix(list-item.scss): the overflow row actions can go offscreen because not enough padding space is saved, and also need to absolute position them

- I compared the layouts of all the list stories and made sure there were no regressions


tay1orjones:
Travis isn't going to complete because I turned it off for #1788 - I'll create a pr to hotfix the deployment change to `master` so the deploy will work on `master` for this PR.

src/components/List/ListItem/ListItem.jsx
```diff
@@ -32,7 +32,11 @@ const ListItemPropTypes = {
   expanded: PropTypes.bool,
   value: PropTypes.string.isRequired,
   secondaryValue: PropTypes.string,
-  rowActions: PropTypes.arrayOf(PropTypes.node), // TODO
+  /** either a callback render function or a node */
+  rowActions: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.func,
+  ]),
```
tay1orjones, line 39: Is this to fix the dequal max call size w/ useDeepCompareEffect? I think this could be a good overall solution for #1797, and place a deprecating notice on the usage of array of nodes.

src/components/List/ListItem/ListItem.jsx
```diff
@@ -120,7 +124,7 @@ const ListItem = ({
     return nestingLevel > 0 ? (
       <div
         className={`${iotPrefix}--list-item--nesting-offset`}
-        style={{ width: `${nestingLevel * 30}px` }}
+        style={{ width: `${(nestingLevel - 1) * 30}px` }}
```
tay1orjones, line 127: This causes a regression with indentation in hierarchy lists.

image
 `My Reports` should be indented to be aligned with the `Tasks`

tay1orjones (changes requested):

davidicus (approved):
looks good aside from the regression that Taylor mentions.

src/components/List/ListItem/ListItem.jsx
scottdickerson: No they're separate fixes though I think deprecating notice is a good idea

src/components/List/ListItem/ListItem.jsx
```diff
@@ -116,11 +121,22 @@ const ListItem = ({
...
+  }
+
   const renderNestingOffset = () => {
     return nestingLevel > 0 ? (
       <div
         className={`${iotPrefix}--list-item--nesting-offset`}
-        style={{ width: `${nestingLevel * 30}px` }}
+        style={{
+          width: `${
+            nestingLevel === 1 && !isExpandable ? 30 : (nestingLevel - 1) * 30
```
tay1orjones, line 137: This is still regressed. `Request 2` should be indented and is not.

image

What's wrong with the current implementation? The spec calls for each list item to reserve the space for the chevron icon when the list has a hierarchy.

tay1orjones (changes requested):

tay1orjones (approved):

PR #1814, opened 2020-11-17
fix(listitem): fix row actions issue

scottdickerson:
Closes #

**Summary**

- Fixes two issues,:
one the row actions cause the list values to overlap and the padding wasn't giving enough space
two fixes the warning thrown by the DataSeriesFormItem

- Check the updated story to see the longer list item and notice the proptypes error goes away on the new data series form

### DIFF
--- a/src/components/List/ListItem/ListItem.jsx
+++ b/src/components/List/ListItem/ListItem.jsx
@@ -3,7 +3,6 @@ import { DragSource } from 'react-dnd';
 import classnames from 'classnames';
 import { Draggable16, ChevronUp16, ChevronDown16 } from '@carbon/icons-react';
 import PropTypes from 'prop-types';
-import isEmpty from 'lodash/isEmpty';
 import warning from 'warning';
 
 import { EditingStyle } from '../../../utils/DragAndDropUtils';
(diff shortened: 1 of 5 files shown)


src/components/List/ListItem/ListItem.jsx
```diff
@@ -167,9 +167,11 @@ const ListItem = ({
       </div>
     ) : null;
 
+  const hasRowActions =
```
bjornalm, line 170: nitty gritty, but can't this just be `typeof rowActions === 'function' || rowActions?.length`?

src/components/List/ListItem/_list-item.scss
```diff
@@ -175,7 +175,7 @@
         text-overflow: ellipsis;
         padding-right: $spacing-09;
         &__with-actions {
-          padding-right: 0;
+          padding-right: $spacing-06; // save room for the action button
```
bjornalm, line 178: You need even more room. The `bx--overflow-menu` is 2rem so you'll need at least `$spacing-07`

bjornalm (changes requested):

src/components/List/ListItem/_list-item.scss
scottdickerson: for some reason that seems to work for me, I think the ellipses padding helps

bjornalm (approved):
Nice job!

tay1orjones (approved):
looks great!

PR #1830, opened 2020-11-24
fix(TableCard): format timestamps in expanded rows

### LINKED ISSUE #1831: [TableCard] expanded rows don't format timestamps
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

If timestamps are added to expanded rows, they are not being formatted

### To Reproduce

1. Go to expanded rows TableCard and add a timestamp
2. See that the timestamp is unformatted

### Expected behavior

Should format timestamps similar to how we format it in the columns

sstone2423:
Closes #1831

**Summary**

- Timestamps were not being accounted for or formatted in expanded rows
- The diff is much worse than it looks. This was only a few lines that were actually changed. See my comment in the code

- Format timestamps if `type === 'TIMESTAMP` in `expandedRows` prop
- Add missing `linkTemplate` proptypes to TableCard
- Move util functions to separate file as TableCard file was getting very large
- Updated story titles to be consistent with `with <feature>` titles. Some were using `table with` titles causing the names to not line up alphabetically, making it difficult to find the desired stories
- Remove `isRequired` for ListHeader title

- Go to /?path=/story/watson-iot-tablecard--with-row-expansion-and-timestamp  , expand any of the rows, and see that the timestamp is formatted


src/components/TableCard/TableCard.jsx
```diff
@@ -764,7 +604,13 @@ const TableCard = ({
                               {item ? item.label : '--'}
                             </p>
                             <span key={`${item.id}-value`}>
-                              {item ? dataItem.values[item.id] : null}
```
sstone2423, line 767: This is the actual fix

bjornalm (approved):
Nice work

scottdickerson (approved):
LGTM

tay1orjones (approved):

PR #1834, opened 2020-11-25
feat(DashboardEditor): add isSubmitLoading callback

sstone2423:
**Summary**

- Adds `isSubmitLoading` prop for `loading` state on the submit / save button in DashboardEditor

- rename `submitDisabled` to `isSubmitDisabled`
- Add `isSubmitLoading` prop to DashboardEditor and DashboardEditorHeader

- Use the `isSubmitLoading` knob in any of the DashboardEditor stories and see that the loading state in the submit button swaps to loading correctly


bjornalm:
There seems to more changes in this PR than expected, e.g. findMatchingThresholds in the TableCard

sstone2423:
> There seems to more changes in this PR than expected, e.g. findMatchingThresholds in the TableCard

This was due to working off the same branch as another fix. Those changes are no longer shown

src/components/DashboardEditor/DashboardEditorHeader/DashboardEditorHeader.jsx
```diff
@@ -200,15 +201,16 @@ const DashboardEditorHeader = ({
...
             {mergedI18n.headerCancelButton}
           </Button>
         )}
         {onSubmit && (
           <Button
             size="field"
-            disabled={submitDisabled}
-            onClick={() => onSubmit(dashboardJson)}>
+            disabled={isSubmitDisabled}
+            onClick={() => onSubmit(dashboardJson, setIsSubmitLoading)}
```
bjornalm, line 212: Interesting, why a `setIsSubmitLoading` callback to manage an internal state instead of an `isLoading` prop like the `isSubmitDisabled`? I would choose prop over state, but perhaps there is a reason I don't see.

bjornalm (approved):
Looks good, just a question of prop vs state

src/components/DashboardEditor/DashboardEditorHeader/DashboardEditorHeader.jsx
sstone2423: I was thinking that the `isLoading` is directly related to `onSubmit`, but I see the we only use the prop in other examples like ComposedModal. I'll swap this to a prop instead to stay consistent

sstone2423: Swapped to the `isSubmitLoading` prop

JoelArmendariz (approved):

tay1orjones (approved):

Issue #1786, opened 2020-11-11
[Component Proposal] Image gallery modal


davidicus:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

The image card needs the ability to search and choose from a list of photos in the Dashboard editor flow. The image gallery component will be in a Modal and should have:
- search functionality
- icon tabs that switch between gallery and list view
- checkmark selected state for list (does the gallery also contain a checkmark?)

Link to [invision](https://ibm.invisionapp.com/share/KWO0FZ5BHQ9#/screens/319748234)
[Internal issue](https://github.ibm.com/wiotp/monitoring-dashboard/issues/1152) for reference

![image](https://user-images.githubusercontent.com/1590966/98839338-66a4ff80-2413-11eb-93bc-10a3a4596770.png)

Issue #1741, opened 2020-10-29
[CardEditor] ImageCard editing support


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

`ImageCard`s should be editable in the `DashboardEditor` via `CardEditor`. Similar in scope to the [TimeSeriesCard](https://next.carbon-addons-iot-react.com/?path=/story/watson-iot-experimental-cardeditor--for-timeseries)

**Additional context** 
- [Internal issue 1152](https://github.ibm.com/wiotp/monitoring-dashboard/issues/1152)
- Related: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1713

### Specific timeline issues / requests

Marked as a `val:1`, adding to the core team's scope for November.

davidicus:
I am closing this as all parts are complete outside the hotspot editor work that is being tracked [here](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1753).

PR #1844, opened 2020-12-01
feat(imagegallery): link the image gallery to the dashboardeditor

scottdickerson:
Partial/further work of #1786, #1741 

**Summary**

- feat(ImageGallery): links the imageGallery to be used in the DashboardEditor

- fix(ImageCardFormContent): temporarily hide the Edit Image button as the hotspot editing isn't ready yet
- feat(DashboardEditor): show/hide and update the cards based on the ImageGalleryModal
- fix(ImageGalleryModal): should search on id by default since src usually is a base64 encoded string

- Verify you can select images from the ImageGallery in the DashboardEditor


tay1orjones (approved):

davidicus (approved):
These changes look fine. There is something going on with the ability to zoom images chosen from the gallery but I do not think they stem from these changes. Once checks pass we can merge

PR #1854, opened 2020-12-02
fix(TableCard): encode variables for linkTemplate

### LINKED ISSUE #1853: [TableCard] row-specific linkTemplate variables could add reserved characters to URL
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Theres little control over row-specific variables as they're added in the `renderDataFunction` and can only use the current rows formatting. For example, if a timestamp is already being formatted in another column, the variable will also be formatted, which could cause problems in the URL. For example `7/20/2020 02:00` is not valid in URLs, but it still needs to be formatted in the other column

### To Reproduce

1. Go to /?path=/story/watson-iot-tablecard--with-row-specific-link-variables
2. Add a `hour` variable to the current linkTemplate
3. See that the URL now has the formatted hour instead of the timestamp (or any other valid format)

### Expected behavior

Should attempt to encode the variables so the URL will be valid

### Environment/versions:

- OS: MacOS
- Browser:chrome
- `carbon-addons-iot-react` version: v2.115.0

sstone2423:
Closes #1853

**Summary**

- Theres little control over row-specific variables as they're added in the renderDataFunction and can only use the current rows formatting. For example, if a timestamp is already being formatted in another column, the variable will also be formatted, which could cause problems in the URL. For example 7/20/2020 02:00 is not valid in URLs, but it still needs to be formatted in the other column
- This PR solves this issue by first encoding the variable, then passing it to the URL

- Go to /?path=/story/watson-iot-tablecard--with-row-specific-link-variables and see that the `Link` values are being formatted with URL encoding

### DIFF
--- a/src/components/TableCard/TableCard.story.jsx
+++ b/src/components/TableCard/TableCard.story.jsx
@@ -71,6 +71,7 @@ export const WithLinks = () => {
   );
   const cardVariables = object('Dynamic link variable', {
     assetId: '11112',
+    company: 'ibm',
   });
   const tableLinkColumns = [
     {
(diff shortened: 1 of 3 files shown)


tay1orjones (approved):

scottdickerson (approved):
LGTM

Issue #1702, opened 2020-10-19
[deps] upgrade to Carbon version ^10.23.0


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Umbrella issue tracking updating Carbon dependencies to version ^10.23.0

One specific item that sort-of blocks this is our consumption of Carbon's storybook stories. In the newer versions, the Carbon team is updating their stories to use .mdx documentation files. These .mdx files aren't exported, and so cause those stories to fail in our storybook. I've opened an issue to see if they'll consider exporting them https://github.com/carbon-design-system/carbon/issues/7192. Until then we'll have to copy/paste these stories and their .mdx counterparts and manually keep them in sync with new versions of Carbon. I believe we actually already copied over the first couple of these to upgrade to 10.17.x

Related items to be completed as part of this upgrade

- [ ] [[AccordionItemDefer] deprecate in favor of Carbon's deferComponentRender](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1703)
- [ ] [[Breadcrumb] refactor hasOverflow to use implementation from Carbon](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1737)
- [ ] [[FileUploaderItem] error props do not render](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1728)
  - [ ] remove temporary fix in `_file-uploader.scss`
- [ ] Overflow menu background fix introduced in https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1859 should be removed

### Specific timeline issues / requests

Anticipating this should be released near/at end of October 2020.

### Available extra resources

The core team will take this on.

PR #1859, opened 2020-12-03
fix(overflow): removed default background color for button

### LINKED ISSUE #1852: [OverflowMenu] background color is not correct
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Somewhere in the november sprint the overflow menu background color is now different. You can see the issue currently between master and next storybooks.

Invalid background color:
https://next.carbon-addons-iot-react.com/?path=/story/overflowmenu--basic

Valid background color:
https://carbon-addons-iot-react.com/?path=/story/overflowmenu--basic

image

marcJV:
Closes #1852

**Summary**

- Removed the background color from the overflow button

- _overflow-menu.scss

- head on over to ?path=/story/overflowmenu--basic and verify the background color was removed.
- Should look exactly like: https://carbon-addons-iot-react.com/?path=/story/overflowmenu--basic


sstone2423 (approved):
Nice. Did you find out why it happened? It seems like a band-aid fix

tay1orjones:
Yeah, would be nice to know why this happened. There are some other differences. 

image


Left is master, right is next. There’s a missing `display: inline-block`, even though there’s also a `display: flex`. `appearance` is missing. We now have a `font-family` declaration whereas master doesn't. I fear we might need to sift through recent PR deploy previews in attempt to pin down when this was introduced 😣

marcJV:
Narrowed it down to `v2.141.0-next.9` so far for the background color, still looking to see exactly what caused this and if this fixes all of the issues

tay1orjones:
It sounds like this is a side effect of some duplicate/mismatching dependency versions of @carbon/colors and/or carbon-components through the @carbon/charts package.

I think we can merge this as a bandaid fix with an eye to take another look when we update carbon versions #1702

tay1orjones (approved):

PR #1860, opened 2020-12-03
feat(imagecardformcontent): support clearing the image

scottdickerson:
Closes #

**Summary**

- feat(imagecardformcontent): support a button that clears the selected image
- refactor(dashboardeditor): better support selecting the image from the gallery

- Verify the story selecting an image from the gallery works
- Verify the new story of CardEditor for Image sends the onChange event correctly

### DIFF
--- a/src/components/CardEditor/CardEditForm/_card-edit-form.scss
+++ b/src/components/CardEditor/CardEditForm/_card-edit-form.scss
@@ -167,11 +167,10 @@
   }
   &-clear-button {
     position: absolute;
-    right: $spacing-05;
-    margin-bottom: 0;
-    padding: 0 $spacing-02;
-    bottom: $spacing-04;
+    right: 0;
+    top: 1.25rem;
     color: $ui-05;
+    background-color: $ui-02;
     &:hover,
     &:active {
       color: $ui-05;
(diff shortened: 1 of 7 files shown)


src/components/CardEditor/CardEditForm/_card-edit-form.scss
```diff
@@ -157,13 +157,31 @@
...
   &-btn {
     width: 100%;
   }
+  &-clear-button {
+    position: absolute;
+    right: $spacing-05;
+    margin-bottom: 0;
+    padding: 0 $spacing-02;
+    bottom: $spacing-04;
+    color: $ui-05;
```
tay1orjones, line 174: I think this positioning might be redundant if we used the `size="field"` prop on the button. I believe it's intended to size buttons for use in fields like this. Nothing to hold this up for, but would be nice to in the future have it "full bleed" without any padding like other buttons. pagination for example

image

tay1orjones (approved):
It's an edge case we might not need to address right now, but If the filename is longer than the input can display, should we truncate it and provide a tooltip on hover?

The close/remove button is transparent currently so I think the text will just run behind it

JoelArmendariz (approved):
LGTM

scottdickerson:
thanks @tay1orjones I made that change

PR #1867, opened 2020-12-07
fix(hierarchylist): change defaultselectedid onselect behavior

### LINKED ISSUE #1866: [HierarchyList] Prevent onSelect when defaultSelectedId changed
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Currently, when `defaultSelectedId` is changed or initially set, `onSelected` is called. However, as the name applies, `onSelected` should only be triggered by an user. This can cause issues for example when ´defaultSelectedId` of `HierarchyList` is updated and thus triggering `onSelected` which causes the need to catch such non-user triggered calls.

jonguenther:
Closes #1866

**Summary**

- change `HierarchyList` behvaior to not trigger `onSelect` if `defaultSelectedId` is changed

- separate `onSelect` callback from actual selection

- existing tests for `defaultSelectedId` should still work


scottdickerson (approved):
LGTM

davidicus (approved):

PR #1868, opened 2020-12-07
fix(dashboardeditor): pass along src for new uploads

tay1orjones:
**Summary**

- For new uploads the src was never passed along in the callback due to it not being in availableImages. This makes one minor change to only check card changes against available images if the image is not flagged as new.

- If you debug the imagecard story on a new upload, you'll see that the `src` is passed through whereas before it was not


scottdickerson (approved):
LGTM

davidicus (approved):

PR #1869, opened 2020-12-07
fix(dashboardeditor): keep state of images to upload

tay1orjones:
**Summary**

- Newly uploaded images need to be added to a local state and then passed back when onSubmit is called. This way the images are uploaded at once on save rather than during the editing process.


scottdickerson (approved):
LGTM

davidicus (approved):

PR #1882, opened 2020-12-11
1560/custom card gallery

### LINKED ISSUE #1560: [CardGalleryList] List variant with large items, pictograms

marcJV:
Closes #1560

**Summary**

- Added ability to have custom cars in CardGalleryList

- Added `@carbon/pictograms-react` dependency
- CardGalleryList now accepts an `icons` object to give custom icons to list items
- CardGalleryList now accepts custom item via `supportedCardTypes` with appropriate titles in `i18n`
- ~Missing icons will get the `<Analyze />` icon from pictograms-react~

- Check out https://deploy-preview-1882--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-experimental-cardeditor--with-no-card-defined-gallery-view-and-custom-cards and observe the new icons and text. Icons and i18n can accept cards, but will not be shown unless defined in supportedCardTypes as well.


package.json
```diff
@@ -133,6 +133,7 @@
     "@carbon/icons-react": "10.17.0",
     "@carbon/layout": "10.12.0",
     "@carbon/motion": "10.6.0",
+    "@carbon/pictograms-react": "^11.0.0",
```
tay1orjones, line 136: Let's pin this like the other Carbon dependencies. 
```suggestion
    "@carbon/pictograms-react": "11.0.0",
```
will need a `yarn install` after to regen the lock file.

src/components/CardEditor/CardGalleryList/CardGalleryList.jsx
```diff
@@ -78,8 +84,12 @@ const CardGalleryList = ({ supportedCardTypes, onAddCard, i18n }) => {
       items={supportedCardTypes.map((cardType) => ({
         id: cardType,
         content: {
-          value: mergedI18n[`cardType_${cardType}`] || cardType,
-          icon: iconTypeMap[cardType],
+          value: mergedI18n[cardType] || cardType,
+          icon: (
+            <div className={`${iotPrefix}--card-gallery-list__icon`}>
+              {mergedIcons[cardType] || <Analyze />}
```
tay1orjones, line 90: I think it's okay to mirror how we had it before and just use `mergedIcons[cardType]` without the `Analyze` fallback. 
```suggestion
              {mergedIcons[cardType]}
```

src/components/CardEditor/_card-editor.scss
```diff
@@ -31,3 +31,8 @@
     }
   }
 }
+
+.#{$iot-prefix}--card-gallery-list__icon > svg {
+  height: 34px;
+  width: 40px;
+}
```
tay1orjones, line 38: Will this contort icons that don't match the aspect ratio of the `@carbon/icons-pictogram` svgs?

tay1orjones (changes requested):
A couple minor things - otherwise looks good!

src/components/CardEditor/_card-editor.scss
marcJV: image

Tested out a few icons; Code16 and Basketball32 to test if it squashes or looks alright. Otherwise I could have it up to the developer to add their own style to the icons. Which way is more preferable?

tay1orjones: Looks fine to me - the style is always overridable if the consumer needs to. 👍

tay1orjones (approved):

src/components/CardEditor/CardEditor.jsx
```diff
@@ -72,6 +72,7 @@ const propTypes = {
    */
   onValidateCardJson: PropTypes.func,
   supportedCardTypes: PropTypes.arrayOf(PropTypes.string),
+  icons: PropTypes.objectOf(PropTypes.node),
```
scottdickerson, line 75: this is a minor point, but since you're specifically looking for keys that match the existing DASHBOARD_EDITOR_CARD_TYPES, it would be better to use the PropTypes.shape({}) to communicate this.  At least a comment indicating that each key should match one of the card types

src/constants/LayoutConstants.js
```diff
@@ -297,4 +297,6 @@ export const DASHBOARD_EDITOR_CARD_TYPES = {
   SIMPLE_BAR: 'SIMPLE_BAR',
   GROUPED_BAR: 'GROUPED_BAR',
   STACKED_BAR: 'STACKED_BAR',
+  LIST: 'LIST',
+  ALERT: 'ALERT',
```
scottdickerson, line 301: I think ALERT shouldn't be added, but instead should be treated as a 'CUSTOM' I didn't think we were going to explicitly support `ALERT` type cards

scottdickerson (changes requested):
a few questions before integration

scottdickerson (approved):
LGTM

PR #1893, opened 2020-12-29
fix(ValueCard): don't abbreviate if precision is set

### LINKED ISSUE #1892: [ValueCard] values not showing precision
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Precision can be set to 4, but the whole numbers are still being truncated. If precision is set, it should be assumed that the user wants to see all numbers

### To Reproduce

1. Go to ValueCard, medium size and add 2-3 attributes with precison: 4 and values > 1000
2. See that the value will be truncated to 1K

### Expected behavior

If I set precision to 4 I should be able to see 4 digits of information in whole numbers and decimal numbers.

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context

internal issue: https://github.ibm.com/wiotp/monitoring-dashboard/issues/1694

sstone2423:
Closes #1892

**Summary**

- If precision is set, the value will not be truncated. Instead, the font-size will adjust according to the length of the value to attempt to keep the value on 1 line and stay readable

- Update story to use long values for precision testing

- Go to /?path=/story/watson-iot-valuecard--medium-vertical-multiple in the netlify and use the knobs to try different large values with varied precisions. Font sizes should adjust according to the value length

### DIFF
--- a/src/components/Dashboard/__snapshots__/Dashboard.story.storyshot
+++ b/src/components/Dashboard/__snapshots__/Dashboard.story.storyshot
@@ -20962,7 +20962,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Dashb
                             color={null}
                           >
                             <span
-                              className="ValueRenderer__AttributeValue-f4stpz-1 ciICRO"
+                              className="ValueRenderer__AttributeValue-f4stpz-1 houHFr"
                               size="SMALLWIDE"
                               title="1709384.1 people"
                               value={1709384.1}
@@ -21419,7 +21419,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Dashb
                             color={null}
                           >
                             <span
-                              className="ValueRenderer__AttributeValue-f4stpz-1 ciICRO"
+                              className="ValueRenderer__AttributeValue-f4stpz-1 houHFr"
                               size="SMALLWIDE"
                               title="1709384.1 people"
                               value={1709384.1}
@@ -21876,7 +21876,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Dashb
                             color={null}
                           >
                             <span
-                              className="ValueRenderer__AttributeValue-f4stpz-1 ciICRO"
+                              className="ValueRenderer__AttributeValue-f4stpz-1 houHFr"
                               size="SMALLWIDE"
                               title="1709384.1 people"
                               value={1709384.1}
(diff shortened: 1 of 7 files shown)


bjornalm:
Looks like comfort level is truncated, or am I misunderstanding this?
image

sstone2423:
> Looks like comfort level is truncated, or am I misunderstanding this?

Odd, theres a few special rules for MEDIUMTHIN size and smaller. This might need a UX opinion on whether we should also use the font size fix on the smaller sizes as well

Here's the font size rules for the smaller cards for example:

```
if (isMini) {
    return 1;
  }
  if (isSmall) {
    return 2;
  }
```

JoelArmendariz (approved):
LGTM pending the UX question.

tay1orjones:
@CynZhang @dianatran18 could you provide guidance on the truncation UX concern above?

tay1orjones (approved):
same - looks good pending the ux concern

dianatran18:
I might need a little more context of the use case for this card from @CynZhang.

From what I would have expected, the labels should sit on top of the value, leaving the full width for the long number. 

Example: 

Screen Shot 2021-01-05 at 11 33 09 PM

(that's a hot average temperature)


Generally for really long numbers in smaller card sizes, we talked about abbreviating the number (for example: 300,000 down to 300K) instead of having the number wrap and shrink in size, as that would cause a lot of varying text sizes, and if you have a lot of varying sizes next to each other it starts to look a little odd. At that point, you're not even using the Carbon text styles right?

There's an exception for abbreviating numbers in small spaces [per IBM's guidance](https://www-03preprod.ibm.com/support/knowledgecenter/ibm_style/numbers.html#abbreviations), but overall, the reason why we bumped up the [card sizes](https://pages.github.ibm.com/ai-applications/design/components/card/usage#sizing) in the updated card design is to reduce the amount of abbreviations that could potentially happen, especially when you responsively resize your screen to be smaller.

sstone2423:
Oh that would make sense to move the label up. Here is the internal issue @dianatran18 @CynZhang 
https://github.ibm.com/wiotp/monitoring-dashboard/issues/1694

sstone2423:
Putting this PR in draft until the design conversation is complete

CynZhang:
thumbs up on @dianatran18 new layout proposal. 
@sstone2423 what would happen if the card width went smaller than the length of the number? Would it auto-scale the font size?

sstone2423:
@CynZhang The current fix would not adjust for card width, but would adjust for card size changes. It will continue to wrap if its a MEDIUM card or larger. 

Could you add a new design spec with the different card sizes that I could work off?

sstone2423:
Closing this PR in favor of creating a new one with the new design spec for ValueCard formatting

PR #1903, opened 2021-01-04
fix(timeseriescard): use dataSourceId as column name with dataFilter

### LINKED ISSUE #1902: [TimeSeriesCard] column name is incorrect when dataFilter is applied
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When `dataFilter` is enabled, the first found matchingDataSource's label is used for the column name, although there can be multiple values if there is more than one dataFilter

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-timeseriescard--data-filter
2. Click on isExpanded knob
3. See that the column is using the first dataFilter label, but not the second

### Expected behavior

Should be a general name as all of the values fall into a single column

### Environment/versions:

- OS: MacOS
- Browser:chrome

sstone2423:
Closes #1902

**Summary**

- If `dataFilter` is enabled, use the `dataSourceId` instead of the `label` as the `matchingDataSource` will only give the first match AND all dataFilters of the same dataSource will fall into the same column
- Updated all knobs using `isExpandable` to `isExpanded` as thats the actual prop name

- Go to /?path=/story/watson-iot-timeseriescard--data-filter , enable the  `isExpanded` knob and see that the `temperature` column name is `temperature` and values for both data filters are still showing


JoelArmendariz (approved):
LGTM

PR #1905, opened 2021-01-04
fix(TableCard): use table loading state rather than card

### LINKED ISSUE #1904: [TableCard] isLoading is using the card loading state rather than table loading state
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

If `isLoading` is set in TableCard, the Card loading state is used rather than the table

### To Reproduce

1. Go to any TableCard implementation
2. Add `isLoading` prop
3. See the card loading state

### Expected behavior

When isLoading is set on TableCard, use the table loading state

### Environment/versions:

- OS: MacOS
- Browser: chrome

sstone2423:
Closes #1904

**Summary**

- When `isLoading` is set in TableCard, set `view: { table: { loadingState: isLoading } }` instead of passing it to Card
- Also fixes a bug where `loadingState.rowCount` was not being respected

- Add `isLoading` knob to all TableCard stories
- Add `isLoading` story to TableCard

- Go to ?path=/story/watson-iot-tablecard--with-is-loading and see that the table loading state is being used and there are 7 rows
- Go to  /?path=/story/watson-iot-table--with-loading-state and see that 7 rows are shown


JoelArmendariz (approved):
LGTM

tay1orjones (approved):

PR #1918, opened 2021-01-06
fix(dashboardeditor): pass through pii props to the child components

scottdickerson:
Closes #

**Summary**

- All of these fixes are just to pass all the i18n props from the DashboardEditor layer down through to the underlying card components

- just passing and adding any missing i18n props down

- To verify it, look at the new i18n story for DashboardEditor and add and manage each of the card types to see the translated strings


src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -154,6 +156,30 @@ const DataSeriesFormItemModal = ({
   const { id, type } = cardConfig;
   const baseClassName = `${iotPrefix}--card-edit-form`;
 
+  const handleTranslation = useCallback(
+    (idToTranslate) => {
```
davidicus, line 160: can we align the use of these arguments. In CardEditFormContent.jsx we are using `id` and here we are using `idToTranslate`. I prefer the more explicit option.

davidicus (approved):
There is just the one comment about alignment of argument names but otherwise looks good.

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
scottdickerson: sure

sstone2423 (approved):

PR #1928, opened 2021-01-11
feat(hotspotmodal): integrate the hotspot modal into the ImageCard Editor

scottdickerson:
Closes #

**Summary**

- Enables the HotSpotEditorModal inside the ImageCardFormItems so you can edit the hotspots of an image

- feat(CardEditFormContent): pass the availableDimensions and dataItems to the ImageCardFormContent
- feat(ImageCardFormContent): add the HotspotEditorModal and local state to manage it
- fix(DashboardEditorCardRenderer): pass along all the iconProps so that testcases can see the title 
- feat(HotspotEditorModal): set default icons and colors to make it easier to use downstream, add 6 new icons to the selector
- fix(ImageCard): merge the default i18n with the passed i18n props
- chore(yarn): upgrade the caniuse version to stop the command line warnings

- To verify, go to the DashboardEditor story, add an ImageCard, select an image and then click 'Edit Image' to use the HotspotModalEditor. You should see the updates in the Action panel when saving the Card after modifying the hotspots


davidicus (approved):

sstone2423 (commented):
When swapping from the 'Tootip' and 'Datasource' tabs, theres a slight width change in the image. You can see it by swapping back and forth between the tabs

I don't really understand the x-coordinate and y-coordinate dropdowns. I would imagine those to be text inputs so that I could manually set the coordinates, but I'm only see data items as options. In addition, I can select 2 different data items for each coordinate which is confusing.

I'm not sure I understand the 'Labels' tab either. I created a hotspot, but theres nothing thats showing for Labels. 

If a hotspot is selected, and then you click the 'Labels' tab, it closes the tooltip instead of just opening the Labels tab

I'm not seeing any actions get called for saving or change the hotspots

scottdickerson:
@sstone2423 I see the slight width change but dont' think that's speciically related to this track can you open a separate issue?.  The others are sort of as designed

sstone2423 (approved):

PR #1929, opened 2021-01-11
fix(editor): value cards need to set precision at the attribute level

### LINKED ISSUE #1911: [ValueCardFormSettings]: the precision value should not be global across the value card, but set and managed per attribute
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

If you use the DashboardEditor to manage a ValueCard you'll notice that the precision is in the Settings pane and is global to the whole Value Card.  We need this to be specific for each attribute.

It should be listed here in the details pane for a given attribute
image

JoelArmendariz:
Closes #1911 

**Summary**

Value cards in the editor were setting decimal precision at the card level, rather than the attribute level.

- Move `decimalPrecision` dropdown to the `DataSeriesFormItemModal`

Make sure that when editing a `ValueCard` in the `DashboardEditor`, you can set the decimal precision per dataItem, and that the selection appears in the JSON config.


scottdickerson (approved):
LGTM

tay1orjones (approved):

PR #1931, opened 2021-01-11
1580 table show column totals for all data in a table

### LINKED ISSUE #1580: [Table] Show column totals for all data in a table
### Summary
NOTE: Pending PAL Design review.  Not yet ready for development.
A user can see the total of all the values in a table column (includes values on all pages).  A user can choose to show or hide the totals row.

**Additional context**
Proposed approach for showing totals
Interactive ptype: https://f0uxik.axshare.com/

- Option to show / hide total row - if user option, accessible from the overflow menu.
- Option to display the word "totals" - if shown, first column is not totaled.
- Totals row is frozen, rows scroll under it
- Thicker line to separate totals row from regular rows
- Totals in bold
- Numeric, non-date columns are totaled
- When user exports table data, totals are included

Question:
- Would the app team be able to choose which columns are totaled?  We would probably want a generic option in some cases, but select specific columns in other cases.

upcoming release?

This is scheduled ~~as a stretch goal for the October Health release.  It's more likely~~ to be part of the Jan release of Health.

date, the better!_

### Available extra resources

kevinsperrine:
Closes #1580 

**Summary**

- Added column aggregation to table footer

- feat(table): show column totals wip by bjorn
- feat(table): column aggregation styles, testing, snapshot updates
- fix(table): prettier on test helper file

-Check the "Aggregate Column Values" story in the Table folder.


davidicus:
@kevinsperrine Looking good. One point is that I think we need to have an option in the table action bar overflow menu to toggle this feature on and off.

src/components/Table/Table.jsx
```diff
@@ -692,6 +739,23 @@ const Table = (props) => {
...
+                  'hasRowSelection',
+                  'hasRowExpansion',
+                  'hasRowActions'
+                ),
+              }}
+              tableState={{
+                aggregations,
+                ordering: view.table.ordering,
+              }}
+              testID={`${id}-table-foot`}
```
davidicus, line 756: will want these to be `data-testid`

src/components/Table/Table.story.jsx
```diff
@@ -3812,4 +3812,27 @@ storiesOf('Watson IoT/Table', module)
         propTablesExclude: [StatefulTable],
       },
     }
-  );
+  )
```
davidicus, line 3815: Since we have so many Table stories let's see if we can add this feature to an existing story

src/components/Table/Table.jsx
kevinsperrine: @davidicus Sure thing. Can you explain the differences between when you all use `testID` as shown here: https://github.com/carbon-design-system/carbon-addons-iot-react/blob/next/docs/guides/testing.md#configuration versus using the full `data-testid` prop?

kevinsperrine:
> @kevinsperrine Looking good. One point is that I think we need to have an option in the table action bar overflow menu to toggle this feature on and off.

Can do. I wasn't sure if that was something we would handle internally, or leave that to the end-user since it can be controlled via the hasAggregation.

kevinsperrine:
@davidicus Just realized the overflow menu could go two ways. Should it be placed in the column overflow menu to toggle it on a per-column basis or in the TableToolbar and turn off the whole row?

src/components/Table/Table.story.jsx
kevinsperrine: > Since we have so many Table stories let's see if we can add this feature to an existing story

@davidicus Merged into the stateful table example with overflow menu, etc.

kevinsperrine:
@davidicus Just assumed this was for the TableToolbar and added the stuff necessary to toggle the aggregation row. See https://deploy-preview-1931--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--simple-stateful-example-with-column-overflow-menu-and-aggregate-column-values

src/components/Table/Table.jsx
davidicus: Hey Kevin. Sorry, I followed the props down. My initial review I thought you were just adding test id as attribute here. I see in the Table footer component you pass it down to an actual data-testid. You can ignore. Thanks

davidicus:
> @davidicus Just realized the overflow menu could go two ways. Should it be placed in the column overflow menu to toggle it on a per-column basis or in the TableToolbar and turn off the whole row?

I believe it should be for the whole table.

marcJV:
Code looks good! Could we have a knob (or another story if easier) that adds pagination? With pagination it looks like it is moved to the top like in the prototype here. 
image

kevinsperrine:
> Code looks good! Could we have a knob (or another story if easier) that adds pagination?

@marcJV Added. Should be available on the preview shortly.

marcJV (approved):

src/components/Table/Table.story.jsx
```diff
@@ -3781,7 +3781,7 @@ storiesOf('Watson IoT/Table', module)
     }
   )
   .add(
-    'Simple Stateful Example with column overflow menu',
+    'Simple Stateful Example with column overflow menu and aggregate column values',
```
sstone2423, line 3784: This is nitpicky, but could you remove the first three words from this to make it easier to read? Theres a ton of table stories and it can be difficult to find certain ones
```suggestion
    'with column overflow menu and aggregate column values',
```

sstone2423 (approved):
Looks good

davidicus (approved):

PR #1932, opened 2021-01-11
fix(BarChartCard): use series for sample data generation

sstone2423:
**Summary**

- `series` was not being used to generate sample data in the dashboard preview which was leading to data items not matching correctly in the generation logic

- Swapped sample data generation for dashboard previews to use series similar to the `isEditable` sample generation
  - This should really be rewritten to have a single generatedata function with a boolean for `isEditable` vs the `isPreview`
- Removed `hideHeader` boolean for custom cards as we shouldn't assume the user wants it
- Added a test to cover default card renders

- Go to any DashboardEditor story, add a BarChartCard, select a time range, select a groupBy, select a data item, then see that the data renders correctly


JoelArmendariz (approved):
LGTM

davidicus (dismissed):
Looks good but you are getting an error for not meeting threshold for testing. Once your test pass we can merge.

scottdickerson:
@davidicus can you rereview this today?

scottdickerson (approved):
LGTM

PR #1934, opened 2021-01-12
feat(hotspot): need to expose the onFetchDynamicDemoHotspots method to the DashboardEditor

scottdickerson:
Closes #

**Summary**

- just need to lift up the prop for onFetchDynamicDemoHotspots all the way out to the DashboardEditor component, so consumers can easily override it.  Also created a default implementation at the HotspotEditorModal level in case they haven't overridden it.
This PR depends on the other hotspot PRs to be integrated first.

- Verify that in the DashboardEditor story where we haven't explicitly set a onFetchDynamicDemoHotspots method that a default dynamic hotspot is returned in the center of the picture


src/components/CardEditor/CardEditForm/CardEditForm.jsx
```diff
@@ -191,6 +191,8 @@ const CardEditForm = ({
   currentBreakpoint,
   availableDimensions,
   testID,
+  // eslint-disable-next-line react/prop-types
```
JoelArmendariz, line 194: Is there a reason this doesn't need to be in the proptypes definition?

sstone2423 (approved):

src/components/CardEditor/CardEditForm/CardEditForm.jsx
scottdickerson: well, they're just kind of passthroughs, I documented them up at the top layer and bottom layer where they're used, but there are like 5 different levels of prop passthrough unfortunately

davidicus (approved):

PR #1936, opened 2021-01-12
fix(dashboardeditor): hide value card font size setting

sstone2423:
**Summary**

- Hide the font size setting for ValueCards in the dashboard editor until https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1892 is fixed

- Go to any DashboardEditor story and see that the ValueCard font size setting is hidden


JoelArmendariz:
I think maybe we should hide the tab completely, rather than having the option go to it and it being empty? I know it's temporary, but it could be strange to have an empty tab like that.

JoelArmendariz (approved):
Looks good!

scottdickerson (approved):
LGTM

PR #1938, opened 2021-01-12
feat(editor): add aggregation

JoelArmendariz:
**Summary**

The editor needs to be able to configure aggregation data for dataItems inside of each card.

Update the `DataSeriesFormItemModal` to be able to show and edit aggregators
Update `DataSeriesFormContent` to be able to handle aggregators

Go to the dashboard editor stories, add or edit dataItems inside the cards. Add and Edit the aggregation methods and grains, and confirm that the changes persist.
Make sure that previous editor features are still working properly and that the JSON is being updated.

Relevant Design Doc: https://ibm.invisionapp.com/share/JCO0QWFDX9H#/319800395_Summary_Dashboard


src/components/CardEditor/CardEditForm/CardEditForm.jsx
```diff
@@ -157,6 +159,44 @@ export const basicCardValidation = (card) => {
...
+/**
+ * removes properties needed for the editor that we don't want the user to be able to modify
+ * @param {Object} card JSON currently being edited
+ * @returns {Object} card with omitted attributes
+ */
+export const hideCardPropertiesForEditor = (card) => {
+  let attributes;
+  let series;
+  if (card.content?.attributes) {
+    attributes = card.content?.attributes?.map((attribute) =>
```
sstone2423, line 171: the null checks for `card.content.attributes` isn't need here since its already checked

src/components/CardEditor/CardEditForm/CardEditForm.jsx
```diff
@@ -157,6 +159,44 @@ export const basicCardValidation = (card) => {
...
+      omit(attribute, [
+        'aggregationMethods',
+        'aggregationMethod',
+        'grain',
+        'uuid',
+      ])
+    );
+  }
+  if (card.content?.series) {
+    series = card.content?.series?.map((attribute) =>
```
sstone2423, line 181: the null checks for `card.contentseries` isn't need here since its already checked

src/components/CardEditor/CardEditForm/CardEditForm.jsx
```diff
@@ -157,6 +159,44 @@ export const basicCardValidation = (card) => {
...
+      omit(attribute, [
+        'aggregationMethods',
+        'aggregationMethod',
+        'grain',
+        'uuid',
+      ])
+    );
+  }
+  return omit(
```
sstone2423, line 190: does this return work if `attributes` and `series` are both defined?

src/components/CardEditor/CardEditForm/CardEditForm.test.jsx
```diff
@@ -90,4 +97,58 @@ describe('CardEditForm', () => {
       expect(mockSetError).toBeCalledWith('1234 is not valid JSON');
     });
   });
+  describe('hideCardPropertiesForEditor', () => {
```
sstone2423, line 100: can you add a test for `series` and `attributes` both being hidden at the same time?

src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
```diff
@@ -122,17 +139,7 @@ const CardEditFormContent = ({
 
   const handleTranslation = useCallback(
```
sstone2423, line 140: Since the function is defined outside of the function, this useCallback and function definition is no longer needed

src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
```diff
@@ -99,11 +100,27 @@ const defaultProps = {
   dataItems: [],
   currentBreakpoint: 'xl',
   availableDimensions: {},
+  isSummaryDashboard: false,
   dataSeriesItemLinks: null,
 };
 
+export const handleTranslationCallback = (idToTranslate, mergedI18n) => {
```
sstone2423, line 107: This can just be called `handleTranslation`

src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
```diff
@@ -99,11 +100,27 @@ const defaultProps = {
...
+  const { openMenuText, closeMenuText, clearAllText } = mergedI18n;
+  switch (idToTranslate) {
+    default:
```
sstone2423, line 110: Can you move this default to the bottom of the switch?

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -144,20 +168,28 @@ const DATAITEM_COLORS_OPTIONS = [
...
 }) => {
   const mergedI18n = { ...defaultProps.i18n, ...i18n };
-  const { id, type } = cardConfig;
+  const { id, type, content } = cardConfig;
   const baseClassName = `${iotPrefix}--card-edit-form`;
 
+  const isTimeBasedCard =
+    type === CARD_TYPES.TIMESERIES ||
+    type === CARD_TYPES.TABLE ||
+    content?.type === BAR_CHART_TYPES.SIMPLE ||
```
sstone2423, line 190: I think you need to also check for `timeDataSourceId` for the bar charts

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -273,128 +402,159 @@ const DataSeriesFormItemModal = ({
...
+                  !availableDimensions[selectedDimensionFilter]),
+            })}>
             <Dropdown
-              id={`${id}_data-filter-value`}
+              id={`${id}_data-filter-key`}
               label=""
               translateWithId={handleTranslation}
               direction="bottom"
-              items={availableDimensions[selectedDimensionFilter]}
+              items={[mergedI18n.none, ...Object.keys(availableDimensions)]}
```
sstone2423, line 496: I've seen `mergedI18n.xxxx` be used as selectable items a lot which could cause issues when the user saves the dashboard. I think its good to translate the labels, but the underlying item that gets selected should always be the same so that we never have to worry about locales in the underlying data

I may be reading it wrong though

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
```diff
@@ -308,39 +326,85 @@ const DataSeriesFormItem = ({
...
+              id={`${cardConfig.id}_dataSourceIds-combobox`}
+              items={formatDataItemsForDropdown(validDataItems)}
+              itemToString={(item) => item.id}
+              titleText={mergedI18n.dataItemEditorDataItemTitle}
+              addToList={false}
+              placeholder={mergedI18n.filter}
+              // clears out the input field after each selection
+              selectedItem={{ id: '', text: '' }}
+              onChange={(selectedItem) => {
+                // ignore the value added by the "enter" keypress
```
sstone2423, line 368: What does this mean? Does enter not work then?

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
```diff
@@ -308,39 +326,85 @@ const DataSeriesFormItem = ({
...
+              key={`data-item-select-${removedDataItems.length}-selected_card-id-${cardConfig.id}`}
+              id={`${cardConfig.id}_dataSourceIds-combobox`}
+              items={formatDataItemsForDropdown(validDataItems)}
+              itemToString={(item) => item.id}
+              titleText={mergedI18n.dataItemEditorDataItemTitle}
+              addToList={false}
+              placeholder={mergedI18n.filter}
+              // clears out the input field after each selection
+              selectedItem={{ id: '', text: '' }}
+              onChange={(selectedItem) => {
```
sstone2423, line 367: Can you abstract this to its own function since theres so much happening there?

src/components/CardEditor/CardEditForm/CardEditFormItems/ValueCardFormItems/ValueCardFormSettings.jsx
```diff
@@ -43,6 +43,10 @@ const propTypes = {
 const defaultProps = {
   cardConfig: {},
   i18n: {
+    xAxisLabel: 'X-axis label',
```
sstone2423, line 46: It looks like a lot of these aren't needed here anymore

src/components/CardEditor/CardEditForm/_card-edit-form.scss
```diff
@@ -115,12 +115,23 @@
     align-items: flex-end;
     padding-bottom: 1rem;
     &--item {
+      .iot--color-dropdown {
+        margin-bottom: 1.25rem;
```
sstone2423, line 119: Use carbon style variables if possible

sstone2423 (changes requested):

sstone2423 (commented):
Can you create an issue in PAL and monitor to support the reorder / drag data items in the design?

src/components/CardEditor/CardEditForm/CardEditForm.jsx
JoelArmendariz: Any card will only have one or the other in the contents section, so one will always be undefined.

src/components/CardEditor/CardEditForm/CardEditForm.test.jsx
JoelArmendariz: This case wouldn't happen in our current card schemas.

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
JoelArmendariz: This check isn't actually that specific. It's just checking for those card types as they are. Not necessarily if they are currently time-based. Just able to be time based.

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
JoelArmendariz: Enter does work, but without this check, a duplicate item gets generated in the list. May be something to do with the underlying implementation of that ComboBox feature paired with how we're using it.

src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
JoelArmendariz: I left it this way to keep it consistent with the other places that similar translation functions are used. I think perhaps we should either use `useCallback` for all of them, or none of them.

src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
JoelArmendariz: I think this name is fine if we decide to keep these translation functions as `useCallback`s

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
JoelArmendariz: I have updated the dropdowns that would place the selected item into the json payload (the grains dropdown for example) to use the `{ id: '', text: '' }` convention so that the user can specify their own translation while we reference a source of truth with the `id`. Note: this required an update to the `aggregationMethods` property on the `dataItems` proptype to accomplish.

src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
sstone2423: Okay its just redundant code so we can refactor all the other spots later

sstone2423 (approved):

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
```diff
@@ -254,19 +265,23 @@ const DataSeriesFormItem = ({
...
   // determine which content section to look at
-  const dataSection =
+  const data =
     cardConfig.type === CARD_TYPES.TIMESERIES ||
     cardConfig.type === CARD_TYPES.BAR
       ? cardConfig?.content?.series
       : cardConfig?.content?.attributes;
 
+  // initialize items with a unique id if not present
+  const dataSection = data?.map((item) =>
```
davidicus, line 281: should this be momoized to not create new unique ids on each render?

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
```diff
@@ -377,54 +451,89 @@ const DataSeriesFormItem = ({
...
-                    dataItem.color ||
-                    DATAITEM_COLORS_OPTIONS[i % DATAITEM_COLORS_OPTIONS.length],
-                }}
-              />
-            ) : null,
+            icon:
+              cardConfig.type === CARD_TYPES.TIMESERIES ||
+              cardConfig.type === CARD_TYPES.BAR ? (
+                <div
+                  style={{
```
davidicus, line 458: Do we have to inline these styles? We should add classes and update custom css properties

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
```diff
@@ -377,54 +451,89 @@ const DataSeriesFormItem = ({
...
-                    setSelectedDataItems(
-                      filteredItems.map((item) => item.dataSourceId)
+              !isComplexDataSeries && [
+                <Button
+                  key={`data-item-${dataItem.dataSourceId}_edit`}
+                  renderIcon={Edit16}
+                  hasIconOnly
+                  kind="ghost"
+                  size="small"
+                  onClick={() => {
```
davidicus, line 477: This seems like a lot of logic for an inline function. Can we pull out into a function that we pass?

src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
```diff
@@ -377,54 +451,89 @@ const DataSeriesFormItem = ({
...
+                  iconDescription={mergedI18n.edit}
+                />,
+              ],
+              <Button
+                key={`data-item-${dataItem.dataSourceId}_remove`}
+                renderIcon={Subtract16}
+                hasIconOnly
+                kind="ghost"
+                size="small"
+                onClick={() => {
```
davidicus, line 517: same comment on pulling this into a function that we pass into component.

src/components/CardEditor/CardEditForm/_card-edit-form.scss
```diff
@@ -115,12 +115,23 @@
     align-items: flex-end;
     padding-bottom: 1rem;
     &--item {
+      .#{$iot-prefix}--color-dropdown {
+        margin-bottom: 1.25rem;
```
davidicus, line 119: we should use tokens

src/components/CardEditor/CardEditForm/_card-edit-form.scss
```diff
@@ -115,12 +115,23 @@
...
     &--item-half {
+      display: flex;
+      flex-direction: column;
       margin-right: 1rem;
       width: 50%;
+      &-label {
+        margin-bottom: 1rem;
+      }
+      &-content {
+        margin-bottom: 0.6rem;
```
davidicus, line 133: is it necessary to not use standard token size here?

davidicus (changes requested):
Just a few minor comments

src/components/CardEditor/CardEditForm/_card-edit-form.scss
JoelArmendariz: This was needed to line up the `read-only` form options with the dropdowns on the same horizontal line.

JoelArmendariz:
@davidicus, updated this PR to address your comments.

davidicus (approved):

PR #1941, opened 2021-01-15
feat(draganddrop): update react-dnd, refactor to use new provider

tay1orjones:
Closes #1772 
Closes #1670 

**Summary**

Fix a bug with `react-dnd` by updating the dependency and refactor how we use it.

**Change List (commits, features, bugs, etc)**

- update `react-dnd` to latest, remove other `react-dnd-*` deps that are no longer needed
- add new hook and component to provide a single shared react-dnd context/provider both in this library and for consumers
- refactor usages of DragDropContext to use the new component/hook
- refactor tests to wrap isolated components in the drag and drop context
- add new documentation about drag and drop

**Acceptance Test (how to verify the PR)**

- status checks should pass
- table view dropdown and drag and drop column reordering should now work [in the story](https://deploy-preview-1941--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-create-save-views)

https://user-images.githubusercontent.com/3360588/104786137-c2607180-5751-11eb-820f-69de5ef4a01a.mov

Issue #1737, opened 2020-10-29
[Breadcrumb] refactor hasOverflow to use implementation from Carbon


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

The Carbon team is updating the breadcrumb to have first-party support for the overflow menu. We'll need to make some changes to our `hasOverflow` implementation to take advantage of the feature now provided by the base component.

Note that this is _only support for the overflow menu_ - we'll need to retain our feature of automatically adding/removing items in to and out of the overflow menu on browser resize.

**Additional context** 
Carbon PR: https://github.com/carbon-design-system/carbon/pull/7085

### Specific timeline issues / requests

This will probably be contained as part of an upgrade of our Carbon dependencies.

Issue #1728, opened 2020-10-27
[FileUploaderItem] error props do not render


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When `invalid`, the `FileUploaderItem` doesn't display the `errorSubject` and `errorBody`.

This component is re-exported from Carbon unmodified by us. It looks like [they have fixed this bug in v10.21.0](https://github.com/carbon-design-system/carbon/pull/6867/files#diff-bb03860d794f52933ed278feedb6d4b5262d2966fa00aed81a2708c86a753446), we'll get it next time we update out Carbon deps #1702 

### To Reproduce

1. Go to [the storybook](https://carbon-addons-iot-react.com/?path=/story/fileuploader--fileuploaderitem)
1. Click to enable the invalid knob true
1. See that no error message is present, whereas [in Carbon it works](https://react.carbondesignsystem.com/?path=/story/fileuploader--file-uploader-item)

### Environment/versions:

- OS: macos
- Browser: chrome
- `carbon-addons-iot-react` version: `v2.139.0`

tay1orjones:
You can work around this issue by applying the fix from Carbon manually in your codebase until we update and publish the fix.

Issue #1703, opened 2020-10-19
[AccordionItemDefer] deprecate in favor of Carbon's deferComponentRender


tay1orjones:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Carbon v10.23.0 should contain https://github.com/carbon-design-system/carbon/pull/7079 which introduces a new HOC, `deferComponentRender`, that will replace the need for our `AccordionItemDefer`. The HOC can be used with any component to defer rendering on initial mount to prevent performance bottlenecks! 🎉 

- [ ] Modify our exports to export `deferComponentRender` from Carbon
- [ ] Ensure we pull through any related documentation on `deferComponentRender`
- [ ] Update our Accordion documentation (if not automatically pulled through via Carbon's storybook documentation)
- [ ] Modify `AccordionItemDefer` to use `deferComponentRender` under the hood
- [ ] Add a deprecation notice to `AccordionItemDefer` directing users to refactor to `deferComponentRender`

### Specific timeline issues / requests

It looks like this will probably ship in 10.23.0 which is slated for end of October/early November 2020.

### Available extra resources

We might roll this into the Carbon upgrade itself https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1702, which a member of the core team will work through.

kevinsperrine:
@tay1orjones Digging into this w/ regard to https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1943; however, it seems https://github.com/carbon-design-system/carbon/pull/7079 was simply closed and never merged.

tay1orjones:
Yeah this can be closed since the HOC didn't pan out.

PR #1943, opened 2021-01-19
chore(carbon): upgrade to the latest 10.26 and 7.26 version of carbon

### LINKED ISSUE #1702: [deps] upgrade to Carbon version ^10.23.0

scottdickerson:
Closes #1702

**Summary**

- Updates Carbon components to the latest version, unfortunately we can no longer import their stories and use them locally.

- (*.story): remove any local stories that were pointing to Carbon component stories as they no longer can be imported downstream
- *(.test) a few minor updates to testcases to handle new version of Carbon, for one any component rendering a Tab needs to mock the scrollIntoView dom method, need to workaround the double call onChange bug in TableToolbarSearch
- *.storyshots): it seems like they've removed the tabIndex = 0 by default from buttons

- Regression test the existing storybook stories to make sure they still function correctly


tay1orjones:
The related issue this PR should close is https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1702 - there are some line items there that can be addressed as part of the upgrade:

> Related items to be completed as part of this upgrade
> 
> - [ ] [[AccordionItemDefer] deprecate in favor of Carbon's deferComponentRender](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1703)
> - [ ] [[Breadcrumb] refactor hasOverflow to use implementation from Carbon](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1737)
> - [ ] [[FileUploaderItem] error props do not render](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1728)
>   - [ ] remove temporary fix in `_file-uploader.scss`
> - [ ] Overflow menu background fix introduced in https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1859 should be removed

Also in that issue I outlined the problem with Carbon story consumption you ran into here. I think the stories need to be copied from Carbon over into our codebase and kept current/updated when upgrading to further versions of Carbon. This could be automated via a script to be ran as part of the upgrade. Otherwise there won't be a way for consumers to find documentation for components that we re-export from Carbon with the specific version that we have pinned in package.json.

scottdickerson:
thanks @tay1orjones rather than copy the storybooks in the short term I was thinking of deferring it until we upgrade our storybook so that we can do the remote reference as the long term strategy.  Then just planning on upgrading the storybook ASAP

scottdickerson:
I'd like to undo the local workarounds but leave the larger refactors open and put them into next Sprint @davidicus

scottdickerson:
surprisingly we still needed the fix for #1859 even after the upgrade.

yarn.lock
```diff
@@ -17124,18 +17140,6 @@ react-dnd-html5-backend@11.1.3:
   dependencies:
     dnd-core "^11.1.3"
 
-react-dnd-test-backend@^11.1.3:
```
tay1orjones, line 17127: Thanks for this - I think this wasn't removed in my PR #1941 because of an issue with node-gyp on my machine.

src/components/HotspotEditorModal/HotspotEditorModal.test.jsx
```diff
@@ -241,9 +241,6 @@ describe('HotspotEditorModal', () => {
     // The initial dynamic hotspot title is showing
     const titleInputElement = screen.getByDisplayValue('dynamic test title');
 
-    // Click anywhere to remove focus the selected hotspot
```
davidicus, line 244: are these changes part of the carbon update?

src/index.js
```diff
@@ -191,8 +191,6 @@ export {
   OrderedList,
   OverflowMenu,
   OverflowMenuItem,
-  PageSelector,
-  Unstable_Pagination, // eslint-disable-line camelcase
```
davidicus, line 195: Does this have a new component that we can import or is the functionality now built into regular pagination?

src/components/HotspotEditorModal/HotspotEditorModal.test.jsx
scottdickerson: yes, they changed the text on their Tooltip component to Focus sentinel @davidicus

src/index.js
scottdickerson: whoops this is a good catch, thanks will push the fix @davidicus

package.json
```diff
@@ -3,9 +3,14 @@
...
   },
-  "workspaces": [
-    "packages/react"
-  ],
+  "workspaces": {
+    "packages": [
+      "packages/react"
+    ],
+    "nohoist": [
+      "**"
```
kevinsperrine, line 11: TL:DL; @storybook/addon-docs does hot hoist properly in lerna and gets stuck in the packages/react/node_modules which breaks everything. This prevents the hoisting and fixes the issue.

However, this is like using a sledgehammer on a finishing nail, but it was a seemingly never-ending cycle of adding individual packages to nohoist and running tests again. Things continued to fail in weird ways at various points in either webpack compilation or testing.

kevinsperrine:
@davidicus upgraded carbon to 10.27.0 and 7.27.0 respectively. Stripped out the mdx from the carbon stories and removed all the `@storybook/addon-docs` stuff from the previous commit. There was only one test (FlyoutMenu) that failed in the upgrade. The Tooltip now requires the onChange handler in some situations, so I added onChange as an optional prop and pass it down to Tooltip.

davidicus:
Change log can be found [here](https://github.com/carbon-design-system/carbon/releases/tag/v10.27.0)

davidicus (approved):

stuckless (commented):
I'm seeing lots of "red" changes to the public API snapshot.  Does that mean there are property API changes as part of this PR?

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -22074,83 +22229,33 @@ Map {
       "requireTitle": Object {
         "type": "bool",
       },
-      "wrapperClassName": Object {
-        "type": "string",
-      },
-    },
-  },
-  "PageSelector" => Object {
```
davidicus, line 22082: Experimental and was removed.

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -20424,14 +20527,12 @@ Map {
         ],
         "type": "oneOfType",
       },
-      "inline": Object {
```
davidicus, line 20427: This will set off warning of mismatched prop type

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -22964,14 +23015,11 @@ Map {
   },
   "Tab" => Object {
     "defaultProps": Object {
-      "href": "#",
```
davidicus, line 22967: Looks like they are moving some defaults to other areas of the functional component. 
image

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -23907,7 +23974,6 @@ Map {
   "TimePickerSelect" => Object {
     "defaultProps": Object {
       "disabled": false,
-      "hideLabel": true,
```
davidicus, line 23910: default is removed since they are deprecating the prop but is still available

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -24394,15 +24468,19 @@ Map {
     },
   },
   "DropdownSkeleton" => Object {
-    "defaultProps": Object {
```
davidicus, line 24397: another deprecated prop that is still available but default is removed.

davidicus:
> I'm seeing lots of "red" changes to the public API snapshot. Does that mean there are property API changes as part of this PR?

Hey @stuckless. Most of the deletions are changes in proptype from string to node which will not require any changes for consumers. There were some areas were they were deprecating props and removed them from defaults. Those will have to be handled. I tried to leave notes on areas I felt needed attention.

jkeen (approved):
LGTM 🚢

PR #1949, opened 2021-01-21
feat(imagecard): thresholds should change the icon for a hotspot

### LINKED ISSUE #1899: [HotspotEditorModal] improve how thresholds are managed
### What package is this for?

- [x] React
- [ ] Angular

### Summary
Improve card global threshold management or only use thresholds per hotspot which would mean that card global threshold management functions should be removed.

**Additional context** 
The thresholds can be managed per hotspot which is now the default behaviour when there is no `cardConfig.thresholds` prop available. Then all hotspots will be exported with their own thresholds like the ImageHotspot (and indirectly ImageCard) expects.

If the prop cardConfig.thresholds is present it will be used instead and existing + new thresholds will be placed there on Save. However, this leads to some strange situations in the UI, e.g. If both hotspot A and B use data source X and the user changes the threshold for X when editing A. What should happen?

I'm not sure why we need to support keeping the thresholds in the cardConfig.thresholds prop since the Monitor card documentation shows the thresholds per hotspot and the ImageCard and ImageHotspot components in PAL also expects the thresholds to be defined per hotspot. If we choose to only keep thresholds per hotspot then we can remove the `thresholdsHelperFunctions.js` and its usage, if we want to support the global thresholds we need to define how it should work for scenarios like above and implement them.

upcoming release?

date, the better!_

### Available extra resources

scottdickerson:
Closes #1899

**Summary**

- If a threshold is matched for an attribute the icon for the hotspot should be overridden, in addition, there was no code in the ImageCard actually using the cardConfig.thresholds so all the utility methods to keep them in-sync with attribute-level thresholds was removed

- fix(editorUtils): remove global threshold management code
- fix(HotspotEditorModal): remove global threshold management code
- refactor(thresholdHelperFunctions): delete the helper methods now that they're no longer required to set global thresholds
- feat(ImageCard): set default function to render Icons
- feat(ImageHotspots): add icon determination code that examines the thresholds

- Verify in the basic story for ImageCard that the second hotspot icon is overridden by the threshold


sstone2423 (approved):

davidicus (approved):

davidicus:
@scottdickerson Not sure why we have failures in the formatting. I tried to run locally and it said everything was formatted properly. I am going to merge in master and hopefully it builds properly.

PR #1951, opened 2021-01-21
fix(hotspoteditormodal): make the user hold CTRL key before adding hotspots

### LINKED ISSUE #1897: [HotspotEditorModal] only add the hotspot if the user holds CTRL key before clicking
### What package is this for?

- [x] React
- [ ] Angular

### Summary
1. New hotspots should only be added if the user clicks on the image while holding down the ctrl key. Currently any click will add a new hotspot which causes hotspots to be accidentally added.
2. The info text should be updated to reflect this
3. (possibly optional) There has been requests to see the coordinates next to the mouse when hovering the image so that a person defining dynamic hotspots would now what the x/y values should be.

upcoming release?

date, the better!_

### Available extra resources

scottdickerson:
Closes #1897

**Summary**

- Update the ImageHotspots component to only add the hotspot if the CTRL key is held when the mouse is clicked

- feat(HotspotEditorTooltipTab): update the infoText
- feat(HotspotTextStyleTab): update the infoText
- feat(ImageCard): keep track of whether ctrl key is active (this is to change the cursor), attach key listeners, flip the cursor only when CTRL is pressed, prevent the default contextMenu from popping in edit mode, only add the hotspot if the mouseevent has the ctrlKey to true

- Go to the hotspot editor modal story try and add a hotspot without pressing CTRL


sstone2423 (approved):

packages/react/src/components/ImageCard/ImageHotspots.test.jsx
```diff
@@ -307,8 +307,9 @@ describe('ImageHotspots', () => {
           selectedHotspots={[]}
         />
       );
-
-      userEvent.click(screen.getByAltText(testImageText));
+      userEvent.click(screen.getByAltText(testImageText), {
+        ctrlKey: true,
+      });
       expect(onAddHotspotPosition).not.toHaveBeenCalled();
```
davidicus, line 313: This test looks like it is clicking with the ctrl button pressed but the assertion is that `onAddHotSpotPosition` has not been called. Shouldn't that be the opposite?

scottdickerson: in the first case the ImageHotspots isn't editable

davidicus (approved):

PR #1958, opened 2021-01-21
fix(dashboardgrid): don't regenerate layout if one already exists

### LINKED ISSUE #1959: [DashboardGrid] adding a new card removes the existing layout
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

If a new card is added to a layout, it removes / regenerates the entire layout causing all other custom cards to go back to their default positions

### To Reproduce

1. Go to http://127.0.0.1:3000/?path=/story/watson-iot-experimental-dashboardeditor--default
2. Add 2 cards
3. Move them into custom positions
4. Add a new card
5. See that the layouts of the first 2 cards has been reset

### Expected behavior

Adding or removing cards should not regenerate the layout

### Environment/versions:

- OS: [e.g. MacOS, Windows]

sstone2423:
Closes #1959 

**Summary**

- Layouts were being re-generated if cards were not found in the layout which was not needed as react-grid-layout will regenerate the layouts itself

- Lower the test threshold because the removed branch was covered with tests, causing the total coverage to go down by 1-2% (its essentially the same test threshold)
- Removed styled component from DashboardGrid
- Removed `isArray(children)` check because the prop type is required to be an array already

- Go to any DashboardEditor story, add 2 cards with different titles (to help tell which is which), move one or both of the cards into new positions, then add a new card. The layout of the first 2 cards should remain the same. Continue to do so with more cards for further testing


JoelArmendariz (approved):
LGTM

packages/react/jest.config.js
```diff
@@ -21,9 +21,9 @@ module.exports = {
       lines: 80,
     },
     './src/components/Dashboard/DashboardGrid.jsx': {
-      statements: 73,
-      branches: 61,
-      lines: 73,
+      statements: 72,
```
davidicus, line 24: 🙈

davidicus (approved):

packages/react/jest.config.js
sstone2423: I know 😞

PR #1965, opened 2021-01-22
test: fix missing prop warnings

sstone2423:
Closes #

**Summary**

- This fixes a small chunk of our test warnings, but there are still a LOT more
- Need to review our tests as some of them have no real purpose but to fool the coverage into thinking its a tested component, such as testing that a default prop is defined

- Added WizardInline to jest ignore as it will be deprecated in the next release
- Added loading story for List

- builds should pass

### DIFF
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/TableCardFormItems/TableCardFormContent.test.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/TableCardFormItems/TableCardFormContent.test.jsx
@@ -24,6 +24,7 @@ const commonProps = {
   },
   onChange: jest.fn(),
   setSelectedDataItems: jest.fn(),
+  translateWithId: jest.fn(),
 };
 
 describe('TableCardFormContent', () => {
--- a/packages/react/src/components/List/ListHeader/ListHeader.test.jsx
+++ b/packages/react/src/components/List/ListHeader/ListHeader.test.jsx
@@ -1,37 +1,9 @@
-import React from 'react';
-import { render, screen } from '@testing-library/react';
-import '@testing-library/jest-dom/extend-expect';
-
 import ListHeader from './ListHeader';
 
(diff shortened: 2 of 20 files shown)


packages/react/src/components/List/ListHeader/ListHeader.test.jsx
```diff
@@ -1,37 +1,9 @@
-import React from 'react';
-import { render, screen } from '@testing-library/react';
-import '@testing-library/jest-dom/extend-expect';
-
 import ListHeader from './ListHeader';
 
 describe('ListHeader', () => {
```
scottdickerson, line 3: Why did we remove these testcases, they're valid ideas to verify I think

sstone2423: I just think we can write better tests than 'should render' as thats already tested by using the full List component with a header. I re-added the tests

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/TableCardFormItems/TableCardFormContent.test.jsx
```diff
@@ -49,7 +50,7 @@ describe('TableCardFormContent', () => {
     expect(screen.queryByText('temperature')).toBeDefined();
     fireEvent.click(screen.queryByText('temperature'));
     // the selection state of the box should be updated
-    expect(screen.queryAllByTitle('Clear all selected items')).toHaveLength(1);
+    expect(document.querySelector('[aria-label="Clear Selection"]')).toBeTruthy();
```
sstone2423, line 53: This is due to `translateWithId` being a mocked function

davidicus: Can this be scoped to the container?

davidicus (approved):

PR #1969, opened 2021-01-25
Include uuid with dataSourceIds in dashboard editor for line, bar, image, and value

JoelArmendariz:
**Summary**

We need to be able to keep track of dataItem edits based on a unique ID that can be passed to the consumer.
There will be a subsequent PR for Table cards.

 - Add new `dataItemId` prop for dataItems, so that we can keep track of the original dataItem name as the dataSourceIds will be uniquely generated and passed to the consumer.
 - Generate a unique version of the dataSourceIds for each item that is being used on cards in the editor

 - Add, remove, and update metrics in the editor and confirm that they are still applying changes


packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
```diff
@@ -258,9 +258,14 @@ const DataSeriesFormItem = ({
       ? cardConfig?.content?.series
       : cardConfig?.content?.attributes;
 
-  // initialize items with a unique id if not present
+  // initialize items with a unique dataSourceId if not present
   const dataSection = useMemo(
-    () => data?.map((item) => (!item.uuid ? { ...item, uuid: uuid.v4() } : item)),
+    () =>
+      data?.map((item) =>
+        item.dataSourceId?.length > 36 // TODO: check if there is uuid present in the dataSourceId string
```
scottdickerson, line 265: could we just compare the dataSourceId to the dataItem name to determine this?  if they're the same, then we should generate the UUID

JoelArmendariz: Good point. I will update this

scottdickerson (approved):
LGTM

davidicus (approved):

PR #1975, opened 2021-01-27
feat(rulebuilder): intial addition of rule builder

### LINKED ISSUE #1919: [Component Proposal] RuleBuilder editor
Part of #1445 

### What package is this for?

- [x] React
- [ ] Angular

### Summary

This component will live inside the rule builder page and provide the functionality to add rules and groupings of rules that can be saved as a 'filter'.

image

### Additional context
See how this part fits into rule builder flow in the [Invision prototype](https://ibm.invisionapp.com/share/2YO0PTTRHKU#/screens/319789254)

kevinsperrine:
Closes #1919 

**Summary**

- initial addition of the rule builder 

-addition of a RuleBuilderEditor component that allows nested rule and groups.

- checkout the story path here: `?path=/story/watson-iot-experimental-rulebuilder--rule-builder-editor`


packages/react/src/components/RuleBuilder/RuleBuilderHeader.jsx
```diff
@@ -0,0 +1,82 @@
...
+    addRule: 'Add rule',
+    addGroup: 'Add group',
+  },
+  groupLogic: 'ALL',
+  testID: 'rule-builder-header',
+};
+
+const propTypes = {
+  id: PropTypes.string.isRequired,
+  groupLogic: GroupLogicPropType,
```
davidicus, line 24: if we could get some comments for some of the less obvious props as to what they do

packages/react/src/components/RuleBuilder/RuleBuilderEditor.jsx
```diff
@@ -0,0 +1,143 @@
...
+  addRule,
+  filterRulesById,
+  findRulePathById,
+  generateRuleGroup,
+  getRuleByPath,
+  updateRuleAtPath,
+} from './utils';
+
+const propTypes = {
+  defaultRules: RuleGroupPropType,
```
davidicus, line 18: If we could get a small comment for the lest obvious props.

packages/react/src/components/RuleBuilder/RuleBuilderHeader.jsx
kevinsperrine: resolved with 7fac216

sstone2423 (changes requested):
Looks great. Could you add a basic story where theres only a single group at first? I'm a little bit confused on what the overall results are. Could you add some kind of story that shows the output or results similar to the invision so that we know it works for the use case?

sstone2423 (changes requested):
It looks like the `RuleBuilderEditor` need to be exported as well

packages/react/src/components/RuleBuilder/Rule.jsx
```diff
@@ -0,0 +1,234 @@
...
+});
+
+RuleGroupPropType.rules = PropTypes.arrayOf(PropTypes.oneOf([RulesPropType, RuleGroupPropType]));
+
+const propTypes = {
+  rule: PropTypes.oneOfType([RulesPropType, RuleGroupPropType]).isRequired,
+  columns: TableColumnsPropTypes.isRequired,
+  onAddRule: PropTypes.func.isRequired,
+  onRemoveRule: PropTypes.func.isRequired,
+  onChange: PropTypes.func.isRequired,
```
stuckless, line 46: What does the event look like?

packages/react/src/components/RuleBuilder/Rule.jsx
```diff
@@ -0,0 +1,234 @@
...
+export const RuleGroupPropType = PropTypes.shape({
+  id: PropTypes.string.isRequired,
+  groupLogic: GroupLogicPropType.isRequired,
+});
+
+RuleGroupPropType.rules = PropTypes.arrayOf(PropTypes.oneOf([RulesPropType, RuleGroupPropType]));
+
+const propTypes = {
+  rule: PropTypes.oneOfType([RulesPropType, RuleGroupPropType]).isRequired,
+  columns: TableColumnsPropTypes.isRequired,
```
stuckless, line 43: Why are we using the table column props here?  We should define our own properties for this.  Does the rule builder support the size property?  what properties do we expect to use here?

packages/react/src/components/RuleBuilder/RuleBuilderEditor.jsx
```diff
@@ -0,0 +1,146 @@
...
+  getRuleByPath,
+  updateRuleAtPath,
+} from './utils';
+
+const propTypes = {
+  // the rules passed into the component. The RuleBuilder is a controlled component, so
+  // this works the same as passing defaultValue to a controlled input component.
+  defaultRules: RuleGroupPropType,
+  // the columns of the table to be used in the column dropdown.
+  columns: TableColumnsPropTypes.isRequired,
```
stuckless, line 22: define the props that needed for this component

packages/react/src/components/RuleBuilder/RuleBuilderEditor.jsx
```diff
@@ -0,0 +1,146 @@
...
+  onChange: PropTypes.func,
```
stuckless, line 23: what does this event look like

stuckless (changes requested):
I don't feel like this component is very re-usable and seems to be solving a very specific and rigid use case.

The rule editor should accept more information around column metadata and make better decisions around what to render.  For example, the columns should be providing a "type" such as integer, double, string, date, time, datetime, at the very minimum and then use a specific field renderer for each type that can support type validation, etc.  The rule builder could accept a functional render callback that given a column it will return a render function for the given column, where the default is managed internally for the given types.

The allowed types should be configurable.  ie, while I listed the given types above, at a minimum, Maximo/Graphite has 15 types.

The list of operands, for equal, greater than, etc, should be a configurable list, per column.  ie, not every column is going to support greater than.  And what about Not Equal?  What about ranges? or In?  This list needs to be configurable per column, although there can be a default that can be passed in if you are not using per column lists.  Does "Less than" make sense for a text field?

When I use this component in storybook things flicker and jump around.  This component is so close to working in RTL mode (good job), but, there is 1 small issue with the tree lines that needs to be addressed.

All in all this is a good start, but, this component needs more flexibility to be used as a common component.

kevinsperrine:
@stuckless, Thanks. That's all very helpful info! I didn't have any specs to build from except for a design screenshot, and since I'm new to this project, I don't have any background knowledge on Graphite/Maximo. I'll talk all this over with @davidicus in our call today and make the necessary adjustments.

packages/react/src/components/RuleBuilder/utils.js
```diff
@@ -0,0 +1,123 @@
+export const findRulePathById = (arr, id, parentIndex = []) => {
```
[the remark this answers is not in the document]
kevinsperrine: @sstone2423 Sure thing. They're added in the newest push.

packages/react/src/components/RuleBuilder/RuleBuilderEditor.jsx
kevinsperrine: added an onChange action to the stories, so show the rule tree.

packages/react/src/components/RuleBuilder/RuleBuilderEditor.jsx
kevinsperrine: @stuckless added custom prop types instead of using the table columns. This also includes a render function and per-column operand values.

packages/react/src/components/RuleBuilder/Rule.jsx
kevinsperrine: @stuckless I don't know what you mean by "size property". I don't see that used in reference to the table anywhere. Can you point me in the right direction?

packages/react/src/components/RuleBuilder/RuleBuilderEditor.jsx
stuckless: you can document the onChange here, since, this is what people will see
```
/* Function accepting the current rule */
onChange: PropTypes.func
```
ie, don't be afraid to add comments to the code... they are useful ;)  The comment for coumns is fine, but event handled should be documented in the code as to what they accept.

packages/react/src/components/RuleBuilder/Rule.jsx
stuckless: My point is not that you should be using it but rather... are you using all the props from TableColumnPropTypes here, or just a couple.  ie, we should define just that props you are using.  I said "size" but I was thinking of "width" of just a random property on the column that you are likely not using (nor should you) but you define it as something that you accept.

packages/react/src/components/RuleBuilder/RuleBuilderEditor.jsx
kevinsperrine: Got it. It's honestly been a few years since using prop-types, so I'm a bit out of the game. I've been used to being able to define callbacks and their params in TypeScript. I'll go through and add some comments to more thoroughly document these events.

kevinsperrine: addressed in 2899f10 push.

davidicus (approved):
This component is still very much in flight so I think it is good to go as is. Since this will be used internally there is no need to export.

sstone2423 (approved):

PR #1979, opened 2021-01-28
Fix alertcard editor

scottdickerson:
Closes #

**Summary**

- Set of fixes related to the DashboardEditor needed to allow Alert Cards to be edited

- fix(CardEditFormContent): need to set the key to the card.id so the component is regenerated every time we switch cards and resets its state
- fix(DataSeriesFormContent): don't render if there are no data items to select (this is true for Alert Cards)
- fix(DashboardEditor): call the onCardChange as new cards are added so the consumers can change the default JSON
- fix(DashboardEditorCardRenderer): the callback function for custom cards was not being called correctly
- fix(index): export DASHBOARD_EDITOR_CARD_TYPES so other consumers can use them

- You can see the testcases have changed to cover the new fixes, you can also verify in CardEditor stories that if no data items are passed, the DataItems section doesn't render
- Verify in DAshboardEditor that as new cards are added, the onCardChange callback is called


davidicus (approved):

packages/react/src/components/Table/Table.story.jsx
```diff
@@ -2910,6 +2910,30 @@ WithLoadingState.story = {
   name: 'with loading state',
 };
 
+export const WithCustomTooltip = () => (
```
davidicus, line 2913: Are these changes meant to be included with this PR. Seem unrelated and adds another story. 😬

JoelArmendariz (approved):
LGTM

packages/react/src/components/Table/Table.story.jsx
scottdickerson: yeah I needed it to understand how to use Table tooltips :)

PR #1981, opened 2021-01-28
feat(SuiteHeader): Adding support for different types in some i18n properties in SuiteHeader component

blechner:
**Summary**

- Some changes in the way that data is passed to the `SuiteHeader` component will require that `i18n.surveyTitle` and `i18n.profileLogoutModalBody` be strings instead of functions. So, to keep the component backwards compatible, those properties now have proptypes `PropTypes.oneOfType([PropTypes.string, PropTypes.func])` instead of only `PropTypes.func`. And the component is now handling the properties differently depending on the type.

- Added a new test for `SuiteHeader` where `i18n.surveyTitle` and `i18n.profileLogoutModalBody` are passed as functions (besides the existing test where all `i18n` props are passed as strings) to increase coverage.


sstone2423 (approved):

bryancboyd (approved):

scottdickerson (approved):
LGTM

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -7554,10 +7570,6 @@ Map {
         "isRequired": true,
         "type": "func",
       },
-      "suiteName": Object {
```
davidicus, line 7557: These are breaking changes in the API?

davidicus (approved):

PR #1982, opened 2021-01-28
feat(editor): include onCardJsonPreview callback for the consumer

JoelArmendariz:
**Summary**
Need to include a callback for the consumer to make changes to the JSON config before it gets shown in the JSON editor.

 - Adds new prop `onCardJsonPreview`


scottdickerson (approved):
LGTM can you add some testcases that verify that the onCardJsonPreview callback is called with the correct parameters when the callback function exists?

scottdickerson:
@davidicus can you take a look at this one?

sstone2423 (approved):

davidicus (approved):

PR #1990, opened 2021-01-29
fix(cardeditform): custom cards do not yet support Settings panels in our DashboardEditor

scottdickerson:
Closes #https://github.ibm.com/wiotp/monitoring-dashboard/issues/1906

**Summary**

- we haven't yet built a way for Custom Card types to send along the Setting form they support, until we do, we should hide the Settings tab from CardEditForm

- fix(CardEditForm): hide the Settings tab for the Custom card type

- Verify in the DashboardEditor story that if you add a custom card there is no Setting tab showing in the Form

### DIFF
--- a/packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
+++ b/packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
@@ -204,11 +204,12 @@ class FilterHeaderRow extends Component {
                 <div />
               ) : column.options ? (
                 column.isMultiselect ? (
-                  <MultiSelect
+                  <MultiSelect.Filterable
                     key={columnStateValue}
                     className={`${iotPrefix}--filterheader-multiselect`}
                     id={`column-${i}`}
                     aria-label={filterText}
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
```diff
@@ -204,11 +204,12 @@ class FilterHeaderRow extends Component {
                 <div />
               ) : column.options ? (
                 column.isMultiselect ? (
-                  <MultiSelect
+                  <MultiSelect.Filterable
                     key={columnStateValue}
                     className={`${iotPrefix}--filterheader-multiselect`}
                     id={`column-${i}`}
                     aria-label={filterText}
+                    placeholder={column.placeholderText || 'Choose an option'}
```
davidicus, line 212: we should this with  all our i18n strings moving forward. It would save us from having to define big objets that muddy up the component and we wouldn't need to merge default with props.

davidicus (approved):

JoelArmendariz (approved):

PR #1993, opened 2021-02-01
fix(DashboardEditor): check min and max before applying the number input value

sstone2423:
Closes #

**Summary**

- When NumberInput has a min or max, the input value needs to be checked before applying the `onChange`

- Added a function to check if a number value is valid when optionally given a min or max
- Used the function to check if the number value is valid before applying the onChange

- Go to any DashboardEditor story, add a ValueCard, go to the settings tab in the CardEditForm, and type in valid and invalid numbers to see the input respond correctly. It should only save numbers that are valid (between 16 and 54). If an invalid number / string is given, it will show the invalid text indication and will not save the number


scottdickerson (approved):
LGTM

JoelArmendariz (approved):
LGTM

PR #1999, opened 2021-02-02
Fix(timeseriescard/barchartcard): support the decimalPrecision prop to control the y-axis and tooltip values

### LINKED ISSUE #1998: [DashboardEditor] Allow TimeSeries Cards to control whether they show the zero on x and y axis in their settings
### What package is this for?

- [x] React
- [ ] Angular

### Summary

As you add TimeSeriesCards to the DashboardEditor, you should be allowed to control whether the zero shows on the axes or not.  By default it probably shouldn't (which it currently is doing), 

To see this in action, add aa new Time Series Card from the Dashboard Editor and show the JSON, note the XAxis and YAxis options.  However there's no reference to them in the Settings menu

scottdickerson:
Closes #1998

**Summary**

Two major fixes here:
- To expose the includeZeroOn*Axis props in the Settings panel of the DataSeriesFormSettings for TimeSeriesCards
- To actually respect the decimalPrecision prop in BarChartCard and TimeSeriesCard that is being passed from the DashboardEditor

- fix(BarChartCard): listen for and use the decimalPrecision property when formatting the axis, the tooltip and the table card values
- feat(DataSeriesFormSettings): ask for the includeZeroOn*Axis property in the form
- fix(editorUtils): default the includeZeroOn*Axis property to false to match the propType for TimeSeriesCard
- fix(TimeSeriesCard): respect the decimalPrecision prop in the x-axis tooltip and table

- Add a new BarChartCard, TimeSeriesCard through the DashboardEditor control and set the decimal Precision prop on the Settings tab, you should see the tooltips and y-axis ticks show that precision level.
- In the TimeSeriesCard Settings tab, set the includeZeroOn*Axis properties and pop up the JSON Code preview to make sure they're being set correctly.


packages/react/src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -152,15 +154,25 @@ const BarChartCard = ({
...
 
   if (!isAllValuesEmpty) {
-    tableColumns = tableColumns.concat(
-      generateTableColumns(
-        timeDataSourceId,
-        categoryDataSourceId,
-        type,
-        uniqueDatasets,
-        i18n.defaultFilterStringPlaceholdText
+    tableColumns = tableColumns
```
davidicus, line 157: Should we memoize these functions?

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormSettings.jsx
```diff
@@ -118,6 +120,40 @@ const DataSeriesFormSettings = ({ cardConfig, onChange, i18n }) => {
           value={content?.decimalPrecision}
         />
       </div>
+      <div className={`${baseClassName}--input--toggle-field ${baseClassName}--input`}>
+        <span>{mergedI18n.includeZeroOnXaxis}</span>
+        <ToggleSmall
+          data-testid="includeZeroOnXaxis toggle"
```
davidicus, line 126: did we mean to have a space in the test-id?

davidicus (approved):
looks pretty good. just a few comments

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormSettings.jsx
scottdickerson: good point I'll remove

JoelArmendariz (approved):
LGTM

PR #2003, opened 2021-02-04
fix(hierarchylist): fire onSelect for defaultSelectedId

### LINKED ISSUE #2004: [HierarchyList] onSelect is not being called for `defaultSelectedId`
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

If `defaultSelectedId` is provided on initial render, the `onSelect` does not get called

### To Reproduce

1. Go to HierarchyList defaultSelectedId story locally
2. Add a breakpoint / log on the user provided `onSelect`
3. See that it doesn't get fired on initial render

### Expected behavior

Should always fire `onSelect` if a new item is selected

### Environment/versions:

- OS: MacOS
- Browser:chrome

sstone2423:
Closes #2004 

**Summary**

- `onSelect` was not being called if `defaultSelectedId` was used

- Add knobs to all stories for all relevant props
- Add a test to check if `onSelect` gets called when `defaultSelectedId` is set

- Unfortunately the `onSelect` action knob doesn't seem to work on initial render so the best way to test is it run locally, go to HierarchyList - defaultSelectedId story, add a breakpoint or console log for `onSelect` and see that it gets fired on the initial render

### DIFF
--- a/packages/react/src/components/List/HierarchyList/HierarchyList.jsx
+++ b/packages/react/src/components/List/HierarchyList/HierarchyList.jsx
@@ -258,6 +258,10 @@ const HierarchyList = ({
         if (!isEqual(selectedIds, [defaultSelectedId])) {
           // If the defaultSelectedId prop is updated from the outside, we need to use it
           setSelected(defaultSelectedId);
+
+          if (onSelect) {
+            onSelect(defaultSelectedId);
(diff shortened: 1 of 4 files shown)


sstone2423:
Closing because I guess this is by design

PR #2007, opened 2021-02-05
fix(HierarchyList): add isLargeRow prop

sstone2423:
**Summary**

- `isLargeRow` and `className` wasn't available to HierarchyList
- Hide the reordering modal if editing is disabled in the list as its just extra, unused elements

- adds `isLargeRow` and `className` prop to HierarchyList which is then passed to List
- removes HierarchyList from experimental category (not sure why the other PR didn't save)
- Conditionally renders the list reorder modal dependent on whether `editingStyle` is given

- Go to any HierarchyList story and turn the `isLargeRow` knob to true


davidicus (approved):
This looks good. Not quite sure how the Rulebuilder test was changed but it is failing now. I know it's not part of this change set

scottdickerson (approved):
LGTM don't know what's up with the failure below being tagged in this PR

PR #2008, opened 2021-02-05
feat(suiteheader): Support for loading states and extensibility for Help, Profile and Applications menu items

blechner:
**Summary**

- loading states for the `SuiteHeader` component - shows skeleton components when data is not available in the header menus.
- extensibility for the `SuiteHeader` component - adds supports for custom help menu items, custom profile links and custom applications in the app switcher

- tests have been created to test loading states
- an existing story has been extended to include custom entries in the SuiteHeader component action items

Loading states:

image

image

image

Extensibility:

image

image

image

### DIFF
--- a/packages/react/src/components/SuiteHeader/SuiteHeader.jsx
+++ b/packages/react/src/components/SuiteHeader/SuiteHeader.jsx
@@ -356,9 +356,21 @@ const SuiteHeader = ({
                           },
                           content: mergedI18N.about,
                         })
-                    : Array.from({ length: 6 }, (x, i) => ({
-                        content: <SkeletonText key={`$help-skeleton-${i}`} />,
-                      })),
+                    : [
+                        {
+                          metaData: {
+                            element: 'div',
+                          },
+                          content: (
+                            <div
+                              className={`${settings.iotPrefix}--suite-header-help--loading`}
+                              data-testid="suite-header-help--loading"
+                            >
+                              <SkeletonText paragraph lineCount={6} />
+                            </div>
+                          ),
+                        },
+                      ],
                 },
                 {
                   label: 'user',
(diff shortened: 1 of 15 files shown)


cgirani (approved):

packages/react/src/components/SuiteHeader/SuiteHeader.jsx
```diff
@@ -29,7 +31,7 @@ const ROUTE_TYPES = {
   SURVEY: 'SURVEY',
 };
 
-export const SuiteHeaderRoutePropTypes = PropTypes.shape({
```
davidicus, line 32: How come you are removing the `PropTypes.shape`? Setting it this way will throw an error, will it not? I believe proptypes need to be a function.

davidicus: Never mind I see you are adding the shape below.

blechner: @davidicus I did this because I saw other examples where the proptype objects are exported as plain objects, and then the component importing it would define its own proptypes as `PropTypes.shape(ImportedObjectFromAnotherComponent)`. [Here](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/next/packages/react/src/components/Header/Header.jsx#L18), for example. If this is not a good practice, I can revert the change and export the objects as the return value of `PropTypes.shape()`. Besides, I am not sure why this particular proptype is being exported, nobody imports it.

blechner: Oh, I forgot to answer your question, it doesn't throw errors when the main component proptypes object is defined because when I removed the `PropTypes.shape()` from te object being exported, I also added the `PropTypes.shape()` when defining the current component's propTypes object:
It was like this before:
```
  /** URLs for various routes on Header buttons and submenus */
  routes: SuiteHeaderRoutePropTypes.isRequired,
```
And now it is defined like this:
```
  /** URLs for various routes on Header buttons and submenus */
  routes: PropTypes.shape(SuiteHeaderRoutePropTypes),
```

davidicus: No, it is fine. I just had missed where you added the `PropTypes.shape` further down the file

davidicus (approved):

PR #2012, opened 2021-02-08
fix(HierarchyList): scroll only if needed and support secondary nodes

sstone2423:
**Summary**

- Scrolling was always happening on item selection regardless of whether the item was already in view or not causing some weird jumping
- `secondaryValue` was not accepting nodes in the list

- Added the `scroll-into-view-if-needed` package to check if the node is visible or not, before attempting to scroll to it
- Allowed `secondaryValue` to accept nodes
- Added string type check for searching secondary values
- Updated 1 HierarchyList story to use a custom component as `secondaryValue`

- Go to /?path=/story/watson-iot-hierarchylist--with-default-selected-id in chrome, firefox, and safari and ensure the item is scrolled to on initial render, then continue to select other items and ensure no scrolling occurs
- Go to /?path=/story/watson-iot-hierarchylist--with-mixed-hierarchies, open the 'Tasks' category, and see that the InlineLoading component is successfully rendering, then use the search bar and ensure searching still works (this ensures I didn't break the search by adding the node)


packages/react/src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -206,7 +208,7 @@ const HierarchyList = ({
 }) => {
   const [expandedIds, setExpandedIds] = useState(defaultExpandedIds);
   const [searchValue, setSearchValue] = useState('');
-  const [filteredItems, setFilteredItems] = useState(cloneDeep(items));
+  const [filteredItems, setFilteredItems] = useState(() => cloneDeep(items));
```
scottdickerson, line 211: This is very smart!

scottdickerson (approved):
LGTM

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
davidicus: I definitely think it's good to instantiate the state this way but what is the reason for the cloneDeep? Especially now.

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -40,7 +41,7 @@ const propTypes = {
   /** Buttons to be presented in List header */
   buttons: PropTypes.arrayOf(PropTypes.node),
   /** ListItems to be displayed */
-  items: PropTypes.arrayOf(PropTypes.any).isRequired,
+  items: PropTypes.arrayOf(PropTypes.shape(ListItemPropTypes)),
```
davidicus, line 44: This is a breaking change of an item that we just took out of experimental. We should make this an enum with the deprecation warning for the old way included. I see it's no longer required so what behavior is changed by this new shape?

davidicus (changes requested):
We need continue to support the old prop type since this will be in a minor release

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
sstone2423: This is to keep the prop items intact as they'll get mutated a bunch inside of the component

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
sstone2423: This enforces the same prop types as the List below expects so the `any` was always wrong and would throw prop types warnings regardless

davidicus: Yeah I see down in the list now. Thanks.

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
sstone2423: Removed as its actually not needed

davidicus (approved):

PR #2015, opened 2021-02-10
fix(valuecard): precision is at the attribute level not card level

scottdickerson:
Closes #

**Summary**

- Value card was ignoring the precision value passed in all cases, and unfortunately it was being passed at the card level rather than the attribute level.

- fix(Attribute): move precision to the Attribute level
- fix(ValueCard): remove default precision property at the card level
- fix(CardPropTypes): add precision to the Attribute section

http://127.0.0.1:3000/?path=/story/watson-iot-valuecard--medium-2
- If you go to the Value card story for Medium 2 and change the precision setting inside the attribute, you'll see it get updated, this doesn't work in the previous build of addons

### DIFF
--- a/packages/react/src/components/ValueCard/Attribute.jsx
+++ b/packages/react/src/components/ValueCard/Attribute.jsx
@@ -15,6 +15,8 @@ const propTypes = {
   attribute: PropTypes.shape({
     label: PropTypes.string,
     unit: PropTypes.string,
+    // decimal precision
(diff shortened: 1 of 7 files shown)


JoelArmendariz (approved):
LGTM

sstone2423 (approved):
Good catch!

davidicus (approved):

PR #2017, opened 2021-02-10
fix(Hierarchylist): style, i18n, selection

sstone2423:
**Summary**

- Merge i18n prop with the default props to ensure all get passed through
- Add `hasDeselection` prop to remove the ability to deselect an item once its selected
  - When `hasDeselection` is false, don't allow deselection and don't refire `onSelect`
- Padding on sides should be 16px instead of 8
- Padding should be reduced when an item is select to account for the selected border

- Added tests and story knobs for new prop
- Added knob for search i18n

- Go to /?path=/story/watson-iot-hierarchylist--stateful-list-with-nested-searching , select an item, then try to select it again. It should not get deselected and it should not fire `onSelect`
- Go to the same story and attempt to edit the i18n search prop. You should see the changes live in the search bar


scottdickerson (approved):
LGTM

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -87,6 +90,7 @@ const defaultProps = {
   hasSearch: false,
   hasPagination: true,
   hasMultiSelect: false,
+  hasDeselection: false,
```
davidicus, line 93: We are changing the default  behavior for this component which could lead to breaks for consumers. Can we rework this so the old behavior is maintained? `hasNoDeselection` seems gross but could be an alternative. Otherwise we need to default this to true and add to our V3 issues checklist to change this to false in the new version.

davidicus (changes requested):
Just the one comment on the breaking change. Thank goodness we removed this from experimental 🙄

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
sstone2423: Yeah I'm not a fan of `hasDeselection` as a name either honestly. This was a mistake, meant it to be defaulted to true

packages/react/src/components/List/HierarchyList/HierarchyList.story.jsx
```diff
@@ -120,6 +124,7 @@ export const WithDefaultSelectedId = () => (
       isLoading={boolean('isLoading', false)}
       isLargeRow={boolean('isLargeRow', false)}
       onSelect={action('onSelect')}
+      hasDeselection={boolean('hasDeselection', false)}
```
davidicus, line 127: since the default is true the story should as well.

davidicus (commented):
Just the one change for the story and this should be good to go.

davidicus (approved):

PR #2018, opened 2021-02-10
feat(SelectUsers) Add SelectUsers modal

### LINKED ISSUE #1924: [Component Proposal] Select Users Modal
Part of #1445 

### What package is this for?

- [x] React
- [ ] Angular

### Summary

A new component that will allow a user to pick from a list to be added to another inside a modal.
image

### Additional context
See how this part fits into rule builder flow in the [Invision prototype](https://ibm.invisionapp.com/share/2YO0PTTRHKU#/screens/319789254)

jkeen:
Closes #1924 

**Summary**

Created `SelectUsers` component, allowing nested selection of users and groups for integration into the rule builder flow

- Added `SelectUsers` component
- Added some basic tests

It was unclear what the requirements and expectations for user input are, so some assumptions were made re: when groups should be selectable (assumption: top-level groups/categories are not selectable unless it's a flat list of users). That logic is pretty straightforward to change,though and depending on requirements we could even pass in some functions that determine whether or not to enable/disable the row or show the actions.


davidicus:
@dianatran18 UX question. In [this](https://deploy-preview-2018--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-selectusersmodal--default-modal) story we have different groups (roles/teams). What should we do if an individual in a group is added and then later on the entire group added? If we delete the group (with the individual inside) do we pull that person out individually as well? If a person is in two groups and one group is deleted do we delete the person from inside the second group? I think we need some guidance on how to handle these scenarios. Perhaps something we can discuss in next office hours?

davidicus:
@jkeen I would move this to an experimental section of the storybook in so we can still merge while we iron out the UX.

packages/react/src/components/SelectUsersModal/SelectUsersModal.jsx
```diff
@@ -0,0 +1,266 @@
...
+    selected: PropTypes.string,
+    addUser: PropTypes.string,
+    removeUser: PropTypes.string,
+    recent: PropTypes.string,
+  }),
+  users: PropTypes.arrayOf(GroupShape).isRequired,
+  initialSelectedUsers: PropTypes.arrayOf(GroupShape).isRequired,
+
+  /** Should the dialog be open or not */
+  open: PropTypes.bool, // eslint-disable-line react/boolean-prop-naming
```
davidicus, line 42: the convention we have for props like this now is to ad the prefix `is` so `isOpen`. This will probably allow you to remove the disable eslint comment.

packages/react/src/components/SelectUsersModal/_select-users-modal.scss
```diff
@@ -0,0 +1,46 @@
...
+  .bx--modal-container .bx--modal-content {
+    padding-right: $spacing-05;
+  }
+
+  .#{$iot-prefix}--list {
+    background-color: #f4f4f4;
+    height: 100%;
+  }
+  .#{$prefix}--search-input {
+    background-color: #fff;
```
davidicus, line 13: should be using carbon color tokens with the gray10 theme. Can find tokens [here](https://www.carbondesignsystem.com/guidelines/color/usage)

packages/react/src/components/SelectUsersModal/_select-users-modal.scss
```diff
@@ -0,0 +1,46 @@
+@import '../../globals/vars';
+
+.#{$iot-prefix}--select-users-modal {
+  .bx--modal-container .bx--modal-content {
+    padding-right: $spacing-05;
+  }
+
+  .#{$iot-prefix}--list {
+    background-color: #f4f4f4;
```
davidicus, line 9: color token

davidicus (changes requested):
Just a few minor comments. Can we log out the object that is passed when we hit submit? I am not seeing the returned object in the action panel just the event.

packages/react/src/components/SelectUsersModal/SelectUsersModal.jsx
```diff
@@ -0,0 +1,266 @@
...
+        if (isInSelected || isChildOfSelected || isCategory) {
+          return null;
+        }
+
+        return () => [
+          <Button
+            key={`${username}-list-item-button-${depth}`}
+            style={{ color: 'black' }}
+            role="button"
+            aria-label="Add"
```
scottdickerson, line 169: this is translatable content, probably should come from the mergedI18n

packages/react/src/components/SelectUsersModal/SelectUsersModal.jsx
```diff
@@ -0,0 +1,266 @@
...
+      renderDisabledState: () => false,
+      renderRowActions: ({ user, depth }) => {
+        const { username } = user;
+
+        if (depth === 0) {
+          return () => [
+            <Button
+              key={`${username}-list-item-button-${depth}`}
+              style={{ color: 'black' }}
+              aria-label="Remove"
```
scottdickerson, line 193: translatable content again

packages/react/src/components/SelectUsersModal/SelectUsersModal.jsx
```diff
@@ -0,0 +1,266 @@
...
+            />,
+          ];
+        }
+
+        return null;
+      },
+    });
+  };
+
+  const flattenUsers = (results, user) => {
```
scottdickerson, line 209: I'm not sure how many users you're thinking of managing but you might need to memoize and useCallback if you're rendering a bunch

scottdickerson (approved):
LGTM with aa few minor comments

davidicus (approved):

kevinsperrine:
@davidicus This was merged into #2066 with a few minor adjustments for integration. I'll note them in that PR.

PR #2026, opened 2021-02-11
fix(repo): adding a 'browser' field to avoid webpack default config issue

### LINKED ISSUE #2025: [REPO] add "browser" entry point to our package.json
Some consumers are being served the wrong package because of webpack's [default](https://webpack.js.org/configuration/resolve/#resolvemainfields) resolution in their config which prioritizes `module` over `main`.

image

Adding the `browser` field should resolve this issue for consumers without them having to modify their configs

davidicus:
Closes #2025

**Summary**

- Adding `browser` entry point will let webpack bundler's resolve our package properly in development

- added browser entry to package.json


scottdickerson (approved):
LGTM

cal-smith (approved):
Given that this has been tested a few times this makes sense to me 👍

PR #2036, opened 2021-02-12
Release 2.143.0

### LINKED ISSUE #2028: [Release] February - 2.143.0
Merge next into master to trigger 2.143.0 release on @latest

- [x] merge `next` into `master `to publish the next version of our main channel.
- [x] merge `master` into `next` to properly bump up the version of our next channel
- [x] Create an overview of the features and bugs that were completed during the sprint of both core and outside contributors.
    - https://github.com/carbon-design-system/carbon-addons-iot-react/pulls?q=is%3Apr+is%3Aclosed+merged%3A2021-01-01..2021-02-12
    - #2027 was the last commit pulled in

davidicus:
Closes #2028 

**Summary**

Bring in all the changes from the `Next` development branch into our `Master` branch. Please DO NOT MERGE. I will merge this PR in and update the next branch the night of the final approval. 

- pulled in `next`
- fixed a memory leak in one of our stories

- Check that stories all look fine
- Check that project still builds


scottdickerson (approved):
LGTM

jkeen:
Shouldn't most of these commits be git moves/renames and not git deletions + git additions? Seems like switching to a monorepo shouldn't wipe out all that commit history along with the ability to do git blames, should it?

jkeen:
Also, should #2039 be added to this release? The problem described in there (which, I would love some verification to make sure it isn't a just-me issue) if released is going to encourage more people to `--no-verify` around those pre-commit hooks, because they consistently fail from not being able to find the changed files in the monorepo setup

davidicus:
> Shouldn't most of these commits be git moves/renames and not git deletions + git additions? Seems like switching to a monorepo shouldn't wipe out all that commit history along with the ability to do git blames, should it?

I did notice that git blame was missing! good catch.

cal-smith:
> > Shouldn't most of these commits be git moves/renames and not git deletions + git additions? Seems like switching to a monorepo shouldn't wipe out all that commit history along with the ability to do git blames, should it?
> 
> I did notice that git blame was missing! good catch. @cal-smith any ideas on how to easily reconcile this?

I don't know there's an easy way to fix this at this point (though I'm all ears!). As far as I can tell git itself has no notion of what a move/rename is, `git mv` is just a convince to represent a system level `mv` + `git add newfile; git rm oldfile`. In this case, due to the significance of the changes happening I _believe_ we're seeing the rename/move auto-detection break down in the GitHub UI (and maybe git itself). The changes to the files _should_ still be tracked (since git doesn't actually track files), but it does make navigating the history harder.

Again, if anyone has ideas for how we can resolve this without blowing away the history we've accrued since the changeover I'm 100% on board!

jkeen:
@cal-smith @davidicus I dug a little deeper and did some testing on a branch I made. Renames and add/deletions do show up differently in the github PR diff (the moves show up as single line renames without showing file contents), but in either case when looking at the history of the file on Github things get cut off at the commit where the file got moved ([example](https://github.com/carbon-design-system/carbon-addons-iot-react/commits/release-2-143-0/packages/react/src/components/AddCard/AddCard.jsx)). It looks like git does the right thing internally if someone doesn't specify `git mv` and just moves them, though, because in both cases it looks like git blame works ([example](https://github.com/carbon-design-system/carbon-addons-iot-react/commits/release-2-143-0/packages/react/src/components/AddCard/AddCard.jsx))

On the command line it turns out that `git log --follow /path/to/file/name` will show the log entries _including_ moves. Did not know about --follow until now. There's also an advanced option in GitLens for VSCode that enables that option for file history in there, too.

So in short, I think we're good on this branch as-is? Sorry to raise the alarm bells

cal-smith:
> So in short, I think we're good on this branch as-is? Sorry to raise the alarm bells

No need to apologize! There's a few things we're going to try to get the file history back, which would be nice. It's good to know that at the very least `blame` and `log` already handle the renames reasonably well :+1:

jkeen (approved):

PR #2039, opened 2021-02-12
fix(build): keep lint-staged and husky definition on the top level to avoid pre-commit problems

### LINKED ISSUE #2037: [Build] pre-commit step `lint-staged` fails with current monorepo setup
### What package is this for?

- [x] React

### Describe the bug

`lint-staged` fails when committing changed files with an error similar to: 

`[error] No files matching the pattern were found: "/path/to/carbon-addons-iot-react/packages/react/packages/react/src/components/SelectUsersModal/SelectUsersModal.jsx"`

Notice the duplicate `packages/react` part of the path. 

### To Reproduce

1. Modify a file in the source tree
2. `git add` that file to stage it
2. Try to commit, letting the pre-commit hooks run
3. Notice the [error] and failing state of the pre-commit hook

### Environment/versions:

- OS: MacOS, Node 14.15.4

### Additional context
image

jkeen:
Closes #2037 

**Summary**

The current setup kept erroring/failing on pre-commit hooks, complaining that it couldn't find a file at an invalid path. According to `lint-staged` docs, the setup for monorepos should only be on the root level, and we've got specifications in package.json at the root as well as in `packages/react`

Remove duplicate setup for inner package file. 

1. Modify a file in the source tree, adding a line that will fail lint, e.g. `let foo = 'bar'`
2. git add that file to stage it
3. git commit -m "fix(build): test that this works", letting the pre-commit hooks 
4. Commit should fail from bad lint. 
5. Comment out the line we added in step 1 : `// let foo = 'bar'`
6. git commit -m "fix(build): test that this works", letting the pre-commit hooks 
7. Commit should succeed


jkeen:
I did a fresh clone of my repo and this problem disappeared. If that sounds weird, I agree, and I would love to dedicate a month to figuring out why that happened. Alas, no.

PR #2043, opened 2021-02-16
feat(dashboardeditor): improve performance while editing

### LINKED ISSUE #2000: [DashboardEditor] If I have lots of cards with lots of datapoints that are timeseriescards entering a title is slow
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

To see the 🐛 you need to create a Dashboard in the DashboardEditor with lots of TimeSeriesCards then try and edit one of their titles.  The rerender cycle causes major performance issues

scottdickerson:
Closes #2000

**Summary**

- summary_here

- fix(dataSeriesFormContent): if the field is supposed to end up in the groupBy section, don't set a unique dataSourceId
- feat(DashboardEditor): allow resizing of cards inside Dashboard Editor
- refactor(DashboardEditor): move most of the card decisions down to DashboardEditorCardRenderer so that I can memoize and greatly improve performance
- fix(TimeSeriesCard): add memoization to prevent the carbon charts code from rerendering
- fix(BarChartCard): add memoization to prevent the carbon charts code from rerendering

- Inside the DashboardEditor component you should be able to resize cards
- If you have lots of cards inside your DashboardEditor, you should be able to quickly edit the title of one of them (before the performance was too slow)

### DIFF
--- a/packages/react/src/components/DashboardEditor/DashboardEditor.jsx
+++ b/packages/react/src/components/DashboardEditor/DashboardEditor.jsx
@@ -2,7 +2,6 @@ import React, { useState, useEffect, useMemo, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import { InlineNotification, SkeletonText } from 'carbon-components-react';
 import classnames from 'classnames';
-import { Warning24 } from '@carbon/icons-react';
 
 import { settings } from '../../constants/Settings';
 import { DASHBOARD_EDITOR_CARD_TYPES, CARD_TYPES } from '../../constants/LayoutConstants';
@@ -16,8 +15,7 @@ import {
   getDuplicateCard,
   renderBreakpointInfo,
   DataItemsPropTypes,
-  validThresholdIcons,
-  validHotspotIcons,
+  renderDefaultIconByName,
 } from './editorUtils';
 
 const { iotPrefix } = settings;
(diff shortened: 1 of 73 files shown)


packages/react/src/components/DashboardEditor/DashboardEditor.jsx
```diff
@@ -281,7 +278,30 @@ const defaultProps = {
   breakpointSwitcher: null,
   supportedCardTypes: Object.keys(DASHBOARD_EDITOR_CARD_TYPES),
   renderHeader: null,
-  renderIconByName: null,
+  renderIconByName: (iconName, iconProps) => {
```
sstone2423, line 281: Could this get defined in its own function? Its pretty big to go inside the defaultProps

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.jsx
```diff
@@ -164,65 +126,225 @@ const renderCustomCard = (props) => {
...
-            ...omit(originalDynamicHotspot, 'type', 'x', 'y'),
-          }))
-        )
+
+    const handleOnCardMouseDown = useCallback(
+      (e) => {
+        handleOnClick(setSelectedCardId, cardConfig.id);
+        cardConfig.onMouseDown(e);
+      },
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
sstone2423, line 228: Whats being disabled here? I think all of these should be able to be memoized

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.jsx
```diff
@@ -164,65 +126,225 @@ const renderCustomCard = (props) => {
...
-    case CARD_TYPES.BAR:
-      return (
-        <EditorBarChartCard
-          {...others}
-          dataItems={dataItems}
-          availableDimensions={availableDimensions}
-        />
+    // If the renderCardPreview function returns a valid value for the card, use it,
+    if (
+      typeof renderCardPreview === 'function' &&
```
sstone2423, line 293: should we just remove this? We're not using it and it allows consumers to stray from the design

scottdickerson: I don't think we can since we published a prop type with it.

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.jsx
scottdickerson: believe it or not, it's a bug with the rule, I don't want the callback to trigger for any change of the cardConfig, only those two fields, but if I don't disable, it will switch it to 'cardConfig'

packages/react/src/components/DashboardEditor/DashboardEditor.jsx
scottdickerson: good idea, I moved it out

sstone2423 (approved):

packages/react/src/components/RuleBuilder/utils.test.js
```diff
@@ -10,7 +10,7 @@ import {
 let tree;
 
 const NEW_RULE_MATCH = expect.objectContaining({
-  id: expect.stringMatching(/[a-zA-Z0-9]{10}/),
+  id: expect.stringMatching(/[a-zA-Z0-9]{9,10}/),
```
davidicus, line 13: There has been another fix just merged that addresses this in a slightly different way, FYI

davidicus (approved):

packages/react/src/components/RuleBuilder/utils.test.js
scottdickerson: ok fixed this

PR #2044, opened 2021-02-17
feat(dashboardeditor): scroll to added cards

### LINKED ISSUE #1987: [DashboardEditor] If I add a card that's on the bottom and is scrolled off-screen, the Dashboard Grid should scroll to maeke is visible
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Add a card to a Dashboard with lots of cards on it.  By default the card is added to the bottom but the user might not realize it because it's scrolled offscreen

sstone2423:
Closes #1987 

**Summary**

- When a card is added, scroll to the bottom of the dashboard grid so that the card is visible. This is done by checking if the scroll height of the grid container has changed, meaning a card has been added, then scrolling to the bottom

- Re-adds dashboard-grid's style file as it got lost in the monorepo merge
- Update the styling for DashboardEditor to only show the background color when breakpoints are being used
- Fix the height of the grid container in DashboardEditor to not have any scroll by default

- Go to any DashboardEditor story, add cards until the visible grid is full, then continue to add cards and ensure you can always see the newly added card
- While in those stories, ensure the editor doesn't start with a scrollbar when its empty

### DIFF
--- a/packages/react/src/components/Dashboard/_dashboard-grid.scss
+++ b/packages/react/src/components/Dashboard/_dashboard-grid.scss
@@ -1,0 +1,17 @@
+.#{$iot-prefix}--dashboard-grid {
+  position: relative;
+  .react-grid-item.cssTransforms {
+    transition-property: none;
+  }
+
+  .react-resizable-hide .react-resizable-handle {
+    /* workaround to hide the resize handles in react-grid-layout */
+    display: none;
+  }
+}
+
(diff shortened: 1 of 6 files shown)


scottdickerson (approved):
Looks good just need to make sure that when the DashboardEditor supports resizing (once my latest track gets integrated) that the resize handles still show

davidicus (approved):

PR #2047, opened 2021-02-17
fix(repo): swap styled components to scss

### LINKED ISSUE #1610: [Pull Request Template] add a Regression Test field / question
### What package is this for?

- [x] React
- [x] Angular

### Summary

The current PR template has an Acceptance Test field/question currently that is helpful to remind the dev to give the reader some idea on how to test the PR. I think it would be beneficial (and I've added it myself before)  to have a similar 'Regression Test' field to remind the user to think about how they could have broken old functionality. An example could be:

It doesn't necessarily have to be specific instructions as it could be fully tested for, but it helps to jog the dev's brain

upcoming release?

N/A

### Available extra resources

Can add myself if others think this would be valuable

sstone2423:
Closes #1610 
Part of #370 

**Summary**

- Adds a Regression question for PRs
- Swaps styled components from FileDrop, WizardInline, Header, ComposedModal, ComposedStructuredList, ImageCard
- Moved imported style sheets to their respective sections (experimental vs carbon vs addons

- updated snapshots

- Go to stories for FileDrop, WizardInline, Header, ComposedModal, ComposedStructuredList, ImageCard and see if anything looks out of place


packages/react/src/components/ComposedStructuredList/ComposedStructuredList.jsx
```diff
@@ -80,26 +36,38 @@ const StructuredList = ({
     );
   }
   return (
-    <Fragment>
-      <StructuredListWrapperStyled
+    <>
+      <StructuredListWrapper
         {...StructuredListWrapperProps}
+        className={classnames({
+          [`${iotPrefix}--composed-structured-list__wrapper`]: isFixedWidth,
```
JoelArmendariz, line 43: For future implementations, it would probably be better for these one-off class conditions to be set with css variables.

JoelArmendariz (approved):
LGTM

packages/react/src/components/ComposedStructuredList/_composed-structured-list.scss
```diff
@@ -0,0 +1,32 @@
...
+    caption-side: bottom;
+    display: table-caption;
+  }
+
+  &__loading {
+    padding-top: $spacing-05;
+  }
+
+  &__list-cell {
+    min-width: var(--width);
```
davidicus, line 26: these should probably be more descriptive in the future. While I do not think it should have any conflicts since we are defining the value on the component it may be more self documenting if it were to have `list-cell-width` or some such

packages/react/src/components/Header/_header.scss
```diff
@@ -126,3 +126,16 @@ button.#{$prefix}--header__menu-item {
     outline: none;
   }
 }
+
+.#{$iot-prefix}--header--story-test-class {
```
davidicus, line 130: Not sure about shipping styles that are just for the storybook to consumers. We can put these in the storybook css file.

davidicus (changes requested):
Just the one comment on moving the story style to storybook css file.

packages/react/src/components/ComposedStructuredList/_composed-structured-list.scss
sstone2423: Yep definitely makes sense. This component will be deprecated in the next release as well

packages/react/src/components/Header/_header.scss
sstone2423: Moved to story scss file

davidicus (approved):
LGTM

PR #2042, opened 2021-02-16
Update the button component with testID and fix for Legend colors that are not rendering in the editor.

herleraja:
**Related Issues**
- Closes https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/1942
- Closes https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/1807
- Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1829

**Summary**

- Adding `testID` to the Button component and fix for Legend colors that are not rendering in the editor.

**Change List (commits, features, bugs, etc)**

- Added `testID` to all the ComposedModal ( primary and secondary buttons) and Button components.
- Updated vscode `launch.json` to help in the debugging process.
- Contain fix for the Legend colors rendering in the editor.
- added default `hour` interval to time series card.

**Acceptance Test (how to verify the PR)**

- Make sure all the components load without any issues.

PR #2052, opened 2021-02-18
feat(use-popover-positioning): custom hook for auto-positioning popover elements

### LINKED ISSUE #972: [Component Proposal] usePopoverPosition - hook to improve contextual awareness rendering for tooltip, flyout, etc

kevinsperrine:
Closes #972 

**Summary**

Simple wrappers around the carbon tooltip and overflow menu allow them
to use the usePopoverPosition to automatically change direction or flip
properties to recalculate the menuOffset and ensure the element
stays in the viewable area. The FlyoutMenu component was also modified
to use this functionality.

- added usePopoverPositioning hook
- simple wrappers around the carbon OverflowMenu and Tooltip components to implement it
- added the hook to the FlyoutMenu also

- auto positioning stories should automatically redraw overflow, tooltip, and flyout menus to not overflow the bounds of the screen


packages/react/src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -126,7 +59,7 @@ const DefaultFooter = ({ setIsOpen, onCancel, onApply, i18n }) => (
     <Button
       className={`${iotPrefix}--flyout-menu__submit`}
       aria-label={i18n.applyButtonText}
-      testId="flyout-menu-apply"
+      data-testid="flyout-menu-apply"
```
herleraja, line 62: There is a change in https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2042 Which will effect this PR.

i.e. You should use `testID` instead of `data-testid` for PAL `Button`component.

kevinsperrine: Thanks, @herleraja. I've updated it in the newest push. I'm still learning when/where to use `data-testid`, `testID`, or `testId`. Just don't know carbon well enough yet. I'll get there!

packages/react/src/components/OverflowMenu/OverflowMenu.story.jsx
```diff
@@ -41,6 +43,21 @@ const props = {
   }),
 };
 
+OverflowMenu.displayName = 'OverflowMenu';
+
+export default {
+  title: 'OverflowMenu',
```
davidicus, line 49: For these extended OverflowMenu and Tooltip the stories should be in the IoT section.

kevinsperrine: @davidicus Moved these into IoT

packages/react/src/hooks/usePopoverPositioning.jsx
```diff
@@ -0,0 +1,194 @@
...
+  const offBottom = buttonRect.bottom + tooltipRect.height > windowHeight;
+  const T = offTop ? 'top' : '';
+  const R = offRight ? 'right' : '';
+  const B = offBottom ? 'bottom' : '';
+  const L = offLeft ? 'left' : '';
+
+  return [T, R, B, L].filter(Boolean).join('-');
+};
+
+export const usePopoverPositioning = ({
```
sstone2423, line 22: Could you add jsdoc comments for the props and what it does / how it fixes the positioning issue. This all looks good, but I can see it being difficult to figure out whats happening in the future

sstone2423 (approved):
Just 1 comment to try and add more information about the hook

packages/react/src/hooks/usePopoverPositioning.jsx
kevinsperrine: @sstone2423 Lots of docs added in this (https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2052/commits/aa7e3eaf1b7a26dea4dc6dabe961c26a69f91059) newest commit.

davidicus (approved):

PR #2053, opened 2021-02-18
fix(rulebuilder): pad id generator to fix intermittent failing test

### LINKED ISSUE #2040: [RuleBuilder] Race condition in RuleBuilder test
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

A test for a RuleBuilder utility function intermittently fails. 
image

kevinsperrine:
Closes #2040 

**Summary**

- summary_here

- items_here

- tests here

### DIFF
--- a/packages/react/src/components/RuleBuilder/utils.js
+++ b/packages/react/src/components/RuleBuilder/utils.js
@@ -132,7 +132,7 @@ export const getRuleByPath = (arr, path) => {
  */
 export const generateRule = () => {
   return {
-    id: Math.random().toString(36).substring(2, 12),
+    id: Math.random().toString(36).substring(2, 12).padStart(10, '0'),
     columnId: '',
     operand: '',
     value: '',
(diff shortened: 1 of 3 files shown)


herleraja (approved):
LGTM

packages/react/src/components/RuleBuilder/utils.js
```diff
@@ -132,7 +132,7 @@ export const getRuleByPath = (arr, path) => {
  */
 export const generateRule = () => {
   return {
-    id: Math.random().toString(36).substring(2, 12),
+    id: Math.random().toString(36).substring(2, 12).padStart(10, '0'),
```
sstone2423, line 135: This seems unnecessary just to get a test to pass. Maybe its better to remove the length check in the test instead unless you're something specific is happening that makes it need 10 numbers

sstone2423 (changes requested):

packages/react/src/components/RuleBuilder/utils.js
kevinsperrine: @sstone2423 Good point. Used that approach in newest push.

sstone2423 (approved):

davidicus (approved):

PR #2058, opened 2021-02-22
refactor(colordropdown): swap to combobox

sstone2423:
Closes #

**Summary**

- Use a ComboBox instead of a Dropdown so that users can add valid colors to the selectable item list

- **Note**: There is two issues with this refactor in that we lose the simple icon dropdown  and the ability to render the icon of the selected item. I think the second issue is okay at the moment as we can still see the colors if users expand the list box, but the loss of the icons is a bit more problematic

- Swapped all existing uses of ColorDropdown (mainly in DashboardEditor which is also experimental) to use the new combobox prop model
- Updated all tests to use updated prop model
- Updated ColorDropdown to show as many knobs as possible

- Go to the default ColorDropdown story and ensure that you can select, deselect, filter by typing and by arrow keys, selecting by clicking and entering all of the existing options
- Go to the preset selection story and ensure 'yellow' is preselected and that you're able to 'x' the option and filter on other options
- Go to custom color story and ensure you can add new custom colors to the existing list by typing 'enter'. Try different combinations of invalid and valid css color styles (hex, name, rgb). Ensure that you can't add an invalid color such as 'abc'
- Go to /?path=/story/watson-iot-experimental-hotspoteditormodal--empty , add a tooltip, then test all of the above issues
- Go to /?path=/story/watson-iot-experimental-hotspoteditormodal--empty , add a label, then test all of the above issues
- Go to /?path=/story/watson-iot-experimental-dashboardeditor--default , add a timeseriescard, add a data item, edit that data item, and test the above issues in the modal
- Go to /?path=/story/watson-iot-experimental-dashboardeditor--default , add a grouped bar chart, add a data item, add a groupby, edit that data item, and test the above issues in the modal
- Go to /?path=/story/watson-iot-experimental-dashboardeditor--default , add a table, add a data item, edit that data item, add a threshold, and test the above issues in the modal


kevinsperrine:
@sstone2423 One thing I encountered is, in the default story, if I type to filter and select a color, shouldn't I also be able to choose that color by hitting <kbd>Enter</kbd>? For example, type "red" into the box, use the arrows to select "red90", and then hit <kbd>Enter</kbd>. I expected it to select red90, but instead it adds "red" to the list as a custom color and the box is empty.
colorbox

sstone2423:
@kevinsperrine Yep thats definitely an issue. Good catch

sstone2423:
Closing until we get back to this

Issue #2062, opened 2021-02-24
[AppShell] Boolean to hide side-nav while still showing the hamburger


JordanWSmith15:
### What package is this for?

- [X ] React

### Summary

Please describe your request in one or two sentences.
Boolean property on the application shell to hide the side-navigation while still showing the hamburger (three line) icon and passing a state from it

### Specific timeline issues / requests

We are in the process of implementing the MAS header in Manage for 8.4 via an iFrame. However, we are creating our own side-navigation control to match the PAL, and need to render it instead. 

What we need is, a boolean to hide the PAL side-menu, but not the hamburger (three line) icon. When true, the side-nav is not shown, but the hamburger is visible and has it's state exposed so that we can use it. We need notification of that state so that we can when the user toggles the button, we can open and close our own custom menu into a pinned position.  

![Screen Shot 2021-02-24 at 8 33 04 AM](https://user-images.githubusercontent.com/21066152/109008351-3ceaae00-767b-11eb-90f1-1f94c0b14904.png)

bryancboyd:
@blechner I just prototyped this in `2062-hamburger-only`.  Could you take over?

PR #2065, opened 2021-02-24
feat(suiteheader): Adding support for custom side nav panel logic.

blechner:
Closes [#2062](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2062)

**Summary**

- Including an extra flag that, when true, will force the side nav button to show even when native side nav menu is not configured. The idea is that a custom side nav can be implemented, and toggled according to the custom function property passed to `onSideNavToggled`.

- Implemented extra story with the scenario explained above.
- Included an extra test to keep coverage at an acceptable level.


packages/react/src/components/SuiteHeader/SuiteHeader.test.jsx
```diff
@@ -77,6 +78,16 @@ describe('SuiteHeader', () => {
     );
     expect(screen.getByRole('banner', { name: 'main header' })).toBeInTheDocument();
   });
+  it('renders with left sidenav toggle button (custom side nav support)', () => {
+    const mockOnSideNavToggled = jest.fn();
+    const wrapper = mount(
```
davidicus, line 83: All new test should be written using React testing library over enzyme.

davidicus (commented):
Just a small comment about the test

packages/react/src/components/SuiteHeader/SuiteHeader.test.jsx
blechner: Sure, thanks for the heads up. I implemented with enzyme because I found an example of what I wanted to do in another unit test. But I'll convert the test to only use render/screen, instead of mount 👍

kevinsperrine (approved):
LGTM

stuckless (approved):

davidicus (approved):

PR #2078, opened 2021-03-01
fix(tablecard): in dashboard editor

scottdickerson:
Closes #

**Summary**

- There were two problems in using the TableCards in the DashboardEditor
- The first was that in certain instances a pageSize of 0 was being passed, which causes the Pagination component from carbon to cycle indefinitely.  Also there was a small bug where the whole table rows wouldn't render and there was whitespace underneath
- The second was that if the columns in the table changed, it would render duplicate columns due to a bug with the underlying Table where if you change the order and number of columns it renders duplicates

- fix(TableCard): use key to throw away the table and rerender if the pageSize or the number of columns change.
- fix(TableCard): we can fit two more rows in the editable card, fix the numberOfRowsPerPage calculation
- fix(timeSeriesUtils): increase the generated number of rows to 100 so we can handle all the possible sizes of a table card

- Verify the old TableCard stories
- Verify that if you add a TableCard through the DashboardEditor, and change the columns, it renders successfully with the correct columns. Make sure if you change the size of the TableCard, the new number of rows per page renders

### DIFF
--- a/packages/react/src/components/DashboardEditor/__snapshots__/DashboardEditor.story.storyshot
+++ b/packages/react/src/components/DashboardEditor/__snapshots__/DashboardEditor.story.storyshot
@@ -1422,7 +1422,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT Exper
                               <span
                                 className="bx--pagination__text bx--pagination__items-count"
                               >
-                                1–10 of 10 items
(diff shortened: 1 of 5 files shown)


JoelArmendariz (approved):
LGTM

davidicus (approved):

PR #2082, opened 2021-03-01
fix(dashboardeditor): data filter was not showing

scottdickerson:
Closes #

**Summary**

- The DashboardEditor was using an incorrect property for the isSummaryDashboard

- fix(dataSeriesFormModal): align the fields at their bottom
- fix(dataSeriesFormModal): sort the items and protect the null pointer
- fix(dashboardEditor): move isSummaryDashboard prop

- Verify if you popup the DataItemModal for the SummaryDashboard story, that the dataFilter option shows for the ValueCard

### DIFF
--- a/packages/react/src/components/BarChartCard/barChartUtils.js
+++ b/packages/react/src/components/BarChartCard/barChartUtils.js
@@ -220,7 +220,7 @@ export const formatChartData = (
     series = series.map((item) => omit(item, 'label'));
   }
 
-  let data =  values ?? [];
+  let data = values ?? [];
   if (!isNil(values) && !isEmpty(series)) {
     data = [];
     // grouped or stacked
--- a/packages/react/src/components/BarChartCard/barChartUtils.js
+++ b/packages/react/src/components/BarChartCard/barChartUtils.js
@@ -220,7 +220,7 @@ export const formatChartData = (
     series = series.map((item) => omit(item, 'label'));
   }
 
-  let data = values;
+  let data = !isNil(values) ? values : [];
   if (!isNil(values) && !isEmpty(series)) {
     data = [];
     // grouped or stacked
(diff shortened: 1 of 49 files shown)


packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -515,7 +515,7 @@ const DataSeriesFormItemModal = ({
               onChange={({ selectedItem }) => {
                 if (selectedItem !== mergedI18n.none) {
                   const dataFilter = {
-                    [selectedItem]: availableDimensions[selectedItem][0],
+                    [selectedItem]: availableDimensions[selectedItem].sort()[0],
```
sstone2423, line 518: Is this to resort the options after selection?

sstone2423 (approved):

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
scottdickerson: it's to select the first option automatically @sstone2423.  I had added sorting to the values before rendering in the list.
I needed to change this because the values were now sorted below and the non-first entry was being selected

packages/react/src/components/BarChartCard/barChartUtils.js
```diff
@@ -220,7 +220,7 @@ export const formatChartData = (
     series = series.map((item) => omit(item, 'label'));
   }
 
-  let data = values;
+  let data = !isNil(values) ? values : [];
```
davidicus, line 223: Can we use the native js `??` here?
```suggestion
  let data =  values ?? [];
```

davidicus (commented):
Looks good. Just one small comment about `isNil`

davidicus (approved):

PR #2083, opened 2021-03-01
fix(dashboardeditor): groupby should not be removed when data item is selected

sstone2423:
internal issue: https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/2010

**Summary**

- If a groupby was already in use, and the user removed a selected data item, it would also remove the groupBy. Needed to change the logic to only check if the `filteredGroupBy` was empty, meaning theres no longer any groupBy elements selected

- Remove omit with no second argument
- Added test to cover this case

- Go to /?path=/story/watson-iot-experimental-dashboardeditor--summary-dashboard-with-initial-value , edit the first card, add 2 data items, and 2 groupBy's. First check that the dataSource is correct and showing all data items and groupBy's in the JSON editor. Next, remove 1 data item and recheck the JSON. Next, remove 1 groupBy and recheck the JSON. Continue removing all items and try different combinations to make sure it is always removed, added, or stays the same correctly


scottdickerson (approved):
LGTM

davidicus (approved):

PR #2086, opened 2021-03-03
fix(datetimepicker): cant use multiple datetimepicker on same page #2085

### LINKED ISSUE #2085: [DateTimePicker] Can't use multiple DateTimePicker on same page
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

DateTimePicker has hard coded ids in code. E.g.  iot--date-time-picker__wrapper, relative, absolute, last-number, last-interval, relative-to-when, relative-to-time, date-picker-input-start, date-picker-input-end, start-time and end-time.
This causes issues while using multiple DateTimePicker on same page.

### To Reproduce

1. Go to 'http://health-team1.rtp.raleigh.ibm.com:9080/maximo_123/oslc/graphite/relengineer/index.html#/health'
2. Login using maxadmin/maxadmin
3. Click on any asset.
4. Expand the DateTimePicker in Health history.
5. Click on "Custom Range".
6. Select "Absolute".
7. Note that it works fine.
8. Next expand Asset timeline accordion item.
9. Expand the DateTimePicker in Asset timeline.
10. Click on "Custom Range".
11. Select "Absolute".
12. Note that the calendar is not displayed inside the DateTimePicker.
13. Additionally, if you leave the DateTimePicker in Health history expanded (in custom range), any action performed on DateTimePicker in Asset timeline changes the one in Health history.

### Expected behavior

It should be possible to use multiple DateTimePicker components on same page.

### Environment/versions:

- OS: All
- Browser: All
- `carbon-addons-iot-react` version: 2.143.0-next.40

### Additional context

Screen Shot 2021-03-02 at 11 33 19 PM
Screen Shot 2021-03-02 at 11 33 32 PM

upcoming release?

Needed for upcoming Health release. DCUT is March 15.

navjeets:
Closes #2085 

**Summary**

- Fixes issues with using multiple DateTimePicker component on same page.

- Added a new required prop "id". It is appended to all the hard coded ids to make them unique. This ensures that there are no duplicate ids if there's more than one DateTimePicker component on the same page.

- Tested the fix in Health application

### DIFF
--- a/packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
+++ b/packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
@@ -5938,6 +5938,10 @@ Map {
         ],
         "type": "shape",
       },
+      "id": Object {
+        "isRequired": true,
+        "type": "string",
+      },
       "intervals": Object {
         "args": Array [
           Object {
--- a/packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
+++ b/packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
@@ -5641,6 +5641,7 @@ Map {
         "toLabel": "to",
         "toNowLabel": "to Now",
       },
+      "id": undefined,
       "intervals": Array [
         Object {
           "label": "minutes",
@@ -5939,7 +5940,6 @@ Map {
         "type": "shape",
       },
       "id": Object {
-        "isRequired": true,
         "type": "string",
       },
       "intervals": Object {
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -18,6 +18,7 @@ import {
 import moment from 'moment';
 import { Calendar16 } from '@carbon/icons-react';
 import classnames from 'classnames';
+import uuid from 'uuid';
 
 import TimePickerSpinner from '../TimePickerSpinner/TimePickerSpinner';
 import { settings } from '../../constants/Settings';
@@ -172,7 +173,7 @@ const propTypes = {
   /** The language locale used to format the days of the week, months, and numbers. */
   locale: PropTypes.string,
   /** Unique id of the component */
-  id: PropTypes.string.isRequired,
+  id: PropTypes.string,
 };
 
 const defaultProps = {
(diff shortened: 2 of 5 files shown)


packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -5938,6 +5938,10 @@ Map {
         ],
         "type": "shape",
       },
+      "id": Object {
+        "isRequired": true,
+        "type": "string",
+      },
```
kevinsperrine, line 5944: Should we make the id required and make a public API change, or, by default, generate a few characters as a hash as carbon does with other components like the tooltip, and override that with the ID if one is given?

kevinsperrine (changes requested):
Everything looks good to me except my one comment about the API change.

navjeets:
@kevinsperrine I can remove the isRequired and add the following
if (!id) {
  id = Date.now() + Math.random();
}

OR just add a static variable "uniqueId"
if (!id) {
  DateTimePicker.uniqueId++;
  id = DateTimePicker.uniqueId;
}

Please let me know.

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
davidicus: Yeah we should not require this and instead allow for the id prop to override the default like Kevin suggest. We do  have uuid package that we use in certain components.

davidicus:
> @kevinsperrine I can remove the isRequired and add the following
> if (!id) {
> id = Date.now() + Math.random();
> }
> 
> OR just add a static variable "uniqueId"
> if (!id) {
> DateTimePicker.uniqueId++;
> id = DateTimePicker.uniqueId;
> }
> 
> Please let me know.

I think instead of doing a check we can just include the default id as part of default arguments of the functional component. 
```
const DateTimePicker = ({
 ...
  id = uniqueId,
  ...others
}) => {
  // component code
}

**Comment by kevinsperrine** [comment]
> I think instead of doing a check we can just include the default id as part of default arguments of the functional component.
> 
> ```
> const DateTimePicker = ({
>  ...
>   id = uniqueId,
>   ...others
> }) => {
>   // component code
> }
> ```

@navjeets , I agree with David here, a simple default argument like this should suffice.

```
id = Math.random().toString(36).slice(0, 6)
```

navjeets:
@davidicus @kevinsperrine I removed the isRequired from "id". Using the uuid.v4() as default value instead.

davidicus (approved):

kevinsperrine (approved):

PR #2095, opened 2021-03-05
chore(timeseriescard): raise testing coverage

### LINKED ISSUE #2093: [TimeSeriesCard] raise test coverage
Raise test coverage to acceptable levels and remove the exception in jest.config.js

kevinsperrine:
Closes #2093 

**Summary**

- added test cases with all the edge case props to get coverage on branches
- added test case checking the CSV download works to increase function coverage
- fixed an issue where the tooltip was incorrect if there was no date available
- fixed two issues where it was assumed series was always an array, but proptypes specify it can also be an object
- still couldn't get to 80 on functions, because most of them require the actual chart to be rendered
- removed the other exception and bumped functions to 75%

- added test cases

- All tests pass, coverage is now increased above 80%, except for functions which is 75%.

### DIFF
--- a/packages/react/jest.config.js
+++ b/packages/react/jest.config.js
@@ -14,7 +14,7 @@ module.exports = {
       functions: 80,
       lines: 80,
     },
-    './src/components/**/!(ColumnHeaderSelect|WizardInline|WizardHeader|FilterHeaderRow|TableToolbar|RowActionsCell|RowActionsError|TableHead|StatefulTable|StatefulTableDetailWizard|CatalogContent|FileDrop|Dashboard|CardRenderer|ImageHotspots|PageHero|PageTitle|EditPage|AsyncTable|ImageCard|ImageUploader|TableHead|ColumnResize|ColorDropdown|TimeSeriesCard|BarChartCard|DashboardGrid|DashboardEditor).jsx': {
+    './src/components/**/!(ColumnHeaderSelect|WizardInline|WizardHeader|FilterHeaderRow|TableToolbar|RowActionsCell|RowActionsError|TableHead|StatefulTable|StatefulTableDetailWizard|CatalogContent|FileDrop|Dashboard|CardRenderer|ImageHotspots|PageHero|PageTitle|EditPage|AsyncTable|ImageCard|ImageUploader|TableHead|ColumnResize|ColorDropdown|BarChartCard|DashboardGrid|DashboardEditor).jsx': {
       statements: 80,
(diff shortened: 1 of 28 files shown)


sstone2423 (approved):

packages/react/jest.config.js
```diff
@@ -101,9 +101,7 @@ module.exports = {
     './src/components/Page/PageTitle.jsx': { branches: 75 },
     './src/components/ImageCard/ImageCard.jsx': { branches: 71 },
     './src/components/TimeSeriesCard/TimeSeriesCard.jsx': {
-      statements: 77,
-      functions: 70,
-      lines: 77,
+      functions: 75,
```
davidicus, line 104: We are still leaving an exception here?

davidicus: We should increase in order to remove all exceptions from this file

davidicus (commented):
Seems good except the remaining exception.

packages/react/jest.config.js
kevinsperrine: @davidicus The exceptions are all related to state dependant hover functions like [this one](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/5711c716475fe2c33b441450af0d9a8fad0e6934/packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx#L361), or [ref based useEffects](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/5711c716475fe2c33b441450af0d9a8fad0e6934/packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx#L414) that require the actual chart to be loaded and not just mocked. I wasn't sure what level of refactoring we wanted to employ. I can extract those hover functions, pass the `alertRanges` in, and return the hover function. That would extract those enough to be testable, and, I think, put us over the mark.

kevinsperrine: @davidicus Extracted those methods and tested their outputs. That raises our coverage high enough to remove the exception without worrying about not mocking the chart.

davidicus (approved):

kevinsperrine:
Block waiting to hear about whether this upgrade to react-grid-layout [(diff)](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2104/files#diff-1f344ac391eeecc21ec0f01fb07430a47f4b80d20485c125447d54c33c4bbfc4R145) was an intentional change and if I need to include the updated lock file in this PR.

PR #2113, opened 2021-03-11
fix(dataseriesformitemmodal): grain selector is showing incorrectly for certain card types

scottdickerson:
Closes #

**Summary**

- fix(DataSeriesFormItemModal): table, simple bar, and stacked bar charts may not be time-based in certain cases.  We should be more careful showing the grain and only show it for the real time-based cases
- fix(DataSeriesFormItemContent): when selecting an attribute for a simple bar chart, we should also generate a unique id so that it's clear this was selected by the user.  This allows a non-mean aggregator to be selected for that attribute

- Verify in the DashboardEditor story that if you popup the DataSeriesFormModal for a Simple Bar, Stacked Bar or Table Card attribute that is NOT time-based, that the grain selector is hidden and no grain is set for the item.
- Here are the scenarios, a Simple Bar that's grouped by dimension NOT time
- A Stacked Bar that's grouped by dimension NOT time
- A table card whose autogenerated TIMESTAMP column has been removed
- Verify when selecting a SimpleBar chart attribute, that the dataSourceId contains a UUID

### DIFF
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
@@ -176,7 +176,8 @@ const DataSeriesFormItemModal = ({
 
   const isTimeBasedCard =
     type === CARD_TYPES.TIMESERIES ||
-    type === CARD_TYPES.TABLE ||
+    (type === CARD_TYPES.TABLE &&
+      content?.columns?.find((column) => column.type === 'TIMESTAMP')) ||
     (content?.type === BAR_CHART_TYPES.SIMPLE && content?.timeDataSourceId) ||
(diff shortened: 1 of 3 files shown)


kevinsperrine (approved):
LGTM. The bar/stacked charts are good. The grain selector is hidden as expected and uuid appended to the datasourceid when selecting something like manufacturer or device. The only thing I noticed on the table, and this may be completely unrelated as I'm not too familiar with what you're trying to accomplish here, is that anytime I changed the group by field or added a new data column the timestamp column was automatically reinserted on each change.

davidicus (approved):

scottdickerson:
@kevinsperrine yeah this is non-ideal honestly but we have a separate issue to track for table card editing, we were trying to make the table card easy to use for time based tables (the large majority of all tables) but we need to do a better job of allowing users to make a non time-based table without having to keep removing the auto-added timestamp field.

PR #2121, opened 2021-03-15
feat(suiteheader): Including the 'extraContent'  property to both Header and SuiteHeader.

### LINKED ISSUE #2116: [App Shell] Suite Header tag to display system status
### What package is this for?

- [X ] React

### Summary

In Maximo Manage, there is an existing feature to display the system's current state (Prod, Test, Admin Mode, etc). We need this capability exposed. Desire is to do this with a Tag.
Screen Shot 2021-03-11 at 3 13 17 PM

We need this for MAS 8.4.
Currently, when we attempt to place a tag inside of the header within our own repo, a log error occurs. 
image

### Available extra resources
We can place the tag ourselves but it is currently giving use a log error

blechner:
Closes #2116 

**Summary**

- In order to include a property to allow a custom element to be rendered to the right side of SuiteHeader's `appName` string, I needed to add support for a custom element in the underlying `Header` component. I called the property `extraContent`. If this is not a suitable name or if there is a naming convention for this kind of component property, please let me know and I'll change it.

image

- The commit includes a property in both `Header` and `SuiteHeader`, called `extraContent`, that allows a custom element to be passed and rendered to the right side of `SuiteHeader`'s `appName` string (which translates to an element rendered to the right side of `Header`'s `subtitle` string.

- Included new stories for both `Header` and `SuiteHeader` with a `Tag` being passed as `extraContent` (which is the desired scenario described by the issue).


packages/react/src/components/Header/Header.jsx
```diff
@@ -52,6 +52,8 @@ const propTypes = {
   appName: PropTypes.string.isRequired,
   /** Optional prop that provides additional app information */
   subtitle: PropTypes.string,
```
davidicus, line 54: If this is a one off thing for just this one product should we just open up the subtitle to allow for any node?

blechner: I had considered this approach in the beginning, but then I realized that `SuiteHeader` would need both properties , `appName` (which is mapped to `Header`'s `subtitle`) and the `extraContent`, because `SuiteHeader` also uses `appName` independently in other places (where `extraContent` is not used). Besides, I thought it was safer to just include an optional property in an existing component (`Header`) than repurposing an existing property that might be used in many other projects that are already consuming the `Header` base component.
But if you think it is better change te PropType of `subtitle` to accept any generic element, I think I can do that.

blechner: Ok, so, I'll try to reuse the `Header`'s `subtitle` property instead of creating the `extraContent` property. And `SuiteHeader` will have the `extraContent` to separate `appName` and the custom element. PAL `Header`'s `subtitle` property will be `oneOfType(PropTypes.string, PropTypes.element)`

blechner: @davidicus done! I've removed `extraContent` from `Header`. Please let me know if the this approach is acceptable. Thanks.

davidicus: @blechner This change is fine as long as there is no expectation on clicking on the tag and having it be different than the title link. These will both be wrapped in the link tag and clicking on either will trigger that anchor. Which is why I suggested just keeping your original implementation in my comments 7 days ago. Sorry if that was not clear. @JordanWSmith15 do you want to confirm that this is acceptable for your use case?

JordanWSmith15: Only click event we need in the header is a "Skip to main content" link, which the header seems to already support. So, I think we are good!

davidicus (approved):
Jordan has confirmed that this is acceptable behavior so approving.

packages/react/src/components/Header/Header.jsx
blechner: @davidicus yes, your comment was clear, but I talked with @bryancboyd and we thought that the approach to repurpose the subtitle prop would be less "invasive" to a common component that many other projects might be using, because of what you had commented earlier: "The extra prop is probably the easiest but still feels wrong if this is not a pattern we want teams to pick up."

kevinsperrine (approved):

PR #2126, opened 2021-03-18
fix(chartcards): deep merge the default content props

sstone2423:
**Summary**

- Default props only get shallow merged by default so the chart cards `content` object was not being respected. This PR merges the default props using https://lodash.com/docs/4.17.15#defaultsDeep

- Moved `showTimeInGMT` and `tooltipDateFormatPattern` outside of content as it was defined in the prop types (I'm not sure if this was ever working?)

- Go to stories for BarChartCard, TimeSeriesCard, and PieChartCard and ensure that `truncation` is being defaulted in the underlying chart component's `options.legends` prop . This prop is only defaulted so if it exists in the chart component, the defaultsDeep worked


packages/react/src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -54,28 +80,30 @@ const BarChartCard = ({
   className,
   domainRange,
   timeRange,
+  showTimeInGMT,
+  tooltipDateFormatPattern,
   ...others
 }) => {
-  const { noDataLabel } = i18n;
+  // need to deep merge the nested content default props as default props only uses a shallow merge natively
+  const contentWithDefaults = useMemo(() => defaultsDeep(content, defaultProps.content), [content]);
```
scottdickerson, line 88: this is bad because it mutates content, need to do `defaultsDeep({},content,defaultProps.content)

scottdickerson:
I'll make the requested change above

scottdickerson (approved):

packages/react/src/components/PieChartCard/PieChartCard.jsx
```diff
@@ -248,5 +253,5 @@ const PieChartCard = ({
 };
 
 PieChartCard.propTypes = propTypes;
-PieChartCard.defaultProps = defaultProps;
+PieChart.defaultProps = defaultProps;
```
kevinsperrine, line 256: Was this a mistake? I noticed it b/c it removes a bunch of defaultProps from the publicAPI change.

kevinsperrine (changes requested):

packages/react/src/components/PieChartCard/PieChartCard.jsx
sstone2423: Definitely a mistake. I'll try and change today

sstone2423: Fixed

kevinsperrine (approved):

davidicus (approved):

PR #2158, opened 2021-03-26
feat(PageTitleBar): add scroll states

JoelArmendariz:
**Summary**
There are new requirements for the `PageTitleBar` in regards to scrolling. 
Here is a link to the designs: https://pages.github.ibm.com/ai-applications/design/components/page-header/usage/

 - Add `DYNAMIC` `STATIC` `CONDENSED` and `STICKY` headerModes

 - `STATIC` should be the `PageTitleBar` as is
 
image

 - `DYNAMIC` should start as a `STATIC` mode, but should collapse the title bar to the condensed version when scrolled.
 
image

 - `CONDENSED` a sticky, smaller version of the title bar
 
image

 - `STICKY` should be just like `STATIC` but with a white background, and sticky to the top
 
image


sstone2423 (approved):

kevinsperrine (approved):

packages/react/src/components/PageTitleBar/_page-title-bar.scss
```diff
@@ -1,12 +1,31 @@
...
+      margin: -$spacing-05 0 $spacing-06 $spacing-07;
+    }
+    &--condensed {
+      &-static {
+        margin-left: $spacing-03;
+      }
+      &-before {
+        margin-top: $spacing-07;
+        margin-left: $spacing-03;
+        transition: all 240ms;
```
davidicus, line 19: is this custom transition or is this something we can use the Carbon motion mixin? https://www.carbondesignsystem.com/guidelines/motion/overview/

davidicus (commented):
Looks good. There is just the one comment about the motion mixin.

packages/react/src/components/PageTitleBar/_page-title-bar.scss
davidicus: Examples of its use [here](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/06202605e433edd3a3a936c783cd895ded205289/packages/react/src/components/ProgressIndicator/_progress-indicator.scss#L88)

davidicus (approved):

packages/react/src/components/PageTitleBar/_page-title-bar.scss
JoelArmendariz: This is a custom transition, but I have added the duration tokens. Thanks!

PR #2186, opened 2021-03-31
style(page-title-bar): fix offsets

JoelArmendariz:
**Summary**

The sticky header styles where ignoring the suite header because the offset was only setting the position for the dynamic transition from `STATIC` to `CONDENSED` in the `DYNAMIC` headerMode.

 - Create a new `stickyHeaderOffset` prop and rename the `headerModeDynamicOffSet` offset to `dynamicTransitionOffset`
 - Pass the offsets to the proper `top` attribute

Not much of a change to see in the storybook. This basically just allows the consumer to specify where the sticky headers should "stick".
Make sure that the different headerModes still work properly in the `PageTitleBar` `with dynamic scrolling` story.

### DIFF
--- a/packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
+++ b/packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
@@ -4517,10 +4517,10 @@ Map {
       "collapsed": false,
       "content": undefined,
       "description": null,
+      "dynamicTransitionOffset": 48,
       "editable": false,
       "extraContent": undefined,
       "headerMode": "STATIC",
-      "headerModeDynamicOffSet": 0,
       "i18n": Object {
         "editIconDescription": "Edit page title",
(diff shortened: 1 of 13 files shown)


sstone2423 (approved):

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -4517,17 +4517,18 @@ Map {
       "collapsed": false,
       "content": undefined,
       "description": null,
+      "dynamicTransitionOffset": 48,
       "editable": false,
       "extraContent": undefined,
       "headerMode": "STATIC",
-      "headerModeDynamicOffSet": 0,
```
davidicus, line 4523: This is technically a breaking change but since its unlikely anyone has picked up this feature we can let it slide. Let's try to avoid this in the future.

davidicus (approved):

PR #2197, opened 2021-04-05
feat(card): add card implementation for Angular

### LINKED ISSUE #2196: Card implementation for Angular
From #1952 

Implementation of the card component for Angular

cal-smith:
Closes #2196 

- Adds a suite of card components and directives
- **Does not** implement any of the Chart cards (out of scope for this PR)
- Updates the Angular storybook
- basic tests


packages/angular/src/card/card-toolbar-action.directive.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  constructor(@Optional() protected overflowMenuRef: OverflowMenu) {}
+
+  ngOnInit() {
+    if (this.overflowMenuRef) {
+      this.overflowMenuRef.triggerClass =
+        'iot--card--toolbar-action iot--card--toolbar-svg-wrapper bx--btn--icon-only bx--btn bx--btn--ghost';
+      this.overflowMenuRef.flip = true;
+      this.overflowMenuRef.offset = { x: 4, y: 0 };
+      this.classList = '';
```
theiliad, line 45: this might just be some angular-related logic that I don't understand, but looks like we're immediately setting `this.classList` to `''`?

packages/angular/src/card/card-toolbar-action.directive.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   *
+   * @param overflowMenuRef optional ref to the OverflowMenu instance this directive may be attached to
+   */
+  constructor(@Optional() protected overflowMenuRef: OverflowMenu) {}
+
+  ngOnInit() {
+    if (this.overflowMenuRef) {
+      this.overflowMenuRef.triggerClass =
```
theiliad, line 41: `this.overflowMenuRef.triggerClass = this.classList`?

packages/angular/src/card/card.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Get the height of just the content area as a formatted string
+   *
+   * @returns the height as a string ex. `'200px'`
+   */
+  getContentHeight() {
+    if (!this.height) {
+      return '';
+    }
+    return `${this.height - 48}px`;
```
theiliad, line 48: what's `48`? maybe good to call it something so we understand what amount of height we're deducting here

packages/angular/src/package.json
```diff
@@ -22,6 +22,6 @@
     "@carbon/icons": "10.20.0",
     "@carbon/telemetry": "0.0.0-alpha.6",
     "flatpickr": "4.6.1",
-    "@ai-apps/styles": "0.0.0"
+    "@ai-apps/styles": "^2.145.0-next.16"
```
theiliad, line 25: is there any point in locking this dep to a specific version?

theiliad (changes requested):
Overall some angular code that looks good to me but I can't guarantee that it actually will perform well since **I'm not** as familiar

davidicus (approved):
Looks ok as far as I can tell. May want someone more familiar with angular to review.

packages/angular/src/package.json
cal-smith: Nope, but this _should_ be managed by lerna automatically

packages/angular/src/card/card.service.ts
cal-smith: Good idea ... I'll add it as a field to the service, that way we can configure it

packages/angular/src/card/card-toolbar-action.directive.ts
cal-smith: Yeah that should work actually!

packages/angular/src/card/card-toolbar-action.directive.ts
cal-smith: Yeah, we don't want to duplicate the styles for the container as well as the button in this case, so this just disables the styles on the host element ... Less Angular related, and more specific to what we're doing with the overflow menu style overrides

theiliad (approved):
LGTM

packages/angular/src/card/card-title.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, HostBinding, Input } from '@angular/core';
+
+@Component({
+  selector: 'ai-card-title',
+  template: `
+    <div class="iot--card--title--text" [attr.title]="text">
```
zvonimirfras, line 6: Should we put an ngIf to allow them to completely replace this with custom content from ng-content?

packages/angular/src/package.json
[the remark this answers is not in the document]
cal-smith: We _could_ but lerna manages this for us anyway so I'm not sure it's necessary

packages/angular/src/card/card-title.component.ts
cal-smith: I figure that if they need something that custom they can build the structure themselves, eventually this will have some extra features (like carbon tooltips for truncated text) and trying to merge that with a fully custom content will become a bit of a pain.

The `ai-card-header` (and I mean, really all of these components) is built with the expectation that consumers could pass in whatever they want ... I fully expect consumers to build their own header components from these building blocks.

So yeah, I suppose we could do that, but we might not have to?

packages/angular/src/card/card.stories.ts
```diff
@@ -0,0 +1,202 @@
...
+              <button aiCardToolbarAction>
+                <svg ibmIcon="popup" size="16"></svg>
+              </button>
+            </ai-card-toolbar>
+          </ai-card-header>
+          <ai-card-content>
+            Card content
+          </ai-card-content>
+        </ai-card>
+        <ai-card [defaultHeight]="400" style="margin-bottom:50px">
```
zvonimirfras, line 86:
```suggestion
        <ai-card [defaultHeight]="400" style="margin-bottom: 50px">
```

packages/angular/src/package.json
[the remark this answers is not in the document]
zvonimirfras: Oh, right...

packages/angular/src/card/card-title.component.ts
zvonimirfras: Ah, true. Maybe if it was a more complicated component.

zvonimirfras (approved):

PR #2199, opened 2021-04-06
chore: remove netlify.toml

cal-smith:
The in-repo netlify.toml conflicts with the netlify config in the dashboard which makes deploying the Angular netlify site impossible. This removes that, and we'll just need to update the config in the netlify dashboard after merging.

The `ai-apps-pal-angular` netlify app deploy will fail, since it can't find the right directory to publish, this is expected and not a blocker for this PR.


davidicus (commented):
Will this affect the PR build previews?

cal-smith:
> Will this affect the PR build previews?

Nope! Everything is already configured in the netlify UI, so the config file is redundant at this point anyway

davidicus (approved):

Issue #2077, opened 2021-03-01
[Table] Allow HTML or component column titles


gillibrand:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

I have at least three uncovered use case for title columns:

1) Use an icon only for the visible column "name". This is for columns that themselves just show status icons, so a full name is wasteful of space. See the "has comments" column below.
2) Columns with icon decorators. See below for an "Embedded In" link column that show an orange left-arrow in indicate it's an incoming use case.
3) Long tooltips on columns where embedded help is required.

I think all of these can just be solved by the app if columns can provide either their own HTML or a custom React component for their titles. Only text labels are supported currently.

![image](https://user-images.githubusercontent.com/805573/109544360-4fe6ef00-7a7c-11eb-91dc-a3b056626f06.png)

### Specific timeline issues / requests

Elm 7.0.3. Around June.

davidicus:
@dianatran18 Do we have any design guidance around this?

dianatran18:
1. For icon-only columns, this is the guidance we have around that right now: 
https://pages.github.ibm.com/ai-applications/design/components/datatable/columnalignment#icon-only-columns
I think having text in the icon-only column header makes it more clear than just having an icon, as icons can be ambiguous and may not always be clear to users. Especially since the rows in the column are all icons, having a descriptive column header is valuable. 

2. Similar to the first answer, I think relying on the icon alone to describe "incoming" in this use case is a little ambiguous and not very clear. Having icons in the column header also starts to add more clutter to an already dense data table, and starts to compete with other icons associated with the column header such as the sort icon. I think using the tooltip to help further describe the column is valuable and works better than the icon. In this specific use case, the color of the icon also conflicts with hover/sort states for the column header, and wouldn't pass accessibility against any column header background color. 

3. This we did just add recent guidance on, which you can find here: 
https://pages.github.ibm.com/ai-applications/design/components/datatable/general#column-header
This was feedback we received from MAS to try to create more alignment on how we display tooltips in column headers. Similar to everything described above, we're trying to move away from using info icons in the column header, as it adds more clutter and also competes with other icons such as the sort icon. We are recommending to use the [definition tooltip](https://www.carbondesignsystem.com/components/tooltip/usage/#definition-tooltip) styling Carbon provides to display tooltips. 

@davidicus I believe there's a related issue or conversation somewhere about using the definition tooltip in column headers right?

kevinsperrine:
[...]

PR #2203, opened 2021-04-08
feat(TablePropTypes) Allow non string elements in the table header

herleraja:
Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2077

**Summary**

- allow non-string elements in the table header

- feat(TablePropTypes) allow non-string elements in the table header

- able to view work with an existing table and other add-on components.


packages/react/src/components/Table/StatefulTable.test.jsx
```diff
@@ -347,9 +347,9 @@ describe('stateful table with real reducer', () => {
     let viewProps;
     const onUserViewModified = jest
       .fn()
-      .mockImplementation(({ view, columns, state: { currentSearchValue } }) => {
+      .mockImplementation(({ view, state: { currentSearchValue } }) => {
         viewProps = {
-          columns,
+          columns: initialState.columns,
```
sstone2423, line 352: what does this change?

packages/react/src/components/Table/TablePropTypes.js
```diff
@@ -88,7 +88,7 @@ export const TableRowPropTypes = PropTypes.arrayOf(
 export const TableColumnsPropTypes = PropTypes.arrayOf(
   PropTypes.shape({
     id: PropTypes.string.isRequired,
-    name: PropTypes.string.isRequired,
+    name: PropTypes.node.isRequired,
```
sstone2423, line 91:
```suggestion
    name: PropTypes.oneOfType([PropTypes.node, PropTypes.string].isRequired,
```

Might be better to explicitly allow both strings and nodes still

herleraja: Done!

davidicus (commented):
We should change the example in storybook to use the definition tooltip

herleraja:
> We should change the example in storybook to use the definition tooltip

I tried with `TooltipDefinition` but seems like the focus was taken by header (sorting)

* custom header without focus

image

* custom header with focus (looks like we are trying too many things in the table header)
image


So I'm going to keep it the same for the time being.

packages/react/src/components/Table/StatefulTable.test.jsx
herleraja: We can not compare (isEqual) on two react component if there is a cyclic dependency(https://formidable.com/blog/2018/introducing-react-fast-compare/)

I tried to skip the test case failure by skipping the column value.

davidicus:
@herleraja having an icon in the header probably does not allow for enough room and will provide poor usability. We need to figure out how to provide a definition tooltip to trigger the tooltip. 

image


If you look at the column in the image you can not even see the icon. There is also black text in the filter.

herleraja:
> @herleraja having an icon in the header probably does not allow for enough room and will provide poor usability. We need to figure out how to provide a definition tooltip to trigger the tooltip.
> 
> image
> 
> If you look at the column in the image you can not even see the icon. There is also black text in the filter.

@davidicus I agree!

herleraja:
@davidicus I think the storybook is one example. Maybe we can merge this PR so that others can customize their table header as per their requirement.

herleraja:
Not needed anymore. 

https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2272 covers the above usecase.

PR #2215, opened 2021-04-13
chore: cache dependencies and optimize storybook builds

cal-smith:
This PR adds caching to all the Github workflows, based on https://github.com/actions/cache/blob/main/examples.md#node---lerna

It also adds a `--build` flag to the `storybook.js` script - this allows us to set up the whole monorepo but only build the storybook for the framework we specify. This way we can avoid building both storybooks when we're deploying to netlify.

It also, also, `nohoist`s storybook ... until we get everything updated to the latest this will avoid some of the weird issues we've been having with the storybook builds.


davidicus (approved):

davidicus:
code looks ok but there are a lot of test failures

cal-smith:
> code looks ok but there are a lot of test failures

The netlify stuff seems to be something on their end ... It looks like the react accessibility tests are failing for PageTitleBar, but it may just be a dependency issue

zvonimirfras (approved):

PR #2216, opened 2021-04-13
fix(suiteheader): redirect statements that open links in new tabs were not using the correct parameters.

blechner:
**Summary**

`SuiteHeader` component had an issue in redirect statements that were supposed to open links in new tabs: the second argument of `window.open()` was `'blank'` instead of `'_blank'`.

Also, commenting out the logic around `document.referrer` in `SuiteHeader` as apps importing the component may have a referrer policy that prevents complete URLs to be available, which could cause redirects to the wrong URL.

- Some tests have changed to check for the correct parameters passed to `window.open()`.


davidicus (approved):

packages/react/src/components/SuiteHeader/SuiteHeader.jsx
```diff
@@ -347,13 +347,13 @@ const SuiteHeader = ({
...
                         if (isAdminView) {
                           // Only use referrer URL if it is not the login screen.
-                          if (document.referrer !== '' && document.referrer.indexOf('auth') < 0) {
-                            href = document.referrer;
-                            routeType = ROUTE_TYPES.REFERRER;
-                          } else {
-                            href = navigatorRoute;
-                            routeType = ROUTE_TYPES.NAVIGATOR;
-                          }
+                          // if (document.referrer !== '' && document.referrer.indexOf('auth') < 0) {
```
davidicus, line 350: can we lose the commented out code?

yarn.lock
```diff
@@ -24016,15 +24016,15 @@ react-sizeme@^2.5.2, react-sizeme@^2.6.7:
     shallowequal "^1.1.0"
     throttle-debounce "^2.1.0"
 
-react-sizeme@^2.6.3:
-  version "2.6.3"
-  resolved "https://registry.yarnpkg.com/react-sizeme/-/react-sizeme-2.6.3.tgz#82659d64249c0bf8fc1ddc60b4965d3fb58a5680"
+react-sizeme@^3.0.1:
```
kevinsperrine, line 24019: Why did react-sizeme get bumped here?

kevinsperrine: after checking the various card/dashboard editor/hotspot modal stories, it doesn't seem like it breaks anything, but a second set of eyes would be nice.

kevinsperrine (approved):

yarn.lock
blechner: I don't know what caused this change, I only ran `yarn install` and `yarn test -u` before committing the code.

packages/react/src/components/SuiteHeader/SuiteHeader.jsx
blechner: Done!

yarn.lock
kevinsperrine: interesting. this might be a hold-over from the recent carbon upgrade. @jkeen, did you run into any issues with react-sizeme being bumped?

blechner: please let me know if you want me to revert the change in `yarn.lock`.

PR #2218, opened 2021-04-14
fix(tablecard): do not preformat the timestamp columns only do it on render

scottdickerson:
Closes #

**Summary**

- The tablecard was formatting timestamps inside the 'data' prop being passed to the table instead of formatting the data in a renderDataFunction.  This causes problems for two reasons.  One if you preformat the data then you do it for the whole table, not just the visible table cells so it's a big performance hit.  Two is that if you preformat the data, then any other cell of the table that needs to access the raw timestamp data in the row cannot get to it.  Instead they receive the preformatted localized string, which leads to problems downstream

- fix(TableCard): remove the formatting of timestamp data from the data generation step, and instead move the formatting to the renderDataFunction for those columns.

- Verify the actual data prop passed to the table contains the real timestamp

- Verify the TableCard stories still show the formatted timestamps correctly inside TIMESTAMP columns
- None of the storybook stories should change (except the Dashboard one where I fixed the metadata for the Tablecard)


sstone2423 (approved):
Looks good, but it looks like you picked up some of @cal-smith 's commits here. Just need to double check that those are correct

cal-smith:
> Looks good, but it looks like you picked up some of @cal-smith 's commits here. Just need to double check that those are correct

I haven't fully fixed the tests yet, so those commits will still include the broken `nohoist` ... I should have it resolved soon - locally I'm down to a single failing test suite.

packages/react/src/components/TableCard/TableCard.jsx
```diff
@@ -361,6 +361,13 @@ const TableCard = ({
 
   const newColumns = thresholds ? columnsUpdated : columnsWithFormattedLinks;
 
+  const filteredTimestampColumns = useMemo(
+    () =>
+      columns
+        .map((column) => (column.type && column.type === 'TIMESTAMP' ? column.dataSourceId : null))
```
davidicus, line 367:
```suggestion
        .map((column) => ( column?.type === 'TIMESTAMP' ? column.dataSourceId : null))
```

davidicus (approved):
Looks good. Just one comment about consistency but not huge deal.

PR #2227, opened 2021-04-20
fix(SimplePagination) : Add total items count

herleraja:
Closes #

**Summary**

- add total item counts to the simple pagination component and hide custom range links if not required.

- fix(SimplePagination) : Add total items count
- fix(DateTimePicker): hide custom range link if not required
- fix(List): return the parent id

- DateTimePicker and SimplePagination component works without any issues

### DIFF
--- a/packages/react/src/components/List/List.jsx
+++ b/packages/react/src/components/List/List.jsx
@@ -184,7 +184,7 @@ const List = forwardRef((props, ref) => {
           editingStyle={editingStyle}
           secondaryValue={secondaryValue}
           rowActions={rowActions}
-          onSelect={editingStyle ? () => {} : () => handleSelect(item.id)}
+          onSelect={editingStyle ? () => {} : () => handleSelect(item.id, parentId)}
           onExpand={toggleExpansion}
           onItemMoved={onItemMoved}
           itemWillMove={itemWillMove}
--- a/packages/react/src/components/SimplePagination/SimplePagination.jsx
+++ b/packages/react/src/components/SimplePagination/SimplePagination.jsx
@@ -37,7 +37,7 @@ const SimplePaginationDefaultProps = {
   prevPageText: 'Prev page',
   totalItemsText: 'Items',
   totalItems: undefined,
-  testID: `${iotPrefix}--simple-pagination`,
+  testID: `${iotPrefix}-simple-pagination`,
 };
 
 /** This is a lighter weight pagination component than the default Carbon one */
@@ -60,14 +60,14 @@ const SimplePagination = ({
   const handlePrev = () => onPage(page - 1);
 
   return (
-    <div className={`${iotPrefix}--simple-pagination-container`} data-testid={testID}>
+    <div className={`${iotPrefix}-simple-pagination-container`} data-testid={testID}>
       {totalItems ? (
-        <span className={`${iotPrefix}--simple-pagination-page-label`} maxpage={maxPage}>
+        <span className={`${iotPrefix}-simple-pagination-page-label`} maxpage={maxPage}>
           {`${totalItems} ${totalItemsText}`}
         </span>
       ) : null}
-      <div className={`${iotPrefix}--simple-pagination-page-bar`}>
-        <span className={`${iotPrefix}--simple-pagination-page-label`} maxpage={maxPage}>
+      <div className={`${iotPrefix}-simple-pagination-page-bar`}>
+        <span className={`${iotPrefix}-simple-pagination-page-label`} maxpage={maxPage}>
           {pageText ? `${pageText} ${page}` : pageOfPagesText(page, maxPage)}
         </span>
         {maxPage > 1 ? (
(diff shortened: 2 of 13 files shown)


packages/react/src/components/SimplePagination/SimplePagination.jsx
```diff
@@ -48,56 +60,63 @@ const SimplePagination = ({
...
+          {pageText ? `${pageText} ${page}` : pageOfPagesText(page, maxPage)}
+        </span>
+        {maxPage > 1 ? (
+          <>
+            <div
               className={
                 hasPrev
-                  ? `${iotPrefix}-simple-pagination-caret`
-                  : `${iotPrefix}-simple-pagination-caret-disabled`
+                  ? `bx--pagination__button bx--pagination__button--backward ${iotPrefix}-addons-simple-pagination-button`
```
sstone2423, line 78: Use the included `prefix` for carbon's `bx` and the `addons` string isn't needed as the `iotPrefix` handles our unique prefixing. We should also try to follow BEM by adding `--` between the prefix and component name, and adding `__` between the main component and subcomponent names
```suggestion
                  ? `${prefix}--pagination__button bx--pagination__button--backward ${iotPrefix}--simple-pagination__button`
```

packages/react/src/components/SimplePagination/SimplePagination.jsx
```diff
@@ -48,56 +60,63 @@ const SimplePagination = ({
...
+                  : `bx--pagination__button bx--pagination__button--backward ${iotPrefix}-addons-simple-pagination-button-disabled`
```
sstone2423, line 79: We should also try to follow BEM by adding `--` between the prefix and component name, and adding `__` between the main component and subcomponent names. For specific behaviors or differences in the subcomponent, add `--`
```suggestion
                  `${prefix}---pagination__button ${prefix}--pagination__button--backward ${iotPrefix}--simple-pagination__button--disabled`
```

sstone2423: It looks like the component was previously written like this , but it wouldn't hurt to update to the proper standard

packages/react/src/components/SimplePagination/SimplePagination.story.jsx
```diff
@@ -25,6 +25,8 @@ export const Default = () => (
       page={number('page', 1)}
       maxPage={number('maxPage', 4)}
       onPage={action('onPage')}
+      totalItems={number('totalItems', 10)}
+      totalItemsText={text('total tems display text (totalItemsText)', 'Assets')}
```
sstone2423, line 29: Small typo on `items`

sstone2423 (changes requested):

sstone2423 (approved):

scottdickerson (approved):
LGTM

packages/react/src/components/SimplePagination/SimplePagination.jsx
```diff
@@ -48,56 +60,63 @@ const SimplePagination = ({
...
-            }
-            role="button"
-            tabIndex={hasPrev ? 0 : -1}
-            onClick={hasPrev ? handlePrev : undefined}
-            onKeyDown={hasPrev ? (evt) => handleEnterKeyDown(evt, handlePrev) : undefined}
-          >
-            <CaretLeft16
-              dir="ltr"
-              description={prevPageText}
+    <div className={`${iotPrefix}--simple-pagination-container`} data-testid={testID}>
```
kevinsperrine, line 63: Should these class name changes happen in a minor version? It could potentially create a lot of issues for people using xpath tests.

packages/react/src/components/List/List.jsx
```diff
@@ -184,7 +184,7 @@ const List = forwardRef((props, ref) => {
           editingStyle={editingStyle}
           secondaryValue={secondaryValue}
           rowActions={rowActions}
-          onSelect={editingStyle ? () => {} : () => handleSelect(item.id)}
+          onSelect={editingStyle ? () => {} : () => handleSelect(item.id, parentId)}
```
davidicus, line 187: This seems unrelated to the rest of these changes. Was this intentional?

packages/react/src/components/SimplePagination/SimplePagination.jsx
```diff
@@ -48,56 +60,63 @@ const SimplePagination = ({
...
+          {pageText ? `${pageText} ${page}` : pageOfPagesText(page, maxPage)}
+        </span>
+        {maxPage > 1 ? (
+          <>
+            <div
               className={
                 hasPrev
-                  ? `${iotPrefix}-simple-pagination-caret`
-                  : `${iotPrefix}-simple-pagination-caret-disabled`
+                  ? `bx--pagination__button bx--pagination__button--backward ${iotPrefix}-addons-simple-pagination__button`
```
davidicus, line 78: Is there a reason we are changing these styles. I agree that we shouldn't these. Let's keep these in and open an issue to remove these in V3 and add the version 3 tag

packages/react/src/components/SimplePagination/_simple-pagination.scss
```diff
@@ -1,29 +1,34 @@
 @import '~carbon-components/scss/globals/scss/vars';
 @import '../../globals/vars';
 
-.#{$iot-prefix}-simple-pagination-container {
+.#{$iot-prefix}--simple-pagination-container {
   display: flex;
   height: $spacing-09;
-  justify-content: flex-end;
+  //justify-content: space-between;
```
davidicus, line 7: Can we lose this commented out code?

davidicus (changes requested):
Just a few comments

packages/react/src/components/List/List.jsx
herleraja: This was a fix for select function. In some place we are sending parentid back and some place we aren't.  So it's best to send this information so that we can use it in a callback function.

packages/react/src/components/SimplePagination/SimplePagination.jsx
herleraja: Sure. Let me revert these changes.
@sstone2423 FYI. Let's not edit styles property till V3

davidicus (approved):

PR #2228, opened 2021-04-20
fix(progressindicator): add title attribute for text ellipsis

sstone2423:
**Summary**

- Adds a title attribute to the ProgressIndicator step labels so that the user can hover over the ellipsis and see the hidden text for long strings

- Updated the tests to reflect the change
- Swapped the label element to a span from a p element as its only a label

- Go to /?path=/story/watson-iot-progressindicator--presentation-vertical and hover over the first step's label that is very long. You should see a tooltip on hover of the full text

- Tests should still pass

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,21 +3,30 @@ name: Test
 on: [push, pull_request]
 
 jobs:
-  test:
-    name: '`yarn test`'
+  test-react:
+    name: Test React
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
 
+      - uses: dorny/paths-filter@v2
+        id: changes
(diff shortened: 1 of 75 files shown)


kevinsperrine (approved):
LGTM. I don't know if the change from a `p` tag to a `span` tag causes any xpath issues for down-stream testing though. @davidicus?

sstone2423:
Solid point. I can swap it back if so

herleraja (approved):
LGTM

davidicus:
> Solid point. I can swap it back if so @kevinsperrine

Yeah. Unfortunately, until we get al the test IDs in place and everyone can stop using xpath we have to be careful with markup changes. Let's switch that back and this will be good to go.

davidicus (approved):

PR #2230, opened 2021-04-20
Flyout menu tooltip option

JoelArmendariz:
**Summary**
We need to add the ability to show the onHover tooltips for the `FlyoutMenu` trigger button.

 - Add `hideTooltip` prop. (defaults to true)
 - Pass the rest of the `buttonProps` through to the trigger button

Go to any `FlyoutMenu` story and use the knobs for `hideTooltip`, `buttonProps`, and `iconDescription`. See that the component is updated accordingly.

Make sure basic functionality is still working properly.

image

### DIFF
--- a/packages/react/src/components/FlyoutMenu/FlyoutMenu.jsx
+++ b/packages/react/src/components/FlyoutMenu/FlyoutMenu.jsx
@@ -207,8 +207,12 @@ const FlyoutMenu = ({
   ) : (
     <DefaultFooter setIsOpen={setIsOpen} onCancel={onCancel} onApply={onApply} i18n={i18n} />
   );
+
   return (
     <div
+      style={{
+        '--tooltip-visibility': iconDescription ? 'visible' : 'hidden',
+      }}
       ref={buttonRef}
       className={classnames(
         [`${iotPrefix}--flyout-menu`],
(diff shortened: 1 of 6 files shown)


herleraja (approved):
LGTM

kevinsperrine (approved):
All good. I checked it out locally since the netlify deploys are borked at the moment. @davidicus Is anyone looking into that on the Angular side?

packages/react/src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -220,6 +225,7 @@ const FlyoutMenu = ({
       )}
     >
       <Button
+        {...buttonProps}
```
davidicus, line 228: What are we spreading here? Can we not call it out specifically?

davidicus (commented):
This looks good just the one comment on what we are spreading to the button. Seems unrelated to the rest of the changes.

packages/react/src/components/FlyoutMenu/FlyoutMenu.jsx
JoelArmendariz: This was so that the user could also specify the `tooltipPosition` from the default `top`. I can make that prop explicit instead, just thought there might be other use-cases where the button's other props would come into play.

davidicus (approved):

PR #2231, opened 2021-04-20
fix(table): include search.defaultValue in filtering logic

### LINKED ISSUE #1901: [Table] isFiltered does not account for Search filter default value
### What package is this for?

- [ x ] React
- [ ] Angular

### Describe the bug

In graphite we cannot use the toolbar.search.value to set the current search, and as such we use the toolbar.search.defaultValue.  If we set the .value then we can no longer use the search as it re-renders with every keystroke.

But, using defaultValue with a search that returns so results shows the "no data" message instead of the "no results" message.  I think this is because the isFiltered check only looks at .value and NOT at .defaultValue.

kevinsperrine:
Closes #1901 

**Summary**

- include search.defaultValue to determine if the "no data" or "no results" message should be displayed

- included search.defaultValue in isFiltered logic on the table

- The [no data story](https://deploy-preview-2231--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--with-no-data) should still show "no data"
- The [no results story](https://deploy-preview-2231--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--with-no-results) should still show "no results"
- In the [Stateful Example with Create & Save Views](https://deploy-preview-2231--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-create-save-views) change the "defaultSearchValue" knob to "test" and then change to the "My view 1" view and it should show the "no results" message

### DIFF
--- a/packages/react/src/components/Table/Table.jsx
+++ b/packages/react/src/components/Table/Table.jsx
@@ -4,7 +4,6 @@ import merge from 'lodash/merge';
 import pick from 'lodash/pick';
 import useDeepCompareEffect from 'use-deep-compare-effect';
 import { Table as CarbonTable, TableContainer, Tag } from 'carbon-components-react';
-import isNil from 'lodash/isNil';
 import uniqueId from 'lodash/uniqueId';
 import classnames from 'classnames';
 import { useLangDirection } from 'use-lang-direction';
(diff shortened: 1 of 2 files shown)


herleraja (approved):
LGTM

davidicus (approved):

PR #2232, opened 2021-04-20
fix(list-item): add aria-label and title to expand button

### LINKED ISSUE #2094: [List] with hierarchy accessibility violation
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The List with hierarchy has accessibility violations using the IBM Equal Access Accessibility Checker. The expand list icon does not have an accessible name. 

`Interactive component does not have a programmatically associated name`

### To Reproduce

Checked the list with hierarchy component in Storybook and ran the accessibility scan using the IBM Equal Access Accessibility Checker extension. Seeing the same issues while using the List component in my codebase 

### Environment/versions:

- OS: MacOS
- Browser: Chrome
- `carbon-addons-iot-react` version: v2.114.5

### Additional context

Screen Shot 2021-03-05 at 10 13 48 AM

upcoming release?

date, the better!_

kevinsperrine:
Closes #2094 

**Summary**

- added aria-label and title attributes to the ListItem when it's expandable. 
- ❓❓❓ QUESTION: The current tests are checking for "expand" when the item is expanded and "close" when the item is closed. I would expect these to be opposite. ie. If the item is expanded, the aria-label should be "Close" to explain the purpose of that button, no?

- Running the accessibility checker should not show errors for `Interactive component does not have a programmatically associated name` on the expand/contract buttons.

### DIFF
--- a/packages/react/src/components/List/ListItem/ListItem.jsx
+++ b/packages/react/src/components/List/ListItem/ListItem.jsx
@@ -121,7 +121,12 @@ const ListItem = ({
   itemWillMove,
   dragPreviewText,
 }) => {
-  const handleExpansionClick = () => isExpandable && onExpand(id);
+  const handleExpansionClick = (event) => {
+    event.stopPropagation();
+    if (isExpandable) {
+      onExpand(id);
+    }
+  };
 
   if (__DEV__ && Array.isArray(rowActions)) {
     warning(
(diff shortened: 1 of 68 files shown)


herleraja (approved):
LGTM

packages/react/src/components/List/ListItem/ListItem.jsx
```diff
@@ -149,12 +149,14 @@ const ListItem = ({
         className={`${iotPrefix}--list-item--expand-icon`}
         onClick={handleExpansionClick}
         data-testid="expand-icon"
+        aria-label={expanded ? i18n.expand : i18n.close}
```
davidicus, line 152: yes like you mentioned. This looks like it should be flipped.

davidicus (changes requested):

davidicus (approved):

PR #2241, opened 2021-04-21
fix(dashboard-editor): always use dataSourceId in selectedData array

### LINKED ISSUE #2051: [DashboardEditor] The Data Item selector uses a different label for each data item than ends up in the data section￼
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

*Labels passed in getValidDataItems are not being used by the selector even though they are used in the Data Item section

This leads to some strange behavior where once you select a data item, when it shows in the bottom area for each data item, it doesn't match
image

To see the error, select a DataItem from the picker for a TimeSeriesCard in the DashboardEditor:
https://next.carbon-addons-iot-react.com/?path=/story/watson-iot-experimental-dashboardeditor--default

Notice the labels don't match between the selector and the section

kevinsperrine:
Closes #2051 

**Summary**

- always use the dataSourceId in the selectedItems array to ensure the proper labels are used by the dropdown and in the list of selected data points

- use dataSourceId when adding new data items to be used.

- In the [DashboardEditor default story](https://deploy-preview-2241--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-experimental-dashboardeditor--default) add a TimeSeriesCard and ensure the labels always match in the dropdown and in the list of data items for the card.


davidicus (approved):

scottdickerson (approved):
LGTM

PR #2254, opened 2021-04-22
feat(dashboard-editor): add decimal precision to ImageCards

### LINKED ISSUE #2045: [DashboardEditor] Image Card Hotspot Modal cannot set precision or aggregate
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When editing an image card in the DashboardEditor you should be allowed to set a custom aggregator/grain and decimal precision for each attribute (just like you can do in a Value Card) However those additional capabilities haven't been added

### To Reproduce

image

kevinsperrine:
Closes #2045 

**Summary**

- Adds precision to data items on ImageCards in the DashboardEditor

- display precision dropdown on ImageCards
- add precision to the reducer to save its state
- update styles to fix increment/decrement on thresholds
- update test cases to account for the extra buttons in the modal

- Add an image card to the DashboardEditor
- Edit the image
- add Pressure as a data item
- edit Pressure
- Ensure Aggregate/Precision fields are available and persist after change/save

### DIFF
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
@@ -341,8 +341,11 @@ const DataSeriesFormItemModal = ({
               />
             </div>
           )}
-          {type === CARD_TYPES.IMAGE && (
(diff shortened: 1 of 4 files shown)


davidicus (approved):

scottdickerson (approved):
LGTM

PR #2259, opened 2021-04-22
fix(dashboard-editor): internationalization fixes

JoelArmendariz:
**Summary**
Need to pass through a `locale` prop to the dashboard editor so that the cards render the correct date labels.
Also fixes a bug where the `translateWithId` callback wasn't returning anything.

Pass through a new `locale` prop for the `DashboardEditor`.
Fix `translateWithId` to return the new strings.
Small fix for the dataSeriesItems edit button tooltip position that was hiding under its parent.

Pass a locale of `de` (or any other locale that formats dates differently) to the `DashboardEditor` and confirm that the date labels show the new locale format.

Make sure by default, nothing has changed in the cards

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,20 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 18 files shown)


sstone2423 (approved):

kevinsperrine (approved):
I pulled it down locally and it all works as expected. It would be nice to get that locale prop added as a knob in storybook though.

scottdickerson (approved):
Looks good to me

davidicus (approved):

davidicus:
@jkeen FYI

PR #2261, opened 2021-04-23
chore(react): remove momentjs, replace with dayjs

### LINKED ISSUE #2187: [Repo] Evaluate the continued use of moment library
We need to evaluate the continued use and possible removal of moment.js in favor of native [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) Object. Moment states that their library is not tree shaking friendly. Using native methods will allow us to have a smaller package size for users

More information on the reason for switching can be found [here](https://momentjs.com/docs/#/-project-status/) from the Moment project status page. 

If found that we can still serve our use cases using native function this issue will track its removal from the project. If need be this can be converted to an epic since our use of moment is pretty widespread. 

Other resources:
[You Probably Don't Need Moment.js Anymore](https://dockyard.com/blog/2020/02/14/you-probably-don-t-need-moment-js-anymore)
[You don't (may not) need Moment.js](https://github.com/you-dont-need/You-Dont-Need-Momentjs/blob/master/README.md)
[Why you shouldn't use Moment.js...](https://inventi.studio/en/blog/why-you-shouldnt-use-moment-js)
[4 alternatives to moment.js for internationalizing dates](https://blog.logrocket.com/4-alternatives-to-moment-js-for-internationalizing-dates/)

jkeen:
Closes #2187 

**Summary**

- Removed moment and replaced with dayjs

- Dayjs has basically the same API as moment, although it requires you to manually include the plugins you need in order to keep bundle size small. Also dayjs is immutable. 
- Date mocking changes in test environment and for storybook snapshots


packages/react/src/components/ComboChartCard/__snapshots__/ComboChartCard.story.storyshot
```diff
@@ -693,9 +693,9 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT Exper
                             >
                               <span
                                 className=""
-                                title="10/28/2018 12:34"
+                                title="10/28/2018 07:34"
```
kevinsperrine, line 696: This all looks great, but I am curious why these snapshots changed. I would have expected them to stay the same if it were a seamless replacement.

jkeen:
I changed the implementation of date mocking which changes a bunch of snapshots: 

The old implementation (`mockDate('2018-10-28T12:34:56Z')`) would return the static date for anything passed into `new Date()`, meaning that parsing dates using the format `new Date("2021-04-29T05:00:00Z")` would ignore the input entirely and return `new Date("2018-10-28T12:34:56Z")`. That caused snapshots for components that were expressing a range like "20 minutes ago" in the DateTime component to all be the same date, as well as sample dates that may have gone through some parsing/transformation to be output as that same static mocked date (if it passed through the Date constructor, which varies between implementations of moment and dayjs.) Basically, the old way wasn't consistently testing the thing it should be testing, imo. 

Using the new implementation (mockdate), new Dates/Date.now return the specified point in time while arguments passed into Date are respected, and ranges relative to that point in time render correctly.

sstone2423 (approved):
Looks great

scottdickerson (approved):
LGTM

PR #2262, opened 2021-04-23
feat(flyout-menu): add flyout menu

### LINKED ISSUE #2204: [Menu] Design specs
### What package is this for?

- [x] React
- [ ] Angular

[Menu@2x.pdf](https://github.com/carbon-design-system/carbon-addons-iot-react/files/6278945/Menu%402x.pdf)

Menu

### Available extra resources
Component guidance reference:
https://pages.github.ibm.com/ai-applications/design/components/menus/overview

### LINKED ISSUE #2205: Menu specs
### What package is this for?

- [ ] React
- [x] Angular

[Menu@2x.pdf](https://github.com/carbon-design-system/carbon-addons-iot-react/files/6278959/Menu%402x.pdf)

Menu

### Available extra resources
Component guidance reference:
https://pages.github.ibm.com/ai-applications/design/components/menus/overview

### LINKED ISSUE #2249: Add filter flyout menu
https://pages.github.ibm.com/ai-applications/design/components/datatable/filtering#filter-flyout

zvonimirfras:
In addition to adding a flyout menu to angular implementation, it updates the react implementation to match the desired design.

Closes #2249 
Closes #2204 
Closes #2205


davidicus:
@dianatran18 do you want to do a review of these style changes for the flyoutmenu? https://next.carbon-addons-iot-react.com/?path=/story/watson-iot-flyout-menu--default-example

davidicus (approved):
it looks fine from the React stories. I have asked Diana if she wants to give a review of the style changes.

dianatran18:
I do want to do a review 😅 I'll put together some comments but before I do what's the max size you currently have for that "large"flyout?

zvonimirfras:
> I do want to do a review 😅 I'll put together some comments but before I do what's the max size you currently have for that "large"flyout?

Can't tell who this is for 🤔

dianatran18:
> > I do want to do a review 😅 I'll put together some comments but before I do what's the max size you currently have for that "large"flyout?
> 
> Can't tell who this is for 🤔

Whoever has the answer. 

FYI @Najil-K can we do an audit of this menu as well?

Najil-K:
> > > I do want to do a review 😅 I'll put together some comments but before I do what's the max size you currently have for that "large"flyout?
> > 
> > 
> > Can't tell who this is for 🤔
> 
> Whoever has the answer.
> 
> FYI @Najil-K can we do an audit of this menu as well?

@dianatran18 Ok👍

Najil-K:
[UI audit - Flyout Menu (v1).pdf](https://github.com/carbon-design-system/carbon-addons-iot-react/files/6516420/UI.audit.-.Flyout.Menu.v1.pdf)
.
.
.
.
UI audit - Flyout Menu (v1) @2x

packages/angular/src/flyout-menu/flyout-menu-pane.component.ts
```diff
@@ -0,0 +1,197 @@
...
+ * [See demo](../../?path=/story/components-flyout-menu--basic)
+ *
+ * html:
+ * ```
+ * <ai-flyout-menu-pane>
+ *	options
+ * </ai-flyout-menu-pane>
+ * ```
+ *
+ * <example-url>../../iframe.html?id=components-flyout-menu--basic</example-url>
```
cal-smith, line 33: I don't think `example-url` does anything at the moment, with the switch to tsdoc ... I bet we could make a webcomponent that renders the example though 🤔

packages/angular/src/flyout-menu/flyout-menu-pane.component.ts
```diff
@@ -0,0 +1,197 @@
...
+      class="bx--tooltip bx--tooltip--shown iot--flyout-menu--body"
+      [ngClass]="{
+        'iot--flyout-menu--body__bottom-start': position === 'bottom-start',
+        'iot--flyout-menu--body__bottom-end': position === 'bottom-end',
+        'iot--flyout-menu--body__top-start': position === 'top-start',
+        'iot--flyout-menu--body__top-end': position === 'top-end',
+        'iot--flyout-menu--body__left-start': position === 'left-start',
+        'iot--flyout-menu--body__left-end': position === 'left-end',
+        'iot--flyout-menu--body__right-start': position === 'right-start',
+        'iot--flyout-menu--body__right-end': position === 'right-end',
```
cal-smith, line 52: Do we need all these? Or is this something our dialog component can handle already?

packages/angular/src/flyout-menu/flyout-menu.component.ts
```diff
@@ -0,0 +1,89 @@
...
+      <button
+        aria-label="Helpful description"
+        data-testid="flyout-menu-button"
+        tabindex="0"
+        class="
+        iot--flyout-menu--trigger-button
+        iot--btn
+        bx--btn
+        bx--btn--ghost
+        bx--btn--icon-only
```
cal-smith, line 51: Would the `ibmButton` directive help cut down on the classlist here?

packages/angular/src/flyout-menu/flyout-menu.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  selector: 'ai-flyout-menu',
+  template: `
+    <ng-template #templateRef let-tooltip="tooltip">
+      <div class="bx--tooltip__content">
+        <div class="iot--flyout-menu--content">
+          <ng-content></ng-content>
+        </div>
+        <div class="iot--flyout-menu__bottom-container">
+          <ng-content select="[cancelButton]"></ng-content>
+          <ng-content select="[applyButton]"></ng-content>
```
cal-smith, line 30: If we always have cancel/apply it may be easier to just include them here and have a `@Output` for `cancel` and `apply`

cal-smith: Or make a footer/button container component that can include them and fully pass of the responsibility to the app ... How would the app signal that the dialog should close in this case? Or is that something that makes more sense for the component to control anyway?

packages/angular/src/flyout-menu/flyout-menu.stories.ts
```diff
@@ -0,0 +1,38 @@
...
+  .addDecorator(withKnobs)
+  .add('Basic', () => ({
+    template: `
+      <div style="position: absolute; left: 50%; top: 50%;">
+        <ai-flyout-menu [flip]="flip" [placement]="placement">
+          <div class="title">
+            Filter
+            <a ibmLink (click)="clearFilterClicked($event)" class="clear-flyout" href="#">Clear</a>
+          </div>
+          Columns
```
cal-smith, line 27: Would it make sense to wrap this in a container `ai-flyout-content` component?

packages/angular/src/flyout-menu/flyout-menu.stories.ts
```diff
@@ -0,0 +1,38 @@
...
+          <button ibmButton="secondary" cancelButton>Cancel</button>
+          <button ibmButton applyButton>Apply</button>
```
cal-smith, line 29: What content is allowed in the footer area? How customizable is it? We may want a `ai-flyout-footer` wrapper for these (plus that's easier for people to understand ... the hidden attribute thing trips people up fairly frequently)

packages/angular/src/flyout-menu/index.ts
```diff
@@ -0,0 +1,3 @@
+export { FlyoutMenu } from './flyout-menu.component';
+export { FlyoutMenuPane } from './flyout-menu-pane.component';
+export { FlyoutMenuModule } from './flyout-menu.module';
```
cal-smith, line 3: Also make sure there's a `package.json` in the `flyout-menu` directory - that way it's exposed as a separate entrypoint ... Also, it should be exported from the root `index.ts`

cal-smith (changes requested):
Overall looks good! Just a few questions/conversations in a few areas, but otherwise it seems like a solid first pass

packages/angular/src/flyout-menu/flyout-menu-pane.component.ts
zvonimirfras: These are iot prefix so I don't think dialog would handle them.

packages/angular/src/flyout-menu/flyout-menu.stories.ts
zvonimirfras: What ever you put into the menu becomes its content so for now I don't see the benefit of a separate component.

packages/angular/src/flyout-menu/flyout-menu.component.ts
zvonimirfras: They are optional so in addition to events, we'd also need inputs to enable/disable them.
Selectors allow full flexibility, so that's what I went with instead.

As the designs evolve, we might want to revisit that.

cal-smith: Sorry, I guess I wasn't super clear, how about something like - 

```html
<ai-flyout-footer>
  <button ibmButton="secondary">Cancel</button>
  <button ibmButton="primary">Apply</button>
</ai-flyout-footer>
```

That way we don't have to do any extra work (the styles should handle laying out the buttons) and the consumer gets the same flexibility in specifying the button control without having to add random attributes to the buttons (plus they can add N extra buttons, which may be useful?)

If we wanted to be really helpful we could _also_ add a `ai-flyout-footer-buttons` component that wraps up the footer + the default set of buttons, but then we run into the points you mentioned (events + inputs + g11n concerns)

packages/angular/src/flyout-menu/flyout-menu-pane.component.ts
cal-smith: Right, but what styles do they actually apply? If it's just some `left`/`right` we can probably skip them, if it has more specific style hooks (content alignment or something?) then we'll have to keep them. As-is if it's just `left`/`right` I'm guessing the dialog is overriding it with inline styles

cal-smith:
Added some clarifying comments @zvonimirfras ... Basically ready to go once we sort those out 👍

packages/angular/src/flyout-menu/flyout-menu-pane.component.ts
zvonimirfras: they do more than that 👍

packages/angular/src/flyout-menu/flyout-menu.component.ts
```diff
@@ -0,0 +1,81 @@
...
+ */
+@Component({
+  selector: 'ai-flyout-menu',
+  template: `
+    <ng-template #templateRef let-tooltip="tooltip">
+      <div class="bx--tooltip__content">
+        <div class="iot--flyout-menu--content">
+          <ng-content></ng-content>
+        </div>
+        <ng-content select="ai-flyout-menu-footer"></ng-content>
```
cal-smith, line 25: Maybe also add `.iot--flyout-menu__bottom-container` so users can pass in a custom footer with the iot base class?
```suggestion
        <ng-content select="ai-flyout-menu-footer, .iot--flyout-menu__bottom-container"></ng-content>
```

It's tricky, since we don't have a content component, so we have to use selectors to put stuff in the right place ... but that should work for most cases?

cal-smith (approved):
One suggest, but otherwise everything looks good!

There's some weirdness in the flyout menu button when it opens:

image

But it seems like that's just the styles (react avoids that because it immediately focuses the cancel button)

Also a couple type errors in the build, nothing crazy though: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2262/checks?check_run_id=2776576326

PR #2266, opened 2021-04-26
Add angular tabs

### LINKED ISSUE #2211: Tabs part 1 implementation for Angular
From #1952

Implementation of the Tabs component for Angular

cal-smith:
Closes #2211

**Summary**

Adds an first pass of an implementation of the tabs component for Angular.

Things still missing/not working:

- tab header actions (needs ground up implementation)
- container tabs (not tested)
- right aligned tab dropdown by default (needs upstream fix)
- other misc design issues

Otherwise all the groundwork is here and ready to be consumed - The general architecture of a `TabController` driving everything should serve us well for this use case, and as we expand into more complex domains (like dashboards). As usual, this implementation is intended to be fairly minimal and composable - it's expected that teams will use these building blocks to create more complex components, and as the Angular PAL matures we expect we will provide more complete "pattern" components that expand on the functionality provided here.


packages/angular/src/tabs/tab-action.directive.ts
```diff
@@ -0,0 +1,21 @@
+import { Component, Directive, ElementRef, OnInit } from '@angular/core';
+import { Button } from 'carbon-components-angular';
+
+@Directive({
+  selector: '[aiTabAction]'
```
zvonimirfras, line 5: Neat little helper 👍

zvonimirfras (approved):

PR #2273, opened 2021-04-27
Refactor react build, update storybook

cal-smith:
This ensures that the output of the react build is flattened (no surprise node_modules directories, or src directories in the output) and updates storybook to be on par with the Angular package. The storybook update allows us to remove a number of the `nohoist` rules, which simplifies our package layout and reduces the number of duplicate packages installed (should be a tiny little bit faster!).


kevinsperrine (changes requested):
Looks like the section title structure changed for Storybook 6. The section title for Carbon components is gone and is now first in the list, and the Getting Started section is under a `0` title and 
Screen Shot 2021-04-28 at 9 21 52 AM

kevinsperrine:
I can't speak to the actual build output changes, but everything else except the storybook sections seems fine. "Show Info" buttons are still working, Readme tabs still work for overflow/table components, and knobs/actions/etc all seem unchanged.

cal-smith:
> Looks like the section title structure changed for Storybook 6. The section title for Carbon components is gone and is now first in the list, and the Getting Started section is under a `0` title.
> Screen Shot 2021-04-28 at 9 21 52 AM

Great catch, I'll take a look at this ... I think the title structure may have changed, but it should be a fairly straight forward update 🤔

cal-smith:
> > Looks like the section title structure changed for Storybook 6. The section title for Carbon components is gone and is now first in the list, and the Getting Started section is under a `0` title.
> > Screen Shot 2021-04-28 at 9 21 52 AM
> 
> Great catch, I'll take a look at this ... Hopefully it's a fairly straight forward update 🤔

Well that was anything but straight forward. You'll notice the story roots are now numbered - that was the only way I could get it to behave as expected (backed up by a number of similar issues on the storybook repo) ... it seems like we'll be able to clean it up once storybook v6 is released, so this is really a stop-gap anyway

kevinsperrine (approved):

davidicus (commented):
@cal-smith I was unable to build the project as is. Part of the error was not being able to find the package `@babel/helper-builder-react-jsx`.  Once added I was able to build the project. 
Another area of concern that I saw was the following:
image

I started a sample app and used a sample bar chart and it loaded fine but am not sure if there is some functionality that is going to break.  I will try to test the current build to see if they are  part of this refactor or not.

davidicus:
Running the build again without the extra babel package worked. So not sure why the intermittent failure.

davidicus:
Testing the build on next it does look like this error was the present before these changes.

kevinsperrine:
> Testing the build on next it does look like this error was the present before these changes.

Is that at all related to the weird failures I'm seeing on [this PR](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2277/checks?check_run_id=2477129205)

davidicus:
> > Testing the build on next it does look like this error was the present before these changes.
> 
> Is that at all related to the weird failures I'm seeing on [this PR](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2277/checks?check_run_id=2477129205)

Sorry, @kevinsperrine I said error but it is actually a warning, which didn't fail the build.

cal-smith:
> > Testing the build on next it does look like this error was the present before these changes.
> 
> Is that at all related to the weird failures I'm seeing on [this PR](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2277/checks?check_run_id=2477129205)

It's plausible ... I'd do a `yarn why` on the packages giving you trouble and see where they're getting placed ... Ideally everything should be installed to the root, so if it's getting split for some reason that could be where the problems coming from.

kevinsperrine:
> > > Testing the build on next it does look like this error was the present before these changes.
> > 
> > 
> > Is that at all related to the weird failures I'm seeing on [this PR](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2277/checks?check_run_id=2477129205)
> 
> It's plausible ... I'd do a `yarn why` on the packages giving you trouble and see where they're getting placed ... Ideally everything should be installed to the root, so if it's getting split for some reason that could be where the problems coming from.

No such luck. yarn why shows it all being in the root, but it has something to do with the namedExports config in rollup, I think. It's unable to find getMenuOffset in `lib`, but if I switch it to `es` it finds it, and then chokes on `invariant` b/c in Carbon it's imported as `import invariant from 'invariant';`, but technically invariant doesn't have a default export, so it fails to rollup. No idea why it's happen on PR #2277, but not other places.
```js
[!] Error: 'getMenuOffset' is not exported by node_modules/carbon-components-react/lib/components/OverflowMenu/OverflowMenu.js
https://rollupjs.org/guide/en#error-name-is-not-exported-by-module-
src/components/OverflowMenu/index.jsx (4:9)
2: import { OverflowMenu as CarbonOverflowMenu } from 'carbon-components-react';
3: import PropTypes from 'prop-types';
4: import { getMenuOffset } from 'carbon-components-react/lib/components/OverflowMenu/OverflowMenu';
            ^
```

cal-smith:
> > > > Testing the build on next it does look like this error was the present before these changes.
> > > 
> > > 
> > > Is that at all related to the weird failures I'm seeing on [this PR](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2277/checks?check_run_id=2477129205)
> > 
> > 
> > It's plausible ... I'd do a `yarn why` on the packages giving you trouble and see where they're getting placed ... Ideally everything should be installed to the root, so if it's getting split for some reason that could be where the problems coming from.
> 
> No such luck. yarn why shows it all being in the root, but it has something to do with the namedExports config in rollup, I think. It's unable to find getMenuOffset in `lib`, but if I switch it to `es` it finds it, and then chokes on `invariant` b/c in Carbon it's imported as `import invariant from 'invariant';`, but technically invariant doesn't have a default export, so it fails to rollup. No idea why it's happen on PR #2277, but not other places.
> 
> ```js
> [!] Error: 'getMenuOffset' is not exported by node_modules/carbon-components-react/lib/components/OverflowMenu/OverflowMenu.js
> https://rollupjs.org/guide/en#error-name-is-not-exported-by-module-
> src/components/OverflowMenu/index.jsx (4:9)
> 2: import { OverflowMenu as CarbonOverflowMenu } from 'carbon-components-react';
> 3: import PropTypes from 'prop-types';
> 4: import { getMenuOffset } from 'carbon-components-react/lib/components/OverflowMenu/OverflowMenu';
>             ^
> ```

Oh weird ... I'm curious if this is something we should be poking Carbon to deal with upstream, or if we're just missing a rollup config (I know some tools have a "synthetic" default export for cases like that). I'm also not sure why it's only showing up in #2277, from what I can see there's nothing in there that would directly cause the issue. It's possible it's an edge case in `carbon-components-react` that no ones hit yet?

Looking at the source for invariant - https://github.com/zertosh/invariant/blob/master/invariant.js - That `module.exports` is definitely going to give us a bad time ... `defaultIsModuleExports` may help? https://github.com/rollup/plugins/tree/master/packages/commonjs#defaultismoduleexports

kevinsperrine:
> > > > > Testing the build on next it does look like this error was the present before these changes.
> > > > 
> > > > 
> > > > Is that at all related to the weird failures I'm seeing on [this PR](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2277/checks?check_run_id=2477129205)
> > > 
> > > 
> > > It's plausible ... I'd do a `yarn why` on the packages giving you trouble and see where they're getting placed ... Ideally everything should be installed to the root, so if it's getting split for some reason that could be where the problems coming from.
> > 
> > 
> > No such luck. yarn why shows it all being in the root, but it has something to do with the namedExports config in rollup, I think. It's unable to find getMenuOffset in `lib`, but if I switch it to `es` it finds it, and then chokes on `invariant` b/c in Carbon it's imported as `import invariant from 'invariant';`, but technically invariant doesn't have a default export, so it fails to rollup. No idea why it's happen on PR #2277, but not other places.
> > ```js
> > [!] Error: 'getMenuOffset' is not exported by node_modules/carbon-components-react/lib/components/OverflowMenu/OverflowMenu.js
> > https://rollupjs.org/guide/en#error-name-is-not-exported-by-module-
> > src/components/OverflowMenu/index.jsx (4:9)
> > 2: import { OverflowMenu as CarbonOverflowMenu } from 'carbon-components-react';
> > 3: import PropTypes from 'prop-types';
> > 4: import { getMenuOffset } from 'carbon-components-react/lib/components/OverflowMenu/OverflowMenu';
> >             ^
> > ```
> 
> Oh weird ... I'm curious if this is something we should be poking Carbon to deal with upstream, or if we're just missing a rollup config (I know some tools have a "synthetic" default export for cases like that). I'm also not sure why it's only showing up in #2277, from what I can see there's nothing in there that would directly cause the issue. It's possible it's an edge case in `carbon-components-react` that no one has hit yet?
> 
> Looking at the source for invariant - https://github.com/zertosh/invariant/blob/master/invariant.js - That `module.exports` is definitely going to give us a bad time ... `defaultIsModuleExports` may help? https://github.com/rollup/plugins/tree/master/packages/commonjs#defaultismoduleexports

Yeah, I have no idea what caused it either; however, by switching it to the follow it worked just fine. 🤷‍♂️ 

```jsx 
import * as OM from 'carbon-components-react/lib/components/OverflowMenu/OverflowMenu';
const { getMenuOffset } = OM;
```

davidicus:
I was able to verify that the builds work in a test application. Where are we leaving off with the comments above? Is this something we need to add to this PR?

davidicus (approved):

PR #2274, opened 2021-04-27
feat(list-builder): new list-builder component

### LINKED ISSUE #2101: [List builder]
### What package is this for?

- [X] React
- [ ] Angular

### Summary

IBM Maximo MRO Inventory Optimization will shortly require the List builder component as specified in the design https://pages.github.ibm.com/ai-applications/design/components/listbuilder/overview/

We would like to be able to start using this component in April.

### You will consider this component complete when...

- [X] It's available as an export of this library
- [ ] Additional criteria you may have [e.g. Fully responsive, Accessible to
      WCAG 2.0, etc.]

### Available extra resources

It's possible that our team can help to implement this component,

# Update: 

Our select users modal seems to fit requirement based on [chat]([url](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2101#issuecomment-796804927)) below. Updating this issue to export the component for use.

kevinsperrine:
Closes #2101 

**Summary**

- Pulled code out of the SelectUsersModal to turn it into a generic 2-column ListBuilder component

- initial addition of ListBuilder component
- add ListBuilder into SelectUsersModal
- fix/add items necessary to allow ListBuilder to work inside the SelectUsersModal

- New ListBuilder stories to function as titled
- SelectUsersModal should work as expected

- previous RuleBuilder/SelectUsersModal tests should pass
- RuleBuilder stories should match what's on next currently


packages/react/src/components/ListBuilder/_list-builder.scss
```diff
@@ -0,0 +1,26 @@
+@import '../../globals/vars';
+
+.#{$iot-prefix}--list-builder__container {
+  width: 100%;
+  display: grid;
+  grid-template-columns: 50% 50%;
+
+  // /* Hide the unused nested modal reordering dialog that throws off the scrolling */
```
sstone2423, line 8: Should this get removed?

sstone2423 (approved):
Looks great. Just 1 comment

packages/react/src/components/ListBuilder/_list-builder.scss
kevinsperrine: Yep. Removed. Thanks.

davidicus:
Does the spec call for "-" icon to remove items from the list?
image

kevinsperrine:
> Does the spec call for "-" icon to remove items from the list?
> image

@davidicus Hmm. Yep, that spec does, but the user modal (that this was extracted from) uses the arrows. https://ibm.invisionapp.com/share/2YO0PTTRHKU#/screens/319789283 Should I use the minus and change the user modal or make it a prop to change the icon?

bjornalm:
image
I noticed there is a scrollbar on the topmost list item. I don't think it was introduced in this PR but since we are building a component from it now we probably want to get rid of it.

packages/react/src/components/ListBuilder/ListBuilder.jsx
```diff
@@ -0,0 +1,219 @@
...
+  return (
+    <div className={`${iotPrefix}--list-builder__container`}>
+      <div className={`${iotPrefix}--list-builder__all`} data-testid={`${testID}__all`}>
+        <HierarchyList
+          title={allListTitle(itemCount ?? allListItems.length)}
+          items={allListItems}
+          hasSearch
+          hasPagination={false}
+          searchId={`${iotPrefix}--list-builder__all--search`}
+          i18n={{
```
bjornalm, line 196: How do we change the other i18n strings in the HierarchyList if we for instance want to support another language?

kevinsperrine:
> image
> I noticed there is a scrollbar on the topmost list item. I don't think it was introduced in this PR but since we are building a component from it now we probably want to get rid of it.
> 
> image
> There is also a scrollbar at the bottom of the right list.
> 
> image
> In the listbuilder story both lists have horizontal scrolling

@bjornalm What browser are you seeing the horizontal scroll bars? I don't have those. I see the vertical on in the user modal, though.

packages/react/src/components/ListBuilder/ListBuilder.jsx
kevinsperrine: Good catch. I'll add expand/close. The others aren't used b/c they aren't exposed.

bjornalm:
...
> > In the listbuilder story both lists have horizontal scrolling
> 
> @bjornalm What browser are you seeing the horizontal scroll bars? I don't have those. I see the vertical on in the user modal, though.

Chrome 90.0.4430.93 on MacOS 11.2.2, but now it doesn't appear anymore.

bjornalm (approved):
Nice work

davidicus (changes requested):
I think we are missing the export from index.js

kevinsperrine:
> I think we are missing the export from index.js

Yup. Totally forgot about that. Fixed now.

davidicus (approved):

PR #2277, opened 2021-04-29
feat(header): automatic overflow handling for header actions

### LINKED ISSUE #1845: [ui-shell] Small breakpoint issues
### What package is this for?

- [ ] React

### Describe the bug

• On small screens, text wraps instead of ellipses
• Hamburger and other icons reduce in width
Screen Shot 2020-12-01 at 3 08 00 PM

### To Reproduce

Not reproducable at [IoT storybook](https://carbon-addons-iot-react.com/?path=/story/watson-iot-header--header-subtitle) because there are not enough icons. But, with MAS fully hooked up, the number of icons cause issues. Example can be seen in Graphite which uses the full MAS set of icons
1. Go to 'https://pages.github.ibm.com/maximo-app-framework/graphite/reference-app/#/about'
2. Make screen small or go into responsive mode
3. See error

### Expected behavior

While we don't necessarily support mobile in all apps, it is inevitable that a customer will report this as a bug. Clients and business partners may also choose to create their own simple Graphite/Maximo application for usage on the phone. The masthead should hide some of the less necessary text, or truncate down appropriately with the Carbon smaller breakpoints

Would be nice to fix this for MAS 8.3

kevinsperrine:
Closes #1845 

**Summary**

- add overflow menu check to the HeaderActionGroup component that automatically moves all HeaderActions into an overflow menu when they run out of space. 
- Adds `shortAppName` prop to the header to use in place of the long name on tablet/mobile breakpoints
- Adds `inOverflow` prop to HeaderAction, HeaderActionMenu, and HeaderActionPanel components to render only the label in the overflow menu without the icon
- Adds `defaultExpanded` prop to HeaderAction to force a submenu to be open when clicked from the overflow menu
- Adds `onClose` prop to HeaderAction to callback to the HeaderActionGroup to close the overflow menu and submenu properly when closing.

See [invision prototype](https://ibm.invisionapp.com/share/6KO0UT1TZUY#/screens/319846452_4-_Responsive_Behaviours) for design specs

- The `small breakpoint` story should collapse all icons into an overflow menu when you shrink the window
- Clicking "help" should open the side panel
- Clicking "user" should replace the overflow menu with the user menu
- Clicking any of the other items should trigger actions in the action tab.

- All tests pass and other Header/SuiteHeader stories function as expected


kevinsperrine:
@JordanWSmith15 I'll much appreciate your 👀  on this PR, please. 

You can see it in action [here](https://deploy-preview-2277--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-header--small-breakpoint) just shrink the window down to see the icons collapse into an overflow menu.

JordanWSmith15:
@kevinsperrine I did some testing on this, and it does behave much better.
The only issue I see is the text still does not truncate. If it is long, it will blow out the container and make the three-dot menu disappear:
Screen Shot 2021-04-30 at 10 23 04 AM
Out of the box, we clearly would not ship long strings like this, but clients have the ability to configure the header, and someone will inevitably open an issue with support. We should add knobs in storybook to test this. Ideally, the text should truncate to an ellipsis (...). At minimum, it should not push the overflow menu out of the container, but instead just wrap or flow behind it.

kevinsperrine:
> @kevinsperrine I did some testing on this, and it does behave much better.
> The only issue I see is the text still does not truncate. If it is long, it will blow out the container and make the three-dot menu disappear:
> Screen Shot 2021-04-30 at 10 23 04 AM
> Out of the box, we clearly would not ship long strings like this, but clients have the ability to configure the header, and someone will inevitably open an issue with support. We should add knobs in storybook to test this. Ideally, the text should truncate to an ellipsis (...). At minimum, it should not push the overflow menu out of the container, but instead just wrap or flow behind it.

Thanks, @JordanWSmith15. I've added truncation to the header, as well as knobs (on the small breakpoint story) for testing long names.

bjornalm:
I understand this is a bit tricky, but now the header is truncated before the overflow menu is shown. 
image

davidicus (approved):

packages/react/src/components/Header/Header.test.jsx
```diff
@@ -285,4 +286,99 @@ describe('Header', () => {
     fireEvent.click(menuItem);
     expect(screen.getByLabelText('user')).toBeTruthy();
   });
+
+  it('should move icons into overflow menu when area too small', () => {
+    Object.defineProperty(window, 'innerWidth', { writable: true, configurable: true, value: 375 });
```
bjornalm, line 291: Should we restore this once done testing?

packages/react/src/components/Header/HeaderAction/HeaderAction.jsx
```diff
@@ -94,21 +112,25 @@ const HeaderAction = ({ item, index, testID }) => {
...
           />
         ) : (
           // otherwise render a submenu type dropdown
           <HeaderActionMenu
             className={`${carbonPrefix}--header-action-btn`}
             key={`menu-item-${item.label}`}
             aria-label={item.label}
-            renderMenuContent={() => item.btnContent}
+            renderMenuContent={() =>
+              isExpanded ? <Close16 fill="white" description="close-icon" /> : item.btnContent
```
bjornalm, line 124: Is there perhaps a carbon color that we should use here, instead of just white?
Also, the description should probably use i18n.

packages/react/src/components/Header/HeaderAction/HeaderActionPanel.jsx
```diff
@@ -46,7 +54,13 @@ const HeaderActionPanel = ({ item, index, onToggleExpansion, isExpanded, focusRe
         onClick={() => onToggleExpansion()}
         ref={focusRef}
       >
-        {item.btnContent}
+        {inOverflow ? (
+          item.label
+        ) : isExpanded ? (
+          <Close16 fill="white" description="close-icon" />
```
bjornalm, line 60: Same comments as for previous close icon

packages/react/src/components/Header/HeaderActionGroup.jsx
```diff
@@ -16,19 +24,146 @@ const propTypes = {
...
+  const [overflowItems, setOverflowItems] = useState([]);
+  const breakpoint = useRef(null);
+  const [overflowOpen, setOverflowOpen] = useState(false);
+  const [showMenu, setShowMenu] = useState(false);
+  const [menu, setMenu] = useState(null);
+
+  const checkForOverflow = useCallback(() => {
+    /* istanbul ignore else */
+    if (overFlowContainerRef.current) {
+      // the first button in the action group
```
bjornalm, line 37: Can we skip the comment and change the code to say the same thing?

packages/react/src/components/Header/HeaderActionGroup.jsx
```diff
@@ -16,19 +24,146 @@ const propTypes = {
...
+                      ],
+                      overFlowContainerRef.current.parentNode
+                    );
+                  }
+
+                  return (
+                    <OverflowMenuItem
+                      title={child.label}
+                      key={`${child.label}-${i}`}
+                      className=""
```
bjornalm, line 110: empty className

packages/react/src/components/Header/HeaderActionGroup.jsx
```diff
@@ -16,19 +24,146 @@ const propTypes = {
...
+                  ) {
+                    ReactDOM.createPortal(
+                      [
+                        <HeaderAction
+                          item={child}
+                          index={i}
+                          key={`header-action-item-${child.label}-${i}`}
+                          testID={`header-action-item-${child.label}`}
+                          // used to render only the label in the overflow menu instead of the icon
+                          inOverflow
```
bjornalm, line 99: why not use the name "renderLabel" for this prop instead?

packages/react/src/components/Header/_header.scss
```diff
@@ -31,6 +58,56 @@
...
     outline: none;
   }
+
+  .#{$iot-prefix}--header__overflow-menu-container {
+    position: relative;
+  }
+
+  .#{$prefix}--overflow-menu {
+    &:hover {
+      background-color: $gray-90;
```
bjornalm, line 68: Not sure we need to support this, but is there a non color specific carbon name for the gray-90 that works with the different themes? E.g. `$hover-secondary`. Perhaps this is a question for the design team. It would be nice if the invision mocks came with the name of the colors.

bjornalm (approved):
Nice work! Just a few comments/questions.

davidicus (changes requested):
Bjorn has a few comments that should be addressed

packages/react/src/components/Header/_header.scss
kevinsperrine: Nope, not that I can find. grepping through all the stuff just shows carbon using the raw variables in ui-shell, too. 
```
./node_modules/@ai-apps/styles/scss/vendor/carbon-components/scss/components/ui-shell/_navigation-menu.scss:    background-color: $ibm-color__gray-90;
./node_modules/@ai-apps/styles/scss/vendor/carbon-components/scss/components/ui-shell/_theme.scss:/// TODO needs color var gray-90-hover
./node_modules/@ai-apps/styles/scss/vendor/carbon-components/scss/components/ui-shell/_theme.scss:$shell-header-bg-06: $carbon--gray-90;
./node_modules/@ai-apps/styles/scss/vendor/carbon-components/scss/components/ui-shell/_theme.scss:$shell-panel-bg-04: $carbon--gray-90;
./node_modules/@ai-apps/styles/scss/vendor/carbon-components/scss/components/ui-shell/_theme.scss:/// TODO needs color var gray-90-hover
```

packages/react/src/components/Header/HeaderAction/HeaderAction.jsx
kevinsperrine: Good call. Replaced with carbon white, and added description via i18n prop

packages/react/src/components/Header/HeaderActionGroup.jsx
kevinsperrine: Done.

packages/react/src/components/Header/HeaderActionGroup.jsx
kevinsperrine: Absolutely. #namingishard.

packages/react/src/components/Header/HeaderActionGroup.jsx
kevinsperrine: deleted.

packages/react/src/components/Header/HeaderAction/HeaderActionPanel.jsx
kevinsperrine: same as other.

packages/react/src/components/Header/Header.test.jsx
kevinsperrine: good eye.

davidicus (approved):

packages/react/src/components/OverflowMenu/index.jsx
```diff
@@ -1,12 +1,15 @@
 import * as React from 'react';
 import { OverflowMenu as CarbonOverflowMenu } from 'carbon-components-react';
 import PropTypes from 'prop-types';
-import { getMenuOffset } from 'carbon-components-react/lib/components/OverflowMenu/OverflowMenu';
+// no idea why import { getMenuOffset } won't work, but it causes an 'getMenuOffset' is not exported by node_modules/carbon-components-react/lib/components/OverflowMenu/OverflowMenu.js error
+import * as OM from 'carbon-components-react/lib/components/OverflowMenu/OverflowMenu';
```
herleraja, line 5: This may have caused some issues downstream. Wild card imports are not suggested.

PR #2281, opened 2021-04-30
Dashboard editor custom card side bar

JoelArmendariz:
**Summary**
Expands the support of CustomCard types in the dashboard editor to allow the consumer to specify what they want to render in both the content editor and the settings editor.

 - Use two new properties on the payload of the `CUSTOM` card type (`renderEditContent` & `renderEditSettings`) to render custom editor elements.
 - Include the `icons` prop used in the `CardGalleryList` as a top level prop for use with the Custom card type.

Go to the `DashboardEditor` `With custom card` story and see that there is a new type in the gallery list called `My custom type`. 
Verify that the custom content and settings editors on the right properly update the CustomCard content in the dashboard.

Make sure the other defined cards still act as expected.

image


scottdickerson:
You attached the DashboardEditor control callbacks to the actual card itself so that multiple different custom cards can be used in the DashboardEditor with different Content and Settings Panels (smart!), but does it actually work?  Can you extend the Dashboard Custom Card story to show these two types of custom cards in the gallery and verify that each one can use a different set of forms to edit?

JoelArmendariz:
> @JoelArmendariz You attached the DashboardEditor control callbacks to the actual card itself so that multiple different custom cards can be used in the DashboardEditor with different Content and Settings Panels (smart!), but does it actually work? Can you extend the Dashboard Custom Card story to show these two types of custom cards in the gallery and verify that each one can use a different set of forms to edit?

@scottdickerson I have added a 2nd custom card to the story with its own edit panels. You can go into each and update them separately.

scottdickerson (approved):
LGTM

davidicus (approved):
Any ideas why those snapshots are changing for ImageCard and HeirarchyList

PR #2283, opened 2021-05-03
fix(ListHeader): should not have loading state

sstone2423:
**Summary**

- If List is asynchronously searching for items, the searchbar (and the rest of the header) should not change on `isLoading`. If the loading state triggers, the search input value and focus is lost, causing a bad UX

- Remove isLoading prop from ListHeader

- Go to /?path=/story/watson-iot-hierarchylist--stateful-list-with-nested-searching , turn the isLoading knob to true and see that the search bar is not in a loading state

- Should not affect existing functionality


kevinsperrine (approved):

herleraja (approved):
LGTM

davidicus (approved):

PR #2284, opened 2021-05-03
fix(simplepagination): add aria-label

sstone2423:
**Summary**

- The `description` prop doesn't seem to have any affect on `svg` elements so it has been swapped to use `aria-label` so that it can be queried and read by screen readers

- The next and previous button should now have `aria-label` set

- Shouldn't affect existing functionality

### DIFF
--- a/packages/react/src/components/List/HierarchyList/__snapshots__/HierarchyList.story.storyshot
+++ b/packages/react/src/components/List/HierarchyList/__snapshots__/HierarchyList.story.storyshot
@@ -427,14 +427,14 @@ exports[`Storybook Snapshot tests and console checks Storyshots Watson IoT/Hiera
                 tabIndex={-1}
               >
                 <svg
-                  aria-hidden={true}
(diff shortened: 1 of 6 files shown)


packages/react/src/components/SimplePagination/_simple-pagination.scss
```diff
@@ -7,7 +7,7 @@
   width: 100%;
   border: rem(1px) solid $ui-03;
   background-color: $ui-01;
-  align-items: center;
+  //align-items: center;
```
kevinsperrine, line 10: remove the comment?

kevinsperrine (approved):

packages/react/src/components/SimplePagination/_simple-pagination.scss
sstone2423: Meant to remove that. Thanks!

bjornalm (approved):
Nice work

PR #2288, opened 2021-05-03
fix(table): set tabindex to -1 on batch actions when hidden

### LINKED ISSUE #1980: [Table] Can <Tab> to hidden batch action buttons
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When the batch action buttons are hidden, then still have tabIndex="0" and can be <Tab>ed to. This makes keyboard nav harder and allows those actions to be triggered with no selection.

### To Reproduce

Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--with-multi-select-and-batch-actions
Deselect all rows.
Shift-Tab to Filter button.
Shift-Tab again. Visual focus is lost.
Hit spacebar. JavaScript console show an action event.
Dump document.activeElement in the console to see that a batch action button is focued. (Be sure the pick the right iframe in JS console first).

CodeSandbox is broken for me with CORS errors loading Carbon since about Jan 25 2021.

### Expected behavior

Visually hidden buttons should not be accessible via the keyboards. All buttons/controls could be set to tabIndex=-1, but making the entire batch toolbar area display:none seems simpler.

kevinsperrine:
Closes #1980 

**Summary**

- set tabindex to -1 on batch action buttons when batch actions are hidden
- fix data-testid typo on TableBatchActions bar.

- follow steps on #1980 to ensure that batch actions are not available to tab to.


herleraja (approved):
LGTM

jkeen:
Isn't it more appropriate to set disabled on the buttons in this case, given that they shouldn't be able to be used and are visually hidden (but not display: none probably due to effects?). Would a screen reader still think these buttons are good enough to interact with or no? Either way pointer-events: none and tabindex=-1 kinda gets to the same place as disabled but in a less direct way imo

kevinsperrine:
> Isn't it more appropriate to set `disabled` on the buttons in this case, given that they shouldn't be able to be used and are visually hidden (but not `display: none` probably due to effects?). Would a screen reader still think these buttons are good enough to interact with or no? Either way `pointer-events: none` and `tabindex=-1` kinda gets to the same place as `disabled` but in a less direct way imo

@jkeen maybe. I'm honestly not sure great a11y stuff, so I don't know if a screen reader would still access it in disabled state, or if it's tabbable in disabled state. My reasoning with tabindex was because we didn't want it to be tabbable since it's visually hidden in this scenario.

jkeen:
> My reasoning with tabindex was because we didn't want it to be tabbable since it's visually hidden in this scenario.

screen reading aside (because I don't know for sure either) disabled elements automatically aren't triggerable, clickable, focusable, tabbable, etc, which might make that route more semantically correct markup wise? This definitely isn't a hill I'll die on but it just seems like we've got two properties (tabindex + pointer-events) doing a partial job where only disabled is needed

kevinsperrine:
> it just seems like we've got two properties (tabindex + pointer-events) doing a partial job where only disabled is needed

good point. I'll play around with a screenreader and make any adjustments.

kevinsperrine:
@jkeen Disabled works and the screen reader skips over it just the same as with tabindex=-1. However, carbon uses tabindex=-1 also for the Cancel button that gets automatically added by the [TableBatchActions component](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/DataTable/TableBatchActions.js#L65). I have no idea why they make that choice though. 🤷‍♂️  These are the kinds of issues why I've never understood a11y stuff. I feel like there are too many ways to approach it and I never know which one is preferred in which case.

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -211,15 +211,21 @@ const TableToolbar = ({
...
       shouldShowBatchActions={hasRowSelection === 'multi' && totalSelected > 0}
       totalSelected={totalSelected}
       translateWithId={(...args) => tableTranslateWithId(i18n, ...args)}
     >
       {batchActions.map(({ id, labelText, ...others }) => (
-        <TableBatchAction key={id} onClick={() => onApplyBatchAction(id)} {...others}>
+        <TableBatchAction
+          key={id}
+          onClick={() => onApplyBatchAction(id)}
+          tabIndex={hasRowSelection === 'multi' && totalSelected > 0 ? 0 : -1}
```
bjornalm, line 225: We write `hasRowSelection === 'multi' && totalSelected > 0` 3 times here including the prop `shouldShowBatchActions`. Perhaps it is better to place it in a const.

bjornalm (changes requested):

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
kevinsperrine: good call. I didn't do it initially, since it made the diff look like way more changes than it was. Done now.

davidicus (approved):

PR #2295, opened 2021-05-06
feat(Tablecard): allow passing of filters

### LINKED ISSUE #1409: [TableCard] need to be able to pass a custom column filter to a TableCard
### Describe the bug

If a user has a TableCard that needs custom filters based on the data, allow them to render custom filter choices

if a tablecard has 'discrete' values like status we should be able to do an exact match on a column (from a pick list) rather than a wildcard match

jkeen:
Closes #1409

**Summary**
Pass filter options from TableCard along to underlying Table. 

** Acceptance test **
https://deploy-preview-2208--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-tablecard--with-custom-filters
- only one table row should be rendered. Clearing filters should show all rows.

### DIFF
--- a/packages/react/src/components/TableCard/TableCard.story.jsx
+++ b/packages/react/src/components/TableCard/TableCard.story.jsx
@@ -108,10 +108,10 @@ WithLinks.story = {
 
   parameters: {
     info: {
-      text: `<p>Links can added by providing a linkTemplate prop to the content.columns[i] property. 
-                2 additional properties can be configured within the linkTemplate object: href and target</p> 
+      text: `<p>Links can added by providing a linkTemplate prop to the content.columns[i] property.
+                2 additional properties can be configured within the linkTemplate object: href and target</p>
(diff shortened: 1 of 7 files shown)


packages/react/src/components/TableCard/TableCard.story.jsx
```diff
@@ -1015,6 +1015,34 @@ EditableWithExpandedRows.story = {
...
+      <TableCard
+        title="Open Alerts"
+        content={{
+          columns: tableColumns,
+        }}
+        values={tableData}
+        filters={[{ columnId: 'alert', value: 'failure' }]}
+        size={CARD_SIZES.LARGE}
+      />
+      );
```
kevinsperrine, line 1037: have an extra `);` here.

kevinsperrine (dismissed):

bjornalm (approved):
Nice work

davidicus (approved):

PR #2296, opened 2021-05-06
fix(table): need to handle null values in filtering

scottdickerson:
Closes #https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/2132

**Summary**

- If a column had null values in it (only currently happens with numeric values I believe) then attempting to perform a column filteration would throw an error to the console and crash the component
``` TypeError: te is null
    defaultComparison tableReducer.js:43
    filterData tableReducer.js:76
    filterData tableReducer.js:59
    filterData tableReducer.js:56
    filterSearchAndSort tableReducer.js:122
    tableReducer tableReducer.js:156
    React 2
    StatefulTable StatefulTable.jsx:64
    React 7
    unstable_runWithPriority scheduler.production.min.js:19
    React 5
```

- story(Table): change the default generated values to contain some null values in the number column
- fix(tableReducer): verify the toString actually exists before attempting to sort on both values, handle null values in defaultComparison function as well as runSimpleFilters function

- Verify that column filtering on the number column in the table stories no longer throws an error

- Verify that sorting and filtering continue to work correctly for other column types

### DIFF
--- a/packages/react/src/components/Table/Table.story.jsx
+++ b/packages/react/src/components/Table/Table.story.jsx
@@ -335,7 +335,7 @@ const getNewRow = (idx, suffix = '', withActions = false) => ({
     date: new Date(100000000000 + 1000000000 * idx * idx).toISOString(),
     select: selectData[idx % 3].id,
     secretField: getString(idx, 10) + suffix,
-    number: idx * idx,
+    number: idx % 3 === 0 ? null : idx * idx,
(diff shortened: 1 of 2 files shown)


herleraja (approved):
LGTM. I needed this too.

bjornalm (approved):

kevinsperrine (approved):

Issue #1019, opened 2020-03-18
[table] column resizing limited by row action column


bjornalm:
### Describe the bug
For a table with row actions, when you hide columns, the right-hand side of the table is still considered occupied, and you cannot drag the furthest-right column to fill the space.

Reason:
A table with row actions will get an additional column appended. This column will be assigned all the remaining width. When normal columns are hidden, this row action column will occupy the new width and since it cannot be resized itself we end up in the problem described above.

See gif:
https://ibm.ent.box.com/s/q36zlk7mfdc56xlh8par51mnan0jdlue/file/631494139544
### To Reproduce

1. Open a table that allows both column resizing and column reordering/hide/show
2. Hide a couple of columns
3. Notice how the row action column fills up the new space
4. Show the hidden columns and notice how the row action column does not shrink and cannot be resized.

### Expected behavior
The complete expected behavior is not 100% clear to me, but perhaps this should be solved in two steps.

a) The row actions column will not be allowed to grow when columns are hidden by the user, i.e. the row actions column will remain the width it had before any user initiated show/hide of columns. The row actions column should also remain the same width when normal columns are resized. This would solve the issue as described and is already implemented in the solution for issue [1016](https://github.com/IBM/carbon-addons-iot-react/issues/1016) 

b) Although the above might solve the immediate problem, we probably want to go beyond this since the row action column could never be made smaller than its initially rendered width. This can become an issue in scenarios like the following:

A table has 2 columns defined, both 100 px wide. The second column is hidden by default. The table is rendered in a container that is 300 px wide. The first column renders 100px wide and the row action column gets the remaining 200 pixels (although it only needs 50px). Now the user toggles the previously hidden second column to be shown. Since the row action column never changes in size the space needed for the second column will be taken from the other columns, which in this case is only the first column. This results in the first column getting a width of 0.

Even if the width distribution algorithm is improved to distribute the available 100px in a smarter way over the two normal columns, it will seem odd that the action column gets to keep its space. Figuring out how and when to "steal" space from the row actions column could be possible... but now it is getting very complicated. One solution is to allow a suboptimal column distribution but also allow the row actions column to be resized... or find a completely different design for this problem.

### Edit:
There has been a lot of changes to column hide/show that we should be aware of but I think this issue still persist. The Action button column is taking up more than 50px. Let's verify that this issue persist and if so apply the solution described by @dianatran18

bjornalm:
[...]

Issue #2061, opened 2021-02-23
[Table] Column resize issue when columns are removed and then readded


RogerRosado:
### What package is this for?

- [X ] React
- [ ] Angular

### Describe the bug

Stateful Example with expansion and column resize
When removing and adding columns with the column selection widget, the columns on the left narrows when they are added back in.

### To Reproduce

1. Go to https://next.carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-expansion-max-pages-and-column-resize
2. Select Column selection widget to activate column headers
3. From left to right deselect each of the columns **except** for the last one
4. Next from left to right add the columns back
5. Will notice that the columns on the left narrows, while the columns towards the right widens

### Expected behavior

Smart evenly spaced columns

### Environment/versions:

Carbon Storybook

https://user-images.githubusercontent.com/56408920/108897065-b040e100-75e3-11eb-8f3f-96092f6dffb4.mp4

### Specific timeline issues / requests

 MAS 8.4 April

JordanWSmith15:
@davidicus This is effecting our Health release for 8.4.   I am not sure if this is a regression or not, but it's having some impact

JordanWSmith15:
![Screen Shot 2021-02-23 at 4 17 58 PM](https://user-images.githubusercontent.com/21066152/108909575-b68a8980-75f2-11eb-86da-a539d4768a0c.png)

RogerRosado:
The issue is now that the columns do not resize when you add the columns back in from left to right. 

Instructions:
1. Remove all columns from left to right and leave the last one "React Node" visible. 
2. Next add the first column back in which is noted as "String".

There is a big space on the right side of the table while the first column is either wrapped or truncated, depending on the wrapCellText setting:

![image](https://user-images.githubusercontent.com/56408920/112037577-3d5e5380-8b18-11eb-8e1e-2bfdaf5cec65.png)

JordanWSmith15:
@davidicus FYI

davidicus:
@JordanWSmith15 @RogerRosado I am not seeing what the issue is. The behavior seems to be consistent to what it was before this issue was fixed. When I add columns back in following the instructions above I am still able to resize the columns. The last column was never able to be resized.

PR #2306, opened 2021-05-07
fix(table): resizable table handles window width changes

### LINKED ISSUE #1974: [Table] Column resize broken is table width changes
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The Table component seems to take a snapshot of column widths on first render. If you change the width of the table after that, then resizing the columns doesn't size them to match your cursor position. In extreme cases the columns "freak out" and gets super huge, crushing other columns.

This affects tables with `props.options.hasResize = true` and a fluid width.

### To Reproduce

- Make a Chrome window half width of the screen.
- View this Storybook: https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--with-resize-and-initial-column-widths-on-simple-stateful-with-row-selection-sort
- Maximize the window to full screen.
- Resize the columns. Ouch.

### Environment/versions:

- OS: all
- Browser: all
- `carbon-addons-iot-react` version: 2.98.0 (at least)

This is a regression compared the the current (non-PAL) table components being used in my app. Ideally this should be fixed before the next release (about 6 months time).

bjornalm:
Closes #1974

**Summary**
Added fix to make sure table columns can be manipulated (resized, hidden, shown) after the window has been resized. The main scenario solution is to add an "expander column" that grows and shrinks when a resizable table is resized.  For scenarios with `useAutoTableLayoutForResize` (i.e. resizable columns + table-layout:auto) where the browser resizes all the columns automatically there will be no expander column and we use a window resize listener to keep the actual column sizes and react code sizes in sync.

This PR also solved #1019

- added expander column to grow and shrink for resizable tables (unless the table uses `useAutoTableLayoutForResize` which makes resizing of all columns automated)
- added window resize listener for resizable tables with `useAutoTableLayoutForResize`. Resize listener measures columns when the window is resized. 
- tested all storybook scenarios 
- added test cases
- added role and aria-label to resize handle for better UI and easier testing
- added dev dependency @testing-library/react-hooks to help with hook testing

1. Test story from the issue and make sure the bug does not appear. Verify that column resize works. https://deploy-preview-2306--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-table--with-resize-and-initial-column-widths-on-simple-stateful-with-row-selection-sort 
2. Test a story with resizable columns + initial column widths.  Verify that each column gets the width assigned and that resizing the browser grows and shrinks the expander column and not the real columns. Verify that column resize works.  https://deploy-preview-2306--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--with-resize-and-initial-column-widths-on-simple-stateful-with-row-selection-sort
3. Test a story with resizable columns that do not have initial widths. Verify that column resize works. Verify that the expander column doesn't occupy any space before the browser window is resized.   https://deploy-preview-2306--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-max-pages-and-column-resize
4. Testa a story with resizable columns but that also has `useAutoTableLayoutForResize`. Make sure resizing works. Verify that all columns are resized when the browser is resized. Verify that there is no expander column. https://deploy-preview-2306--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--with-resize-and-no-initial-column-width-and-auto-adjusted-column-widths
5. Test a story without column resize and make sure it behaves like normal, i.e. columns are resized when the browser window is resized. https://deploy-preview-2306--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-table--stateful-example-with-every-third-row-unselectable

Try **all** (sorry) different table stories to make sure they look and act fine on render and on window resize. Only the last column should be resized for stories with `hasResize` unless `useAutoTableLayoutForResize` is used (which is only one story).


kevinsperrine (commented):
This works great almost everywhere, but I am running into some issues resizing the "expander" after hiding/showing columns in the table. We changed the behavior of this due to a bug report opened while you were out (see  #2061). It now auto-distributes the space available to all columns rather than trying to reinsert the column with the same width as when it was removed. The "expander" column seems to fail to resize correctly after columns have been removed and re-added. 

This is from https://deploy-preview-2306--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-table--stateful-example-with-expansion-max-pages-and-column-resize.

expandable-column-resize

bjornalm:
> This works great almost everywhere, but I am running into some issues resizing the "expander" after hiding/showing columns in the table. We changed the behavior of this due to a bug report opened while you were out (see #2061). It now auto-distributes the space available to all columns rather than trying to reinsert the column with the same width as when it was removed. The "expander" column seems to fail to resize correctly after columns have been removed and re-added.
> 

Good catch. Looks like this bug already exists in 'next', i.e. resizing fails after **all** columns are first hidden and then (some are) shown again. This is bug has probably been there for a while is unrelated to the current changes. I would suggest that we open a new issue to fix that problem.

kevinsperrine:
> > This works great almost everywhere, but I am running into some issues resizing the "expander" after hiding/showing columns in the table. We changed the behavior of this due to a bug report opened while you were out (see #2061). It now auto-distributes the space available to all columns rather than trying to reinsert the column with the same width as when it was removed. The "expander" column seems to fail to resize correctly after columns have been removed and re-added.
> 
> Good catch. Looks like this bug already exists in 'next', i.e. resizing fails after **all** columns are first hidden and then (some are) shown again. This is bug has probably been there for a while and is unrelated to the current changes. I would suggest that we open a new issue to fix that problem.

Perfect.

kevinsperrine (approved):

kevinsperrine:
Great work!

davidicus:
> > This works great almost everywhere, but I am running into some issues resizing the "expander" after hiding/showing columns in the table. We changed the behavior of this due to a bug report opened while you were out (see #2061). It now auto-distributes the space available to all columns rather than trying to reinsert the column with the same width as when it was removed. The "expander" column seems to fail to resize correctly after columns have been removed and re-added.
> 
> Good catch. Looks like this bug already exists in 'next', i.e. resizing fails after **all** columns are first hidden and then (some are) shown again. This is bug has probably been there for a while and is unrelated to the current changes. I would suggest that we open a new issue to fix that problem.

Has this bug been opened?

davidicus (approved):
This PR looks good. Let's open the bug to track the issue mentioned in the thread

bjornalm:
> > > This works great almost everywhere, but I am running into some issues resizing the "expander" after hiding/showing columns in the table. We changed the behavior of this due to a bug report opened while you were out (see #2061). It now auto-distributes the space available to all columns rather than trying to reinsert the column with the same width as when it was removed. The "expander" column seems to fail to resize correctly after columns have been removed and re-added.
> > 
> > 
> > Good catch. Looks like this bug already exists in 'next', i.e. resizing fails after **all** columns are first hidden and then (some are) shown again. This is bug has probably been there for a while and is unrelated to the current changes. I would suggest that we open a new issue to fix that problem.
> 
> Has this bug been opened?

Yes it was opened when it was discovered. Perhaps I should have linked it from this PR for clarity. Here it is
https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2340

PR #2307, opened 2021-05-07
fix(table-body-row): only show hover cursor on checkbox

### LINKED ISSUE #2014: [TableBodyRow] Cursor style indicates action on hover
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When the table has `isSelectable: true` and you hover a Table Row the cursor changes to a pointer indicating an action like selecting the row. The row is only selectable when clicking on the checkbox. There should be a conditional check of the single selectable prop before adding the styles to change the cursor. The offending line is found [here](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/670bcfbd280c3627758f38ddb01499cdcdd71a52/packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx#L136). The condition should be changed to look for single row  select and not just `isSelectable`

### To Reproduce

1. Go to https://next.carbon-addons-iot-react.com/?path=/story/watson-iot-table--with-multi-select-and-batch-actions
2. Hover over table row
3. See cursor change
4. Click and see there is no action

kevinsperrine:
Closes #2014 

**Summary**

- adjusted styled-component to only show the pointer cursor on selectable row checkbox elements instead of the whole row
- updated snapshots

- following steps in the original issue the pointer cursor should only be visible on the checkbox element not the whole row

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,73 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.145.0-next.76](https://github.com/carbon-design-system/carbon-addons-iot-react/compare/v2.145.0-next.75...v2.145.0-next.76) (2021-05-10)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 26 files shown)


bjornalm (approved):

kevinsperrine:
> 👍 In the future we can take the opportunity to refactor off of styled components.

Sure thing. I can do it now if you want. That sounds like an undertaking that would make for a good epic.

PR #2308, opened 2021-05-07
Add angular table

### LINKED ISSUE #2242: Add helpers for table content alignment
Content can be left, center or right aligned and there's a bit more to it than just changing text-align so it would be nice to give devs a way to do it in a consistent way that we can easily update in case design changes.

https://pages.github.ibm.com/ai-applications/design/components/datatable/columnalignment/

Donisius:
Closes #2242


cal-smith (changes requested):
This is a good start, but given some of the complexities presented by the PAL table I'm not sure this is how we want to approach the table component for the PAL.

The existing table and model have a number of shortcomings, so I think we'd be best served by trying a ground up rethink.

PR #2309, opened 2021-05-07
fix(combo-box): allow item ids to be numbers

### LINKED ISSUE #2041: [Combobox] Number as item id crashes component on selection
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

In `ComboBox` passing a number as `id` will crash the component on selection with following error:
```
downShiftSelectedItem[currentId].trim is not a function
```

This is probably caused due to the fact that a string is expected as `id` and that numbers do not have `trim()`.

### To Reproduce

1. Go to [this Stackblitz](https://stackblitz.com/edit/carbon-dummy-8mmckh?file=components%2FComboBox.jsx)
2. Select an item from  `ComboBox` and click it
3. See that the component will crash with `downShiftSelectedItem[currentId].trim is not a function`

Stackblitz: https://stackblitz.com/edit/carbon-dummy-8mmckh?file=components%2FComboBox.jsx

### Expected behavior

The component should be able to handle numbers as `id` for items, just like its upstream counterpart in `carbon-components-react`.

### Additional context
The upstream `ComboBox` component does not have any issues with this problem as seen in [this CodeSandbox](https://codesandbox.io/s/heuristic-engelbart-fzklm?file=/src/index.js).

### Environment/versions:

- OS: MacOS
- Browser: Chrome
- `carbon-addons-iot-react` version: `v2.143.0-next.42`

kevinsperrine:
Closes #2041 

**Summary**

- Allows items passed to ComboBox to use numbers as their ids

- Clicking "Option 1" in the Default ComboBox story should fire an onChange action with {id: 0, text: 'Option 1'} without any issues.

- all tests pass and no changes to existing ComboBox stories

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.jsx
+++ b/packages/react/src/components/ComboBox/ComboBox.jsx
@@ -113,13 +113,13 @@ const ComboBox = ({
   const handleOnChange = ({ selectedItem: downShiftSelectedItem }) => {
     const newItem =
       downShiftSelectedItem &&
-      Object.keys(downShiftSelectedItem).reduce(
-        (acc, currentId) => ({
+      Object.keys(downShiftSelectedItem).reduce((acc, currentId) => {
+        const value = downShiftSelectedItem[currentId];
+        return {
           ...acc,
-          [currentId]: downShiftSelectedItem[currentId].trim(),
-        }),
-        {}
-      );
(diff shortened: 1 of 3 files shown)


bjornalm (approved):
Nice work

davidicus (approved):

PR #2310, opened 2021-05-07
fix(imagecard): set of small fixes to imagecard and the dashboardeditor

scottdickerson:
Closes #

**Summary**

- Small set of fixes to the Image Card and default settings for the Image Card inside the Dashboard Editor

- fix(editorUtils): default the displayOption to 'contain' as the card works better that way (and hotspots are located in the correct place)
- fix(ImageHotspots): don't call the callback if it's not set (it's not set if you click without pressing CTRL)
- fix(ImageHotspots): have to put the index in the key because you can have two hotspots with the same x and y value and it will throw an console error

- Verify adding new Image Cards inside the DashboardEditor default their Setting for displayOption correctly
- Verify that if you Edit the Image in the Dashboard Editor story, switch to the Labels tab and then click the image without pressing CTRL that no error is logged to the console
- Verify that the new 'text' hotspot shows in all the existing ImageHotspots stories

- Verify that fixed hotspots render correctly in the existing stories


bjornalm (approved):
Nice job! When I worked on this I think the assumption was that the hotspot positions were unique and could be used as IDs, but if they are not there might be other consequences, e.g. for selecting/deselecting hotspots sharing the same position. Did you look into that?

And perhaps we should make this functionality more explicit by writing at least one unit test so that there is no future confusion about it?

I approved the code changes but I haven't been able to do any manual testing since the build has failed.

kevinsperrine (approved):
deploy preview works as expected and the code looks good, but I do wonder about Bjorn's comment (https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2310#pullrequestreview-655290724) re: two hotspots existing at the same point.

scottdickerson:
In this case the two hotspots just render on top of each other so one is usually not visible, and I believe the top one gets the unselected/deselected focus.  However you can filter the image card by device and in that case you'll be able to see either the top hotspot or the bottom hotspot

scottdickerson:
@davidicus can you take a look at this one, I'm just waiting on the code owner review

davidicus (approved):

PR #2325, opened 2021-05-12
feat(suiteheader) Adding a custom hook to wrap the API call that populates SuiteHeader

blechner:
**Summary**

- Adding a custom hook (`useUiResources`) to wrap the API call that retrieves data to populate the `SuiteHeader`. The old `useSuiteHeaderData` has not been removed because some apps might still be using it, but it should no longer be used.

- No react component has been changed, so no new tests have been added.


packages/react/src/components/SuiteHeader/SuiteHeader.story.jsx
```diff
@@ -579,8 +583,70 @@ HeaderWithDataFetching.story = {
   name: 'Header with data fetching',
 };
 
-*/
```
kevinsperrine, line 582: The end comment here was accidentally removed, so the new story examples are commented out.

packages/react/src/components/SuiteHeader/hooks/useSuiteHeaderData.js
```diff
@@ -1,3 +1,5 @@
+// This custom hook is deprecated, use useUiResources instead
```
kevinsperrine, line 1: If you move this to a docblock statement on the function, then IntelliSense will be able to pick it up.

```
/**
 * @deprecated use useUiResources instead
 */
const useSuiteHeaderData = ({
```
Screen Shot 2021-05-18 at 9 47 24 AM

kevinsperrine (changes requested):

packages/react/src/components/SuiteHeader/SuiteHeader.story.jsx
blechner: @kevinsperrine it was not an accident, I've added some more "unsupported" async stories that use the new auxiliary ways to leverage the `GET /uiresources` API that I've created in this PR. The stories have been commented out on purpose. They make API calls so, if I left them uncommented, the build would fail.

packages/react/src/components/SuiteHeader/hooks/useSuiteHeaderData.js
blechner: Thanks, I didn't know about this `@deprecated` annotation. I'll make the change

packages/react/src/components/SuiteHeader/SuiteHeader.story.jsx
kevinsperrine: @blechner Ah. got it. It's probably better suited to be moved into a README.md file and documented as examples of usage rather than just commented out code in the stories. See the [Table README](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/next/packages/react/src/components/Table/README.md) and stories as an example.

blechner: @kevinsperrine Thanks for the heads up. I'll create a `README.md` file and move the commented out code there.

kevinsperrine (approved):
LGTM.

cgirani (approved):
LGTM 👍

davidicus (approved):

PR #2328, opened 2021-05-13
Master

davidicus:
Updates next with latest version bump. 1.145.0


davidicus:
@cal-smith Are these version bumps right? Do we just bump the angular package on every react release?

cal-smith:
> @cal-smith Are these version bumps right? ~Do we just bump the angular package on every master release? I don't think you all are doing two release types, are you?

This looks fine to me ... There's definitely been changes to the angular package since the last release, so those version bumps make sense to me. We're just following the overall repo release cadence - no sense in doing something different.

cal-smith (approved):

PR #2335, opened 2021-05-17
fix(suiteheader): Making the transition of “loading state” to “data ready state” in SuiteHeader seamless

blechner:
Closes #

Closes an issue where any open panel (profile or app switcher) was automatically closing when SuiteHeader transitioned from "loading" to "data ready" state.

**Summary**

The use of Carbon's `HeaderContainer` to control the state of the side nav panel was causing the entire tree returned by its "render props" function to remount on every update in `SuiteHeader`'s properties. The solution was to remove the `HeaderContainer` component and manage the side nav toggle state in `SuiteHeader`. There was no possible solution to avoid a complete remounting of the components under `HeaderContainer` if we still used the "render props" approach.

Another change that was necessary was to avoid changing the `actionItems` array length. When API data becomes available, if the user is an admin, then `SuiteHeader` will include the "Cog" menu item by pushing a new element at the beginning of the `acionItems` array. This approach makes React think that all components in subsequent positions in the array are new and need to be created from scratch. The solution was to always include the "Cog" menu item, even before API data is received, so that `actionItems[0]` is already created beforehand (with `display: none`), to avoid remounting all action items that come after the "Cog" in the array.

All tests are still passing, which means that side nav functionality did not break after this change.

All tests are still passing, which means that side nav functionality did not break after this change.

Hint when looking at the file changes: add `?w=1` to the URL to ignore spaces and tabs, as most changes in `SuiteHeader.jsx` that show in the diff comparison were caused by the removal of a wrapper component, i.e. the child content of the removed wrapper has only an indentation change.

### DIFF
--- a/packages/react/src/components/SuiteHeader/__snapshots__/SuiteHeader.story.storyshot
+++ b/packages/react/src/components/SuiteHeader/__snapshots__/SuiteHeader.story.storyshot
@@ -4232,6 +4232,45 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/S
       <div
         className="bx--header__global"
       >
+        <button
+          aria-label="Administration"
+          aria-pressed={null}
+          className="bx--header-action-btn admin-icon admin-icon__hidden bx--header__action bx--btn bx--btn--primary bx--btn--icon-only bx--tooltip__trigger bx--tooltip--a11y bx--tooltip--bottom bx--tooltip--align-center"
+          data-testid="menu-item-Administration-global"
+          disabled={false}
+          onClick={[Function]}
+          tabIndex={0}
+          type="button"
+        >
+          <span
+            className="bx--assistive-text"
+          >
+            Administration
+          </span>
+          <span
+            id="suite-header-action-item-admin"
+          >
+            <svg
+              aria-hidden={true}
+              data-testid="admin-icon"
+              description="Settings"
+              fill="white"
+              focusable="false"
+              height={20}
+              preserveAspectRatio="xMidYMid meet"
+              viewBox="0 0 32 32"
+              width={20}
+              xmlns="http://www.w3.org/2000/svg"
+            >
+              <path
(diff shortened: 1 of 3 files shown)


packages/react/src/components/SuiteHeader/__snapshots__/SuiteHeader.story.storyshot
```diff
@@ -4294,6 +4294,45 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/S
       <div
         className="bx--header__global"
       >
+        <button
```
kevinsperrine, line 4297: There's a snapshot failure after the auto-overflow menu PR was merged. We'll also want to check that the overflow functionality still works as expected because it's built to check for overlap between the first icon in the action items and the div containing the title/subtitle. I don't know how it'll work with this new hidden button in the first position.

kevinsperrine: Yeah, we'll need to account for the icon in some way. The deploy preview shows the [loading state story](https://deploy-preview-2335--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-suiteheader--loading-state) always having overflow icon and the [next story](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-suiteheader--loading-state) works as expected and the overflow only appears at smaller screen sizes.

kevinsperrine (changes requested):
Everything looks good except the overflow icon. It would also be nice to simulate this loading experience with a setTimeout promise, so it's more clear exactly what's happening in production during this transition.

kevinsperrine (approved):
LGTM

packages/react/src/components/SuiteHeader/__snapshots__/SuiteHeader.story.storyshot
blechner: Fixed by replacing `display: none` with `visibility: hidden` in the first actionItem in SuiteHeader.

cgirani (approved):
LGTM 👍

PR #2336, opened 2021-05-17
feat(table): add multi-sort columns to table

### LINKED ISSUE #1376: [Table] Multi-sort rows
### Summary

From [Internal Watson-IoT/pal-tracking#164](https://github.ibm.com/Watson-IoT/pal-tracking/issues/164)

> This is an EDS item.
> 
> In complex and large data sets, users often will need to sort based on more than a single column, eg: first sort all files by file type, and then by date. Resulting in groups of files by type ordered individually within that group by date.
> 
> 
> * Using the column overflow (#1375  ) the user can select 'multi-sort' as an action.
>   Image:
>   image
> * This will 'pop' a modal with two inputs: column to sort on (as a dropdown), and sort order. Along with controls to add another column and sort order, and remove a sort when more than one is applied.
>   Image:
>   image
> * The user can use these to build an ordered set of column sorts
>   Image:
>   image
> * Once applied, the columns sort icons will gain a small badge indicating the order in which they are sorted:
>   Image:
>   image
> 
> This is present in spreadsheet tools such as Excel
> https://cdn.ablebits.com/_img-blog/alphabetize-excel/first-sort-level.png

Dependent on https://github.com/IBM/carbon-addons-iot-react/issues/1375

kevinsperrine:
Closes #1376 

**Summary**

Adds the ability to multi-sort columns within a table. A new `option.hasMultiSort` prop is added. The `view.table.sort` prop now accepts both a single sort objects `{ columnId: string; direction: 'ASC' | 'DESC' | 'NONE' }` or an array of sort objects. The order of this array determines the order the multi-sort is applied to the table. There are new callback actions for interacting with the multi-sort column data:  `onSaveMultiSortColumns`, `onCancelMultiSortColumns`, `onAddMultiSortColumn`, and `onRemoveMultiSortColumn`. 

- added TableMultiSortModal component
- updated snapshots
- add documentation
- cleaned up Table stories by adding knobs and removing duplicates
- renamed stories for ListBuilder/MenuButton to add number prefix, so they appear in correct location

- The StatefulTable with multi-sort story should allow selecting and sorting by different columns
- The Table with multi-sort story should show the multi-sort modal and callbacks fire in the action panel when clicking save/cancel or add/remove buttons.

- Existing Table stories using single-column sorting should function as previously expected


packages/react/src/components/Table/tableReducer.js
```diff
@@ -212,21 +222,36 @@ export const filterSearchAndSort = (
...
+
+  const { columnId, direction } = sort;
+  return getCustomColumnSort(columns, columnId)
+    ? getCustomColumnSort(columns, columnId)({ data: searchedData, columnId, direction })
+    : getSortedData(searchedData, columnId, direction);
 };
 
 /** This reducer handles sort, filter and search that needs data otherwise it proxies for the baseTableReducer */
 export const tableReducer = (state = {}, action) => {
+  console.log({ action });
```
bjornalm, line 254: Forgotten console.log on a few places here

kevinsperrine: removed.

packages/react/src/components/Table/tableReducer.js
```diff
@@ -212,17 +222,46 @@ export const filterSearchAndSort = (
...
+  if (isEmpty(sort)) {
+    return searchedData;
+  }
+
+  if (Array.isArray(sort)) {
+    let sortStack = firstBy(() => 0);
+    sort.forEach(({ columnId, direction }) => {
+      const customSortFn = getCustomColumnSort(columns, columnId);
+      if (customSortFn) {
+        const sortedValues = customSortFn({ data: searchedData, columnId, direction }).map(
```
kevinsperrine, line 239: I think it's worth discussing a breaking change in V3 here, so that the customSort function is just a comparator instead of a method that takes the data, column, and direction. It would be much easier (and more efficient) to handle sorting related issues without having to maintain this backwards compatibility. Unless this method was chosen for a specific reason that I'm unaware of.

bjornalm:
There are some layout issues with RTL
image

bjornalm:
I think we should have a story where this is combined with column resize since this change affects the column headers and there might be some unforeseen conflicts.

packages/react/src/components/Table/tableReducer.js
```diff
@@ -212,17 +222,46 @@ export const filterSearchAndSort = (
...
-      ? getCustomColumnSort(columns, columnId)({ data: searchedData, columnId, direction })
-      : getSortedData(searchedData, columnId, direction)
-    : searchedData;
+
+  if (isEmpty(sort)) {
+    return searchedData;
+  }
+
+  if (Array.isArray(sort)) {
+    let sortStack = firstBy(() => 0);
```
bjornalm, line 235: `firstBy(() => 0);` seems a bit odd, is so that you can directly use `thenBy` in the loop below?

Perhaps this multisort can be broken out into its own function in order to improve readability for the 3 return statements of this larger filterSearchAndSort function.

packages/react/src/components/Table/tableReducer.js
```diff
@@ -561,6 +600,63 @@ export const tableReducer = (state = {}, action) => {
       );
     }
 
+    case TABLE_MULTI_SORT: {
```
bjornalm, line 603: use TABLE_MULTI_SORT_MODAL_TOGGLE?

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
```diff
@@ -0,0 +1,235 @@
...
+        return clone;
+      });
+
+      onRemoveMultiSortColumn(index);
+    },
+    [onRemoveMultiSortColumn]
+  );
+
+  const handleSaveMultiSortColumns = useCallback(() => {
+    onSaveMultiSortColumns(selectedMultiSortColumns);
```
bjornalm, line 151: Perhaps we need to check that the selectedMultiSortColumns actually contain columns here. Not sure what the UI should look like, possibly a validation message warning that columns are must be selected. The current code can emit `[{columnId: "select", direction: "ASC"}, {columnId: "", direction: "ASC"}]` which works for the stateful table bit might cause problems for apps dealing with their own states.

bjornalm (changes requested):
Solid work! Very nicely done, I've reviewed this pretty thoroughly now and there are a few issues we should fix.

packages/react/src/components/Table/README.md
```diff
@@ -287,6 +288,45 @@ const columns = [
...
+```jsx
+<Table
+  id="table"
+  columns={columns}
+  data={data}
+  actions={{
+    table: {
+      onAddMultiSortColumn: (index) => {},
+      onRemoveMultiSortColumn: (index) => {},
+      /* ([{columnId: 'select', direction: 'ASC'}, {columnId: 'string', direction: 'ASC'}]) => {} */
```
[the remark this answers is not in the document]
kevinsperrine: Yeah, it's just showing the structure the onSaveMultiSortColumns callback is returning. I added a prefix to explain that.

packages/react/src/components/Table/StatefulTable.jsx
```diff
@@ -292,8 +301,27 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
         callbackParent(onColumnResize, resizedColumns);
       },
       onOverflowItemClicked: (id) => {
+        if (id === 'multi-sort') {
+          dispatch(tableMultiSort());
```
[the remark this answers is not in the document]
kevinsperrine: agreed. renamed.

packages/react/src/components/Table/TableHead/_table-head.scss
```diff
@@ -33,6 +33,14 @@
     }
   }
 
+  .#{$iot-prefix}--table-header-label__sort-order {
+    position: absolute;
+    bottom: $spacing-04;
+    right: $spacing-04;
```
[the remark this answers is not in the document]
kevinsperrine: nice catch. i always forgot about rtl.

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
```diff
@@ -0,0 +1,235 @@
...
+      });
+
+      onAddMultiSortColumn(index);
+    },
+    [onAddMultiSortColumn]
+  );
+  const handleRemoveMultiSortColumn = useCallback(
+    (index) => () => {
+      setSelectedMultiSortColumns((prev) => {
+        const clone = [...prev];
```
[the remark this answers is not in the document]
kevinsperrine: yup, that's better. I was just in the splice headspace from the previous code.

packages/react/src/components/Table/tableReducer.js
davidicus: I think it was a functions so you can sort things that are not comparable like an error icon.

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
```diff
@@ -0,0 +1,235 @@
...
+            ...currentItem,
+            direction: selectedItem.id,
+          },
+        });
+      });
+    },
+    []
+  );
+
+  const multiSortColumns = useMemo(() => {
```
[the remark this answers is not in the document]
kevinsperrine: I think the useMemo makes more sense here than the useCallbacks elsewhere. I add those in basically just out of habit. A table with a large number of columns could see the benefit here with the memo though.

packages/react/src/components/Table/tableReducer.js
kevinsperrine: 🤔  Wouldn't that still be possible with a comparator? They can use the `renderDataFunction` to render the icon, and then `sortFunction` could still just be a [comparator as used here](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/next/packages/react/src/components/Table/Table.story.jsx#L106). It seems like the only thing the full function is accomplishing to me is cloning to not mutate the original array.

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.test.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+import TableMultiSortModal from './TableMultiSortModal';
+
+const callbacks = {
+  onAddMultiSortColumn: jest.fn(),
+  onRemoveMultiSortColumn: jest.fn(),
+  onSaveMultiSortColumns: jest.fn(),
+  onCancelMultiSortColumns: jest.fn(),
+};
+describe('TableMultiSortModal', () => {
```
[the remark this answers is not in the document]
kevinsperrine: good call.

packages/react/src/components/Table/tableReducer.js
kevinsperrine: added a comment to explain the firstBy and moved multisort to its own function.

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
kevinsperrine: Yeah, I thought of that. This is one of those weird instances where we're building a library, but managing the application state, too. @davidicus What do you think? Do we need error handling on this modal onSave and if so, I assume that'll need to be exposed to consumers as well? Or, just expose an onSave callback and let them deal with it?

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
```diff
@@ -0,0 +1,235 @@
...
+    return columns
+      .filter((col) => {
+        const columnAlreadySorted =
+          selectedMultiSortColumns.find((sortCol) => sortCol.columnId === col.id) !== undefined;
+
+        return col.isSortable && !columnAlreadySorted;
```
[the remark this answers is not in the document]
kevinsperrine: good call.

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
bjornalm: Another option is to force the dropdown to have a preselected column, so there is no room for empty selections.

kevinsperrine: default to first available column and direction

kevinsperrine:
> I think we should have a story where this is combined with column resize since this change affects the column headers and there might be some unforeseen conflicts.

Added columnSelection and resizing to the stateful table with multi-sort story. Seems fine.

bjornalm (approved):

davidicus:
The outline gets cut out from the new overflow menu.
image

kevinsperrine:
> The outline gets cut out from the new overflow menu.
> image

@davidicus Nice catch. However, it raises a couple of issues. 1) The styling issue already exists for any overflow menu added to the head. see https://deploy-preview-2336--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--simple-stateful-example-with-column-overflow-menu Easy enough to fix with css, but 2) The usage of sorting and overflow menu together causes a `<button>` within a `<button>` warning on the console when in DEV. I can fix it my changing the main button (for sorting) to an `<a role="button"...>`, but that'll almost certainly break downstream tests (it breaks 2 in our code). What do you think the best plan of action is?

kevinsperrine:
> > The outline gets cut out from the new overflow menu.
> > image
> 
> @davidicus Nice catch. However, it raises a couple of issues. 1) The styling issue already exists for any overflow menu added to the head. see https://deploy-preview-2336--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--simple-stateful-example-with-column-overflow-menu Easy enough to fix with css, but 2) The usage of sorting and overflow menu together causes a `<button>` within a `<button>` warning on the console when in DEV. I can fix it by changing the main button (for sorting) to an `<a role="button"...>`, but that'll almost certainly break downstream tests (it breaks 2 in our code). What do you think the best plan of action is?
> 
> EDIT: I could change the code to make the arrows the button and only change sort when clicking the arrows specifically, but that breaks the currently expected behavior of being able to click anywhere on the name of the column to re-sort it.

@davidicus I pushed this up yesterday. I adjusted the css, so that the focus outline on the cell doesn't get hidden, but to fix the button-within-a-button I switched that cell from using a `<button>` when sortable to using an `<a role="button">` instead. It keeps existing functionality and fixes the console warnings; however, it did break 2 tests--both enzyme tests looking for specific parts of the DOM--and it changes the markup from a `button` to an `a` tag. LMK if you think I should revert and switch the button to the sort arrows instead and break expected functionality when using sort and overflow menus together.

davidicus:
@kevinsperrine is there a way to use React.portals to avoid this button within button scenario? I would rather not change the markup if we can help it. Especially since it already broke our test. Chances are it will do the same for others.

kevinsperrine:
> @kevinsperrine is there a way to use React.portals to avoid this button within button scenario? I would rather not change the markup if we can help it. Especially since it already broke our test. Chances are it will do the same for others.

I'll give it a shot.

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
```diff
@@ -0,0 +1,282 @@
...
+    onSaveMultiSortColumns(defaulted);
+  };
+
+  const handleCancelMultiSortColumns = () => {
+    setSelectedMultiSortColumns(sort);
+    onCancelMultiSortColumns();
+  };
+
+  return (
+    <ComposedModal
```
davidicus, line 208: does this need a data-testId?

davidicus, line 208: do we need  a testID prop for this?

stuckless (commented):
The choice to use an overflow that pops up a dialog seems like an odd behaviour (but, I see that is coming straight from PAL, so I guess that is good).  At the very least, IF I have to select the triple dot menu and select multi-sort, then it should open the dialog and configure the entry such that it selects the column that opened the dialog, so that all I need to do is press +.  If existing sorts exist, and I click the triple dot on a column that is not already sorted then it should add a new row and set that selected column in the dropdown.

I see a couple of behavioural issues that PAL design does not address.
1). It would seem that once I create a sort, then clicking a table column will remove ALL sorts.  This does not feel right.  I would expect that clicking a table column that is sorted would just invert the sort of that table column, and not remove all the sorts from the table.  Perhaps if I clicked a table column that is NOT sorted it might remove all other sorts, although I expect that customers would likely expect that action to just add another sorted column automatically.

2). The modal should allow the sorted columns to be reordered.

3). The modal should have a clear/remove all option, no?

@lukefirth do you have any comments on the behaviours here?

davidicus (approved):

davidicus:
@dianatran18 do you have any feedback for @stuckless comments above?

stuckless (changes requested):
Modal in RTL is not correct.  X is in the wrong location and the input components are not aligned correctly.

If I add more than 4 items it blows out of the dialog.

On a small screen if I open the dropdown in the dialog I can't select all the items, and keyboard navigation appears to now work either.  Not sure if it's a dropdown issue or a dialog issue.

packages/react/package.json
```diff
@@ -153,6 +153,7 @@
     "react-visibility-sensor": "^5.0.2",
     "scroll-into-view-if-needed": "^2.2.26",
     "styled-components": "^4.1.3",
+    "thenby": "^1.3.4",
```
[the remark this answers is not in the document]
kevinsperrine: It's only 28K unpacked. It's a 2K module, and yeah, it simplifies the experience significantly when sorting in various directions and orders.

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
```diff
@@ -0,0 +1,282 @@
...
+              onChange={handleSelectMultiSortColumn(index)}
+              titleText={
+                index === 0
+                  ? i18n.multiSortSelectColumnSortByTitle
+                  : i18n.multiSortSelectColumnThenByTitle
+              }
+              initialSelectedItem={getInitialSelectedColumn(columnId)}
+              translateWithId={handleTranslation}
+            />
+            <Dropdown
```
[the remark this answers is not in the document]
kevinsperrine: I'll add this when I add it to the modal as well.

packages/react/package.json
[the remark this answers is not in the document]
stuckless: I would be very conservative about adding new libraries.  I'm not saying, don't, but you should verify that this library is "blessed".  ie, it in the IBM open source list of approved libraries (you can check the wicked system).  When you add a new library, it has to get legal approval eventually.  That is why I question importing a library to do a sorting, when that seems pretty trivial.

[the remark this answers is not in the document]
davidicus: our OM is looking into what is happening with our code scans we are supposed to have for teams to use. We will hopefully have an answer soon on when we can expect them.

dianatran18:
I can provide some feedback, but just a note that the pattern defined here was from the original EDS/IoT PAL, so I wasn't involved in the original pattern/nor have I thoroughly looked into the multi-sort pattern myself.

> 1). It would seem that once I create a sort, then clicking a table column will remove ALL sorts. This does not feel right. I would expect that clicking a table column that is sorted would just invert the sort of that table column, and not remove all the sorts from the table. Perhaps if I clicked a table column that is NOT sorted it might remove all other sorts, although I expect that customers would likely expect that action to just add another sorted column automatically.

I briefly spoke with the the PAL team on this, and I think it's fine that if you click to sort on an already sorted column, it would invert the sort, and not remove all sorts from the table. And agreed that if you clicked on a column that is not already sorted, then it would remove all other sorts.

> 2). The modal should allow the sorted columns to be reordered.
> 
> 3). The modal should have a clear/remove all option, no?

Again I'm not deeply familiar with the pattern defined, but these statements make sense to me. 

I'm fine moving forward with this so we can close this out, but I would ask that the teams that will be implementing this in their products to get feedback on their users on what the expected behavior is for everything listed above.

packages/react/src/components/Table/TableMultiSortModal/TableMultiSortModal.jsx
kevinsperrine: added.

kevinsperrine:
@stuckless @dianatran18 

> Modal in RTL is not correct. X is in the wrong location and the input components are not aligned correctly.
I've fixed the alignment, but the X we will tackle with another issue, as that's in the ComposedModal component not directly related to this feature. https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2396

> If I add more than 4 items it blows out of the dialog.
This has been addressed.  We had to resort to using `Select` components instead of `Dropdown` because there's currently no good solution for handling the overflow of dropdown menus within a modal. There are multiple issues in Carbon. See https://github.com/carbon-design-system/carbon/issues/6288 and https://github.com/carbon-design-system/carbon/issues/7565#issuecomment-786921600. 

> On a small screen if I open the dropdown in the dialog I can't select all the items, and keyboard navigation appears to now work either. Not sure if it's a dropdown issue or a dialog issue.
Keyboard navigation was working, but the modal wasn't grabbing focus, so you had to tab through all the items on the table in the background to get to the modal. The save button now grabs focus when the modal opens, so keyboard navigation is easier.

I also added testIds to the select menus and to the modal itself. Clicking on the sort button (for a column that's selected for multi-sort) will now invert the sort for that column, but clicking sort on a new column not included in multi-sort will reset the sort and only use that single column. The modal window now also has a "Clear sorting" button to wipe it all out.

kevinsperrine:
@stuckless Friendly bump on the updates to this PR. Can you give it another look now with @davidicus's https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2336#issuecomment-859756617 in mind?

stuckless (approved):

davidicus (approved):

PR #2337, opened 2021-05-18
feat(table): improve nested row selections

### LINKED ISSUE #124: [Table]: selection box support indeterminate state based on the selection of their children state

bjornalm:
#124

Closes #124

**Summary**

Improved the selection logic for nested rows with multi select according to specs in the issue comment:

_1. Selection states are none (empty), all (filled), partial (dash). There are selection icons for each.
2. Selecting the parent will select all children
3. Unselecting the parent will unselect all children
4. Changing the selection of a child will change the state of the parent to partial
5. Selecting a partial parent will make the parent and children selected_

I interpreted 4 as unselecting a child will set parent to partial if there still are additional children selected. Selecting a child when all other children are selected will select parent as well. 

- Added logic to the TableComponent that handles selection/deselection of rows and how that affects the parent and children rows.
- Added prop isIndeterminate to the TableBodyRow component
- Extended the onRowSelected callback params to include an array with the ids of all  currently selected rows
- Simplified the StatefulTable and the baseTableReducer since more logic was put in the table.
- Removed isSelectAllIndeterminate & isSelectAllSelected from the table view of the
StatefulTable since it was not use.
- Updated unit tests for tableReducer since the API for tableRowSelect was modified
- Added new unit tests to Table and StatefulTable
- Fixed the story `WithSingleSelectAndNestedTableRows` that despite the name didn't use `hasRowSelection: 'single'`

Test the selection scenarios described in the spec above using the story https://deploy-preview-2337--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--stateful-example-with-row-nesting-and-fixed-columns Verify that automatic selections to parents and children are propagated all the way up and down.

- Verify that single select stories still work
- Verify that all other table stories work as expected

### DIFF
--- a/packages/react/src/components/Table/README.md
+++ b/packages/react/src/components/Table/README.md
@@ -348,7 +348,7 @@ The `Table` component supports row selection when using the options.hasRowSelect
   actions={{
     table: {
       onRowClicked: (rowId: string) => {},
-      onRowSelected: (rowId: string, selected: boolean) => {},
+      onRowSelected: (rowId: string, selected: boolean, selectedIds: array) => {},
     },
   }}
   options={{
@@ -678,7 +678,7 @@ the following props:
       /** rowId is a string */
       onRowClicked: (rowId) => {},
       /** rowId is a string, selected is a boolean */
-      onRowSelected: (rowId, selected) => {},
+      onRowSelected: (rowId, selected, selectedIds: array) => {},
       /** allSelected is a boolean. true is all are selected, false otherwise */
       onSelectAll: (allSelected) => {},
     }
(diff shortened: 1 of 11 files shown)


kevinsperrine:
The various states outlined in 1-5 all work as expected, but is there a spec for the expected behavior when children are selected, and then the select all checkbox is clicked? This behavior seems odd to me. 
row-selection

bjornalm:
> The various states outlined in 1-5 all work as expected, but is there a spec for the expected behavior when children are selected, and then the select all checkbox is clicked? This behavior seems odd to me.

Yes I agree. There is a newly created issue for that bug that I think we should fix.

kevinsperrine (approved):

davidicus (approved):

PR #2343, opened 2021-05-19
feat(tearsheet): new component

hectordavis:
**Summary**
New TearSheet component based on https://pages.github.ibm.com/cdai-design/pal/components/tearsheet/usage

image
image

- Check `TearSheet` component story

### DIFF
--- a/packages/react/src/components/TearSheet/TearSheet.jsx
+++ b/packages/react/src/components/TearSheet/TearSheet.jsx
@@ -70,12 +70,12 @@ const TearSheet = ({
   };
   return (
     <div
-      data-testid={`${iotPrefix}-tear-sheet-${idx}`}
-      className={`${iotPrefix}-tear-sheet ${className || ''}`}
+      data-testid={`${iotPrefix}--tear-sheet-${idx}`}
+      className={`${iotPrefix}--tear-sheet ${className || ''}`}
     >
       <div
-        className={`${iotPrefix}-tear-sheet--header ${
-          headerExtraContent ? `${iotPrefix}-tear-sheet--header__extraContent` : ''
+        className={`${iotPrefix}--tear-sheet--header ${
+          headerExtraContent ? `${iotPrefix}--tear-sheet--header__extraContent` : ''
         } `}
       >
         <Button
@@ -89,11 +89,11 @@ const TearSheet = ({
           testID={`tearSheetCloseBtn${idx}`}
         />
         <h2>{title}</h2>
-        <span className={`${iotPrefix}-tear-sheet--header--description`}>{description}</span>
+        <span className={`${iotPrefix}--tear-sheet--header--description`}>{description}</span>
         {headerExtraContent}
       </div>
 
-      <div className={`${iotPrefix}-tear-sheet--content`}>
+      <div className={`${iotPrefix}--tear-sheet--content`}>
         {typeof children === 'function'
           ? children(idx, openNextSheet, goToPreviousSheet, closeAllTearSheets)
           : children}
--- a/packages/react/src/components/TearSheet/TearSheet.jsx
+++ b/packages/react/src/components/TearSheet/TearSheet.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Close16 } from '@carbon/icons-react';
+import classnames from 'classnames';
 
 import { settings } from '../../constants/Settings';
 import Button from '../Button';
@@ -71,12 +72,12 @@ const TearSheet = ({
   return (
     <div
       data-testid={`${iotPrefix}--tear-sheet-${idx}`}
-      className={`${iotPrefix}--tear-sheet ${className || ''}`}
+      className={classnames(`${iotPrefix}--tear-sheet`, className)}
     >
       <div
-        className={`${iotPrefix}--tear-sheet--header ${
-          headerExtraContent ? `${iotPrefix}--tear-sheet--header__extraContent` : ''
-        } `}
+        className={classnames(`${iotPrefix}--tear-sheet--header`, {
+          [`${iotPrefix}--tear-sheet--header__extraContent`]: !!headerExtraContent,
+        })}
       >
         <Button
           hasIconOnly
--- a/packages/react/src/components/TearSheet/TearSheet.jsx
+++ b/packages/react/src/components/TearSheet/TearSheet.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { cloneElement, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { Close16 } from '@carbon/icons-react';
 import classnames from 'classnames';
@@ -22,23 +22,25 @@ const propTypes = {
   /** onClose optional function. When provided, it will run when the TearSheet is closed */
   onClose: PropTypes.func,
   /** closeAllTearSheets function is provided via TearSheetWrapper component. It should't be overridden */
-  closeAllTearSheets: PropTypes.func.isRequired,
+  closeAllTearSheets: PropTypes.func,
   /** openNextSheet function is provided by TearSheetWrapper component and it shouldn't be overriden. This function can be accessed via children prop to open the next sheet */
-  openNextSheet: PropTypes.func.isRequired,
+  openNextSheet: PropTypes.func,
   /** goToPreviousSheet function is provided by TearSheetWrapper component and it shouldn't be overriden. This function can be accessed via children prop to go back to the previous sheet or close the current one if its index is 0 */
-  goToPreviousSheet: PropTypes.func.isRequired,
+  goToPreviousSheet: PropTypes.func,
   /** i18n messages */
   i18n: PropTypes.shape({
     close: PropTypes.string,
   }),
-  /** children prop can be a node or a funcion. If it is a function, the following props are passed from TearSheetWrapper: idx, openNextSheet, goToPreviousSheet, closeAllTearSheets  */
-  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+  children: PropTypes.node,
 };
 
 const defaultProps = {
   idx: undefined,
   title: 'Title',
   description: 'Description',
+  closeAllTearSheets: undefined,
+  openNextSheet: undefined,
+  goToPreviousSheet: undefined,
   headerExtraContent: null,
   className: undefined,
   onClose: undefined,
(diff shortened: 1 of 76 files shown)


packages/react/src/components/TearSheet/TearSheet.story.jsx
```diff
@@ -0,0 +1,151 @@
...
+              </Button>
+              <div
+                style={{
+                  display: 'flex',
+                  justifyContent: 'flex-end',
+                  borderTop: '1px solid #e0e0e0',
+                }}
+              >
+                <Button
+                  onClick={() => goToPreviousSheet()}
```
kevinsperrine, line 64: Can we also fire the onClose event here? I know it's just the story, but currently, if you click the "Close" button (not the close 'x') on the first tearsheet, clicking the "Show TearSheet" button will not show it anymore.

packages/react/src/components/TearSheet/tear-sheet-wrapper.scss
```diff
@@ -0,0 +1,42 @@
+.tear-sheet-wrapper {
+  top: 0;
+  left: 0;
+  height: 100%;
+  width: 100%;
+  opacity: 0;
+  z-index: -1;
+  position: fixed;
+  // stylelint-disable-next-line declaration-property-unit-blacklist
+  transition: linear 250ms all;
```
kevinsperrine, line 10: Can we use the [@carbon--motion](https://github.com/carbon-design-system/carbon/blob/main/packages/motion/scss/motion.scss#L58) package and/or duration tokens to handle the animations, make them match the [spec](https://pages.github.ibm.com/cdai-design/pal/components/tearsheet/style#motion), and get rid of all the unit blacklists? See the [progress indicator](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/ddbd45c952a6f36dfa4ccfbf0f54c175db0f43b6/packages/styles/src/components/ProgressIndicator/_progress-indicator.scss#L88) or [row actions cell](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/ddbd45c952a6f36dfa4ccfbf0f54c175db0f43b6/packages/react/src/components/Table/TableBody/RowActionsCell/_row-actions-cell.scss#L60) as examples.

packages/react/src/components/TearSheet/tear-sheet.scss
```diff
@@ -0,0 +1,38 @@
...
+  height: 100%;
+  background-color: $ui-01;
+
+  & * {
+    max-width: 100%;
+  }
+
+  &--header {
+    padding-left: $spacing-07;
+    padding-right: $spacing-07;
```
kevinsperrine, line 11: Can we add a little more padding to the right side when in RTL mode?
Screen Shot 2021-05-20 at 4 18 12 PM

packages/react/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,44 @@
...
+const DEBOUNCE_TIMEOUT = 300;
+
+const useWindowSize = (debounceDelay = DEBOUNCE_TIMEOUT) => {
+  const [windowSize, setWindowSize] = useState({
+    width: window?.innerWidth,
+    height: window?.innerHeight,
+  });
+  const timer = useRef(null);
+
+  const debounce = (fn, delay) => {
```
kevinsperrine, line 12: lodash is available within the project, so you can easily replace this with `import debounce from 'lodash/debounce';`

packages/react/src/index.js
```diff
@@ -117,6 +117,7 @@ export DynamicHotspotSourcePicker from './components/HotspotEditorModal/DynamicH
 export ComboChartCard from './components/ComboChartCard/ComboChartCard';
 export MenuButton from './components/MenuButton/MenuButton';
 export ListBuilder from './components/ListBuilder/ListBuilder';
+export TearSheetWrapper from './components/TearSheet/TearSheet';
```
kevinsperrine, line 120: This is actually exporting the `TearSheet`, not the `TearSheetWrapper`, and we'll need to export both components, right?

kevinsperrine (changes requested):
Thanks, @hectordavis. This looks really great! There are a few items that I think should be addressed, but really solid work overall.

packages/react/src/components/TearSheet/TearSheet.jsx
```diff
@@ -0,0 +1,102 @@
...
+  const onCloseButton = () => {
+    if (onClose) {
+      onClose();
+      goToPreviousSheet();
+    } else {
+      goToPreviousSheet();
+    }
+  };
+  return (
+    <div data-testid={`tear-sheet-${idx}`} className={`tear-sheet ${className || ''}`}>
```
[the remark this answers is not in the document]
kevinsperrine: this applies to all the tear-sheet-* classes, btw.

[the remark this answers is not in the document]
hectordavis: Done 👍

packages/react/src/components/TearSheet/TearSheet.story.jsx
hectordavis: The `onClose` event is only passed to the `TearSheet` component to be used by the `x` button, but I've resolved this issue by adding the `setOpen(false)` to the onClick event of this particular button.

packages/react/src/components/TearSheet/TearSheet.story.jsx
```diff
@@ -0,0 +1,151 @@
...
+      <TearSheetWrapper isOpen={isOpen} onCloseAllTearSheets={() => setOpen(false)}>
+        <TearSheet
+          title="First sheet"
+          description="Generic description"
+          onClose={() => setOpen(false)}
+        >
+          {(idx, openNextSheet, goToPreviousSheet) => (
+            <div
+              style={{
+                width: '1000px',
```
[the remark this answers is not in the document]
hectordavis: Hey, @kevinsperrine.
The design specs state that the tearsheet remains 64px from the left and right sides of the viewport, so I've added a new `--content-max-width` (which is equivalent to the viewport width - 64px from each side) prop to be used by the `max-width` prop in the container scss 👍

packages/react/src/components/TearSheet/tear-sheet-wrapper.scss
hectordavis: Thanks for the heads-up! I've changed them to use the motion package but I couldn't find the TearSheet specifications for the duration, so I picked the ones closer to the duration I already had. Please let me know if I should change them.

packages/react/src/components/TearSheet/tear-sheet.scss
hectordavis: Good catch! I've added more padding to the right side of the header. As for the TearSheet content, it will be up to the user implementing the TearSheet. For this particular story, I've added `2rem` of `margin-right` to the **Open 2nd sheet** button.

packages/react/src/hooks/useWindowSize.js
hectordavis: Done 👍

packages/react/src/index.js
hectordavis: Right! Done 👍

packages/react/src/components/TearSheet/TearSheet.jsx
```diff
@@ -0,0 +1,102 @@
...
+  const onCloseButton = () => {
+    if (onClose) {
+      onClose();
+      goToPreviousSheet();
+    } else {
+      goToPreviousSheet();
+    }
+  };
+  return (
+    <div data-testid={`iot-tear-sheet-${idx}`} className={`iot-tear-sheet ${className || ''}`}>
```
kevinsperrine, line 69: Sorry, there's a variable for this, too, like in scss. 

```
import { settings } from '../../constants/Settings';

<div data-testid={`${iotPrefix}--tear-sheet-${idx}`} className={`${iotPrefix}--tear-sheet ${className || ''}`}>
```

And we do add double hyphens `--` after that prefix to match a BEM-esque structure.

kevinsperrine (approved):
Just update that className prefix with the variable and everything looks good to me. Thanks!

packages/react/src/components/TearSheet/TearSheet.story.jsx
```diff
@@ -0,0 +1,154 @@
+/*
+ * Licensed Materials - Property of IBM
+ * 5737-M66, 5900-AAA
+ * (C) Copyright IBM Corp. 2020 All Rights Reserved.
+ * US Government Users Restricted Rights - Use, duplication, or disclosure
+ * restricted by GSA ADP Schedule Contract with IBM Corp.
+ */
```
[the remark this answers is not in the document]
hectordavis: No problem! Done 👍

packages/react/src/components/TearSheet/TearSheet.jsx
hectordavis: Done 👍

cgirani (approved):
LGTM 👍

bryancboyd (approved):
looks good 👍

packages/react/src/components/TearSheet/TearSheet.jsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+  };
+  return (
+    <div
+      data-testid={`${iotPrefix}--tear-sheet-${idx}`}
+      className={`${iotPrefix}--tear-sheet ${className || ''}`}
+    >
+      <div
+        className={`${iotPrefix}--tear-sheet--header ${
+          headerExtraContent ? `${iotPrefix}--tear-sheet--header__extraContent` : ''
```
davidicus, line 78: We should just bring in the `classname` package for this logic to be consistent with other components. Can add to line 74 as well to not have the extra space when no className is provided.

davidicus (commented):
Small comment on classname logic. I am also curious about the children function call.

packages/react/src/components/TearSheet/TearSheet.jsx
hectordavis: Done 👍  
Added to `TearSheet` and `TearSheetWrapper`

packages/react/src/components/TearSheet/TearSheetWrapper.jsx
```diff
@@ -0,0 +1,180 @@
...
+
+  return (
+    <div
+      className={classnames(`${iotPrefix}--tear-sheet-wrapper`, className, {
+        ...animationClasses?.overlay?.reduce((acc, curr) => ({ ...acc, [curr]: isOpen }), {}),
+      })}
+      style={{ '--window-height': `${windowSize.height}px` }}
+      data-testid={`${iotPrefix}--tear-sheet-wrapper`}
+    >
+      {childrenArray.map((tearSheet, idx, arr) => {
```
davidicus, line 146: We may want to use [`React.Children.map`](https://reactjs.org/docs/react-api.html#reactchildrenmap)

davidicus (approved):

PR #2344, opened 2021-05-19
fix(overflow-menu): getMenuOffset import fix

kevinsperrine:
Closes #

**Summary**

- using a wildcard import to pull in getMenuOffset was causing build errors downstream. Reverted back to the original import method and it magically works now. 🤕


scottdickerson (approved):
LGTM

davidicus (approved):

PR #2345, opened 2021-05-19
feat(image-hotspots): add minimapBehavior prop for better control

### LINKED ISSUE #1356: [imagecard stories] Auto-hiding of minimap should be an option, not the only behavior.
### Describe the bug

- Originally the design shows the minimap 'showing' at all times. 
- There is currently a knob that allows you to specify that the minimap should not show ever (although this seems to be broken).
- Monitor team got valid feedback that in their usecases where the image is generally small, it's beneficial to have the minimap hide automatically when not actively panning. This was then made the default behavior in Monitor, making it impossible to have the original intent (always show minimap). However it also changed the default behavior of the underlying PAL component unilaterally.
- There are usecases that still exist of images at larger sizes that benefit from having the minimap as a way to navigate around it, which is impossible when it is auto-hidden as it cannot be interacted with.

- This bug is to rectify that change.

### To Reproduce

image

1. Go to any of the image hotspot stories (not sure on what the difference is here from image card... 
2. Click on the image and drag to see the minimap.
3. Scroll down to the knobs and click the hide minimap.
4. See error in that there is no way to always show the minimap (when zoomed), nor does the hide minimap knob seem to function, as it will still show on pan even when this is checked.
image

### Expected behavior

- The current 'auto hide minimap' behavior added as an option to its behavior like the 'hide' option already.
- Optimum would be the make the behavior a choice: 
  1. Always show
  2. Show on pan
  3. Always hide

No specific timeline identified currently.

kevinsperrine:
Closes #1356 

**Summary**

- adds a `minimapBehavior` prop for better control over when the minimap is displayed.

- Add `minimapBehavior` prop
- Deprecate `hideMinimap` prop
- Add tests to confirm behavior is as expected
- fixed styles on minimize icon when zoomed in.
- updated snapshots

- in the [https://deploy-preview-2345--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-imagehotspots--landscape-image-landscape-container story](https://deploy-preview-2345--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-imagehotspots--landscape-image-landscape-container)
  - zooming in and panning should show the minimap by default.
  - change the minimapBehavior prop to 'hide' and the minimap should always be hidden. even when panning.
  - change the minimapBehavior prop to 'show' and the minimap should always be visible--even when at default zoom or when panning.

- previously expected behavior (only showing minimap when panning) should be seen in all stories

### DIFF
--- a/packages/react/src/components/ImageCard/ImageHotspots.jsx
+++ b/packages/react/src/components/ImageCard/ImageHotspots.jsx
@@ -44,8 +44,8 @@ const propTypes = {
       }
     }
   },
-  /* 'hide' always hides the minimap, 'show' always shows the minimap, 'onPan' only shows the minimap when panning an image */
-  minimapBehavior: PropTypes.oneOf(['hide', 'show', 'onPan']),
+  /* 'hide' always hides the minimap, 'show' always shows the minimap, 'showOnPan' only shows the minimap when panning an image */
+  minimapBehavior: PropTypes.oneOf(['hide', 'show', 'showOnPan']),
   /** when true activates mouse event based create & select hotspot fuctionality */
   isEditable: PropTypes.bool,
   isHotspotDataLoading: PropTypes.bool,
@@ -99,7 +99,7 @@ const defaultProps = {
   hideHotspots: false,
   /* @deprecated in favor of minimapBehavior */
   hideMinimap: undefined,
-  minimapBehavior: 'onPan',
+  minimapBehavior: 'showOnPan',
   isHotspotDataLoading: false,
   isEditable: false,
   onAddHotspotPosition: () => {},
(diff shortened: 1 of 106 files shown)


bjornalm:
If I go to the story  https://deploy-preview-2345--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-imagehotspots--landscape-image-landscape-container story and zoom in, there is no minimap by default like the acceptance test states.

packages/react/src/components/ImageCard/ImageHotspots.jsx
```diff
@@ -29,7 +29,23 @@ const propTypes = {
...
+        return new Error(
+          `${componentName}: '${propName}' prop is deprecated in favor of 'minimapBehavior="${
+            value ? 'hide' : 'show'
+          }"'.`
+        );
+      }
+    }
+  },
+  /* 'hide' always hides the minimap, 'show' always shows the minimap, 'onPan' only shows the minimap when panning an image */
+  minimapBehavior: PropTypes.oneOf(['hide', 'show', 'onPan']),
```
bjornalm, line 48: onPan is a bit ambiguous, do we show or hide on pan? Can we call is showOnPan instead?

bjornalm (approved):
Nice work, just minor naming issue

packages/react/src/components/ImageCard/ImageHotspots.jsx
kevinsperrine: yep. that's more clear.

PR #2346, opened 2021-05-19
fix(overflow-menu): use es import from carbon

kevinsperrine:
Closes #

**Summary**

- use es not lib when importing overflow menu.


herleraja (approved):
LGTM

herleraja:
can you change in ExampleDropContainerApp `import uid from 'carbon-components-react/lib/tools/uniqueId'`

davidicus (approved):

PR #2348, opened 2021-05-20
fix(table): prevent prop warnings and fix testids

### LINKED ISSUE #2005: [RowActionsCell] Spreading props that cause console warnings
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

If users uses the `isEdit` or `isDelete` row actions it causes a console warning since we are spreading props onto the button. We already omit one prop and we just need to include these other two. 

https://github.com/carbon-design-system/carbon-addons-iot-react/blob/606c8da10af26c61164fe8702ea12caf2343344c/packages/react/src/components/Table/TableBody/RowActionsCell/RowActionsCell.jsx#L159

kevinsperrine:
Closes #2005 

**Summary**

- omit isDelete, isEdit, and isDivider props from being passed to the button
- fix data-testids on TableBody and TableExpandHeader.

- no errors in the console when isEdit or isDelete is included in a table action when _not_ in the overflow menu.

- no other changes to table

### DIFF
--- a/packages/react/src/components/BarChartCard/__snapshots__/BarChartCard.story.storyshot
+++ b/packages/react/src/components/BarChartCard/__snapshots__/BarChartCard.story.storyshot
@@ -2184,7 +2184,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/B
                       </thead>
                       <tbody
                         aria-live="polite"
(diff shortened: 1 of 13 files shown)


herleraja (approved):
LGTM

davidicus (approved):

PR #2350, opened 2021-05-20
chore(list-item): add title attribute to dnd icon

JoelArmendariz:
**Summary**
Need to at the title attribute the `ListItem` drag handles for accessibility.

 - Include the default i18n props along with the user defined i18n.
 - Add `Drag handle` title

Go to the `SimpleList` `List with reorder` story and hover over the drag handle icon. See that the browser tooltip shows.

 Make sure that you can still drag the `ListItems` successfully.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -17,7 +17,7 @@
 # These users own any files in the following directory at the root of
 # the repository and any of its subdirectories.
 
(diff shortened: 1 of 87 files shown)


kevinsperrine (approved):
LGTM. Have a few snapshots that still need to be updated, though.

herleraja (approved):
LGTM

PR #2351, opened 2021-05-20
feat(cardeditor): support custom settings panels for all card types, even default ones

scottdickerson:
Closes #

**Summary**

- Previously only the Custom card types could render custom EditSettingsForm sections.  This adds the capability for all card types

- feat(CardEditFormContent): show any custom sections if they are included
- fix(TimeSeriesCard): PropTypes for oneOf were incorrect
- fix(CardPropTypes): document the existing renderEditContent prop

- Verify the new CardEditor story and make sure the custom section shows for the TimeSeriesCard

- Verify that other card types in the CardEditor stories do not render custom content sections

### DIFF
--- a/packages/react/src/constants/CardPropTypes.js
+++ b/packages/react/src/constants/CardPropTypes.js
@@ -688,4 +688,8 @@ export const CardPropTypes = {
   locale: PropTypes.string,
   /** a way to pass down dashboard grid resize handles, only used by other card types */
   resizeHandles: PropTypes.array,
+  /** Optional callback function that is passed an onChange function and the original cardConfig function.
+   * This allows additional information to be passed to be used in the Card Editor for this type.
+   * You need to return an array of child objects with a header: {title, tooltip: {tooltipText: PropTypes.string}} and content element to render * */
+  renderEditContent: PropTypes.func,
 };
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormContent.jsx
@@ -146,7 +146,6 @@ const CardEditFormContent = ({
   );
 
   const editContentSections = renderEditContent && renderEditContent(onChange, cardConfig);
-
   return (
     <>
       <CommonCardEditFormFields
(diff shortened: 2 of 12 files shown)


packages/react/src/components/StorybookSnapshots.test.js
```diff
@@ -87,6 +87,10 @@ describe(`Storybook Snapshot tests and console checks`, () => {
     // https://stackoverflow.com/questions/53271193/typeerror-scrollintoview-is-not-a-function
     window.HTMLElement.prototype.scrollIntoView = jest.fn();
     window.HTMLElement.prototype.scrollTo = jest.fn();
+    window.getComputedStyle = () => ({
+      getPropertyValue: () => '25',
+    });
+    window.document = new Document();
```
kevinsperrine, line 93: let's reset these after the tests like in the other PR.

kevinsperrine (approved):
LGTM. Just the one note about resetting those mocks in the storybook tests.

packages/react/src/constants/CardPropTypes.js
```diff
@@ -688,4 +688,8 @@ export const CardPropTypes = {
   locale: PropTypes.string,
   /** a way to pass down dashboard grid resize handles, only used by other card types */
   resizeHandles: PropTypes.array,
+  /** Optional callback function that is passed an onChange function and the original cardConfig function.
+   * This allows additional information to be passed to be used in the Card Editor for this type.
+   * You need to return an array of child objects with a header: {title, tooltip: {tooltipText: PropTypes.string}} and content element to render * */
+  renderEditContent: PropTypes.func,
```
herleraja, line 694: Suggestion : 

`renderEditContent` can be renamed into `renderExtraContent`

herleraja (approved):
LGTM

packages/react/src/components/StorybookSnapshots.test.js
scottdickerson: ok I resolved this one thanks

packages/react/src/constants/CardPropTypes.js
scottdickerson: I think we have to leave it with Edit since it was already delivered that way I'm just documenting that property

davidicus (approved):

PR #2354, opened 2021-05-24
fix(side-nav): remove bad props from story and component

### LINKED ISSUE #2352: [SideNav] deprecated prop making it to DOM element, causing warnings
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

We are passing in the prop `translateById` to `SideNav` component. Carbon no longer uses this component and they spread remainder of props to the underlining element. This is causing warnings which have started blowing up downstream consumers test cases. 

We more than likely have an exception in our console spy setup in jest snapshot test file for this warning. Once we remove check to be sure there is no exception and if present remove as well.

kevinsperrine:
Closes #2352 

**Summary**

- translateById was removed from the carbon SideNav, but we were still passing it.

- comment out translateById and link to comment in carbon about re-adding it at some point
- stop passing `onClickSideNavExpand` to SideNav in SideNav stories
- add i18n prop for for the SideNav aria-label

- the SideNav story should show no warnings in the console

- The SideNav should continue to work as expected

### DIFF
--- a/packages/react/src/components/SideNav/SideNav.jsx
+++ b/packages/react/src/components/SideNav/SideNav.jsx
@@ -58,21 +58,20 @@ export const SideNavPropTypes = {
     })
   ).isRequired,
   isSideNavExpanded: PropTypes.bool,
-  /** An array of strings which will be options of switcher */
-  // switcherProps: PropTypes.obj,
   i18n: PropTypes.shape({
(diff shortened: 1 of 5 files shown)


davidicus (approved):

scottdickerson (approved):
LGTM

PR #2355, opened 2021-05-24
fix(tablecard): allow row-specific and cardVariables prop

sstone2423:
**Summary**

- If `cardVariables` was provided to TableCard, row-specific variables were no longer working correctly. This fix splits the two and assumes that all variables within the table rows are row-specific

- Remove the `with links` story in favor of the `row-specific links` story to avoid confusion
- Added a `keysToSkip` argument to `handleCardVariables` and `getCardVariables` to skip over specific keys such as the TableCard `href`
  - also sets up the ability to get better and/or fix this issue: https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1527
- Moved all tableCardUtil tests to the tableCardUtils test file
- Added describe blocks to groups of tests based on the same function and/or topic
- Renamed `propertyVariables` variable in `getCardVariables` to `cardVariables` to better reflect its purpose

- Go to /?path=/story/1-watson-iot-tablecard--with-row-specific-link-variables and check each link's href to ensure that the device id and timestamp is specific to that row's values
- Also check that the title is substituted (the bug here was that these 2 types of variables could not co-exist properly)

- Should be covered by the above tests

### DIFF
--- a/packages/react/src/components/TableCard/TableCard.jsx
+++ b/packages/react/src/components/TableCard/TableCard.jsx
@@ -179,11 +179,12 @@ const TableCard = ({
   // Set the locale
   dayjs.locale(locale);
   /** Searches for variables and updates the card if it is passed the cardVariables prop */
+  // Need to skip the linkTemplate variable for now because we will handle it at render time per row
   const {
     title,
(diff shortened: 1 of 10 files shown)


scottdickerson (approved):
LGTM

davidicus (approved):

PR #2356, opened 2021-05-25
Add getDefaultCard prop to editor

JoelArmendariz:
Closes #

**Summary**
Need to give the consumer the ability to create their own default card configs.

Add `getDefaultCard` prop to editor

Go to the `With get default card` story and see that the card properly renders a custom default config

Make sure that the card configs (found in the json editor modal) are still the correct defaults for mainstream cards.

### DIFF
--- a/packages/react/src/components/DashboardEditor/DashboardEditor.jsx
+++ b/packages/react/src/components/DashboardEditor/DashboardEditor.jsx
@@ -77,6 +77,10 @@ const propTypes = {
    * getValidTimeRanges(card, selectedDataItems)
    */
   getValidTimeRanges: PropTypes.func,
+  /** if provided, determines the default cardConfig for a new card when it is added
+   * getDefaultCard(cardType)
+   */
+  getDefaultCard: PropTypes.func,
   /** an array of dataItems to be included on each card
    * this prop will be ignored if getValidDataItems is defined
    */
@@ -300,6 +304,7 @@ const defaultProps = {
   onEditTitle: null,
   getValidDataItems: null,
   getValidTimeRanges: null,
+  getDefaultCard: null,
   availableImages: [],
   dataItems: [],
   availableDimensions: {},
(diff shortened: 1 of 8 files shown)


packages/react/src/components/DashboardEditor/DashboardEditor.jsx
```diff
@@ -459,11 +465,16 @@ const DashboardEditor = ({
...
-      // notify consumers that the card has been added if they're listening (they might want to tweak the card defaults)
-      const cardConfig = onCardChange
-        ? onCardChange(getDefaultCard(type, mergedI18n), dashboardJson)
+      const defaultCard = customGetDefaultCard // Use the default card specified by the consumer if it exists
+        ? customGetDefaultCard(type)
         : getDefaultCard(type, mergedI18n);
 
+      // notify consumers that the card has been added if they're listening (they might want to tweak the card defaults here)
+      const cardConfig =
+        onCardChange && !customGetDefaultCard
```
scottdickerson, line 474: is this right?  Should we call onCardChange even if they have a customGetDefaultCard or is the comment above the line incorrect?  Seems like they disagree.

scottdickerson (approved):
one small question but otherwise looks good

packages/react/src/components/DashboardEditor/DashboardEditor.jsx
JoelArmendariz: Right, the comment is left over and incorrect. To me it makes sense to call onCardChange when adding a new card IF there is no getDefaultCard passed, but if there is one passed, I would think the consumer would want to handle it there instead. I'll remove the misleading comment.

kevinsperrine (approved):

packages/react/src/components/DashboardEditor/DashboardEditor.jsx
davidicus: Can we fix the comment?

davidicus (approved):

PR #2357, opened 2021-05-26
fix(timeseries): card with thresholds should generate sample values between the thresholds

scottdickerson:
This is a little bit of a workaround for a carbon charts bug, they don't render thresholds if they're off the y-axis graph:
https://github.com/carbon-design-system/carbon-charts/issues/1034

**Summary**

- Set of fixes to the TimeSeriesCard so support the edit/preview case with thresholds

- fix(TimeSeriesCard): sample values generation should not regenerate if the series and thresholds really don't change (do a json comparison). 
- refactor(TimeSeriesCard): remove unused yMaxAdjuster property in options
- workaround(TimeSeriesCard): regenerate the LineChart if the thresholds change. (this is to workaround an opened bug on carbon-charts
- feat(timeSeriesUtils): add thresholds to the generateSampleData function so we can make sure that all the sample data values are close to the threshold values (so the thresholds actually show on the graph)

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx
+++ b/packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx
@@ -422,9 +422,6 @@ const TimeSeriesCard = ({
           ...(thresholds?.some((threshold) => threshold.axis === 'y')
             ? { thresholds: thresholds?.filter((threshold) => threshold.axis === 'y') }
             : {}),
-          // ...(chartType !== TIME_SERIES_TYPES.BAR
-          //   ? { yMaxAdjuster: (yMaxValue) => yMaxValue * 1.3 }
-          //   : {}),
           stacked: chartType === TIME_SERIES_TYPES.BAR && series.length > 1,
           includeZero: includeZeroOnYaxis,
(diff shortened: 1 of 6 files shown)


packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -532,6 +534,7 @@ const TimeSeriesCard = ({
               options={options}
               width="100%"
               height="100%"
+              key={`thresholds-key${thresholds?.length ? JSON.stringify(thresholds) : ''}`} // have to regen the component if thresholds change
```
JoelArmendariz, line 537: I think you could use the `objectAgnosticThresholds` var here instead of re-stringifying the thresholds.

JoelArmendariz (approved):
LGTM with one observation

davidicus (approved):

packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx
scottdickerson: good point! Fixed

PR #2361, opened 2021-05-27
fix(dashboardtemplateeditor): fix approach for editing title

scottdickerson:
Closes #

**Summary**

- There was a pretty serious issue with the DashboardEditor onEditTitle implementation.  Since the DashboardEditor is the component saving the draft state of the template in it's local state, we can't allow the parent component to change the template title or any new draft changes to the template will be lost

- fix(DashboardEditor): don't allow parents to change the title, only allow them to control whether the title is editable or not.   The DashboardEditor needs to change it's local draft state of the template to have the new title
- fix(PageTitleBar): fix the proptype of the PageTitleBar so that it's valid

- Verify the new isEditable DashboardEditor story actually updates the title

- Verify normal DashboardEditor stories


kevinsperrine (approved):

herleraja (approved):
LGTM

davidicus (approved):

PR #2362, opened 2021-06-01
MapCard

### LINKED ISSUE #2157: [Component Proposal] Map Component
### What package is this for?

- [x] React
- [ ] Angular

### Summary
A continuation of the exploration [issue](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1347), this issue is to track the development of the various pieces that are used to interact with the map. Since different map providers have their own unique APIs this component will provide the target element as well as props to power the various controls, labels, and headers but consumers will have to instantiate their own maps and pass the appropriate props. 

We will have working storybook examples using MapBox and OpenLayers map providers. 

**Additional context** 

Design specs for the map component have been published on our design site [here](https://pages.github.ibm.com/ai-applications/design/patterns/maps/overview/).

### LINKED ISSUE #2268: [MapCard] Documentation for component
Documentation should be written for the API of this component. The markdown file should live in the root directory of this component and added to the story.

Docs should include a prop table , common uses, link to design pattern, etc.

### LINKED ISSUE #2269: [MapCard] Testing
Component needs comprehensive testing in a few areas. This issue is considered complete when:

- [ ] data test ids have been provided to appropriate pieces 
- [ ] Unit test are 80%+
- [ ] a11y scans in storybook do not show any violations 
- [ ] a11y unit test has been written using the IBM scan tool
- [ ] RTL has been tested and working properly. 
- [ ] Confirm that all strings are able to be substituted for i18n

bjornalm:
Closes #2157, closes #2269, closes #2268

**Summary**
New MapCard component. This is a continuation of a previous PR from @davidicus. See #2157 for more info.  

✓ map config button should only be available in full page view
✓ full page map should have show the minimize/close to the right, after the settings button
✓ full page map controls should be size 40px instead of 32
✓ full page map should have 32px space between border and map controls instead of 16px 
✓ map watermark to the examples are not positions correctly	
✓ create new scrolling map controls container
✓ flyout controls should have little arrow to indicate flyout                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
✓ legend container should have 80% opacity 
✓ add dropshadow to legend and controls
✓ full bleed legend should should be collapsable
✓ full bleed legend causes card resize handle to not work
✓ scrollbars and white space outside map appear when going to full page view. Probably related to the map settings panel
✓ in open layers example the fullscreen button opens up map configuration
✓ clicking close x on map configuration doesn't close the map
✓ there should be an example of map controls acting like toggle controls, current controls behaves like radio buttons.
✓ add test-ids including parent prefix
✓ add actions for the stories
✓ i18n, proptypes and defaults
✓ export component in index.js
✓ initial code cleanup and lint fixes
✓ renamed MabBoxStory component in MapBoxExample.js to match file name
✓  Changed min size of the map in dashboard grid
✓  added a contentClassName to the Card so that we can set HTML class to the content. This was needed because the dashboard grid removes any classes we add to the card itself.
✓  created a const for the proptype definition of the general Button's "icon"-proptype that is used a lot (often with different proptype definitions)
✓ added MapCard proptypes const to CardPropTypes.js
✓ split example map css & actual MapCard css into different files
✓ put non exported examples in storyFiles
✓ documentation in README.md
✓ add unit tests 

**Left to do / discuss**
- create example with related content as map overlay (panel inside card)
- clean up css classes for consistent BEM
- drag and drop of related content panels inside card (stretch goal)
- Have it work in the actual old deprecated dashboard (stretch goal)
- remove layered map control trigger if there are no map controls?
- map controls should be shown/hidden depending on available height space in card (this can and perhaps also should be handled in the app)?
- full page map should have the settings icon aligned with the map controls below (won't fix due to added card complexity)?


1. Verify the text and code in README.md. I.e. does it make sense, is it clear, spelling, are we using correct language?
2. Does it work as expected in the dashboard grid. See story https://deploy-preview-2362--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-dashboard-grid--dashboard-all-cards-as-resizable
3. Does the MapBox example work as expected https://deploy-preview-2362--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-experimental-☢%EF%B8%8F-mapcard--mapbox-story
4. Does the Open Layers example work as expected https://deploy-preview-2362--carbon-addons-iot-react.netlify.app?path=/story/watson-iot-experimental-☢%EF%B8%8F-mapcard--openlayers-story


sstone2423:
Not sure how this happened, but theres 2 different Experimental storybook categories now

Screen Shot 2021-06-01 at 7 55 35 AM

packages/react/src/components/MapCard/Legend.jsx
```diff
@@ -0,0 +1,107 @@
...
+  titleText,
+  hideLegendText,
+  showLegendText,
+  isCollapsed,
+  onCollapsToggle,
+  testId,
+}) => {
+  const renderLegendKeys = (stop, i) => {
+    return (
+      <div key={i} className={`${iotPrefix}--map-legend-keys`}>
```
sstone2423, line 49: Using indices as the key can have potential bugs. It would be better to use one or both of the stop values 
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

packages/react/src/components/MapCard/MapControls.jsx
```diff
@@ -0,0 +1,167 @@
...
+  const handleLayerClick = () => {
+    setLayersOpen((openState) => !openState);
+  };
+
+  const buttonSize = isExpandedMode ? 'field' : 'small';
+
+  const renderControl = (control, index) => {
+    if (control.group) {
+      const groupClass = `${BASE_CLASS_NAME}-btn-group`;
+      const groupKey = `btn-group-${index}`;
```
sstone2423, line 63: Should consider using something unique to the value rather than the index

packages/react/src/components/MapCard/MapControls.jsx
```diff
@@ -0,0 +1,167 @@
...
+            tooltipPosition={tooltipPosition}
+            visibleItemsCount={control.visibleItemsCount}
+          />
+        ) : null
+      ) : (
+        <div key={groupKey} className={groupClass}>
+          {control.group.map((groupedControl, myIndex) => {
+            return (
+              <Button
+                key={`${groupedControl.iconDescription}-${myIndex}`}
```
sstone2423, line 82: Should consider using something unique to the value rather than the index

packages/react/src/components/MapCard/MapControls.jsx
```diff
@@ -0,0 +1,167 @@
...
+                onClick={groupedControl.onClick}
+              />
+            );
+          })}
+        </div>
+      );
+    }
+    return (
+      <Button
+        key={`${control.iconDescription}-${index}`}
```
sstone2423, line 100: Should consider using something unique to the value rather than the index

packages/react/src/components/MapCard/MapControls.jsx
```diff
@@ -0,0 +1,167 @@
...
+        renderIcon={control.icon}
+        iconDescription={control.iconDescription}
+        onClick={control.onClick}
+      />
+    );
+  };
+
+  const renderLayers = (control, index) => (
+    <Button
+      key={`${control.iconDescription}-${index}`}
```
sstone2423, line 116: Should consider using something unique to the value rather than the index

packages/react/src/components/MapCard/README.md
```diff
@@ -0,0 +1,342 @@
+# `MapCard` component
```
sstone2423, line 1: I'm not able to see this in the storybook. It might be related to the Experimental sections being duplicated?

packages/react/src/components/MapCard/ScrollingControls.jsx
```diff
@@ -0,0 +1,125 @@
...
+          [`--visble-items-count`]: visibleItemsCount,
+          [`--scroll-area-height`]: `${visibleItemsCount * buttonHeightPx}px`,
+        }}
+        className={classnames(`${BASE_CLASS_NAME}__scroll-area`)}
+      >
+        {controls.map((control, i) => {
+          return (
+            <Button
+              className={`${BASE_CLASS_NAME}__btn`}
+              key={`${control.iconDescription}-${i}`}
```
sstone2423, line 87: Should consider using something unique to the value rather than the index

packages/react/src/components/MapCard/storyFiles/Optionsfield.jsx
```diff
@@ -0,0 +1,34 @@
...
+      name: PropTypes.string,
+      property: PropTypes.string,
+    })
+  ).isRequired,
+};
+
+const Optionsfield = ({ changeState, property, options }) => {
+  const renderOptions = (option, i) => {
+    return (
+      <label htmlFor={option.name} key={i} className="toggle-container">
```
sstone2423, line 18: Should consider using something unique to the value rather than the index

packages/react/src/components/MapCard/storyFiles/MapboxExample.jsx
```diff
@@ -0,0 +1,315 @@
...
+      kind: 'icon-selection',
+      selected: selectedControls.includes('toggle3'),
+      icon: Events32,
+      iconDescription: 'Toggle control 3',
+      onClick: () => toggleControlSelection('toggle3'),
+    },
+  ];
+
+  const layeredControls = [
+    {
```
sstone2423, line 182: Is there an option to swap the tooltip direction and alignment for all custom buttons? I can see it being useful to move for situations like this:

Screen Shot 2021-06-01 at 8 17 47 AM

packages/react/jest.config.js
```diff
@@ -36,6 +36,18 @@ module.exports = {
       // https://github.com/carbon-design-system/carbon-charts/issues/594
       functions: 77,
     },
+    './src/components/MapCard/storyFiles/MapboxExample.jsx': {
```
sstone2423, line 39: Should these be a wildcard for `storyFiles` if this is the new pattern we're using?

sstone2423 (changes requested):
Wow great job! This is a massive feature, but your README is very detailed. Just a few nitpicks

packages/react/src/components/MapCard/README.md
bjornalm: Ahh, good catch. I need to attach the readme to the using `withReadme`

davidicus:
Should we automatically close the settings panel when the user exits out of full screen? Right now if they go to full screen, open the settings but do not close, and then close the fullscreen the setting panel stays open without a button to close itself. I think the sizing of the map gets thrown off rom this as well.

bjornalm:
> Should we automatically close the settings panel when the user exits out of full screen? Right now if they go to full screen, open the settings but do not close, and then close the fullscreen the setting panel stays open without a button to close itself. I think the sizing of the map gets thrown off rom this as well.

Yes I noticed that after the first PR push. I fixed it in the story, which is where the states are managed for this.

bjornalm:
> Not sure how this happened, but theres 2 different Experimental storybook categories now
> 
> Screen Shot 2021-06-01 at 7 55 35 AM

Yea, David and I thought that was strange but now I realised it is because these stories are missing the `"2 - "` in the title which makes it a new group. I will fix it.

packages/react/src/components/MapCard/Legend.jsx
bjornalm: Good point!

packages/react/src/components/MapCard/MapControls.jsx
bjornalm: Unfortunately the group does not have an identifier and can't be uniquely identified based on its content either, so we gonna have to stick with using index here. We don't expect the ordering of these map controls to change so it shouldn't be a problem.

packages/react/src/components/MapCard/MapControls.jsx
bjornalm: Yes, we are already using `groupedControl.iconDescription` as part of the key which should be unique, but as a backup we also append the index. I theory this can still go wrong but the risk of that happening is probably negligible.

packages/react/src/components/MapCard/MapControls.jsx
bjornalm: On second thought, I've decided to create a key based on the combined iconDescriptions of the child controls + the index.

packages/react/src/components/MapCard/MapControls.jsx
bjornalm: See previous comment

packages/react/src/components/MapCard/MapControls.jsx
bjornalm: See previous comment

packages/react/src/components/MapCard/ScrollingControls.jsx
bjornalm: See previous comment

packages/react/src/components/MapCard/storyFiles/Optionsfield.jsx
bjornalm: fixed

packages/react/jest.config.js
bjornalm: I was thinking about that too, but I'm not sure this is a new pattern. It is a pretty unusual component we are delivering here.

packages/react/src/components/MapCard/storyFiles/MapboxExample.jsx
bjornalm: Nope, that is currently not possible. I don't know if the alternatives are better in this case:
image

bjornalm: @davidicus Do you think we need to open a new issue for this? It is pretty annoying but I don't see a quick fix.

sstone2423: I think tooltipDirection: bottom, and tooltipAlignment: left could work, but not a deal breaker

sstone2423 (approved):

packages/react/src/components/MapCard/storyFiles/MapboxExample.jsx
```diff
@@ -0,0 +1,315 @@
...
+import mapboxgl from 'mapbox-gl';
+import PropTypes from 'prop-types';
+
+import MapCard from '../MapCard';
+
+import Optionsfield from './Optionsfield';
+import './mapbox-example.scss';
+
+mapboxgl.accessToken =
+  'pk.eyJ1IjoiZGF2aWRpY3VzIiwiYSI6ImNrbTN4OWpsZTBjYm0ybnBsaWZkemV6MmgifQ.jpqC4rJzYG6CY3IXc9NLuw';
```
kevinsperrine, line 24: is hard-coding the access token here, okay?

kevinsperrine (approved):
Everything looks great. Really awesome work! Just a couple of minor questions.

packages/react/src/components/MapCard/storyFiles/MapboxExample.jsx
bjornalm: I don't know, I haven't touched this part actually. @davidicus what do you think?

packages/react/src/components/MapCard/Legend.jsx
```diff
@@ -0,0 +1,106 @@
...
+  /** true if there should be extra wide margins, like when a card is expanded to full page */
+  increasedMargin: PropTypes.bool,
+  titleText: PropTypes.string,
+  hideLegendText: PropTypes.string,
+  showLegendText: PropTypes.string,
+  /** list of text - color pairs for the legend. Each pair is an array where pos 0 is the text and pos 1 the color */
+  stops: PropTypes.arrayOf(PropTypes.array).isRequired,
+  /** callback for when the collapse toggle button is clicked */
+  onCollapsToggle: PropTypes.func.isRequired,
+  testId: PropTypes.string,
```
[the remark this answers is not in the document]
bjornalm: I agree. We should probably fix them all in one go in that case, but I'm not sure if that can break external tests depending on the test id.

packages/react/src/components/MapCard/storyFiles/MapboxExample.jsx
davidicus: This is the access token from their docs. Should be fine.

packages/react/src/components/MapCard/Legend.jsx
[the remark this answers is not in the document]
davidicus: I think testID is always a prop that is used to add data-testids. So I think it is ok. I do not think it's ever applied directly to a DOM element.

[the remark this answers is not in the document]
davidicus: That being said. It would be a breaking change. I will add an issue to the V3 release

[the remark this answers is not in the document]
davidicus: For this PR let's go with the more common one

davidicus (approved):

stuckless (changes requested):
Hey guys, we need to start moving hugely complex and resource dependent components into a separate repo.  Now that you are a lerna repo, there's no reason to push components like this in to the core repo.  You can create a carbon-addons-map repo and put the component there and then build it out.

In graphite, the Map component is a separate repo for this reason.  Not all apps will need or want it, and, making it a dependency of the core components would just bloat the apps that don't need it.

stuckless (changes requested):
Once the map is in a separate repo... I then think you need to better refactor the map code into a map component and a separate MapCard that can consume a map component.  It makes no sense to bring in a map component and ONLY allow it to be used in a dashboard.

davidicus:
This is not a map component Sean. It is a wrapper for map component that provides certain pieces for things like map controls, legends and affordances for drag and drop. Since we must support multiple providers like MapBox, Openlayers, etc. it would be impossible to standardize an api that would work for them all.

davidicus:
Notice that mapbox and openlayers are dev dependencies and not shipped with the package.

stuckless:
@davidicus thanks for the dev dependencies clarification.  I'm still thinking this looks like a dashboard component, in that it has lots of properties that are very specific to the dashboard.  If this were uses outside a dashboard how are those properties interpreted?  I just feel that there is a lot of good things in this that should be separated to be clear that we can use this stand alone and/or in a dashboard card.  ie, what is a MapCard vs Map.  ie, a Card implies to me that there is some decorations here that are card based, but, what if the map were full page?  or in a dialog?

This says that it implements the PAL Interactive Map but it appears to provide a very narrow implementation that is designed for the primary use of a dashboard.  In graphite we have maps in many apps but none are in a dashboard.

So, this is why I say that there really should be 2 components here... a core MapLayout which manages all the stuff you mentioned like layers, legend, etc, and another component MapCard that consumes the MapLayout and the MapCard exposes the card behaviours that are specific to dashboards.

Did you guys talk to the Maximo spatial team that has built out the mapping solutions used in the Graphite mobile and web applications?  Seems like there is a huge overlap here, and it's likely that they could eventually consume what you are doing here, if they are involved at the ground level.  Otherwise, we have 2 teams working on map solutions, which seems like a waste given the limited resources.

davidicus:
@stuckless I believe we had asked Jim in a priority meeting about map devs contributing back to the PAL in some way but had never gotten a response on that. When the push to implement MAP component came we looked to the published map spec. In the fullscreen state of the spec this component still seemed to share many of the same characteristics of a Card.

image

So to avoid duplication of functionality and code that would need to keep parity with each other and to make it easier for the same component to be placed either in a dashboard context or a full screen context keeping it in the Card made the most sense, IMO. 

Since our implementation issue was originally a card and requirements other than "meeting the spec" not clearly defined I figured it prudent to get an experimental component up for people to try and then let the component evolve when we had more clearly defined use cases. 

I completely agree that common components should be created in a sharable environment precisely to avoid duplication of efforts. This is exactly what we are trying to do with the PAL.

packages/react/src/components/MapCard/Legend.jsx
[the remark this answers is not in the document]
bjornalm: Sounds good. Fixed for this PR

PR #2369, opened 2021-06-02
feat(dashboardeditor): add edit data item button to the dashboard editor

herleraja:
**Summary**

- add edit data item button to the dashboard editor

- feat(CardEditor) : added a new edit data items button and moved add cards button from top header to the footer. 
- renamed stories for ListBuilder/MenuButton to add number prefix, so they appear in correct location

- new edit data item button should be added to the dashboard editor
Earlier: 
image

Now : 
image

- Dashboard editor functionality should work as expected


packages/react/src/components/DashboardEditor/DashboardEditor.story.jsx
```diff
@@ -591,6 +591,7 @@ export const SummaryDashboardWithInitialValue = () => (
       onCancel={action('onCancel')}
       onSubmit={action('onSubmit')}
       onLayoutChange={action('onLayoutChange')}
+      onEditDataItems={action('onEditDataItems')}
```
davidicus, line 594: Any idea why we don't see this action logged in the knobs panel of the story?

herleraja: I dont know the exact reason, But even the other button's actions are not getting displayed properly. (example: cancel, delete button)

Whereas in my local machine I can see that it's triggering the action. (only one difference is it seems to be Class Object whereas submit button is array of object)

image

davidicus (approved):

kevinsperrine (approved):

PR #2380, opened 2021-06-08
2366 MapCard related content panels and react-dnd upgrade

### LINKED ISSUE #2366: [MapCard] related content panels
### What package is this for?

- [x] React
- [ ] Angular

### Summary

It should br possible to add panels of related content to the MapCard.

As I understand there is no UI spec for this so I will provide some assumptions/suggestions:

_Basic_
1. The panels should look similar to the non full bleed legend with regards to background color, opacity, padding etc.
2. It should be possible to add one or more panels. We can use a prop called `relatedContentPanels` or something to allow the user to specify multiple panel configuration objects containing id, title and a "render prop" for the content.
3. The panels should stack in the top left corner since that is the only available space in the map. Stacking panels can be achieved by placing them in an absolute positioned container.

_Draggable_
4. The user should be able to drag and drop panels inside the map. 
5. When a panel is repositioned its new position is emitted to the app. The position is in percentage of the map container, similar to how image hotspots work, meaning that a panel with position x:50, y:50 is placed at the centre of the map.
6. If we make "draggable" a property of the panel configuration object it will be possible to combine draggable and non draggable panels on the same map
7. Draggable panels cannot be placed in the absolute positioned container as described in point 3 and automatic stacking/positioning of draggable elements is more complex so I suggest that the app should be responsible for the initial/default position of a draggable panel.

bjornalm:
Closes #2366

**Summary**
Upgrades `react-dnd` to the latest version from 11.1.3 to 14.0.2 and the `react-dnd-html5-backend` from 11.1.3 to 14.0.0.
This was needed to use the latest hook based version of the drop target and being able to add it to the MapCard. The MapCard can now accept multiple refs and an example of how to add custom draggable panels was added to storybook and the readme file. 

1. Upgrade  `react-dnd` and `react-dnd-html5-backend`
2. Updated the useDndProviderElement hook to work with the latest version
3. Added a mergeRefs function to the DragAndDropUtils.js that allows us to use multiple refs for the same component
4. Added a `dropRef` prop to MapCard
5. Created a new demo file called MapboxDragPanelExample.jsx that showcases how to create custom draggable panels
6. Updated tests
7. Updated docs
8. Added MapboxDragPanelExample in a story

image

- Try out the new story with draggable panels https://deploy-preview-2380--carbon-addons-iot-react.netlify.app?path=/story/2-watson-iot-experimental-☢%EF%B8%8F-mapcard--mapbox-drag-panels-story
- Proof read the new section "Custom draggable panels" in the readme

- Try out all stories using drag and drop. That should be any story with either SimpleList or Table column reordering


stuckless (dismissed):
When you drag panels off the left side they become clipped.  When you drag them off the right side they become resized (usually, but also could become clipped).

What is the PAL design behaviour for panels?  Should the just dock top, bottom, left, right once they cross the edge?  I don't think they should resize or become clipped.

davidicus:
> When you drag panels off the left side they become clipped. When you drag them off the right side they become resized (usually, but also could become clipped).
> 
> What is the PAL design behavior  for panels? Should the just dock top, bottom, left, right once they cross the edge? I don't think they should resize or become clipped.

I do not think the spec goes into these specifics regarding this but I agree. We should probably have them stick to the edge if they go off the map. No resize or clipping.

bjornalm:
> > When you drag panels off the left side they become clipped. When you drag them off the right side they become resized (usually, but also could become clipped).
> > What is the PAL design behavior  for panels? Should the just dock top, bottom, left, right once they cross the edge? I don't think they should resize or become clipped.
> 
> I do not think the spec goes into these specifics regarding this but I agree. We should probably have them stick to the edge if they go off the map. No resize or clipping.

@davidicus This should be handled by logic and css outside the MapCard, e.g. setting a fixed size on the panels will prevent them from being resized. This example is the simplest implementation needed to show the basics and verify the concept of adding children as draggable elements. If you want I can expand the example to prevent resize and clipping?

kevinsperrine (approved):

bjornalm:
I updated the example to prevent panel resize and clipping. If a panel is dropped too close to the edge or partly outside the container it will be automatically repositioned to the closest "correct" position at the edge but with the proper padding.

davidicus (approved):

stuckless (commented):
Thx for fixing the drag/resize... but there's still an odd dragging experience that makes for an inconsistent dragging experience.  That is sometimes when I drag something slightly outside the map it docks close to where I dragged it (makes sense), but other time it just resets itself back to where it was.  I think it is using some condition of the mouse pointer (ie, if the pointer is outside map box then reset otherwise figure out best place.  But this logic should relly be applied agiast the drag box and not the mouse pointer.

PR #2386, opened 2021-06-09
feat(side-panel): initial

### LINKED ISSUE #2251: Add filter pane component
https://pages.github.ibm.com/ai-applications/design/components/datatable/filtering#filter-pane

zvonimirfras:
Closes #2251 

**Summary**

- add the initial side panel


packages/angular/src/side-panel/side-panel.component.ts
```diff
@@ -0,0 +1,136 @@
...
+  ChevronLeft16,
+  ChevronRight16,
+  OpenPanelLeft16,
+  OpenPanelRight16,
+} from '@carbon/icons';
+import { IconService } from 'carbon-components-angular';
+
+/**
+ *
+ * [See demo](../../?path=/story/components-side-panel--basic)
```
theiliad, line 21: Why not the full URL?

theiliad (changes requested):
Should we also do something about the broken UI in the story?

image

packages/angular/src/side-panel/side-panel.component.ts
zvonimirfras: Cause it doesn't work on localhost and relative works anywhere.

theiliad: Oh nice, I thought the link is just there in the comments, didn't realize that it'll render somewhere

theiliad (approved):
LGTM

Najil-K:
@zvonimirfras The Slide-in & inline variations come without any dummy content, also the buttons are misaligned. would recommend fixing it to do a design review.

image

image

FYI @dianatran18

zvonimirfras:
Looks like netlify didn't catch all the new styles.
Netlify rebuilt, should work as expected now.

Najil-K:
@zvonimirfras, @dianatran18 

PDF: [UI audit - Date Time Range Selector (v1).pdf](https://github.com/carbon-design-system/carbon-addons-iot-react/files/6703191/UI.audit.-.Date.Time.Range.Selector.v1.pdf)


UI audit - Date Time Range Selector (v1) @2x

cal-smith:
There's something really funky going on with the styles:

image

I guess that's due to the scrollbar?

cal-smith:
Also, we can probably update `carbon-components-angular` and use the fancy new button set component for the action buttons ✨

zvonimirfras:
Issue 1: some title options are demo specific and require js, so you might not be able to test them all by just editing html in browser.

Issue 3: that's a demo thing only because it's the height of the content and as content looses space horizontally, it becomes taller.

The rest should be good to go 👍

packages/angular/src/side-panel/side-panel.stories.ts
```diff
@@ -0,0 +1,139 @@
...
+        (close)="active = !active; close.emit()"
+      >
+        <div aiSidePanelTitle [condensed]="titleCondensed">Filter</div>
+        <div class="panel-content">Content</div>
+        <div aiSidePanelFooter>
+          <button ibmButton="secondary" size="lg">Cancel</button>
+          <button ibmButton size="lg">Initiate</button>
+        </div>
+      </ai-side-panel>
+    </div>
```
cal-smith, line 83: It seems like it should be possible to directly include this template in the story ... We should probably avoid adding custom demo components wherever possible (for demo/doc readability) ... If we really do need it we should add a warning comment to the story template

cal-smith (approved):
Looks good overall, just one demo comment (optional) but otherwise good to go

packages/angular/src/side-panel/side-panel.scss
```diff
@@ -0,0 +1,154 @@
+@import '../../../styles/src/globals/vars';
+@import '../../../styles/src/globals/typography';
```
cal-smith, line 2: This should be imported from the package local vendor folder ... Which isn't published?? (https://unpkg.com/browse/@ai-apps/angular@2.146.0-next.59/) ... I'm checking that out ...

cal-smith (changes requested):
One thing I just caught, and a surprise bug in the publish, but otherwise should be good

packages/angular/src/side-panel/side-panel.scss
cal-smith: If we can wait for #2617 this won't be an issue

cal-smith: #2617 is in, if we update this it will be good to go 👍

cal-smith (approved):

PR #2387, opened 2021-06-09
fix(table): view management

sstone2423:
Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2409, https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2410, https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2411

**Summary**

- When user view management is enabled and the `onUserViewModified` callback is fired for search value changes, the *true* search value is not given back.  Fixed the useEffect to handle this

- TableViewDropdown's `View all` option is basically a 'default' view and should never be able to get overwritten. I've removed the option to be able to 'Save changes' and only provided the 'Save as new view' instead. The power of overwriting default views should be given to the consumer as they would most likely make a separate "default" view that could be overwritten, but still preserve the *true* default meaning, show the original table without any config

- TableViewDropdown's actions should be at the top of the list so that users don't need to scroll the entire list (if there are many views) to get to the actions

- Fixed the existing TableViewDropdown story to use the correct edited prop
- Added additional TableViewDropdown stories to show more use cases
- Organized TableViewDropdown's tests into better defined categories
- Updated and added new tests for TableViewDropdown to reflect the above changes

- Go to http://127.0.0.1:3000/?path=/story/1-watson-iot-table--stateful-example-with-create-save-views , add a breakpoint in the `onUserViewModified` and begin to type into the search bar. You should see the values being logged and the rendered values as the same value. (previously this would stop showing the true values)
- Go to http://127.0.0.1:3000/?path=/story/1-watson-iot-%E2%9A%A0%EF%B8%8F-table-tableviewdropdown--table-view-dropdown-edited and ensure that there is no 'Save changes' button
- Go to http://127.0.0.1:3000/?path=/story/1-watson-iot-%E2%9A%A0%EF%B8%8F-table-tableviewdropdown--table-view-dropdown-edited and ensure that the actions are now at the top of the list

- Ensure that you can still see the 'Save changes' button when using custom views in edited mode with this story: http://127.0.0.1:3000/?path=/story/1-watson-iot-%E2%9A%A0%EF%B8%8F-table-tableviewdropdown--custom-view-selected
- Check the default search value stories. I'm still not really sure what the purpose of that prop was


bjornalm:
Is there a github issue related to this PR?

bjornalm:
### Regarding initialDefaultSearch  
It is a complicated situation. I don't remember all the exact details, but I think the reason for needing it can be traced back to the carbon Search component and that it keeps its own state. Workarounds have been added for it not to loose focus when having externally managed state and to allow the search term be updated via changes to the table props.

I found this issue solved by Taylor that gives a clue of what is going on:
https://github.com/carbon-design-system/carbon-addons-iot-react/issues/1386

Your current PR breaks this story, i.e. it is no longer possible to externally update the search value via the props. (We should really have this in a test rather than a story)
https://deploy-preview-2387--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-pre-filled-search

packages/react/src/components/Table/StatefulTable.jsx
```diff
@@ -322,8 +322,8 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
         toolbar: {
           ...view?.toolbar,
           search: {
-            ...view?.toolbar?.search,
             defaultValue: initialDefaultSearch,
+            ...view?.toolbar?.search,
```
bjornalm, line 326: I think this change prevents the search term from being externally updated via the table props

bjornalm (changes requested):

davidicus:
We really need to start creating issues for complex component changes so we can understand the impact and can better understand the changes in the future. It will also help the reviewers to have the context.   

This will also allow us to discuss the changes you are proposing to make before you have already done the work.

kevinsperrine (approved):
After @bjornalm's comments are addressed, I think this is ready. I don't see any other issues.

packages/react/src/components/Table/TableViewDropdown/TableViewDropdown.jsx
```diff
@@ -102,21 +102,28 @@ const TableViewDropdown = ({
...
+    // their own default views that can be changed if they would like that functionality
+    const dialogItems =
+      selectedViewEdited && selectedViewId && selectedViewId !== 'view-all'
+        ? [saveAsNewItem, saveItem, manageViewsItem]
+        : [saveAsNewItem, manageViewsItem];
 
-    return [viewAllItem, ...views, ...dialogItems];
+    // move the action / dialog items to the top so that they are always easily accessible in the case that there are
+    // many views. The user would need to scroll all the way to the bottom to find the actions
+    return [...dialogItems, viewAllItem, ...views];
```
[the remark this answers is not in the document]
sstone2423: Verified with the Monitor designers. Could you provide the design here @amymillett @CynZhang

packages/react/src/components/Table/StatefulTable.jsx
sstone2423: I had a worry that would happen

sstone2423:
> ### Regarding initialDefaultSearch

> If you need the latest search value, you can always use `onApplySearch`.
> 
> I think your problem is related to _when_ the `onUserViewModified` is called. It is only called when the table is being rendered with modified view props. So typing a single search character like 'x' won't trigger it since a single character won't trigger a search (it only updates the internal state of the TableToolbarSearch). Adding another character so that we now have 'xx' will trigger the `onUserViewModified` since new search result will be available and the table is re-rendered. Typing a third, fourth etc character _might_ trigger the `onUserViewModified`, i.e. only if the search result changes.
> 
> So a possible solution for you would be to trigger the `onUserViewModified` also when `onApplySearch` is triggered, but only if the view has not changed.

I would think `onUserViewModified` should always be fired for search changes as with the other view changes. Its a little bit odd that we don't find it on every keypress, unless the consumer specifically gives a debounce. I'll take a look and see how it works if I called `onUserViewModified` directly after the search

packages/react/src/components/Table/StatefulTable.jsx
scottdickerson: ok I've undone this to address in a different fix to the onUserViewModified callback

packages/react/src/components/Table/TableViewDropdown/TableViewDropdown.jsx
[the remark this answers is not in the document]
scottdickerson: I think this change is good too and verified

scottdickerson (approved):
LGTM

PR #2390, opened 2021-06-09
TearSheet header height change fix

hectordavis:
**Summary**
- This fixes an issue where if a content were dynamically added to the TearSheet header, the TearSheet content would expand beyond its container and get cropped by the bottom of the viewport.

- Check the component story. I've added buttons on the tear sheets to include additional content to their header. The tear sheet size should be readjusted to prevent it from being cropped.

The tests below can be checked via story
- Check the stack feature (one tear sheet on top of the other) to see if they still behave like before. 
- Check if the tear sheet's whole content is displayed

### DIFF
--- a/packages/react/src/components/TearSheet/TearSheetWrapper.jsx
+++ b/packages/react/src/components/TearSheet/TearSheetWrapper.jsx
@@ -38,7 +38,7 @@ const TearSheetWrapper = ({ isOpen, className, onCloseAllTearSheets, children })
   const [containers] = useState(childrenArray.map(() => createRef()));
   const [initialContainersWidth, setInitialContainersWidth] = useState(null);
   const [containersStyles, setContainersStyles] = useState({});
-  const [openTearSheetHeaderHeight, setOpenTearSheetHeaderHeight] = useState(null);
+  const [activeTearSheetHeaderHeight, setActiveTearSheetHeaderHeight] = useState(null);
   const [animationClasses, setAnimationClasses] = useState(
     childrenArray.reduce(
       (acc, c, idx) => ({
@@ -62,18 +62,18 @@ const TearSheetWrapper = ({ isOpen, className, onCloseAllTearSheets, children })
           tearSheetsNodes?.[activeTearSheetIdx]?.children?.[0].getBoundingClientRect().height
         )
       ) &&
-      openTearSheetHeaderHeight !==
+      activeTearSheetHeaderHeight !==
         Math.round(
           tearSheetsNodes?.[activeTearSheetIdx]?.children?.[0].getBoundingClientRect().height
         )
     ) {
-      setOpenTearSheetHeaderHeight(
+      setActiveTearSheetHeaderHeight(
         Math.round(
           tearSheetsNodes?.[activeTearSheetIdx]?.children?.[0].getBoundingClientRect().height
         )
       );
     }
-  }, [activeTearSheetIdx, tearSheetsNodes, openTearSheetHeaderHeight]);
+  }, [activeTearSheetIdx, tearSheetsNodes, activeTearSheetHeaderHeight]);
 
   useEffect(() => (isOpen ? setActiveTearSheetIdx(0) : setActiveTearSheetIdx(null)), [isOpen]);
 
@@ -146,7 +146,7 @@ const TearSheetWrapper = ({ isOpen, className, onCloseAllTearSheets, children })
     windowSize.width,
     windowSize.height,
     windowSize,
-    openTearSheetHeaderHeight,
+    activeTearSheetHeaderHeight,
   ]);
 
   const goToSheet = (idx) => setActiveTearSheetIdx(idx);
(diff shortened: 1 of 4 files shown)


packages/react/src/components/TearSheet/TearSheetWrapper.jsx
```diff
@@ -48,6 +49,32 @@ const TearSheetWrapper = ({ isOpen, className, onCloseAllTearSheets, children })
...
+  const tearSheetsNodes = containers.reduce(
+    (acc, curr) => [...acc, curr?.current?.children[0]],
+    []
+  );
+
+  useEffect(() => {
+    if (
+      activeTearSheetIdx !== null &&
+      !Number.isNaN(
+        Math.round(
```
blechner, line 61: Maybe `Math.round(tearSheetsNodes?.[activeTearSheetIdx]?.children?.[0].getBoundingClientRect().height)` could be a variable.

hectordavis: Good catch. Thanks!

blechner (approved):

packages/react/src/components/TearSheet/TearSheetWrapper.jsx
```diff
@@ -48,6 +49,24 @@ const TearSheetWrapper = ({ isOpen, className, onCloseAllTearSheets, children })
...
   );
 
+  const tearSheetsNodes = containers.reduce(
+    (acc, curr) => [...acc, curr?.current?.children[0]],
+    []
+  );
+
+  useEffect(() => {
+    const tearSheetHeaderHeight = Math.round(
+      tearSheetsNodes?.[activeTearSheetIdx]?.children?.[0].getBoundingClientRect().height
```
kevinsperrine, line 59: This probably makes more sense as a useLayoutEffect since you're calling getBoundingClientRect for measurements. https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

kevinsperrine (approved):
just the one note on useLayoutEffect.

davidicus:
I am not understanding the reason to calculate the height of the header. Can't we have two items in a flex container that can grow with their content? Like this example? https://codesandbox.io/s/green-http-zq8jr

davidicus (changes requested):
Waiting on reply

hectordavis:
Hey, @davidicus 
The tear sheet has a fixed position. If something changes the header height, that's going to affect the rest of the tear sheet because the `--content-max-height` value won't be changed. Now, if we change the window size for instance, this would be recalculated and the tear sheet would adjust, but a change in the header's height won't trigger that. At least it didn't before this PR 🙂 

The tear sheet has a top limit and a bottom limit. The header is taken into consideration as well to decide the max-height of the **content**. After the calculations to get this max-height are done, we need something to trigger them again if something changes (other than what we already checked in the useEffect, like the active tear sheet index, window size, the initial container width, etc.)

The header can't grow upwards, so it'll force the content down causing the tear sheet to be cropped. That's why we need to recalculate the content max-height when the header changes.

It's also worth noting that there's also the case of stacked tear sheets, so they need to keep the same difference in position from each other (top and bottom) even if their headers change dynamically.

In short, I'd say the main point is the fixed positioning.

davidicus:
> Hey, @davidicus
> The tear sheet has a fixed position. If something changes the header's height, that's going to push down the rest of the tear sheet because the `--content-max-height` value won't be recalculated. Now, if we change the window size for instance, then the tear sheet would adjust due to the existing useEffect, but a change in the header's height won't trigger that. At least it didn't before this PR 🙂
> 
> The tear sheet has a top limit and a bottom limit. The header is taken into consideration as well to decide the max-height of the **content** below. After the calculations to get this max-height are done, we need something to trigger them again if something changes (other than what we already checked in the useEffect, like the active tear sheet index, window size, the initial container width, etc.)
> 
> The header can't grow upwards, so it'll force the content down causing the tear sheet to be cropped. That's why we need to recalculate the content max-height if the header is different in size than it was before. Let's pick a TearSheet with a height of 700px (header: 100px, content: 600px). If we add a 50px high content in the header, the expected would be: `header: 150px, content: 550px`, but without the calculation, we have: `header: 150px, content: 600px` resulting in a TearSheet 750px high (with 50px cropped at the bottom).
> 
> It's also worth noting that there's also the case of stacked tear sheets, so they need to keep the same difference in position from each other (top and bottom) even if their headers change dynamically.
> 
> One thing that might come to mind about all of this is: "Why not just use `height: 100%` and / or `max-height: 100%`... Well, it doesn't work. We need the "fixed" `--content-max-height` value to set the tear sheet content to the maximum allowed to work as the design documents and to keep the style when stacking, etc...
> 
> In short, I'd say the main point is the fixed positioning + the tear sheet's content max-height.

@hectordavis I think this can be achieved with styles. The spec is saying tear sheets always are 88px from top of screen and go to bottom of screen. I think the layout and placement can be achieved with styles alone. I may be missing something so perhaps we can hop on a  call.

hectordavis:
Closing this PR as the issue was fixed by PR #2400

PR #2392, opened 2021-06-11
feat(list): Add angular list component 

### LINKED ISSUE #2245: Add list component
https://pages.github.ibm.com/ai-applications/design/components/list/overview/

Donisius:
Closes #2245 

Todo:
- [ ] Add support for single select
- [ ] Add secondary value for list item
- [ ] Add icon support for list item
- [ ] Add support for custom list item/header templates
- [ ] Add documentation
- [ ] Add more storybook examples
- [ ] Create a secondary hierarchical list component that wraps the list component so that list model no longer requires `_displayed` items and can just deal with `items` or maybe deal with filtering through a `hidden` tag which would potentially over-complicate things
- [ ] Make search function in list model easily overridable
- [ ] Find a way to recursively render list items without the need for the `expanded` field in the list model

### DIFF
--- a/packages/angular/src/list/index.ts
+++ b/packages/angular/src/list/index.ts
@@ -1,0 +1,7 @@
+export { ListComponent } from './list.component';
+export { ListHeaderComponent } from './list-header.component';
+export { ListItemComponent } from './list-item.component';
+export { ListModel } from './list-model.class';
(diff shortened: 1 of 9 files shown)


Donisius:
Closing in favor of #2424

PR #2414, opened 2021-06-15
chore(repo): add addon-docs, remove addon-info and storybook-readme

### LINKED ISSUE #963: [Storybook] upgrade to v6 - use addon-docs instead of addon-info

### LINKED ISSUE #2418: [Repo]: Update README.md files to {Component}.md
This is to ensure consistency in the documentation going forward.

kevinsperrine:
Closes #963
Closes #2418 

~Blocked until #2359 and #2336 are merged.~

**Summary**

- removes @storybook/addon-info
- removed storybook-readme
- replaces them with @storybook/addon-docs
- renames all existing MD files to MDX, and uses them as the basis for the new docs tab.
- sets a resolution for webpack 4, because of an incompatibility with an old webpack 4.28 version in the lockfile. 4.41 works fine.

Component w/ stories using `storybook-readme` that were updated
- ListBuilder
- MapCard
- MenuButton
- MultSelect
- OverflowMenu
- Table

- Clicking the new "Docs" tab (beside "Canvas" at the top left of the main storybook window) for any of the stories above should show their current docs
- The "Readme" tab should do longer appear

- All stories should still function as expected


package.json
```diff
@@ -61,20 +61,21 @@
...
     "lint-staged": "^8.1.0",
     "prettier": "^2.1.2",
-    "prettier-config-carbon": "^0.4.0",
-    "arg": "5.0.0"
+    "prettier-config-carbon": "^0.4.0"
   },
   "sideEffects": false,
   "version": "0.0.0-development",
   "resolutions": {
-    "carbon-addons-iot-react/chokidar": "3.3.1",
```
davidicus, line 74: I think is the package resolution syntax we were discussing in scrum. Is there a reason we are removing the scope of our package?

kevinsperrine: That's a good call. I was playing around figuring out how those worked exactly. I'll add it back in.

kevinsperrine: re-added.

packages/react/src/components/MapCard/MapCard.mdx
```diff
@@ -464,4 +464,4 @@ const MapboxDragPanelExample = ({ data }) => {
 ## Feedback
 
 Help us improve this component by providing feedback, asking questions on Slack, or updating this file on
-[GitHub](https://github.com/carbon-design-system/carbon-addons-iot-react/tree/next/packages/react/src/components/MenuButton/README.md).
+[GitHub](https://github.com/carbon-design-system/carbon-addons-iot-react/tree/next/packages/react/src/components/MenuButton/MenuButton.mdx).
```
bjornalm, line 467: This is actually my old copy/paste error :-) It should link to the MapCard not the MenuButton

bjornalm (commented):
Nice work, a small changes needed and there were some questions about dependencies

bjornalm:
We have a problem with the new docs and modals:
image
image

kevinsperrine:
> We have a problem with the new docs and modals:
> image
> image

nice catch! I'll copy the MDX file from Carbon to prevent it and give better docs--this probably something we'll need to do eventually anyway, I assume. We were hoping carbon would bundle the MDX and we could just use then, but I don't know if that's going to happen or not.

packages/react/package.json
```diff
@@ -283,5 +282,9 @@
     "whatwg-fetch": "^3.0.0"
   },
   "sideEffects": false,
-  "version": "2.146.0-next.33"
+  "version": "2.146.0-next.33",
+  "resolutions": {
```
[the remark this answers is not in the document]
kevinsperrine: Yeah, technically they aren't doing anything b/c yarn 1 ignores package level resolutions in a monorepo setup, but @davidicus and I discussed adding them here yesterday just for clarity since they're actually being enforced by the root-level package.json.

package.json
```diff
@@ -75,6 +75,7 @@
     "carbon-addons-iot-react/react-grid-layout": "1.2.2",
     "react-test-renderer": "16.14.0",
     "react-dom": "16.14.0",
-    "react": "16.14.0"
+    "react": "16.14.0",
+    "webpack": "4"
```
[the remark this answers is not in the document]
kevinsperrine: There was an old webpack 4.28 version that caused incompatibilities with the addon-docs library and handling the MDX, so I added the 4 resolution to force it to bump higher and fix that issue. Angular is already using "4" in their project, and we're using it 4.41(?) on our side. I can note that in the summary.

packages/react/src/components/MapCard/MapCard.mdx
kevinsperrine: fixed.

.github/README.MD
```diff
@@ -137,7 +137,7 @@ these packages if you use them directly in your project.
 
 ## 🤲 Contributing
 
-Please check out our [Contribution Guidelines](./.github/CONTRIBUTING.MD) for
+Please check out our [Contribution Guidelines](./CONTRIBUTING.MD) for
```
kevinsperrine, line 140: So this one is tough. By changing this it's fixed when viewing the readme on Github, but will be broken when viewing it on NPM. 

Check the link on github here: https://github.com/carbon-design-system/carbon-addons-iot-react/tree/963-addon-docs

After this change, the NPM link will result in: https://github.com/carbon-design-system/carbon-addons-iot-react/blob/HEAD/CONTRIBUTING.MD which is a 404.

davidicus: Is there a way to use the absolute link in the MD?

kevinsperrine: Yep, but even that has some side effects. It can break local links in docs. Do we point to next or master?

bjornalm:
> > We have a problem with the new docs and modals:
> > ...
> nice catch! I'll copy the MDX file from Carbon to prevent it and give better docs--this probably something we'll need to do eventually anyway, I assume. We were hoping carbon would bundle the MDX and we could just use then, but I don't know if that's going to happen or not.

It is not just the modal, but all stories that use the modal as well :-)

.github/README.MD
kevinsperrine: fixed.

bjornalm (approved):
Nice work!

.github/README.MD
davidicus: Resolved. Adding absolute links

davidicus (approved):

PR #2424, opened 2021-06-21
feat(list): Add angular list

### LINKED ISSUE #2245: Add list component

Donisius:
Closes #2245 

### Angular Todo:
- [x] Add support for nested list items
- [x] Add support for single select for simple and nested list items
- [x] Add support for multi select for simple and nested list items
- [x] Add secondary value for list item
- [x] Add drag and drop reorder for simple and nested list items
- [x] Add support for row actions
- [x] Add an easy way to pass in context for row actions
- [x] Add documentation
- [x] Make an overridable filter function in the list model
- [x] Add support for different list item sizes
- [x] Add empty state
- [x] Add support for disabled list items
- [x] Find a way to recursively render list items without the need for the `expanded` field in the list model
- [x] Update styles #2601
- [ ] Add more storybook examples
- [ ] Add support for custom list header templates
- [ ] Add support for pagination - Need to add simple pagination
- [ ] Add tests

### Changes in react:

**packages/react/src/components/List/List.jsx**

- Increment nesting level of a list item only if there exists at least one expandable item in the list item's hierarchical level.
    - Based on the specs [here](https://pages.github.ibm.com/ai-applications/design/components/list/overview#categorized-checklist-example-with-pagination) and [here](https://pages.github.ibm.com/ai-applications/design/components/list/overview#hierarchical-list-example-with-search-and-selection)

**packages/react/src/components/List/ListItem/ListItem.jsx**

- Nesting offset now has a width of 32px
    - Based on #2601
- Added disabled classes
    - Based on #2601

### Style changes

Changes to styles are based on #2601

**packages/react/src/components/List/ListItem/_list-item.scss**

- Added focus border for list item
- Added disabled handle
- Added disabled expand
- Added focus border for expand icon
- Updated spacings to match new specs


theiliad (approved):

packages/angular/src/list/sample-data.ts
```diff
@@ -0,0 +1,134 @@
...
+        isDraggable: true,
+      }),
+      new AIListItem({
+        value: 'Vancouver',
+        secondaryValue: '2,581,000',
+        isSelectable: true,
+        isDraggable: true,
+      }),
+    ],
+  }),
```
cal-smith, line 47: Very cool

packages/angular/src/list/stories/app-custom-list.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  includes(searchString: string) {
+    return (
+      (this.secondaryValue !== undefined &&
+        this.secondaryValue !== null &&
+        this.secondaryValue.includes(searchString)) ||
+      this.items.some((listItem) => listItem.includes(searchString))
+    );
+  }
```
cal-smith, line 16: Neat!

cal-smith: Would be cool to have a mixed ListItem demo too

cal-smith (changes requested):
We'll also need a `package.json` in the root of the angular component (check out the other components for what that should look like), so that it builds the list as a separate module.

Otherwise, a few comments, but it's looking good!

Took a quick look at the react changes as well, and they seem reasonable to me, but I definitely want someone more familiar with that to check it over.

packages/styles/src/components/GaugeCard/_gauge-card.scss
```diff
@@ -4,8 +4,10 @@
   display: flex;
   flex-wrap: wrap;
   font-size: carbon--type-scale(1);
-  height: 100%;
-  width: 100%;
+  flex-direction: row;
```
davidicus, line 7: @cal-smith It was my understanding that this package was auto generated from styles in the React package. Are we making edits directly to styles package now? Will the React and Angular styles now be divergent?

davidicus: or is this the newly generated styles that have been pulled from React package?

davidicus (approved):
React changes LGTM

packages/styles/src/components/GaugeCard/_gauge-card.scss
cal-smith: I'm guessing it's pulled from styles changed in React, or it hasn't been updated after some changes ... fyi @Donisius

Donisius: Seems to be pulled over from the changes in react. I ran `sync` and `build` in the styles package, is there anything else that needs to be done (besides `postinstall` in angular) @cal-smith?

cal-smith: Nope, that's it! This should be good then :+1:

packages/angular/src/list/list-item/ai-list-item-wrapper.component.ts
```diff
@@ -0,0 +1,81 @@
...
+            (dragOver)="dragOverAbove.emit($event)"
+          ></div>
+          <div
+            aiListTarget
+            targetPosition="below"
+            (dropping)="droppedBelow.emit($event)"
+            (dragOver)="dragOverBelow.emit($event)"
+          ></div>
+        </div>
+        <ng-container [ngTemplateOutlet]="listItem"></ng-container>
```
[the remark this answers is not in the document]
Donisius: I don't think having 2 `ng-content`s works (even if they are mutually exclusive like in this case). The last `ng-content` is always the only one that will work.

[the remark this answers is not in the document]
Donisius: What we can do is check if wrapper is necessary in `ai-list.component.ts`, if that would be less confusing.

packages/angular/src/list/list-item/ai-list-item.class.ts
```diff
@@ -0,0 +1,163 @@
+import { TemplateRef } from '@angular/core';
+
+export class AIListItem {
```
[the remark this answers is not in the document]
Donisius: At the moment all changes do seem to be picked up at least with `expanded`, `selected`, and `indeterminate`. I can add some storybook knobs to test out the other properties and make sure they are also picked up by the change detector.

packages/angular/src/list/list-item/ai-list-item-wrapper.component.ts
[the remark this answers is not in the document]
cal-smith: I'm not sure I follow, I'm not suggesting we _add_ an `ng-content` just that we replace the `ng-container` with the `ng-content` from below ... as it stands we're doing exactly that but with more steps

packages/angular/src/list/list-item/ai-list-item.class.ts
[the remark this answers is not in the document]
cal-smith: Sure ... More just something to consider in the event that change detection related bugs crop up

packages/angular/src/list/list-item/ai-list-item-wrapper.component.ts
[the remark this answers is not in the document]
cal-smith: Ah I found it, it's hidden above `*ngIf="draggable && !disabled; else listItem"` ... Well that makes more sense.

It does make me wonder if here is the right place to be handling that though, if we only need the wrapper _sometimes_ maybe that switch should be somewhere else? Not something I think we need to fix at the moment, but something to think about ... I'll resolve this!

cal-smith (approved):

PR #2425, opened 2021-06-21
fix(datetimepicker): listen to locale and initialize dayjs with it

scottdickerson:
Closes #

**Summary**

- Fixes a bug where the relative dates were not taking locale into account.  I should be able to pass `L` as the dateTimeMask property to the DateTimePicker component and dayjs needs to have the correct locale set.

Notice that the date format here is incorrect for French
image

However, the date inside the Calendar widget is showing the French dates correctly

image

- fix(DateTimePicker): make a story that uses non-default locales and set the locale inside dayjs
- chore(snaps): I have no idea why those two snaps were removed for List and Card, I don't see those stories anymore, but I didn't change them. I believe it was caused by this commit: https://github.com/carbon-design-system/carbon-addons-iot-react/commit/26c5ab798558b9f01a00669836f0e08370585a7d

- Verify the locale story, if you popup the date selector you should see the Day then Month as the French date, and the Calendar should still show the French dates

- Verify default locale is used in other places


bjornalm (approved):
LGTM

kevinsperrine (approved):

sstone2423 (approved):

PR #2427, opened 2021-06-22
fix(hotspoteditormodal): handle element content in hotspots

### LINKED ISSUE #2384: [HotspotEditorModal] handle hotspots that have an element as content
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

A hotspot has a content prop. According to the PropType definition it can be react element (that is directly rendered on the image) or it can be an obj containing the properties needed to render the hotspot using the HotspotContent component. The properties of the hotspot.content object include title, values, attributes etc. When the hotspot.content is a react element the HotspotEditorModal cannot edit these properties. There is an example of a hotspot with a react element as content in the HotspotEditorDataSourceTab.story.jsx which doesn't work.

### To Reproduce

1. Add a hotspot that has an element as content to the cardConfig of a HotspotEditorModal story.
2. Click the hotspot and try to modify the content 

example hotspot:

```
{
 title: 'pressure',
 x: 45,
 y: 25,
 color: '#0f0',
 content: <span style={{ padding: spacing04 }}>Stairs</span>,
}

```

### Expected behavior

We should either not allow the content prop to be a react element, or we should make sure the HotspotEditorModal handles such hotspots and disables the sections that can't be modified.

### Environment/versions:

- OS: [e.g. MacOS, Windows]

bjornalm:
Closes #2384

**Summary**
The Hotspoteditormodal now handles cases where the content of the hotspot is a react element. It does that by disabling the sections of the UI that would conflict with the content property of the hotspot, but leaves the rest of the hotspot configuration editable.

**Note!**
Same issue here as for the latest PRs. There are updated snapshots from a previous commit.

Colordropdown:
- Added disabled prop to 
- updated story to show new prop
- added unit tests
- renamed and fixed test-id prop

HotspoteditorModal:
- updated story with hotspots that has an element as content, both text and fixed
- disabled data source tab when selected hotspot content is an element 
- added unit tests 

HotspotEditorTooltipTab:
- disabled title and description input when selected hotspot content is an element 
- added unit tests 

HotspotTextStyleTab:
- disabled font color and fontsize input when selected hotspot content is an element 
- added unit tests 

1. Go to https://deploy-preview-2427--carbon-addons-iot-react.netlify.app?path=/story/2-watson-iot-experimental-☢%EF%B8%8F-hotspoteditormodal--width-existing-hotspots
2. Select the green hotspot and make sure DataSourceTab, Title and Description are disabled
3. Verify that the other configurations can be changed as normal
4. Select the white text-based hotspot "content is an element" and make sure  DataSourceTab and all font related settings are disabled. 
5. Verify that the other configurations can be changed as normal
6. Repeat the steps above for "fixed" and "text/label" hotspots that you create yourself

Verify that the other HotspotEditorModal stories work


kevinsperrine:
There's a CSS issue with the padding on the number inputs. It's currently set to `padding-right: 8rem` for the `.bx--number input[type=number]` rule, but it seems an override of `5rem` is more appropriate in this case.

Screen Shot 2021-06-22 at 10 48 05 AM

kevinsperrine:
It's also possible to edit the data source _if_ you click the data source tab _first_ before clicking the green hotspot. 
hotspot-editor-modal-content

kevinsperrine (changes requested):
The code all looks good, but there are a few usability items.

For example, when adding a new hotspot to the empty story, and then attempting to set an aggregation on it, the aggregation dropdown doesn't show any options. Plus the other two comments on CSS and Datasource editable when clicking that tab first.

Screen Shot 2021-06-22 at 11 36 27 AM

bjornalm:
> The code all looks good, but there are a few usability items.
> 
> For example, when adding a new hotspot to the empty story, and then attempting to set an aggregation on it, the aggregation dropdown doesn't show any options. Plus the other two comments on CSS and Datasource editable when clicking that tab first.
> 
> Screen Shot 2021-06-22 at 11 36 27 AM

Thanks, yes there are for sure still some UX quirks left in this component. However, the missing aggregations is probably not one of them. Aggregation methods are defined per dataItem and none of our stories have dataItems with `aggregationMethods` defined. We might wanna expand our stories to have some aggregation methods but it is unrelated to this PR.

bjornalm:
> It's also possible to edit the data source _if_ you click the data source tab _first_ before clicking the green hotspot.
> hotspot-editor-modal-content

Good catch! I will fix this

davidicus (approved):

kevinsperrine (approved):

PR #2431, opened 2021-06-23
fix(dashboard-grid): grid should generate layouts for all supported b…

JoelArmendariz:
…reakpoints for each card

Closes: #2432

**Summary**
The dashboard grid was not generating layouts for a breakpoint until the user resized the window to that breakpoint. This caused a problem with the `DashboardEditor` where missing layouts would be saved and not regenerated.

- Check for existing cards and generate layouts for them on render rather than `onLayoutChange`
- Updated the dashboardEditor stories to use proper layouts
- Change the `gridHeight` var to be set dynamically based on number of cards in the dashboard

Go to a dashboard with a grid (dashboardEditor is an easy one) and make sure that the cards render and react to the window changing correctly.
If you look in the React Devtools, you should also see that for `DashboardGrid`, the `layouts` prop is adding layouts for every breakpoint when a new card is added. Not just the current one.
This should be true for a new dashboard in the editor that is adding cards, too.

Make sure dashboards that have cards defined are still rendering as expected according to their given layouts.


sstone2423 (approved):
Great fix for a sneaky bug

kevinsperrine (commented):
I'm seeing this difference when comparing the `with initialValue` between this deploy and next. It seems a side effect of these change, yes?
|this PR| Next |
|:--|:--|
|Screen Shot 2021-06-23 at 12 53 50 PM|Screen Shot 2021-06-23 at 12 54 00 PM|

JoelArmendariz:
> I'm seeing this difference when comparing the `with initialValue` story between this deploy and next. It seems a side effect of these change, yes?
> 
> this PR	Next
> Screen Shot 2021-06-23 at 12 53 50 PM	Screen Shot 2021-06-23 at 12 54 00 PM

So this IS actually a side effect of these changes, but it is an intentional change to the story. I changed the story layouts to make sure that they were all valid and could fit all cards. That one in particular was just me trying to make sure a weird one still worked as expected (with the bottom 2 cards x offset).

kevinsperrine:
> > I'm seeing this difference when comparing the `with initialValue` story between this deploy and next. It seems a side effect of these change, yes?
> 
> So this IS actually a side effect of these changes, but it is an intentional change to the story. I changed the story layouts to make sure that they were all valid and could fit all cards. That one in particular was just me trying to make sure a weird one still worked as expected (with the bottom 2 cards x offset).

Perfect. Thanks, @JoelArmendariz.

kevinsperrine (approved):
LGTM

Issue #2145, opened 2021-03-25
[Card] Design specs


Najil-K:
### What package is this for?

- [x] React
- [ ] Angular

[Cards.pdf](https://github.com/carbon-design-system/carbon-addons-iot-react/files/6399079/Cards.pdf)
![Cards@2x](https://user-images.githubusercontent.com/81316430/116561325-d7879780-a91f-11eb-9c38-0b10228eb73e.jpg)

### Available extra resources

What resources do you have to assist this effort?
Component guidance reference:
https://pages.github.ibm.com/ai-applications/design/components/card/usage

JordanWSmith15:
I assume these updates are targeted to be made against this story:
https://carbon-addons-iot-react.com/?path=/story/watson-iot-card--size-gallery

davidicus:
@Najil-K @dianatran18 Can a card label and subtitle be present at the same time? There are no specs on the treatment of these strings when they are too large for the space. Do we want to provide ellipsis for both label and subtitle? Do we surface the overflown content in a browser tooltip or one of our tool tips like the Title?

dianatran18:
> Can a card label and subtitle be present at the same time? 

No, only one or the other. We currently only have the subtitle in our [card PAL guidance](https://pages.github.ibm.com/ai-applications/design/components/card/usage#anatomy) so we can prioritize that one to prevent confusion. (@Najil-K did the label option come from C&CS?)


> There are no specs on the treatment of these strings when they are too large for the space. Do we want to provide ellipsis for both label and subtitle? Do we surface the overflown content in a browser tooltip or one of our tool tips like the Title?

We're actually working on truncation guidance right now, but it's still a WIP. 

Currently, this is what we have in there:

| UI element  | When to truncate  | When to wrap | How to handle overflow content |
|----------------|------------------------|-------------------|-----------------------------------------|
| Card title | Truncate at the end when the title exceeds one line in the container for only a title, or exceeds two lines in the container for a title and subtitle. | Can be wrapped to no more than two lines. | Display the complete title (and subtitle, if applicable) in a browser tooltip.  |

Do card titles wrap or truncate today @davidicus ?

JordanWSmith15:
Relates to and may fix https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2370

davidicus:
>Do card titles wrap or truncate today @davidicus ?

[...]

PR #2455, opened 2021-06-30
feat(card): add customtoolbarcontent prop

### LINKED ISSUE #2454: [Card] add custom datetimepicker
### What package is this for?

- [x] React
- [ ] Angular

### Summary

 Users need to be able to swap the Card Toolbar Date Picker to use the DateTimePicker.

- [ ] We'll use the timeRangeOptions property as the list of presets that the DateTimePicker will use property, but will need to extend the prop type to also allow the updated offsets to be passed along with the translated message.  
- [ ] We will update the timeRange property to also allow the format the DateTimePicker uses to select dates
- [ ] Under availableActions we'll add a new boolean prop called useDateTimePicker, if this is set and they pass a timeRange property, we need to through
- [ ] On the onCardAction callback, we'll pass DateTimePicker range/values object in this case? or just the start/end date?

**Additional context** 

sstone2423:
Closes #2454

**Summary**

- Adds `customToolbarContent` prop to the Card component that allows custom content to be rendered to the left of the existing toolbar nodes

- Added the `customToolbarContent` prop to the custom card story

- Go to /?path=/story/1-watson-iot-card--implementing-a-custom-card and see that the Button is rendered correctly

- Existing toolbar content should still render (the expand icon in the above story)


packages/react/src/components/Card/Card.jsx
```diff
@@ -229,6 +229,7 @@ export const defaultProps = {
   onBlur: undefined,
   tabIndex: undefined,
   testID: CardWrapper.defaultProps.testID,
+  customToolbarContent: undefined,
```
kevinsperrine, line 232: Can we add this new props to the docs prop table, too?

packages/react/src/components/Card/Card.mdx
```diff
@@ -23,6 +23,7 @@ To develop a custom card component.
 
 - Create a new card component that uses the base Card component
 - If you want to hide the title/toolbar, do not pass a title prop
+- If you want to add custom toolbar content such as a different date range selector, pass the node as a customToolbarContent prop. The node will be placed to the right of the other toolbar content
```
kevinsperrine, line 26: Thanks for adding this!

davidicus:
image
@sstone2423 It looks like your dropdown is being clipped.

davidicus:
@dianatran18 Thoughts on the placement to the right of other action icons?

dianatran18:
> @dianatran18 Thoughts on the placement to the right of other action icons?

Icon buttons (like an overflow menu or expand icon) should sit at the very far _right_ of the card, since icon buttons don't change in size, but I assume the title and the date picker dropdown in this case can vary in size. Those icons shouldn't move from their place, whether there's a dropdown there or not.

I still think the trigger for the date picker would work better as an icon button, and then you display the date range as a subtitle to the header for cards as read-only content so it doesn't clutter the card header.

davidicus:
@CynZhang Would your use case work with @dianatran18 suggestion? At small card sizes this looks like it would break pretty easily. Especially if there were multiple icons on the card. @sstone2423  I think we may want to settle on the design first before pushing any code.

sstone2423:
> image
> @sstone2423 It looks like your dropdown is being clipped.

Looks like this is due to the table in the card content being too small. Just a note, thats purely an example of some custom toolbar content. It could be other things as well. I'll swap those out to ensure the full menu is visible for this example

CynZhang:
The design proposal was to use the icon button to trigger the time picker menu. We needed to deal with the placement of the selected time range on the card because we don't have a subtitle section under the main title yet. @sstone2423 we will need to hide the time range but only show the icon button to save space, and then display the time in the dropdown menu: 

image

dianatran18:
@davidicus if we have [this issue](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2145) as a val 1 can the subtitle variation be prioritized to help solve this issue? It'll be a better overall usability experience to have the date range displayed directly on the card, without a big dropdown in the header area, and also align with what other teams are using the subtitle for anyways

packages/react/src/components/Card/Card.jsx
sstone2423: Added

davidicus:
> @davidicus if we have [this issue](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2145) as a val 1 can the subtitle variation be prioritized to help solve this issue? It'll be a better overall usability experience to have the date range displayed directly on the card, without a big dropdown in the header area, and also align with what other teams are using the subtitle for anyways

Yes I don't see an issue with using the subtitle to display this. @sstone2423 @CynZhang thoughts?

davidicus:
@sstone2423 clicking the custom content in the story mentioned in the PR descriptions does not work. No action or dropdown.

CynZhang:
I think we can take this in two steps. 
1. Provide an icon only version for the time picker (like an overflow menu). The current date picker is a dropdown menu with icon which takes up more space to display on the corner of a card. 
2. Add the subtitle to our dashboard cards and use it to display time range. When a card has a custom time range set by the user, it overrides the dashboard one. Therefore it should be made visible on the card to inform the users. 

image

bjornalm (approved):
From the perspective of adding a customtoolbarcontent prop this code looks ok. Like David implies, it might be nice to connect the custom content button to story action so that we can see that something happened.

kevinsperrine (approved):

sstone2423:
After speaking with @davidicus , this is going to be swapped to having an option to swap the CardRangePicker to be either the current relative CardRangePicker OR the DateTimePicker with the new design spec

scottdickerson:
I'm going to reimplement with the new date picker

PR #2461, opened 2021-07-01
fix(imagehotspots): fixes object-fit contain fill and undefined

### LINKED ISSUE #2417: [ImageHotspots] hotspot positions are incorrect
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
If you add a hotspot to an Image card on the bottom edge via the dashboard editor and save it, the card shows the hotspots inaccurately. I think this is true for image cards with hotspots in general.

### To Reproduce

Go to the dashboard editor => add an image card => edit the image and add a hotspot to each edge => save it and see that the hotspots have moved up.

### Expected behavior
Hotspots should stay accurate.

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context

image

image

 - Here is the image I used to recreate this: 
 - 
test_image

bjornalm:
Closes #2417

**Summary**
Fixes the position of the image and hotspots for different combinations of image sizes, container sizes and objectFit settings for the image element.

HotspotEditorModal:
* added prop displayOption (which is needed when opened by the ImageCardFormContent) with null as default
* updated all stories to have a displayOption knob 
* updated story "With existing Hotspots" with hotspots in each corner of the image so that it is easier to visually identify that hotspot positions are correct.
* modified the default landscape image to be slightly blue in order to make it easy to validate visual changes

ImageCard:
* added missing prop displayOption to content prop
* added ImageCard story that shows hotspots in each corner and lets the user select a displayOption 
* fixed incorrect height and width calculations for the ImageHotspots props
* added key prop regeneration needed for stories that modifies the displayOption
* modified the default landscape image to be slightly blue in order to make it easy to validate visual changes

ImageHotspots:
* modified code to properly position and size the image tag and hotspots div when displayOption is either 'contain', 'fill' or null. This is the main feat of the PR. Required a lot of small changes.
* Updated comment for displayOption prop and removed alternative 'cover' (previously incorrectly called 'fill') since it will not be supported.
* updated stories so that they all have a displayOption knob and also added hotspots in each corner of the image.
* removed hotspotsStyles from the individual cachedHotspots since it didn't make any sense and had no effect
* fixed a story had accidentally flipped height and width properties which caused problems
* changed zooming condition so that it is possible to zoom if height OR width is smaller than original instead of only allowing zooming when both were smaller.
* prevented hotspot to be added if click was on image element but not on "visual" image, i.e. if click was on the "objectFitOffset" area
* made sure the cursor correctly shows crosshair only when on actual visual image and not on "objectFitOffset" area
* added a lot of tests to cover the improved positioning and sizing

ImageCardFormSettings:
* removed Image display option labelled "fill" which was translating to objectFit 'cover'. See discussion under issue.

1. Go to https://deploy-preview-2461--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-imagecard--width-display-options
2. Try all combinations of displayOption, size and zoom and make sure the image and the hotspots are correctly positioned
3. Go to https://deploy-preview-2461--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-imagehotspots--landscape-image-landscape-container
4. Try all the stories with different displayOptions and zoom and make sure the image and the hotspots are correctly positioned
5. Go to https://deploy-preview-2461--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-dashboardeditor--default
6. Create a new image card
7. Select an image from the gallery
8. Click Edit Image and add some hotspots
9. Save and verify that the hotspots are in the correct location
10. Resize the card and verify that hotspots are are in the correct location
11. Repeat from step 5 but change the "Image display options" under the Settings tab before adding the hotspots.
12. https://deploy-preview-2461--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-hotspoteditormodal--empty and add some hotspots
13. Verify that the hotspots appear in the expected position
14. Repeat from step 12 but change the displayOption to test both "fill" and "contain"
15. Try the remaining stories of the HotspotEditorModal and test with different displayOptions. Verify that the hotspots are correctly positioned.

Verify that the stories for ImageCard, ImageHotspots, HotspotEditorModal using displayOption `undefined` produce the same result as they do in branch "next" (except for the additional hotspots added)


kevinsperrine (changes requested):
When adding new hotspots on the dashboard editor story (Part 5 of your acceptance instructions) the new hotspots are adding far to the left of where I'm clicking.

hotspot-locations

davidicus:
I think the failed test are just snapshots

davidicus:
> When adding new hotspots on the dashboard editor story (Part 5 of your acceptance instructions) the new hotspots are adding far to the left of where I'm clicking.
> 
> hotspot-locations

For more info on this when you have the setting to stretch the hotspot is actually rendered where you click. So the error @kevinsperrine mentions is for default or fill values for object-fit

davidicus:
The rest of the acceptance test seems to work as intended, so just that one bug. Great work.

bjornalm:
> When adding new hotspots on the dashboard editor story (Part 5 of your acceptance instructions) the new hotspots are adding far to the left of where I'm clicking.
> 
>

Good catch! I always had my chrome developer tools open, which caused the landscape image in the landscape container to always have vertical offset, instead of the horizontal offset needed for this bug to show itself.

kevinsperrine (approved):
Looks great now. Really solid work on a tricky problem.

davidicus (approved):

PR #2464, opened 2021-07-02
Add browser based testing -cypress

### LINKED ISSUE #1203: [Repo] provide e2e testing framework

davidicus:
Closes #1203 

**Summary**

This PR adds the cypress testing framework to enable browser based  and visual regression testing

Additional issue will need to be opened to complete the writing of visual snapshot testing. 

- Added the cypress framework and some configuration files under the `cypress` directory and the cypress.json file
- added the `cypress-image-diff-js` plugin to allow for visual regression testing. 
- Visual test proof of concept have been written for the `Button` and `CardCodeEditor` components as well as some browser testing that was needed for `CardCodeEditor`. Cypress test will be collocated and have the `.test.e2e.jsx` extension
- added to our `test.yml` file in the github workflows directory to add cypress test runner on PR checks
- added plugins to our `eslintrc` file. 
- ignored cypress test in our `jest.config.js`
- added 8 new scripts to `package.json`
  - `test:e2e:update`: updates base images
   - `test:e2e`: runs cypress test on pre-commit
   - `test:e2e:interactive`:  runs cypress test in a live browser
   - `test:e2e:ci`: runs cypress in CI which records the result to our cypress dashboard account
   - `reports:clean`: cleans up temp folders needed to combine reports. 
    - `reports:copy`: copies both jest and cypress coverage reports to single directory where they can be merged
    - `reports:combine`: Combines coverage reports into one and moves them to our coverage folder
    - `reports`:  runs copy and combine 
- added `webpack-dev-server` in order for cypress to compile and build for testing
- added a `nyc` config to the `package.json`. This is used to combine coverage between jest and cypress. 

test should all pass. If you pull down locally you can run `yarn test:e2e` and all the cypress test should pass. Snapshots will only be ran in ci in order to avoid visual testing failures due to different environments.


.github/workflows/test.yml
```diff
@@ -54,6 +54,41 @@ jobs:
...
+          yarn --frozen-lockfile
+          yarn lerna run --stream postinstall
+          yarn lerna link
+
+      - name: Run tests, collect coverage
+        if: steps.changes.outputs.react == 'true'
+        run: |
+          cd packages/react
+          yarn test:e2e:ci
+        with:
```
kevinsperrine, line 85: The `with:` here is throwing an error and the tests aren't running in CI. https://github.com/carbon-design-system/carbon-addons-iot-react/actions/runs/994948401/workflow

packages/react/cypress.json
```diff
@@ -0,0 +1,19 @@
+{
+  "projectId": "yxfokk",
+  "componentFolder": "src",
+  "env": {
+    "coverage": false,
```
kevinsperrine, line 5: Do we want to open another issue to actually collect coverage and merge it with jest similar to this https://github.com/bahmutov/cypress-and-jest?

packages/react/.eslintrc
```diff
@@ -104,14 +97,12 @@
       "files": [
         "**/*-fixture.{js,jsx}",
         "**/*.test.{js,jsx}",
+        "**/*.test.e2e.{js,jsx}",
```
kevinsperrine, line 100: The overrides in eslintrc takes an array of objects, so you can split off cypress specific file overrides into their own object, too. That'll get rid of needing to add comments to turn off specific rules.

```js
// in .estlintrc overrides
"overrides": [
    {
      // ...
    },
    {
      "files": ["**/*.test.e2e.{js,jsx}"],
      "rules": {
        "jest/expect-expect": "off",
        "testing-library/prefer-screen-queries": "off"
      }
    }
```

packages/react/cypress.json
davidicus: yes I have it as a to do to open this issue.

packages/react/.eslintrc
davidicus: Thanks this was another to do to look into. I will make this change now.

packages/react/cypress.json
davidicus: added to this PR

packages/react/src/components/CardCodeEditor/CardCodeEditor.test.e2e.jsx
```diff
@@ -0,0 +1,69 @@
...
+        onClose={() => {}}
+        language="json"
+        onCopy={() => {}}
+        initialValue="/* write your code here */"
+      />
+    );
+    // wait for script to load. fails intermittently without
+    /* eslint-disable-next-line cypress/no-unnecessary-waiting */
+    cy.wait(200);
+    cy.findByText(/\/\* write your code here \*\//).should('be.visible');
```
bjornalm, line 62: This is strange though right? From what I read the "find" queries in Cypress Testing Library should support cypress "retry-ability" and it would retry until the assertion was true. It should even support setting a retry timeout e.g.

`cy.findLabelText(/Label text/i, { timeout: 1000 }).should('be.visible')`

packages/react/cypress/support/commands.js
```diff
@@ -0,0 +1,3 @@
+const compareSnapshotCommand = require('cypress-image-diff-js/dist/command');
+
```
bjornalm, line 2: Shouldn't `import '@testing-library/cypress/add-commands';` be in here? 

Perhaps it doesn't matter since we import this file to cypress/support/index.js anyway...

https://testing-library.com/docs/cypress-testing-library/intro/

bjornalm (approved):
I don't have much expertise to add on the whole configuration but it looks solid and it runs. Just a couple of questions

packages/react/package.json
```diff
@@ -271,14 +325,14 @@
...
     "sass": "^1.34.0",
-    "sass-loader": "^7.1.0",
     "storybook-addon-rtl": "^0.2.2",
     "style-loader": "^0.23.1",
     "stylelint": "^10.1.0",
     "stylelint-config-recommended-scss": "^3.3.0",
     "stylelint-scss": "^3.10.0",
     "weak-napi": "^2.0.1",
-    "webpack": "^4.28.4",
+    "webpack": "^4.41.5",
```
kevinsperrine, line 334: I get some additional changes in the yarn.lock file when I pulled it down locally and ran `yarn install` might want to confirm everything is in there.

packages/react/package.json
```diff
@@ -34,7 +34,7 @@
     "lint": "yarn lint:javascript && yarn lint:stylelint",
     "lint:javascript": "eslint --ext .jsx --ext .js .",
     "lint:stylelint": "stylelint './src/**/*.scss' --syntax scss --ignorePath .gitignore --custom-formatter ./config/stylelint/sassMsgFormatter.js",
-    "pre-commit": "yarn svg-convert && yarn lint-staged && yarn test:base",
+    "pre-commit": "yarn svg-convert && yarn lint-staged && yarn reports:clean && yarn test:base && yarn test:e2e && yarn reports",
```
kevinsperrine, line 37: whew. our pre-commits are going to take _a lot_ of time as the e2e test coverage grows...

kevinsperrine (approved):
This is great! This will be good for everyone as we build out the coverage going into v3.

packages/react/package.json
```diff
@@ -119,6 +127,41 @@
...
+      "src/**/hooks/*.js",
+      "src/components/SuiteHeader/util/suiteHeaderData.js",
+      "src/components/SuiteHeader/util/uiresources.js",
+      "src/components/FileUploader/stories/*.jsx",
+      "src/components/Table/AsyncTable/*.js?(x)",
+      "src/components/WizardInline/**/*.js?(x)",
+      "src/components/Page/(EditPage|PageHero).jsx",
+      "src/components/Dashboard/(Dashboard|CardRenderer).jsx",
+      "src/components/MapCard/storyFiles/*.jsx",
+      "src/components/BarChartCard/barChartUtils.js",
```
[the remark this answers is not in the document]
davidicus: This may need to be revisited. I had originally started with just copying the items from Jest config but I was noticing threshold failures for these other files. I think as part of our effort to bring the codebase completely up to 90% we can look at these exceptions and either verify that they should be ignored or write test to bring the threshold up properly

packages/react/package.json
```diff
@@ -119,6 +127,41 @@
...
+      "src/components/ComboChartCard/comboChartHelpers.js",
+      "src/components/DashboardEditor/DashboardEditor.jsx",
+      "src/components/Table/Table.test.helpers.js",
+      "src/components/Table/TablePropTypes.js",
+      "src/components/Table/baseTableReducer.js",
+      "src/components/Table/tableReducer.js"
+    ],
+    "branches": 80,
```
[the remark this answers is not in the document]
davidicus: So NYC is the threshold that will run on our coveralls reporting. NYC is the tool that takes coverage from both and combines them into a single report. I had thought about removing the jest config globals but was unsure if we should leave them in. Since most unit test will be done with Jest it may be helpful to have your run fail when your jest test are not above the threshold  so you do not have to run the combine script every time. 

So moving forward you would write your code. Write your test. If you have maxed out the code that can be covered in Jest you can then write an exception for the jest threshold and move to writing your cypress test. Not sure if this the best flow. Definitely open to hearing your thoughts on the dev experience and what your preference would be.

packages/react/src/components/CardCodeEditor/CardCodeEditor.test.e2e.jsx
davidicus: I have tried this and other solutions. Nothing would prevent the intermittent failure that happens when the script takes too long to load. This probably is not fool proof either but its has prevented me from seeing this error for every run I have done since. I have not tried setting it to a super long time so maybe that would work but seems just as flaky as it could still lose the race.

davidicus: I also did not want to hold up the delivery of the framework being bogged down with this specific components test. These were meant to be proof of concept and this component took me a long time because of this weird error. I had seen some ways to spy the script execution and I played around with it a bit but ultimately it was not straight forward and I decided to cut my losses to get the framework in

.github/workflows/test.yml
```diff
@@ -40,12 +40,28 @@ jobs:
...
           cd packages/react
           yarn test:ci
 
+      - name: Run Cypress tests, collect coverage
+        if: steps.changes.outputs.react == 'true'
+        run: |
+          cd packages/react
+          yarn test:e2e:ci
+        env:
+          CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
```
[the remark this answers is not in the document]
davidicus: I am trying to find that out. @scottdickerson do you know if its possible to get multiple access to our repo on the dashboard?

packages/react/package.json
```diff
@@ -49,7 +49,15 @@
     "test:story": "yarn test:engines && cross-env TZ=America/Chicago jest src/components/StorybookSnapshots.test.js --no-watchman --colors",
     "test:clear": "yarn test:engines && cross-env TZ=America/Chicago jest --no-watchman --clearCache --colors",
     "test:engines": "check-node-version --node $(cat ../../.nvmrc) --yarn 1.x",
-    "test:ci": "yarn test --ci --logHeapUsage --coverage"
+    "test:ci": "yarn test --ci --logHeapUsage --coverage",
+    "test:e2e:update": "cypress-image-diff -u",
+    "test:e2e": "cross-env BABEL_ENV='e2e' NODE_ENV='e2e' cypress run-ct --browser chrome --config video=false,screenshotOnRunFailure=false",
+    "test:e2e:interactive": "cross-env BABEL_ENV='e2e' NODE_ENV='e2e' cypress open-ct --browser chrome --config video=false,screenshotOnRunFailure=false",
+    "test:e2e:ci": "cross-env BABEL_ENV='e2e' NODE_ENV='e2e' cypress run-ct  --browser chrome --headless --config --record video=false,screenshotOnRunFailure=false",
+    "reports:clean": "rimraf .nyc_output reports coverage",
```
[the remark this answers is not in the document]
davidicus: That may be a good idea. We were not cleaning our jest coverage folder before. I may want to leave off the coverage directory actually. This script will run on pre-commit. I can see devs wanting to commit some code and still see the coverage report to keep working without having to run all the test again. I had to include the `.nyc_output` because a bug in `nyc` will cause an error if the folder if present before test are ran. I also think reports can be deleted since it's also a temp file.

So I will remove coverage from this script and just have the temp folders get cleaned up. cool?

packages/react/package.json
davidicus: Yes, I was thinking this. I am not sure if it's more annoying to wait for test to pass before pushing up your code or waiting on the actions to run which  may be in a queue and then having to push up more code if there are failures. Thoughts?

kevinsperrine: It's definitely a double-edged sword. There are definitely times I don't break changes up into small commits because the pre-commit takes so long. This is primarily because I view `--no-verify` as dirty. However, since we're limited on the number of actions that can run at a time, I prefer the more immediate feedback. Maybe it would be nice to move these to a `pre-push` hook, so that local dev is faster, but ensures everything is clean before it's pushed up?

packages/react/package.json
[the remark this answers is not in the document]
kevinsperrine: Hmmm...This is a tricky one. Maybe we remove individual thresholds from jest/cypress and then add another script that checks the total coverage from the merged version similar to the `coverage:check` script in [Part 3 of this article](https://dev.to/muratkeremozcan/combined-unit-e2e-code-coverage-case-study-on-a-real-life-system-using-angular-jest-cypress-gitlab-35nk).

packages/react/package.json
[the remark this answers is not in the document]
kevinsperrine: Sounds good to me.

packages/react/package.json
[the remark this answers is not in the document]
davidicus: That is what we are doing already aside from not having jest thresholds removed. Our combine script has the flags `--check-coverage --per-file`. Which will check against the threshold numbers after it combines the reports. 

The issue is that if you are not writing and cypress test and you do not have the coverage generated for them you will have to run the cypress  test just to get the combined report. Not ideal. With jest thresholds in place they can at least be checked for the majority of cases without having to run any additional code.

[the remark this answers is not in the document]
kevinsperrine: Oh, I'm following now. If we move these checks into a `pre-push` hook instead of a `pre-commit` hook like I [mention below](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2464#discussion_r670620018) then I don't see much of an issue having to run both. It gets more out of the way of general development and commits, but forces the coverage before pushes, so you run everything once when attempting to push and then get the feedback to make whatever changes. I think that improves the DX pretty well over the current setup.

packages/react/cypress/support/commands.js
davidicus: Definitely. moved the import here

packages/react/src/components/CardCodeEditor/CardCodeEditor.test.e2e.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+  it('should call the onCopy function when icon is clicked', () => {
+    onCopy = cy.stub();
+    mount(
+      <CardCodeEditor
+        open
+        onSubmit={() => {}}
+        onClose={() => {}}
+        language="json"
+        onCopy={(value) => onCopy(value)}
```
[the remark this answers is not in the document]
davidicus: yes. changed

packages/react/package.json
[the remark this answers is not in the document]
davidicus: Setting testing on a pre-push hook.

.github/workflows/test.yml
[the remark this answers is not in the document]
davidicus: sending you and Bjorn an invite

PR #2467, opened 2021-07-02
fix(datepicker): relative dates while changing the hours minutes throws an exception

scottdickerson:
Closes #2467 

**Summary**

- DateTimePicker was incorrectly parsing the endDate in the case where the hours were changing on the relative date

- fix(DateTimePicker): parse date. The parse date method was parsing too early (before all chars were typed) and using the wrong method to set the hours and minutes of the end date

- Verify that all combinations of using relative dates work correctly

- All the existing relative testcases pass


bjornalm (approved):
LGTM

davidicus (approved):

PR #2475, opened 2021-07-06
chore(repo): deprecate testID and add testId - 2

### LINKED ISSUE #2458: [Repo] Add `data-testid` attribute and `testId` props - 2

- [x] Checkbox
- [x] CodeSnippet
- [x] ColorDropdown
- [x] ComboBox
- [x] ComboChartCard
- [x] ComposedModal
- [x] ComposedStructuredList
- [x] ContentSwitcher

kevinsperrine:
Closes #2458 

Depends on #2457

**Summary**

- Checkbox
- CodeSnippet
- ColorDropdown
- ComboBox
- ComboChartCard
- ComposedModal
- ComposedStructuredList
- ContentSwitcher


packages/react/src/components/ColorDropdown/ColorDropdown.jsx
```diff
@@ -47,8 +48,14 @@ const propTypes = {
   translateWithId: PropTypes.func,
   /** The selected color, use to set initial color */
   selectedColor: colorPropType,
+  // TODO: remove deprecated testID in v3.
+  // eslint-disable-next-line react/require-default-props
+  testID: deprecate(
```
davidicus, line 53: I believe this component is still experimental so we can probably swap out ID without deprecate.

davidicus (approved):
Seems good to go. If you want to wait to remove the deprecation at this point since its already there I am fine with that

kevinsperrine:
> Seems good to go. If you want to wait to remove the deprecation at this point since its already there I am fine with that

Might as well just leave it now. We can take it out with everything else in V3.

bjornalm (approved):
LGTM

PR #2489, opened 2021-07-07
fix(table): wrap the column selection text

### LINKED ISSUE #2482: [Table] Manage Column link no longer visible on screen when resolution set to actual size
Health: https://github.ibm.com/maximo/Asset-Health-Insight/issues/5994

Describe the bug
Manage Column link blows out of the table and causes horizontal scrolling earlier than it shgould

In our reference app, as early as 1100px, the "Manage columns" button begins to cause a horizontal scroll.

In the IoT storybook, it is less noticeable, but only because there are not as many columns:
https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--with-column-selection

From health:
To Reproduce
Log in to Health application
On Assets grid view click on the Manage Columns icon on the top right of page

ISSUE: The Manage Column blue link is not visible on the page
Workaround: Scroll down to the bottom of the page and see there is now a horizontal scroll bar which you can move all the way to the right and when you scroll back up the link is visible... or change the zoom level from 100% to 90% to see the link.

NOTE : This is new behavior in a recent build as I have not changed anything regarding my browser zoom level. Both the Firefox View-Zoom option is set to Actual Size and hamburger menu set to 100%

Expected behavior
Manage Column link should be visible as in previous builds OR be reactive such that when in this scenario it moves to be next to the Total Cost column button or whatever is the last column in that view.

Screenshots

Refapp
iot

kalyanixraut:
Closes #2482 

**Summary**

- Improved the column selection button positioning. Column selection button appears at the end and is not visible on the page when table has many columns until its scrolled and hence its usability is missed.

- added flex-wrap to the table-header-label so that column selection button automatically wraps and is visible even when there are many columns.

- Can be tested in storybook https://deploy-preview-2489--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-resize-and-no-initial-columns

### DIFF
--- a/packages/react/src/components/Table/TableHead/_table-head.scss
+++ b/packages/react/src/components/Table/TableHead/_table-head.scss
@@ -20,7 +20,6 @@
     display: flex;
     user-select: none;
     overflow: hidden;
-    flex-wrap: wrap;
     // allow the multiselect filter to expand to full height and the table to scroll
     .#{$prefix}--multi-select .#{$prefix}--list-box__menu {
       max-height: unset;
--- a/packages/react/src/components/Table/TableHead/_table-head.scss
+++ b/packages/react/src/components/Table/TableHead/_table-head.scss
@@ -20,6 +20,7 @@
     display: flex;
     user-select: none;
     overflow: hidden;
+    flex-wrap: wrap;
     // allow the multiselect filter to expand to full height and the table to scroll
     .#{$prefix}--multi-select .#{$prefix}--list-box__menu {
       max-height: unset;
--- a/packages/react/src/components/Table/TableHead/ColumnHeaderRow/_column-header-row.scss
+++ b/packages/react/src/components/Table/TableHead/ColumnHeaderRow/_column-header-row.scss
@@ -15,6 +15,10 @@
       min-width: 12.75rem;
     }
   }
+
+  .#{$prefix}--table-header-label {
+    flex-wrap: wrap;
+  }
 }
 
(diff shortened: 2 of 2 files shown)


packages/react/src/components/Table/TableHead/_table-head.scss
```diff
@@ -20,6 +20,7 @@
     display: flex;
     user-select: none;
     overflow: hidden;
+    flex-wrap: wrap;
```
bjornalm, line 23: This class is also used by the span elements inside the normal table columns

image


So we should make sure it is a child of the `iot--column-header-row--table-header` selector and added to the file `_column-header-row.scss` instead, e.g.

```
.#{$iot-prefix}--column-header-row--table-header .#{$prefix}--table-header-label {
  flex-wrap: wrap;
}
```

bjornalm (changes requested):
I think that you correctly identified the actual problem, but need to be more specific with the css fix. See comment.

packages/react/src/components/Table/TableHead/_table-head.scss
kalyanixraut: Thanks @bjornalm, for reviewing the PR. I have made the changes as suggested.
https://deploy-preview-2489--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-resize-and-no-initial-columns

bjornalm (approved):

kevinsperrine (approved):

davidicus (approved):

PR #2491, opened 2021-07-07
fix(header): clicking directly on the icon failed to open overflowmenu

### LINKED ISSUE #2488: [Header] Collapsed menu difficult to click
### What package is this for?

- [X] React

### Describe the bug

When the menu icons condense into an overflow menu, the three dot icon cannot be clicked in the middle. You must click on the outside edge. 

### To Reproduce

1. Go to 'https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-suiteheader--header-with-custom-side-nav'
2. Make screen small enough to show the three-dot menu
3. Click in the middle of the menu
4. See that the menu does not open

kevinsperrine:
Closes #2488 

**Summary**

- When clicking directly on the icon the overflow menu would fail to open. This is because the OverflowMenu component triggers the onClick event twice when clicking directly on the icon AND changing the icon based on open/closed state. I tried manually controlling state via the open prop, I tried calling stopPropagation in the onClick of the svg icons, but neither of those solutions work, because the stopPropagation breaks the OverflowMenu from opening. My assumption is the event is triggered twice b/c of the swapping of the icon and that is somehow triggering a re-render that closes the OverflowMenu or the click triggers the outsideClickHandler that closes the menu on outside clicks. 

- added a work-around to ignore the second click and read the OverflowMenu state directly from the `aria-expanded` attribute and set the icon state based on that.

- The `Header` `small breakpoint` story should now open/close the overflow even when clicking directly on the svg icon or on the edges of the button.

- No other changes to the Header OverflowMenu.


bjornalm (approved):
Clever hack, is the root cause in carbon?

kevinsperrine:
> Clever hack, is the root cause in carbon?

I believe so, yes. Eventually, I was going to create a reproduction example and open a bug. I spent hours trying every-which-way I could think of within our code to make it work when swapping the icons, but nothing worked. I think The changing of the icon is surfacing a bug in Carbon.

davidicus (approved):

PR #2494, opened 2021-07-08
feat(datetimepicker): add hasicononly prop

### LINKED ISSUE #2492: [DateTimePicker] support icon only button
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Currently the DateTimePicker only supports the full width dropdown. We have need of the icon only version in cases where its being displayed in an area that is too skinny such as the Card toolbar in smaller card cases

**Design Spec** 

image

sstone2423:
Closes #2492 

**Summary**

- Adds `hasIconOnly` prop to DateTimePicker to optionally only show the icon rather than the selected time range text AND the icon
- Unfortunately this required a fairly large rewrite to use FlyoutMenu as the underlying menu pop out as the current pop out menu was very opinionated and static. It was very difficult to make changes and at the end of the day, FlyoutMenu handles a lot of functionality such as opening the menu in different directions depending on the visibility

- Added Icon Only story
- Remove div card wrappers for all of the stories as they have no purpose
- Memoize the i18n as we do in other components
- Use FlyoutMenu instead of the custom `expanded` box
- Remove a lot of useless style due to the refactor
- Rewrote most of the tests in react-testing-library and fixed multiple tests that were never working properly (not sure how these ever passed)
- Removed all jest fake timers as they had no effect
- Export `PICKER_KINDS` as we needed them in Monitor and were duplicating the code

- Go to /?path=/story/1-watson-iot-datetimepicker--icon-only and verify that only the icon appears on first render (check design spec in the linked issue)
- Click the menu and ensure the menu still looks and acts exactly the same as the regular DateTimePicker. This should have NO functional change to the DateTimePicker and should be purely visible
- Use the RTL knob to ensure everything adjusts correctly
- Check that all other stories were unaffected. Also try the `hasIconOnly` knob throughout the stories to ensure everything works as intended

- This should have NO functional change to the DateTimePicker and should be purely visible


packages/react/src/components/DateTimePicker/_date-time-picker.scss
```diff
@@ -121,70 +125,102 @@
       &.#{$iot-prefix}--date-time-picker__menu-expanded {
         display: block;
       }
+    }
+  }
 
-      .#{$iot-prefix}--date-time-picker__menu-scroll {
```
sstone2423, line 125: This diff looks much worse than it actually is. I split the `.#{$iot-prefix}--date-time-picker__box` into 2 separate classes `.#{$iot-prefix}--date-time-picker__box--icon-only` and `.#{$iot-prefix}--date-time-picker__box--full` for the unique differences. I then took the common nested classes and moved them outside of the scope of the boxes to avoid duplication

packages/react/src/components/DateTimePicker/DateTimePicker.story.jsx
```diff
@@ -46,193 +42,156 @@ export default {
...
+      id="datetimepicker"
+      dateTimeMask={text('dateTimeMask', 'YYYY-MM-DD HH:mm')}
+      relatives={[
+        {
+          label: 'Yesterday',
+          value: RELATIVE_VALUES.YESTERDAY,
+        },
+      ]}
+      hasTimeInput={boolean('hasTimeInput', true)}
+      hasIconOnly={boolean('hasIconOnly', true)}
```
kevinsperrine, line 179: Can we add the onApply/onCancel storybook action here? It's slightly confusing when setting an absolute time range to click "Apply" and the window immediately close. With the iconOnly version we don't have that visual feedback to see the correct range was selected. I can confirm it does by turning `hasIconOnly` off for a second to see the range, but still...

kevinsperrine (changes requested):
Overall this looks really good. Just a few minor visual/RTL issues. 

Not a result of your changes, b/c it exists in next, too, but if `hasTimeInput` is set to false the down chevron is misplaced. It should be near the right-edge of "Yesterday" not all the way to the right.

Screen Shot 2021-07-08 at 4 42 32 PM

When in RTL mode the "to" is out of place. ie. `to Now {DATE}` instead of `Now to {DATE}`

Screen Shot 2021-07-08 at 4 49 53 PM

When in RTL mode the and setting a relative date the number in the `Last` input is out of place. It's by the `-` instead of on the left.

Screen Shot 2021-07-08 at 4 49 35 PM

packages/react/src/components/DateTimePicker/DateTimePicker.story.jsx
sstone2423: Added

sstone2423:
> Not a result of your changes, b/c it exists in next, too, but if `hasTimeInput` is set to false the down chevron is misplaced. It should be near the right-edge of "Yesterday" not all the way to the right.

This is fixed now

> When in RTL mode the "to" is out of place. ie. `to Now {DATE}` instead of `Now to {DATE}`
>
> When in RTL mode the and setting a relative date the number in the `Last` input is out of place. It's by the `-` instead of on the left.

Opened issue for these here: https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2615

scottdickerson (approved):
LGTM

davidicus (changes requested):
Some visual differences between next and this PR
image
image

image
image

image
image


Time not being populated when going from relative where you set a time frame and then switch to absolute
image
image

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -156,86 +153,101 @@ describe('DateTimePicker', () => {
...
+    // expect 20 day range relative to yesterday at 15:30
+    expect(
+      screen.getByText(
+        `${from.format('YYYY-MM-DD')} 15:30 to ${yesterday.format('YYYY-MM-DD')} 15:30`
+      )
+    ).toBeVisible();
+
+    userEvent.click(screen.getByText('Apply'));
     expect(dateTimePickerProps.onApply).toHaveBeenCalled();
+    // calling twice for some reason
```
davidicus, line 203: are comments still needed

packages/react/src/components/DateTimePicker/_date-time-picker.scss
```diff
@@ -109,112 +99,88 @@
...
+  .#{$iot-prefix}--date-time-picker__box--light {
+    background-color: $field-02;
+  }
 
-        .#{$prefix}--fieldset {
-          margin-bottom: $spacing-lg;
-          padding: 0 $spacing-baseline;
+  .#{$iot-prefix}--date-time-picker__menu-scroll {
+    overflow-y: auto;
+    padding-top: 0.75rem;
```
davidicus, line 111: should use token.

packages/react/src/components/DateTimePicker/_date-time-picker.scss
```diff
@@ -37,17 +28,36 @@
     animation: none;
   }
 
+  // move up the tooltip
+  .bx--tooltip[data-floating-menu-direction='bottom'] {
+    margin-top: 0rem;
```
davidicus, line 33: Units not needed here?

packages/react/src/components/DateTimePicker/_date-time-picker.scss
```diff
@@ -109,112 +99,88 @@
...
 
-        .#{$iot-prefix}--date-time-picker__fields-wrapper {
-          display: flex;
-          align-items: flex-end;
-          .#{$prefix}--label {
-            margin-bottom: 0;
-          }
-        }
+    .#{$iot-prefix}--date-time-picker__custom-wrapper {
+      padding-top: 1rem;
```
davidicus, line 123: Tokens

packages/react/src/components/DateTimePicker/_date-time-picker.scss
```diff
@@ -109,112 +99,88 @@
...
+      }
 
-        .#{$iot-prefix}--date-time-picker__listitem {
-          padding: $spacing-04 $spacing-baseline;
-          cursor: pointer;
-          border-left: 4px solid transparent;
-          border-bottom: 1px solid $ui-03;
-        }
+      .#{$prefix}--form-item + .#{$prefix}--form-item {
+        margin-left: 1rem;
```
davidicus, line 136: tokens

scottdickerson:
> Some visual differences between next and this PR
> image
> image
> 
> image
> image
> 
> image
> image
> 
> Time not being populated when going from relative where you set a time frame and then switch to absolute
> image
> image

Thanks for catching all these @davidicus .  I will fix all of them, except this one, which I think is better than the original UX
image

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
scottdickerson: fixed

davidicus:
Deviations from existing
image

- [ ] Clicking anywhere on the time range input box should open the dropdown. Not just the icon. 
- [ ] Cursor should be pointer when hovering entire input. not just icon.
- [ ] When hovered the entire input box should be gray along with icon
- [ ] This input was 41px high and now is 49. 
image
- [ ] Box should span entire length of input
- [ ] There is extra padding above `2021-07-27 11:20 to Now`
image
- [ ] There is extra padding above custom range
image
- [ ] The icon only story renders offscreen
- [ ] `.iot--date-time-picker__box--full` the `--full` modifier was added to existing class but there is no base `.iot--date-time-picker__box` class. This unnecessarily changes the class name which could impact downstream consumers.
Current changes:
image
Next:
image
- [ ] If you try tabbing in the default story. The tab stops have changed. The first one's focus is mostly covered. 
- [ ] when you hit enter after input button is tabbed it places focus on `cancel` button where next maintains the current focus. Not sure which is right but different than next. 
- [ ] There seems to be a lot of internal markup changes as well which will break downstream consumers like Graphite's test.

For the comment about the time not showing up if you go to absolute you have to be on next.  Go to custom range on relative, increment  the "last" field by one minute and then change "relative to" to 1:00. Then click absolute. You will see the time end from the time set in relative. It looks like we have a bug where the from is not being populated but they both should respect the time changes made from the last.

davidicus:
This seems to be a lot of changes to just have the dropdown trigger be only the button.  I am able to get the below result with a few changes of CSS. If we conditionally render the text based off the icon only prop we should be able to do this with minimal changes to the markup. 
image

This component had a lot of weird edge case bugs that have been found and fixed and I am hesitant to change it this drastically. I think we should probably revisit our approach and attempt a mainly css solution. Thoughts?

packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
```diff
@@ -0,0 +1,988 @@
...
+  locale,
+  id = uuid.v4(),
+  hasIconOnly,
+  ...others
+}) => {
+  React.useEffect(() => {
+    if (__DEV__) {
+      warning(
+        false,
+        'The `ColorDropdown` is an experimental component and could be lacking unit test and documentation. Be aware that minor version bumps could introduce breaking changes. For the reasons listed above use of this component in production is highly discouraged'
```
davidicus, line 255: Did we mean to leave this text in the warning here?

packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
```diff
@@ -0,0 +1,1017 @@
...
+      parseDefaultValue(lastAppliedValue);
+      setIsOpen(false);
+
+      if (onCancel) {
+        onCancel();
+      }
+    };
+
+    return (
+      <div className={`${iotPrefix}--date-time-picker__menu-btn-set`}>
```
davidicus, line 678: This should have a data-testid?

packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
```diff
@@ -0,0 +1,1017 @@
...
+                  >
+                    {humanValue}
+                  </TooltipDefinition>
+                </>
+              ) : null}
+            </div>
+          ) : null}
+
+          <FlyoutMenu
+            isOpen={isFlyoutOpen}
```
davidicus, line 762: can we add a data-testid variant here?

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
```diff
@@ -0,0 +1,507 @@
...
+    jest.spyOn(console, 'error').mockImplementation(() => {});
+  });
+
+  afterAll(() => {
+    console.error.mockRestore();
+  });
+
+  it('should not blow up if correct object is passed as default value', () => {
+    mount(
+      <DateTimePicker
```
davidicus, line 52: Can we add a test for the testids themselves? As part of the new testing contract we are ensuring that these selectors will always be available.

packages/react/src/components/DateTimePicker/_date-time-pickerv2.scss
```diff
@@ -0,0 +1,281 @@
...
+      padding: 0 $spacing-baseline;
+
+      &:last-child {
+        margin-bottom: 0;
+      }
+    }
+
+    .#{$iot-prefix}--date-time-picker__custom-wrapper {
+      width: var(--wrapper-width);
+      padding-top: 1rem;
```
davidicus, line 107: needs tokens

packages/react/src/components/DateTimePicker/_date-time-pickerv2.scss
```diff
@@ -0,0 +1,281 @@
...
+  .#{$iot-prefix}--date-time-picker__fields-wrapper {
+    display: flex;
+    align-items: flex-end;
+
+    .#{$prefix}--label {
+      margin-bottom: 0;
+    }
+
+    .#{$prefix}--form-item + .#{$prefix}--form-item {
+      margin-left: 1rem;
```
davidicus, line 176: Need tokens

packages/react/src/components/DateTimePicker/_date-time-pickerv2.scss
```diff
@@ -0,0 +1,281 @@
...
+    }
+  }
+
+  .#{$iot-prefix}--date-time-picker__no-formgroup {
+    margin-bottom: $spacing-05;
+  }
+
+  &.#{$iot-prefix}--date-time-picker__menu-formgroup {
+    display: block;
+    padding-top: 1rem;
```
davidicus, line 186: tokens here

davidicus:
Remaining visual thing would be the extra padding on top of the date here
image

and there are some issue in RTL
image
image

packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
davidicus: I think this should be DateTimePickerV2 instead of ColorDropdown. Does the DateTimePickerV2 need to be experimental?

packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
scottdickerson: sure I'll pass that through

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
scottdickerson: sure, I'm adding for the flyout (other testids are covered it seems under the other testcases

packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
scottdickerson: fixed

davidicus (approved):

PR #2499, opened 2021-07-08
chore(repo): deprecate testID and add testId - 8

kevinsperrine:
Closes #2485  

Depends on #2462 #2475 #2478 #2480 #2481 #2490 #2493

**Summary**

Adds testId prop to element that don't have it, and marks the existing testID prop as deprecated. It also uses the passed testID or testId to create usable test ids for nested components and make them easier to test.

**THIS PR SHOULD NOT CONTAIN ANY BREAKING CHANGES, SO THAT IT CAN BE ROLLED OUT BEFORE V3. IT NEEDS A GOOD EYE TO CONFIRM THAT WITH SO MANY CHANGES.**

Change List (commits, features, bugs, etc)

- PieChartCard
- PrimaryButton
- ProgressIndicator
- RadioButton
- RadioButtonGroup
- RadioTile
- ResourceList
- RuleBuilder
- Search
- SearchFilterButton
- SearchLayoutButton
- SecondaryButton
- Select
- SelectItem
- SelectItemGroup

**Acceptance Test (how to verify the PR)**

- storyshots should now contain more data-testid attributes in their output
- testId should be used on all props for the components mentioned above and not testID
- testID should throw a deprecation warning

**Regression Test (how to make sure this PR doesn't break old functionality)**

- All tests still pass as expected.
- testID or testId should both be valid to maintain existing tests without breaking.

PR #2493, opened 2021-07-08
chore(repo): deprecate testID and add testId - 7

kevinsperrine:
Closes #2484  

Depends on #2462 #2475 #2478 #2480 #2481 #2490

**Summary**

Adds testId prop to element that don't have it, and marks the existing testID prop as deprecated. It also uses the passed testID or testId to create usable test ids for nested components and make them easier to test.

**THIS PR SHOULD NOT CONTAIN ANY BREAKING CHANGES, SO THAT IT CAN BE ROLLED OUT BEFORE V3. IT NEEDS A GOOD EYE TO CONFIRM THAT WITH SO MANY CHANGES.**

Change List (commits, features, bugs, etc)

- NavigationBar
- Notification
- NumberInput
- OrderedList
- OverflowMenu
- OverflowMenuItem
- Page
- PageTitleBar
- PageWizard
- Pagination
- PaginationNav

**Acceptance Test (how to verify the PR)**

- storyshots should now contain more data-testid attributes in their output
- testId should be used on all props for the components mentioned above and not testID
- testID should throw a deprecation warning

**Regression Test (how to make sure this PR doesn't break old functionality)**

- All tests still pass as expected.
- testID or testId should both be valid to maintain existing tests without breaking.

PR #2481, opened 2021-07-06
chore(repo): deprecate testID and add testId - 5

kevinsperrine:
Closes #2477 

Depends on #2462 #2475 #2478 #2480

**Summary**

Adds testId prop to element that don't have it, and marks the existing testID prop as deprecated. It also uses the passed testID or testId to create usable test ids for nested components and make them easier to test.

**THIS PR SHOULD NOT CONTAIN ANY BREAKING CHANGES, SO THAT IT CAN BE ROLLED OUT BEFORE V3. IT NEEDS A GOOD EYE TO CONFIRM THAT WITH SO MANY CHANGES.**

Change List (commits, features, bugs, etc)

- Header
- Hero
- HotspotEditorModal
- Icon
- IconDropdown
- IconSwitch
- ImageCard
- ImageGalleryModal
- InlineLoading
- Link

**Acceptance Test (how to verify the PR)**

- storyshots should now contain more data-testid attributes in their output
- testId should be used on all props for the components mentioned above and not testID
- testID should throw a deprecation warning

**Regression Test (how to make sure this PR doesn't break old functionality)**

- All tests still pass as expected.
- testID or testId should both be valid to maintain existing tests without breaking.

PR #2462, opened 2021-07-02
chore(repo): align usage of testId instead of testID - 1

kevinsperrine:
Closes #2457 

**Summary**

Adds `testId` prop to element that don't have it, and marks the existing `testID` prop as deprecated. It also uses the passed `testID` or `testId` to create usable test ids for nested components and make them easier to test. 

**THIS PR SHOULD NOT CONTAIN ANY BREAKING CHANGES, SO THAT IT CAN BE ROLLED OUT BEFORE V3. IT NEEDS A GOOD EYE TO CONFIRM THAT WITH SO MANY CHANGES.**

**Change List (commits, features, bugs, etc)**

- Accordion
- AddCard
- BarChartCard
- BreadCrumb
- Button
- Card
- CardCodeEditor
- CardEditor

**Acceptance Test (how to verify the PR)**

- storyshots should now contain more `data-testid` attributes in their output
- `testId` should be used on all props for the components mentioned above and not `testID`
- `testID` should throw a deprecation warning

**Regression Test (how to make sure this PR doesn't break old functionality)**

- All tests still pass as expected. 
- `testID` or `testId` should both be valid to maintain existing tests without breaking.

PR #2501, opened 2021-07-09
chore(repo): deprecate testID and add testId - 9

### LINKED ISSUE #2486: [Repo] Add `data-testid` attribute and `testId` props - 9
Part of #1001 

- [x] SelectUsersModal
- [x] SideNav
- [x] SimpleIconDropdown
- [x] SimplePagination
- [x] Skeleton
- [x] SkeletonPlaceholder
- [x] SkeletonText
- [x] Slider
- [x] StructuredList
- [x] SuiteHeader
- [x] Switch

kevinsperrine:
Closes #2486 

Depends on #2462 #2475 #2478 #2480 #2481 #2490 #2493 #2499

**Summary**

- SelectUsersModal
- SideNav
- SimpleIconDropdown
- SimplePagination
- Skeleton
- SkeletonPlaceholder
- SkeletonText
- Slider
- StructuredList
- SuiteHeader
- Switch
- SelectItemGroup

### DIFF
--- a/packages/react/src/components/RuleBuilder/__snapshots__/RuleBuilder.story.storyshot
+++ b/packages/react/src/components/RuleBuilder/__snapshots__/RuleBuilder.story.storyshot
@@ -1009,7 +1009,6 @@ exports[`Storybook Snapshot tests and console checks Storyshots 2 - Watson IoT E
                           className="bx--data-table bx--data-table--no-border"
                           data-testid="edit-table"
                           id="edit-table"
-                          testID="rule-builder-edit-users-table"
                           title={null}
                         >
                           <thead
@@ -1525,7 +1524,6 @@ exports[`Storybook Snapshot tests and console checks Storyshots 2 - Watson IoT E
                           className="bx--data-table bx--data-table--no-border"
                           data-testid="read-table"
                           id="read-table"
-                          testID="rule-builder-read-users-table"
                           title={null}
                         >
                           <thead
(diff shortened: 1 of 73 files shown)


davidicus (approved):

packages/react/src/components/RuleBuilder/__snapshots__/RuleBuilder.story.storyshot
```diff
@@ -1934,6 +1934,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots 2 - Watson IoT E
       </footer>
       <div
         className="iot--select-users-modal"
+        testId="select-users-modal-container"
```
sstone2423, line 1937: Shouldn't this be `data-testId` ?

packages/react/src/components/SelectUsersModal/SelectUsersModal.jsx
```diff
@@ -234,7 +249,7 @@ const SelectUsersModal = ({ isOpen, onClose, onSubmit, users, initialSelectedUse
   const selectedList = displaySelectedUsersList(selectedUsers);
 
   return (
-    <div className={`${iotPrefix}--select-users-modal`}>
+    <div testId={`${testId}-container`} className={`${iotPrefix}--select-users-modal`}>
```
sstone2423, line 252: I think this needs to be `data-testid`

sstone2423 (changes requested):

sstone2423 (approved):

kevinsperrine:
b/c #2506 merged before this one there are no changes here now, since everything here was also merged into that one.

Screen Shot 2021-07-16 at 1 02 27 PM

PR #2506, opened 2021-07-09
chore(repo): deprecate testID and add testId - 10

kevinsperrine:
Closes #2487 

Depends on #2475 #2478 #2480 #2481 #2490 #2493 #2499 #2501

**Summary**

Adds testId prop to element that don't have it, and marks the existing testID prop as deprecated. It also uses the passed testID or testId to create usable test ids for nested components and make them easier to test.

**THIS PR SHOULD NOT CONTAIN ANY BREAKING CHANGES, SO THAT IT CAN BE ROLLED OUT BEFORE V3. IT NEEDS A GOOD EYE TO CONFIRM THAT WITH SO MANY CHANGES.**

Change List (commits, features, bugs, etc)

- Table
- TableCard
- Tabs
- Tag
- TearSheet
- TextArea
- TextInput
- Tile
- TileCatalog

**Acceptance Test (how to verify the PR)**

- storyshots should now contain more data-testid attributes in their output
- testId should be used on all props for the components mentioned above and not testID
- testID should throw a deprecation warning

**Regression Test (how to make sure this PR doesn't break old functionality)**

- All tests still pass as expected.
- testID or testId should both be valid to maintain existing tests without breaking.

PR #2509, opened 2021-07-10
chore(repo): deprecate testID and add testId - 11

### LINKED ISSUE #2508: [Repo] Add `data-testid` attribute and `testId` props - 11
Part of #1001 

- [x] TileCatalogNew
- [x] TileGallery
- [x] TileGroup
- [x] TimePicker
- [x] TimePickerSelect
- [x] TimePickerSpinner
- [x] TimeSeriesCard
- [x] Toggle
- [x] ToggleSmall
- [x] Toolbar
- [x] ToolbarSearch

kevinsperrine:
Closes #2508 

Depends on #2475 #2478 #2480 #2481 #2490 #2493 #2499 #2501 #2506

**Summary**

- TileCatalogNew
- TileGallery
- TileGroup
- TimePicker
- TimePickerSelect
- TimePickerSpinner
- TimeSeriesCard
- Toggle
- ToggleSmall
- Toolbar
- ToolbarSearch


packages/react/src/components/TileGallery/__snapshots__/TileGallery.story.storyshot
```diff
@@ -13,7 +13,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/T
   >
     <div
       className="page-title-bar"
-      data-testid="page-title-bar"
```
kevinsperrine, line 16: Not a breaking change. Just added in a previous testId PR. https://github.com/carbon-design-system/carbon-addons-iot-react/blame/c6d4413d8d07e9183d81faa41b606a33f903e299/packages/react/src/components/TileGallery/__snapshots__/TileGallery.story.storyshot

sstone2423 (approved):

davidicus (approved):

PR #2511, opened 2021-07-10
chore(repo) deprecate `testID` and add `testId` - 12

### LINKED ISSUE #2510: [Repo] Add `data-testid` attribute and `testId` props - 12
Part of #1001 

- [x] Tooltip
- [x] TooltipDefinition
- [x] TooltipIcon
- [x] UIShell
- [x] UnorderedList
- [x] ValueCard
- [x] Walkme
- [x] WizardInline
- [x] WizardModal

kevinsperrine:
Closes #2510 

Depends on #2475 #2478 #2480 #2481 #2490 #2493 #2499 #2501 #2506 #2509

**Summary**

- Tooltip
- TooltipDefinition
- TooltipIcon
- UIShell
- UnorderedList
- ValueCard
- Walkme
- WizardInline
- WizardModal


packages/react/src/components/WizardInline/__snapshots__/WizardInline.story.storyshot
```diff
@@ -6,14 +6,15 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/
...
     className="WizardInline__StyledWizardWrapper-sc-8orpb0-0 hExrwn"
+    data-testid="wizard-inline"
   >
     <div
       className="bx--modal-container"
       data-id="WizardInlineContainer"
     >
       <div
         className="iot--wizard-inline__header page-title-bar"
-        data-testid="page-title-bar"
```
kevinsperrine, line 16: Not a breaking change. This was only added in a previous testId PR. https://github.com/carbon-design-system/carbon-addons-iot-react/blame/9f42ed19798f47c3b6079ff0e091f534546551b6/packages/react/src/components/WizardInline/__snapshots__/WizardInline.story.storyshot

packages/react/src/components/ValueCard/ValueRenderer.jsx
```diff
@@ -59,6 +67,7 @@ const ValueRenderer = ({
   return (
     <div className={`${BASE_CLASS_NAME}__value-renderer--wrapper`}>
       <span
+        data-testid={testId}
```
davidicus, line 70: Do we need test for ValueCard?

kevinsperrine: Yes, yes, we do. I'll get that added.

kevinsperrine: Done.

davidicus (approved):

sstone2423 (approved):

Issue #2435, opened 2021-06-28
[Repo] Upgrade Carbon to fix ref bug in ComboBox


kevinsperrine:
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

Starting in Carbon 10.32/7.32 refs are no longer passed to the ComboBox, so this breaks all custom functionality for tags, etc, in the PAL library. The version needed is unknown, because the issue is still outstanding. 

https://github.com/carbon-design-system/carbon/issues/8744

JordanWSmith15:
@kevinsperrine Is this ready to close? Seems like the proposed carbon upgrade pr has been merged

kevinsperrine:
Yep. This one was fix and merged with the carbon upgrade back before I went on vacation.

PR #2515, opened 2021-07-12
chore(repo): upgrade Carbon to 10.39/7.39

### LINKED ISSUE #2516: [DateTimePicker] absolute custom range breaks when onApply
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When using the DateTimePicker's custom 'absolute' range, if you select a start and end date, then click apply, the component will blow up in local, and cause odd things to happen such as swapping the range back to relative as the inner DatePicker from carbon crashes.

### To Reproduce

1. Go to DateTimePicker
2. Click on 'custom range'
3. Click on 'absolute'
4. Select a start and end date
5. Click 'apply'
6. See error (although this is intermittent)

### Expected behavior

DateTimePicker absolute custom range should always work when applying the dates

### Environment/versions:

- OS: MacOS
- Browser:chrome
- `carbon-addons-iot-react` version: `v2.146.0-next.48`

### Additional context

Internal issue: https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/2591

This bug is fixed in carbon-components-react with this PR: https://github.com/carbon-design-system/carbon/pull/8894 so we just need to upgrade carbon to pick it up

kevinsperrine:
Closes #2516, #2435

**Summary**

This PR upgrades Carbon from 10.31/7.31 to 10.39/7.39. They are all done at once instead of individually over time, because 10.32/7.32 breaks the iot library's ComboBox implementation because it stopped passing refs to the text input. This was not fixed until 10.39.

This breaks down the changes I had to make with each version bump to get our tests to pass.
## 10.32
- Danger buttons now have a new description added to them, so name matching like `getByRole('button', {name: "Delete"})` no longer works, because the buttons now appears in tests as `danger Delete`.
- MultiSelect now use labels instead of titles for options
- 🔥 ComboBox no longer passes ref.

## 10.33
- TableCardFormContent needs label to click edit button in test, not just text.
- ImageTile delete button needs `getByLabelText` instead of just `getByText`
- TableMultiSortModal: replaced `getByText` with `getByRole('button', {name: X})` for the add/remove column buttons
- DashboardEditor: replaced `getByText` with `getByRole('button')`

## 10.34
- No new test failures

### 10.35
- Tooltip: change `getByRole('tooltip')` to `getByRole('dialog').parentNode` when checking `data-floating-menu-direction` attribute
- Header: switch `getByRole('button')` to `getByRole('menuitem')` for items in the overflow menu

### 10.36
- MenuButton requires changing from `ContextMenuItem` to simply `MenuItem` from Carbon. This also applies to other items, such as `ContextMenuDivider`, `ContextMenuRadioGroup`, `ContextMenuSelectableItem`
- StatefulTableDetailWizard, StatefulWizard: change error test to `queryAllByRole('alert')` to find InlineNotifications

### 10.37
- No new test failures

### 10.38
- No new test failures

### 10.39
- No new test failures
- ComboBox now passes ref correctly and our internal refs were updated.

### Visual Changes

- Carbon added `display: inline-flex;` to anchor tags, so those will now follow rules for margin, etc.
- Tooltips have a 0.5rem margin-top applied.

- DateTimePicker should no longer aggressively clear input and maintain proper focus https://github.com/carbon-design-system/carbon/pull/8894
- ComboBox should function as expected. The refs were changed and it was broken in version 10.32. It was fixed in 10.39, and our tests now pass, but it needs a good eye.

- Sorry, but this is a **MAJOR** bump, so we'll need to go through stories with a fine-toothed comb.


stuckless (approved):
This is why I want a 'testing contact' .  Seems like there is a world of hurt coming my way :)

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -26013,396 +26106,197 @@ Map {
...
     },
   },
-  "InlineNotification" => Object {
+  "Pagination" => Object {
     "defaultProps": Object {
-      "hideCloseButton": false,
-      "iconDescription": "closes notification",
-      "notificationType": "inline",
-      "onCloseButtonClick": [Function],
-      "role": "alert",
```
kevinsperrine, line 26090: The loss of all these defaultProps and propTypes seems to be the result of Carbon's new createComponentToggle method used to roll-out v11 updates via feature-flags. https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Notification/index.js#L46

davidicus:
image
image
image
image
image
image

I am seeing some visual regressions from this PR
https://deploy-preview-2515--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-flyout-menu--custom-footer-example
https://deploy-preview-2515--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-gaugecard--with-data-state-no-data
https://deploy-preview-2515--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--stateful-example-with-expansion-max-pages-and-column-resize
https://deploy-preview-2515--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--row-selection-and-batch-actions

davidicus (approved):
Should be good when checks pass

PR #2520, opened 2021-07-12
chore(suite-header): clean up test noise

### LINKED ISSUE #2519: [SuiteHeader] test output cleanup
The jest timers in the new header IdleTimeout code need to be wrapped in act.

kevinsperrine:
Closes #2519 

**Summary**

- The new idle timeout detection in SuiteHeader needed the jest timers wrapped in act to prevent test noise output.

- wrapped jest timers in act
- set the `onIdleTimeoutWarning` to a no-op in the cleanup method of the IdleTimer. This prevents update state errors on unmounted components, because the useEffect cleanup doesn't destroy the class instance--only the timers--so the onIdleTimeoutWarning method can hang around and attempt to call setState on an unmounted component.

- no `act` or `update state` error in SuiteHeader testing output

- IdleTimeout detection still functions as expected.


davidicus (approved):

kevinsperrine:
> Nice work!
> There is still one console error not related to timers though
> image

Yeah, there's a v3 issue for that, because it'll be a breaking change to fix it.

PR #2594, opened 2021-07-14
fix(datetimepicker): pin flatpickr to fix absolute date range

### LINKED ISSUE #2592: [DateTimePicker] Unable to change the selected dates under the absolute option
Unable to change the selected dates under the absolute option. User is only allowed to click once to make the date selection in calendar view. After that, if the user tries to change the selection by clicking else where on the calendar there is no change in the current selection.

1. Go to https://carbon-addons-iot-react.com/?path=/story/watson-iot-datetime-picker--default
2. Click on calendar icon.
3. Click on 'Custom Range'.
4. Select 'Absolute' radio button.
5. Select start date by clicking on calendar.
6. Select end date by clicking on calendar.
7. Try to change the current selection by clicking on a new date.
8. See that nothing happens and the current selection stays unchanged.

Screen Shot 2021-07-13 at 5 02 47 PM

Health defect -> https://github.ibm.com/maximo/Asset-Health-Insight/issues/6896

kevinsperrine:
Closes #2592 

**Summary**

Since Carbon 10.30/7.30 there was an upgrade to flatpickr which bumped it from 4.6.1 to 4.6.9. Since then, absolute DateTimePicker has been broken in our library. It will not allow you to select a new DateTimeRange when clicking days in the calendar. There is a work-around. If you follow the instructions below it will allow you to set a date. However, after much trying to programmatically change focus and/or blur the calendar I was unable to come up with a work-able solution. Our current implementation using the useEffect to set which input is currently focused causes this new version of flatpickr to fire the `onChange` event twice--first with the correct new dates--then again with the original dates thus overwriting the new dates and therefore appears as if no new date is set. I attempted to take out this focus code in the useEffect, but then, we're stuck with only the "end" date being change while the "start" date always stays 04/06/2020 (in the storybook example). 

Edit: 

I think I've narrowed the error down to this change: https://github.com/flatpickr/flatpickr/blame/ac9b752381102118cf4f44b5353fe30811b56209/src/index.ts#L1612 in flatpickr. That relatedTarget check causes the setDate _not_ to fire, so the start date returned to us in the onChange handler is always the same.

Previous functionality is available in this deploy preview https://deploy-preview-2088--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-datetime-picker--selected-absolute

#### Work-around
1. Open the DateTimePicker
2. Click a new day in the calendar.
3. Click away from the calendar (just gray space in the body if using the storybook
4. Click the new day you want to set

- pinned flatpickr to version 4.6.1 (matching the same version used by Angular as well) -- I was unable to get it to resolve correctly when attempting to do it only for the react project.

- Open the DateTimePicker "Selected absolute" story
- Click the DateTimePicker to open it.
- Click "April 8, 2020"
- The Date range should now change to days "6->8" being highlighted.
- Click "April 15, 2020"
- The Date range should now change to days "8->15" being highlighted.
- Click "April 2, 2020"
- The Date range should now change to days "2->15" being highlighted.

- other features work as previously


bjornalm:
I'm not getting to the calendar function when I choose Custom range -> Absolute
image

(Using Chrome on Mac)

bjornalm:
> The Date range should now change to days "1->8" being highlighted.
Nope, it shows 6->8 but I think that is accurate

kevinsperrine:
> > The Date range should now change to days "1->8" being highlighted.
> 
> Nope, it shows 6->8 but I think that is accurate

Yeah, you're right. That was a typo on my part.

kevinsperrine:
> I'm not getting to the calendar function when I choose Custom range -> Absolute from the default story
> image
> 
> This works in next.
> (I'm using Chrome on Mac)

Nice catch. I've fixed it. Just waiting on the deploy preview.

davidicus (approved):

sstone2423 (approved):

PR #2598, opened 2021-07-14
fix(gauge-card): trends were cutoff on bottom of card

### LINKED ISSUE #2407: [GuageCard] content is cut off at smaller sizes
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The extra content like the trend indicator of the gauge card is cut off at smaller sizes

### To Reproduce

1. Go to https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-gaugecard--basic-with-expand
2. See error

image

kevinsperrine:
Closes #2407 

**Summary**

Fixed css and chart size to prevent trends being pushed off the card. I also opened a new issue to tackle auto-adjusting the size for different cards here: #2599 

- changed default radius to reduce the size of the gauge chart and prevent it pushing the trends off the card
- changed how the text is centered in the circle and removed some unnecessary css
- add the 16px bottom padding to match card designs
- added a new GaugeCard story with multiple GaugeCards.

- Nothing should be chopped on the GaugeCard basic story
https://deploy-preview-2598--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-gaugecard--basic
- The multiple story should stack on a `MEDIUM` size card in the `sm` breakpoint
- The multiple story should be inline on a `MEDIUM` size card in the `lg` breakpoint
https://deploy-preview-2598--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-gaugecard--multiple-gauges

- no other changes to GaugeCard


davidicus:
I believe the trend is supposed to go up next to the gauge when there is room as seen here. 
image


This may be a drastic change for applications that are not using these small sizes without the resizing of the gauge.
image

davidicus:
The card is also supposed to be able to support multiple gauges like this. I am not sure why this is not in any story. I wanted to give the heads up. 
image

kevinsperrine:
> The card is also supposed to be able to support multiple gauges like this. I am not sure why this is not in any story. I wanted to give the heads up.
> image

Got thanks. That's good to know. I assumed they were supposed to go below them like the ValueCards. These cards have (seemingly) never handled large sizes well--see below with my changes reverted. The story didn't have the knobs, so I realized it when I added them, and that's why I opened the other issue to track it. I'll revert these changes to fix the layout problems on small cards and match your example above, and we can tackle the different card sizes in the other issue.

Screen Shot 2021-07-15 at 8 26 52 AM

Screen Shot 2021-07-15 at 8 26 20 AM

packages/react/src/components/GaugeCard/GaugeCard.jsx
```diff
@@ -151,7 +151,7 @@ const GaugeCard = ({
                       { [`${iotPrefix}--gauge-value-lg`]: valueLength <= 2 }
                     )}
                     x={gaugeSize / 2}
-                    y="33"
+                    y={gaugeSize / 2 + 8}
```
bjornalm: nit: I would put the stroke value in a const

bjornalm (commented):
I'll revisit after changes as suggested by David

packages/react/src/components/GaugeCard/GaugeCard.jsx
kevinsperrine: this was actually already done. You were reviewing it while I pushed the updates from David's comments. :-)

bjornalm (approved):
Nice work! I like the css cleanup

davidicus:
@JordanWSmith15 @dianatran18 Do you want to take a look at this solution and give your stamp of approval?

JordanWSmith15 (changes requested):
Type looks too big in the gauge:
Screen Shot 2021-07-15 at 11 22 20 AM
Let's reduce the text value class from 1.75 to 1.5. Seems to look better. Class is:  iot--gauge-value-lg tspan:first-of-type

kevinsperrine:
> Type looks too big in the gauge:
> Screen Shot 2021-07-15 at 11 22 20 AM
> Let's reduce the text value class from 1.75 to 1.5. Seems to look better. Class is: iot--gauge-value-lg tspan:first-of-type

Thanks, @JordanWSmith15. The change is in place. Just waiting for the deploy preview to finish.

JordanWSmith15 (approved):

davidicus (approved):

Issue #2258, opened 2021-04-22
Improve table column grouping


zvonimirfras:
https://pages.github.ibm.com/ai-applications/design/components/datatable/groupedcolumns/

We can reuse the code for sterling table and keep the separators between columns, but we need to update the vertical text alignment (should be bottom).

katedraper:
@zvonimirfras - is this an angular request?

zvonimirfras:
@katedraper it is, thanks. Forgot to mark it 🙂

PR #2606, opened 2021-07-15
feat(table): add new table model and multi-line header implementation

### LINKED ISSUE #2395: Data table base API rewrite

zvonimirfras:
Closes #2395 #2258 

**Summary**

Defines the new table model API. This should allow changes to underlying implementation.

Implements a (multi-line) table using that model.


packages/angular/src/table/head/table-head-cell.scss
```diff
@@ -0,0 +1,42 @@
+// TODO add styling hooks to carbon-components-angular so we can remove aiTableHeadCell
+.table-head-cell-text,
+.table-head-cell-icons {
+  top: 0;
+}
+
+.bx--data-table--compact {
```
davidicus, line 7: you may want to use prefix variables as Carbon is doing away with the `bx` in major release.

zvonimirfras: Ah, yes 👍

packages/angular/src/table/head/table-head-cell.component.ts
```diff
@@ -0,0 +1,117 @@
...
+        >
+          {{ column.data }}
+        </span>
+        <ng-template
+          [ngTemplateOutlet]="column.template"
+          [ngTemplateOutletContext]="{ data: column.data }"
+        >
+        </ng-template>
+        <span class="table-head-cell-icons">
+          <svg
```
cal-smith, line 37: Do we have icons we can use with the directive instead of dropping the raw SVG in?

packages/angular/src/table/head/table-head-cell.component.ts
```diff
@@ -0,0 +1,117 @@
...
+        aria-expanded="false"
+        aria-haspopup="true"
+        [ibmTooltip]="column.filterTemplate"
+        trigger="click"
+        [title]="getFilterTitle() | async"
+        placement="bottom,top"
+        [data]="column.filterData"
+      >
+        <svg
+          xmlns="http://www.w3.org/2000/svg"
```
cal-smith, line 93: same icon comment

packages/angular/src/table/head/table-head-cell.scss
```diff
@@ -0,0 +1,51 @@
+@import '../../../../styles/scss/globals/vars';
```
cal-smith, line 1: Same style import comment as the other PR, should be pointing to `../../vendor/@ai-apps/styles/scss`

cal-smith: There's also quite a few styles here, is there anything here we can just pull up into the main PAL styles?

packages/angular/src/table/head/table-head.scss
```diff
@@ -0,0 +1,31 @@
+@import '~carbon-components/scss/globals/scss/vars';
```
cal-smith, line 1: This has a high chance of blowing up in different envs, and the PAL already proxies `carbon-components` so we shouldn't need both this and the PAL import

packages/angular/src/table/head/table-head.scss
```diff
@@ -0,0 +1,31 @@
...
+@import '../../../../styles/scss/globals/vars';
```
cal-smith, line 2: should point to the vendor files

packages/angular/src/table/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './table.module';
```
cal-smith, line 1: I think we need a `package.json` for this folder as well (so it gets built as a separate module)

packages/angular/src/table/table.stories.ts
```diff
@@ -0,0 +1,82 @@
...
+      data: 'Name',
+      rowSpan: 2,
+    }),
+    new TableHeaderItem({ data: 'hwer', colSpan: 2, sortable: false }),
+    null,
+  ],
+  [null, new TableHeaderItem({ data: 'hwer1' }), new TableHeaderItem({ data: 'hwer2' })],
+]);
+
+simpleModel.setData([
```
cal-smith, line 23: The body content isn't showing for me in the story 🤔 Not sure if that's intentional since we're still pretty early or?

cal-smith (changes requested):
Some initial comments, looks good overall though

PR #2612, opened 2021-07-20
fix(timeseriescard): Passing truncation type 'none' to chart tooltip options so that tooltip labels don’t get truncated.

blechner:
Closes #

It closes a label truncation issue found in a product that uses the `TimeSeriesChart`

**Summary**

`carbon-charts` supports tooltip label truncation settings. If not passed, the default is to chop any characters after the 14th character in the tooltip label. `TimeSeriesChart` was not overriding tooltip truncation settings, so now the truncation type is set to `'none'`, which means that the tooltip label can stretch the tooltip width as much as necessary for all the content to show without any cropping.

Before the change:

image

After the change:

image

All unit tests are still passing and and stories are still working.

### DIFF
--- a/packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx
+++ b/packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx
@@ -445,6 +445,9 @@ const TimeSeriesCard = ({
       },
       containerResizable: true,
       tooltip: {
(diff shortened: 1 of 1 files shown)


sstone2423 (approved):

scottdickerson (approved):
LGTM

PR #2613, opened 2021-07-21
Table card scroll

cgirani:
Closes - Monitor reported issue

**Summary**

- Table card not showing a horizontal scrollbar
- Fix Table card height to show a vertical scrollbar

Before:
image

Afte changes:
image

- Fix on table classes used in TableCard
- Use of `prefix` in TableCard file instead of `bx`.
- Update correctly the `numberOfRowsPerPage` for the amount that fits on the page.

- Make sure table card show scrollbar

- N/A

- [x] Make sure all references to `iot` or `bx` class prefix is using the prefix variable

### DIFF
--- a/packages/react/src/components/TableCard/TableCard.jsx
+++ b/packages/react/src/components/TableCard/TableCard.jsx
@@ -32,7 +32,7 @@ import {
 } from './tableCardUtils';
 import ThresholdIcon from './ThresholdIcon';
 
-const { iotPrefix } = settings;
+const { prefix, iotPrefix } = settings;
 
 const StyledStatefulTable = styled(({ showHeader, isExpanded, data, ...rest }) => (
   <StatefulTable {...rest} data={data} />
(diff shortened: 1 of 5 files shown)


packages/react/src/components/TableCard/TableCard.jsx
```diff
@@ -48,10 +48,10 @@ const StyledStatefulTable = styled(({ showHeader, isExpanded, data, ...rest }) =
       bottom: ${(props) => (!props.isExpanded ? '0px' : '25px')};
       ${(props) => (props.isExpanded ? `width: calc(100% - 50px)` : ``)}
     }
-    .bx--data-table-container {
+    .bx--data-table-content {
       ${(props) =>
         props.data && props.data.length > 0 && !props.isExpanded
-          ? `max-height: 435px;`
+          ? `max-height: 523px;`
```
davidicus, line 54: The issue states you are trying to correct for horizontal scroll but these changes seem to be related to height. Does the changing of the class provide the fix?  And are these height changes necessary?

packages/react/src/components/TableCard/TableCard.jsx
```diff
@@ -48,10 +48,10 @@ const StyledStatefulTable = styled(({ showHeader, isExpanded, data, ...rest }) =
       bottom: ${(props) => (!props.isExpanded ? '0px' : '25px')};
       ${(props) => (props.isExpanded ? `width: calc(100% - 50px)` : ``)}
     }
-    .bx--data-table-container {
+    .bx--data-table-content {
```
davidicus, line 51: Since we are touching this file can we use the prefix for these class names?

packages/react/src/components/TableCard/TableCard.jsx
cgirani: The class was incorrect and after having the CSS class fixed. That height was not correct, this is what would render with it:
image

So it's kind of another fix than just the scroll. I can update the PR content, sorry about that.

davidicus (approved):

blechner (approved):

PR #2624, opened 2021-07-29
Align card with updated spec

davidicus:
Closes #2145, Closes #2375  

**Summary**

This PR aligns the Card component with the latest Card Spec and allows for Card Title to Wrap to two lines instead of always truncating. 

**Change List (commits, features, bugs, etc)**

- Change Card component  styles in `_card.scss`
- Created a hook to implement overflow detection since we use in several places, `useHasTextOverflow.jsx`
- Updated test to to check for data-testids and for new elements, (subtitle, footerContent)
- Removed styles from GuageCard, BarChartCard, ImageCard that should have been in `_card.scss`
- added styles to TableCard and MapCard to force fullscreen. (exceptions to the new spec)
- updated snapshots for related components that use Card

**Acceptance Test (how to verify the PR)**

- Card stories should all match new specification outlined #2145 
- In [this](https://deploy-preview-2624--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card--with-ellipsed-title-tooltip-external-tooltip) Card story, both the Title and subtitle should be truncated and have tooltips
- Switch to LargeWide Card size and both title and subtitle should now be fully visible with no tooltip
- MapCard and TableCard stories should all look the same and be fullwidth

**Regression Test (how to make sure this PR doesn't break old functionality)**

- Look at the dashboard editor and different card permutations to make sure they are still functioning as expected. 

**Things to look for during review**

- [ ] Make sure all references to `iot` or `bx` class prefix is using the prefix variable
- [ ] (React) All major areas have a `data-testid` attribute. New test ids should have test written to ensure they are not changed or removed.
- [x] UI should be checked in RTL mode to ensure the proper handling of layout and text.
- [x] All strings should be translatable.
- [x] The code should pass a11y scans (The storybook a11y knob should show no violations). New components should have a11y test written.
- [ ] Unit test should be written and should have a coverage of 90% or higher in all areas.
- [x] All components should be passing visual regression test. For new styles or components either a visual regression test should be written for all permutations or the base image updated.
- [ ] Changes or new components should either write new or update existing documentation.
- [x] PR should link and close out an existing issue

PR #2627, opened 2021-08-02
feat(card): add new datetimepicker to card as option

### LINKED ISSUE #2454: [Card] add custom datetimepicker

scottdickerson:
Closes #2454

**Summary**

- Adds a new option to the Card availableActions that will render the custom range picker

- feat(Card): allow non-boolean values in the available actions
- feat(CardToolbar): add new timepicker if a availableActions.range is of type string and take additional time range options
- feat(cardToolbar): smaller font inside cards
- feat(CardPropTypes): extend the existing props for timeRange and timeRangeOptions and availableActions.range
- fix(TimeSeriesCard): was generating sample values even in the case where we weren't in editable mode

- Verify new card range picker stories send the action correctly

- Verify old card range picker still works in the story.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.146.0-next.70](https://github.com/carbon-design-system/carbon-addons-iot-react/compare/v2.146.0-next.69...v2.146.0-next.70) (2021-07-30)
+
+
+### Features
+
+* **tabs:** add tab actions ([030df06](https://github.com/carbon-design-system/carbon-addons-iot-react/commit/030df060d9629d2835e0ffb7248aa2a326ccef20))
+
+
+
+
+
 # [2.146.0-next.69](https://github.com/carbon-design-system/carbon-addons-iot-react/compare/v2.146.0-next.68...v2.146.0-next.69) (2021-07-27)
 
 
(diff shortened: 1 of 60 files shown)


davidicus:
Now that I see the implementation I am remembering that Diana was saying the date needs to be in the subtitle with the icons remaining icon only. From [discussion thread](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2455#issuecomment-871529449) in the other PR.

Which would require you to consume the changes from [this](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2624) PR to implement.

davidicus (changes requested):
Code looks good but we should follow design's implementation suggestions from [this](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2624) thread

davidicus (approved):

davidicus:
#2630 has been created to align this feature with desired spec.

JoelArmendariz (approved):

PR #2632, opened 2021-08-03
fix(datetimepicker): needs to use the correct locale in the card

scottdickerson:
Closes #

**Summary**

- fixes a bug where the cards were not passing locale down to the CardToolbar to render the DateTimePicker so the calendar was in the wrong format

- fix(Card): pass through the locale to the card
- fix(CardToolbar): localize the dates shown in the human readable date picker

- Verify if you look at the card range picker story and switch the locale, that the human readable date and the Calendar show correct locales in the Absolute Range Picker

- Verify in english


packages/react/src/components/Card/CardToolbar.jsx
```diff
@@ -210,6 +213,8 @@ const CardToolbar = ({
           <DateTimePicker
             id={testId}
             i18n={mergedI18n}
+            dateTimeMask="L HH:mm"
```
kevinsperrine, line 216: Should this mask be a configurable prop also? Secondly, should its default match the current "YYYY-MM-DD HH:mm" format so that the existing titles/snapshots aren't changed?

kevinsperrine (commented):
just a couple of questions about the formatting mask.

bjornalm (approved):
LGTM

packages/react/src/components/Card/CardToolbar.jsx
scottdickerson: sure, this is a good point

stuckless (approved):

packages/react/src/components/Card/CardToolbar.jsx
davidicus: Waiting on these changes to merge. Otherwise we should be good to go.

davidicus (changes requested):

davidicus (approved):

PR #2643, opened 2021-08-06
Empty state fix

### LINKED ISSUE #2642: [Tabs] Icon on second tab doesn't render
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

This one's surprising, an svg icon on the second tab doesn't render!

image

### To Reproduce

Here's how to see it in a storybook
```
export const TabsWithEmptyState = () => {
  return (
    <Tabs>
      <Tab>
        <EmptyState icon="no-result" title="No results found" />
      </Tab>
      <Tab>
        <EmptyState icon="no-result" title="No results found" />
      </Tab>
      <Tab>
        <EmptyState icon="no-result" title="No results found" />
      </Tab>
    </Tabs>
  );
};
```

Tab between the first and second tabs to see it

### Additional context

Working:
image

Broken:
image

davidicus:
Closes #2642 

**Summary**

Icons dynamically generated cannot contain ids or we get duplicated ids when icon is used more than once. Any of the empty state icons require ids in order to provide gradient for the fill. 

To fix the issue I have taken then and put the generated React Icon in a static folder inside of icon directory. 

- Created a `useUniqueId` hook to provide a unique id for Icons
- moved all EmptyState React icon variations into a static directory,
- swapped all references to these components to new location
- added spy for hook so snapshot would not be different each time. 
- updated snapshots
- added a story to showcase same icon being used for multiple instances as seen in the issue

- Make sure EmptyState stories all render 
- make sure any component that uses empty state (Table,List, etc) render correctly 


bjornalm (approved):
I really would have guessed that there was another solution to the problem, but I think you are correct in your analysis. Nice work!

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -31828,12 +31828,6 @@ Map {
   "DashboardIcon" => Object {},
   "DataScientistIcon" => Object {},
   "DataTableIcon" => Object {},
-  "Emptystate404Icon" => Object {},
-  "EmptystateDefaultIcon" => Object {},
-  "EmptystateErrorIcon" => Object {},
-  "EmptystateNoresultsIcon" => Object {},
-  "EmptystateNotauthorizedIcon" => Object {},
-  "EmptystateSuccessIcon" => Object {},
```
kevinsperrine, line 31836: Shouldn't we re-add the modified versions of these back into the export?

kevinsperrine (commented):
Couple of questions.

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
davidicus: yes. good catch.

davidicus: Actually I just did a build and they are there. We should be good. Looking at why they are not present in the snapshot

packages/react/src/components/EmptyState/EmptyState.story.jsx
```diff
@@ -11,7 +12,7 @@ export default {
 
   parameters: {
     component: EmptyState,
-
+    storyshots: { disable: true },
```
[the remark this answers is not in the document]
davidicus: This, I was really burnt out and had no one to bounce any ideas off of. The story where I test multiple instances of the same icon throws a warning but only for the snapshot. Stories work and unit test work. I was out of ideas. We can look at this together today but I turned off to not add more noise in the test output.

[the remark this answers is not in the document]
kevinsperrine: This is fixed with my push to add the body prop to that story.

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -31828,16 +31828,16 @@ Map {
...
   "LineGraphIcon" => Object {},
   "ListIcon" => Object {},
   "SimpleBarGraphIcon" => Object {},
   "ValueKpiIcon" => Object {},
+  "EmptyStateDefaultIcon" => Object {},
+  "EmptyState404Icon" => Object {},
+  "EmptyStateErrorIcon" => Object {},
+  "EmptyStateNoResultsIcon" => Object {},
+  "EmptyStateNotAuthorizedIcon" => Object {},
+  "EmptyStateSuccessIcon" => Object {},
```
kevinsperrine, line 31841: Do we want to match the previous casing? 
ie. `EmptystateNoresultsIcon` instead of `EmptyStateNoResultsIcon`

davidicus: Yeah I suppose I should. This casing make my head hurt.

kevinsperrine: heard that.

packages/react/src/components/EmptyState/EmptyState.jsx
```diff
@@ -72,59 +72,61 @@ const defaultProps = {
...
+        {icon &&
+          React.createElement(typeof icon === 'string' ? icons[icon] : icon, {
+            className: `${iotPrefix}--empty-state--icon`,
+            alt: '',
+            'data-testid': `${testID || testId}-icon`,
+          })}
+        <h3
+          className={`${iotPrefix}--empty-state--title`}
           // TODO: remove deprecated testID in v3.
-          data-testid={`${testID || testId}-action`}
```
davidicus, line 106: This change set is off. I did not make these changes.

kevinsperrine: I think this is just github being weird about the whitespace changes. If you view it with "ignore whitespace changes" on it looks fine.

davidicus: weird. Looking at next branch and this file doesn't have this code either. It appears to be right. https://github.com/carbon-design-system/carbon-addons-iot-react/blob/b9b9ce73515223a1fa7082b5afaebad8b786af92/packages/react/src/components/EmptyState/EmptyState.jsx#L91

kevinsperrine: That's _very_ strange. Github really didn't like you adding that return statement....

kevinsperrine: whitespace diffing is hard apparently.

kevinsperrine (approved):

PR #2647, opened 2021-08-10
fix(bar-chart-card): format x-axis labels to match TimeSeriesCard

### LINKED ISSUE #1572: [BarChartCard] interval='year' formats the axes incorrectly for time-based barChart
### What package is this for?

- [x] React

### Describe the bug

Pass 'year' as the interval prop and notice the x-axis doesn't show the year correctly

image

### LINKED ISSUE #2305: [BarChartCard] xaxis is ignoring the locale property
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
Pass locale french to the Dashboard Editor, notice the x-axis formatting of the line chart uses the correct french months, but the bar chart doesn't

image

kevinsperrine:
Closes #1572, #2305

**Summary**

Copies necessary formatting methods from TimeSeriesCard into BarChartCard to ensure the x-axis formatting matches those of the TimeSeriesCard

**EDIT**: This PR also fixes #2305 where the locales weren't being respected on the x-axis. I've added locale knobs to the BarChartCard story to confirm this.

- Adds ticks property to options being passed to the the chart.
- copies formatTick and maxTicksPerSize from TimeSeriesCard

- I've added a new story to the BarChartCard using time-series data where the number of points and intervals can be changed. As you change intervals from minute, hour, day, month, year. The x-axis label should be formatted the same as it is with TimeSeriesCards.

- Other BarChartCards not using time series data should format as they did previously.


davidicus (approved):
LGTM 👍

packages/react/src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -247,6 +249,49 @@ const BarChartCard = ({
     [categoryDataSourceId, chartData, timeDataSourceId, type, values]
   );
 
+  const previousTick = useRef();
+  dayjs.locale(locale);
+
+  const maxTicksPerSize = useMemo(() => {
```
bjornalm, line 255: nit: perhaps this could be a shared custom hook if it is used in multiple places. Not sure it is worth the trouble..

bjornalm (approved):
Looks good, but just want to make sure the quarter interval is correct?
image

packages/react/src/components/BarChartCard/BarChartCard.jsx
kevinsperrine: I thought about it, but not sure it's worth the time. I think these two places are the only place it's used, and I think, IIRC, the BarChartCard is supposed to replace the TimeSeriesCard at some point, but I could be making that up...

kevinsperrine: On the quarter interval. I'm honestly not sure. It matches the single date that TimeSeriesCard has, but it could be a bug there, too. @scottdickerson what's the expectation for quarterly intervals?

davidicus: @scottdickerson any input here?

davidicus: @kevinsperrine let's fix the conflicts and we will merge this.

Issue #2421, opened 2021-06-18
[MapCard] Error: Can't resolve 'react/jsx-runtime'


kevinsperrine:
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug
The error below is the result of the react-dnd upgrade to 14.0.2 for the new drag-and-drop features of the MapCard. 

```
ERROR in ./node_modules/react-dnd/dist/esm/core/DndProvider.js
Module not found: Error: Can't resolve 'react/jsx-runtime' in '/home/sls/git/maximo-app-framework/graphite-alt-repos/graphite-mas/packages/react-components/node_modules/react-dnd/dist/esm/core'
 @ ./node_modules/react-dnd/dist/esm/core/DndProvider.js 17:0-48 54:9-13
 @ ./node_modules/react-dnd/dist/esm/core/index.js
 @ ./node_modules/react-dnd/dist/esm/index.js
 @ ./node_modules/carbon-addons-iot-react/lib/hooks/useDNDProviderElement.js
 @ ./node_modules/carbon-addons-iot-react/lib/index.js
 @ ./src/ui/common/Loading.js
 @ ./src/ui/TextInput/TextInput.js
 @ ./src/ui/Dialog/DialogController.story.js
 @ ./src/ui sync \.story\.js$
 @ ./config/storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./config/storybook/config.js ./node_modules/webpack-hot-middleware/client.js?reload=true
```

Currently, carbon-addons-iot-react has a peer-dependency of `react@^16.8.6`, but react-dnd (since version 13) requires `react@>=16.14`. Technically, `react@^16.8.6` as a peer-dependency is loose enough to also match `react@16.14.0`, and b/c the `@storybook/angular` package has `react@16.14.0` as a dependency, our dev builds have been using `react@16.14.0` for a couple of months. See PR #2215.

![Screen Shot 2021-06-18 at 11 55 41 AM](https://user-images.githubusercontent.com/609466/122587702-26060800-d02c-11eb-8c98-a99dfcb0987e.png)

```
"@storybook/angular@^6.2.3":
  version "6.2.9"
  resolved "https://registry.yarnpkg.com/@storybook/angular/-/angular-6.2.9.tgz#531280cc87600bebb5833df611948b7e0cb27cc6"
[...]

PR #2651, opened 2021-08-11
fix(repo): pin @carbon/charts versions

### LINKED ISSUE #2639: [Repo] Carbon charts dependency should be pinned to specific version
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

All of our Carbon deps should be pinned to prevent unintended changes between versions. We currently have `^0.41.11` which will pull in the latest patch version. We should only have `0.41.100` (latest release) in our package.json
This will require us to upgrade our peer dep for d3 as well and make sure there are no breaking changes. 

### For instance

>image

Because of a dep change downstream this patch version broke our consumers.

kevinsperrine:
Closes #2639, #2421 

**Summary**

carbon/charts isn't following semver, so patch versions are causing major upgrades and breaking downstream consumers. This pins carbon charts to a specific version and updates are peer deps for react.

- Pins the versions of `@carbon/charts` and `@carbon/charts-react` to prevent the accidental downstream upgrade to `0.41.99+` and forcing a bump in`d3` also. 
- Updates the peer deps for react to 16.14.0 to account for the previous react-dnd upgrade 

- locally run `npx create-iot-react-app <appname>` and drop in this branch into the dependencies to ensure the project builds and the BarChartCard is formatted as expected.

- all other features should work as intended.


kevinsperrine:
@cal-smith @davidicus 

We're going to have to solve version resolutions in the monorepo, or we're eventually going to hit a wall that we can't easily get around. Yarn 1 doesn't support sub-package resolutions. There's [an issue that's been open for years](https://github.com/yarnpkg/yarn/issues/5039). Yarn 2 supposedly supports it, but I haven't tested it yet. In this case, I tried [every possible combination](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md) I could come up with, but the only way to pin the version was to pin it the same across both react and angular packages.

None of these resolutions (in the root package.json) had any effect. The resulting install still contained `@carbon/charts@0.41.101`, because `@carbon/charts-react` uses the `^` in it's dependency on `@carbon/charts`, and yarn ignores all these resolutions.
```js
    "**/carbon-addons-iot-react/@carbon/charts": "0.41.11",
    "**/carbon-addons-iot-react/@carbon/charts-react": "0.41.11",
    "**/carbon-addons-iot-react/**/@carbon/charts": "0.41.11",
    "**/carbon-addons-iot-react/**/@carbon/charts-react": "0.41.11",
    "carbon-addons-iot-react/@carbon/charts": "0.41.11",
    "carbon-addons-iot-react/@carbon/charts-react": "0.41.11",
    "carbon-addons-iot-react/**/@carbon/charts": "0.41.11",
    "carbon-addons-iot-react/**/@carbon/charts-react": "0.41.11"
```

This resolution works, but resolves it for both the react and angular package. react was on `0.41.11`, but angular is using `0.41.31`. I tried bumping react to `0.41.31` to match, but a couple of our tests on the PieChartCard failed in that case, so I kept it at `0.41.11`. I haven't checked this PR on the angular side yet to see if there are any side-effects to pinning an earlier version.
```js
    "@carbon/charts": "0.41.11",
    "@carbon/charts-react": "0.41.11"
```

cal-smith:
> @cal-smith @davidicus 
> 
> We're going to have to solve version resolutions in the monorepo, or we're eventually going to hit a wall that we can't easily get around. Yarn 1 doesn't support sub-package resolutions. There's [an issue that's been open for years](https://github.com/yarnpkg/yarn/issues/5039). Yarn 2 supposedly supports it, but I haven't tested it yet. In this case, I tried [every possible combination](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md) I could come up with, but the only way to pin the version was to pin it the same across both react and angular packages.
> 
> None of these resolutions (in the root package.json) had any effect. The resulting install still contained `@carbon/charts@0.41.101`, because `@carbon/charts-react` uses the `^` in it's dependency on `@carbon/charts`, and yarn ignores all these resolutions.
> ```js
>     "**/carbon-addons-iot-react/@carbon/charts": "0.41.11",
>     "**/carbon-addons-iot-react/@carbon/charts-react": "0.41.11",
>     "**/carbon-addons-iot-react/**/@carbon/charts": "0.41.11",
>     "**/carbon-addons-iot-react/**/@carbon/charts-react": "0.41.11",
>     "carbon-addons-iot-react/@carbon/charts": "0.41.11",
>     "carbon-addons-iot-react/@carbon/charts-react": "0.41.11",
>     "carbon-addons-iot-react/**/@carbon/charts": "0.41.11",
>     "carbon-addons-iot-react/**/@carbon/charts-react": "0.41.11"
> ```
> 
> This resolution works, but resolves it for both the react and angular package. react was on `0.41.11`, but angular is using `0.41.31`. I tried bumping react to `0.41.31` to match, but a couple of our tests on the PieChartCard failed in that case, so I kept it at `0.41.11`. I haven't checked this PR on the angular side yet to see if there are any side-effects to pinning an earlier version.
> ```js
>     "@carbon/charts": "0.41.11",
>     "@carbon/charts-react": "0.41.11"
> ```

We should be good to pin a lower version, I don't think we're really using charts at the moment in Angular ... It seems like a reasonable solution at the moment for us to just match versions of shared dependencies. It's a bit of extra work when updating, but if we stay on top of it, it'll be less painful. Ideally we shouldn't need to resolve different versions for each package, which will help keep install times down too.

sstone2423 (approved):

davidicus (approved):

Issue #2658, opened 2021-08-12
[Table] preserve columns widths on resize


bjornalm:
### What package is this for?

- [x] React
- [ ] Angular

### Summary
Add configurable feature to preserve existing column widths when one column is resized.

See spec [here](https://pages.github.ibm.com/cdai-design/pal/patterns/data-table/resizable-columns/
). We will follow the main principle of respecting the widths of columns not being resized, but not implement the complete UI with this issue.

**Additional context** 
Relates to #2460 and will use the same prop for configuration.

### Available extra resources

What resources do you have to assist this effort?

_Carbon add-ons for Watson IoT is a collaborative system. We encourage teams to
build components and submit them for integration._

PR #2659, opened 2021-08-12
2460 table adding columns ignores the defined widths

### LINKED ISSUE #2460: [DataTable] Adding columns ignores the defined widths
### What package is this for?

- [ ] React

### Summary

In storybook https://carbon-addons-iot-react.com/?path=/story/watson-iot-table--with-resize-on-column-resize-callback-no-initial-column-width-and-column-management
1. Alter the field "The width of the added columns (if any)" to be 500px
2. Click "Add" 
Result:
Column added is still 150px wide
Expected:
Column added matches the desired width, causing a horizontal scroll if necessary

This seems to be effecting our ability to insert columns at a wider size in Health

bjornalm:
Closes #2460

**Summary**
Adding prop `preserveColumnWidths` to prevent column widths from being modified when adding/removing/toggling columns.

The changes are only available for the combination of hasResize:true and preserveColumnWidths:true. This new behavior is hopefully more robust/logical than the original behavior when it comes to handling columns that are missing widths. If explicit column widths are missing during a column modification operation (add/remove/toggle) we will now use the currently rendered width or a default width if needed. But I chose not to add the same robustness for the default preserveColumnWidths:false in order not to break existing functionality. In a major release we can modify the default behaviour so that they are in sync (or perhaps remove the current default and get rid of some complexity).

* Added prop `preserveColumnWidths` to the `Table` and `TableHead` components. When true the existing widths of columns should be preserved when adding/removing/toggling columns. It will also affect the column resize behaviour but that will be addressed in https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2658.
* Modified `TableHead` to handle preserveColumnWidths:true for toggling and external updates of the columns.
* Added more supporting functions in `columnWidthUtilityFunctions.js`
* Added unit tests covering the new functionality
* Enhanced story "with options to explore column settings" to implement column visibility toggle for standard table. Also modified default story settings to be more aligned with how we expect the apps will use this functionality.
* Added a default width of 150px to be used when adding columns or toggling to show columns that do not have widths. 
* Updated the documentation

Use the story https://deploy-preview-2659--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-options-to-explore-column-settings. The knob configuration should be the default one unless stated otherwise. For each test, verify that the actual HTML renders the correct width and that the panel "COLUMNS prop" shows the correct width. No existing columns should grow or shrink (except for the last expander column) during tests, instead the table should grow and shrink in width if needed. 

**Note!**
**You might have to scroll down to see table's the horizontal scrollbar.**
 
- Test adding new column/columns with widths
- Test adding column/columns without width
- Test removing column/columns
- Test toggle hide columns 
- Test toggle show columns with width 
  1. Click the column selection button
  2. Click the button "secret information"
- Test toggle show a column without width
  1. Select `undefined` from knob "initial column width" so that the "secret information" column will not have a width. 
  2. Make sure knob "options.hasColumnSelection" is true
  3. Resize any column to make sure all visible columns get a width in the columns prop. "secret information" columns still has no width.
  4. Click the manage column button and toggle the "secret information" column. It should become visible and get a default width of 150px
- Test adding a column when there are no initial column widths provided to the table
  1. Select `undefined` in the  "initial column width" 
  2. Add a new column with a width
  3. Make sure that all columns now have a width in the "COLUMNS prop" panel but that no columns width visually changed.
- Test adding/removing a hidden column

1. Repeat the same test with knob preserveColumnWidths set to false and make sure the behaviour is the same as in Next when the knobs are the same.

2. Check the other relevant table stories


kevinsperrine (commented):
Really solid work! I think everything is good, but since there are a lot of different combinations and checking-steps, I'm going to come back to it again later today and check once more.

davidicus (approved):
LGTM 👍

kevinsperrine (approved):

PR #2662, opened 2021-08-12
fix(suite-header): truncate long menu item text

### LINKED ISSUE #2221: [AppShell] Strings of the menu items do not truncate
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

Translated text strings in ui shell menus do not ellipse 

### To Reproduce

Extend the string in any of the menu items and it cuts off. The menu items will do this in some translations.
image (21)

MAS 8.5

kevinsperrine:
Closes #2221 

**Summary**

Menu item text wasn't being truncated before.

- add css change to allow truncate of menu text.
- added long text examples to the "Header with custom action items" story.

- long text in the help and profile menus of the "Header with custom action items" story should truncate

- no other changes.


davidicus:
Is there a way to provide the title so users can have a browser tooltip? I know the content is dynamic but I wonder if we use the child prop

kevinsperrine:
> Is there a way to provide the title so users can have a browser tooltip? I know the content is dynamic but I wonder if we use the child prop

Theoretically, we could, but I would think that would fall on the consumers to pass via the metaData props like I change in the story wouldn't it? https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2662/files#diff-2ea9cb012362ff50d44cc770f39dc595015089e7cac8eb04721d463db7be6d86R174

bjornalm (approved):
Code looks good.

We could perhaps check for overflow and automatically add the metadata title if it isn't already present? Chances are that the App developers might miss that detail otherwise.

kevinsperrine:
> Code looks good.
> 
> We could perhaps check for overflow and automatically add the metadata title if it isn't already present? Chances are that the App developers might miss that detail otherwise.

That could work. I'll add that in today.

kevinsperrine:
@davidicus @bjornalm Added fallback titles when an item is truncated if they don't exist and tests to confirm.

davidicus (approved):

PR #2666, opened 2021-08-13
feat(table): preserve columns widths on resize

### LINKED ISSUE #2658: [Table] preserve columns widths on resize

bjornalm:
Closes #2658

**Summary**
Sibling columns widths are now preserved when a column is resized if option `preserveColumnWidths` is true.

- Added `preserveColumnWidths` to ColumnResize component to allow sibling columns to maintain their width when a column is resized.
- Added unit tests and renamed some existing to better show the difference between new and old tests

Go to https://deploy-preview-2666--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-options-to-explore-column-settings use the default knob configuration

1. resize columns and make sure only the resized column changes in width
   * A column can be made wider, all the way to the end of the right side of the visible part of the table
   * A column can be made thinner, but not thinner than the column min width of 62px.
   * Make sure RTL works
2. combine resizing with adding/removing/toggling columns to make sure it all works together
3. uncheck the preserveColumnWidths and verify that the old resize behaviour is still working

1. Test resizing in other table stories and make sure they work as before


kevinsperrine (approved):
Such a better experience. Awesome job. 

Your stuff all works great, but it looks like we have a couple small RTL issues with the pagination we should open an issue for.
Screen Shot 2021-08-13 at 8 50 41 AM

stuckless (approved):

davidicus (approved):

PR #2667, opened 2021-08-13
fix(stateful-table): update pagination properly on deletes

### LINKED ISSUE #1906: after delete items in the last page of StatefulTable, show empty
we are using 
carbon-components@10.17.0 deduped
| +-- carbon-components-react@7.17.0

StatefulTable
There are 2 pages of data, I enter page2, and click all the items on page2 one by one, use toolbar action to delete the selected items, after that, the table show empty, but it is not really, it is just page2 has no data.
The expected behavior is either show previous page, or show current page has no data, and show the pagination.
image

kevinsperrine:
Closes #1906 

**Summary**

- If all items on the last page of a Stateful table were deleted the page would show empty and the pagination bar would disappear. This PR uses the `TABLE_APPLY_ACTION` action to also update the pagination to reflect the changes in data on delete actions.

- update `pagination.totalItems` on delete
- check if the current page is greater than the expected total number of page after the deletion and update the current page if necessary
- pass the updated totalItems to the view.pagination prop instead of the initial totalItems passed to the table.

- On the [simple stateful table](https://deploy-preview-2667--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--simple-stateful-example) story, turn on pagination, go to page 10, and delete all items on that page. Those 10 items should be deleted and then you'll be moved to page 9 where the pagination bar will show 81-90 of 90 items.

- Page should not change if deleting items (event full pages) when not on the last page of the table.

### DIFF
--- a/packages/react/src/components/Table/StatefulTable.jsx
+++ b/packages/react/src/components/Table/StatefulTable.jsx
@@ -347,7 +347,7 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
   const filteredTotalItems =
     filteredCount && filteredCount !== currentlyLoadedDataCount
       ? filteredCount
-      : totalItems || currentlyLoadedDataCount;
+      : view?.pagination?.totalItems || totalItems || currentlyLoadedDataCount;
 
   return filteredData ? (
     <Table
--- a/packages/react/src/components/Table/StatefulTable.jsx
+++ b/packages/react/src/components/Table/StatefulTable.jsx
@@ -50,7 +50,7 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
     options,
     view: {
       toolbar: { customToolbarContent },
-      pagination: { totalItems },
+      pagination: { totalItems: initialTotalItems },
     },
     view: initialState,
     actions: callbackActions,
@@ -81,7 +81,7 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
         data: initialData,
         isLoading,
         view: initialState,
-        totalItems: totalItems || initialData.length,
+        totalItems: initialTotalItems || initialData.length,
         hasUserViewManagement,
       })
     );
@@ -347,7 +347,7 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
   const filteredTotalItems =
     filteredCount && filteredCount !== currentlyLoadedDataCount
       ? filteredCount
-      : view?.pagination?.totalItems || totalItems || currentlyLoadedDataCount;
+      : view?.pagination?.totalItems || initialTotalItems || currentlyLoadedDataCount;
 
   return filteredData ? (
     <Table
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Table/tableReducer.js
```diff
@@ -368,15 +368,27 @@ export const tableReducer = (state = {}, action) => {
       const data = filteredData || state.data;
       // only update the data and filtered data if deleted
       if (action.payload === 'delete') {
+        const { selectedIds } = state.view.table;
+        const { pagination } = state.view;
+        const totalItems = pagination.totalItems - selectedIds.length;
+        const numberOfPages = totalItems / pagination.pageSize;
```
bjornalm, line 374: This can lead to number of pages being in decimal form. We should probably round up

packages/react/src/components/Table/StatefulTable.jsx
```diff
@@ -347,7 +347,7 @@ const StatefulTable = ({ data: initialData, expandedData, ...other }) => {
   const filteredTotalItems =
     filteredCount && filteredCount !== currentlyLoadedDataCount
       ? filteredCount
-      : totalItems || currentlyLoadedDataCount;
+      : view?.pagination?.totalItems || totalItems || currentlyLoadedDataCount;
```
bjornalm, line 350: I don't understand this. Isn't the `totalItems` is destructed from `view.?pagination?.totalItems` already?

bjornalm (changes requested):
The bug described in the issue is solved but unfortunately we are introducing a new issue with these changes. If we go to the last page 10 like in the example but only remove 5 items we should stay on page 10 and it should only show 5 items. However now we end up on page 9.5 and we see 10 items.

packages/react/src/components/Table/tableReducer.js
kevinsperrine: 🤦‍♂️ I definitely didn't check that on Friday. Thanks.

packages/react/src/components/Table/StatefulTable.jsx
kevinsperrine: Yeah, it took me a little bit to figure this one out, too. The [defaultProps function in table](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/next/packages/react/src/components/Table/Table.jsx#L322) passes the length of the data from the initial data as the length of the totalItems, so the initial data passed always overrides the length of the loaded (and in this case changed) data. Thus, the `totalItems` destructed from the merging of that data is always 100 in this case. This may not be the best way to fix it, IDK. But that's the situation.

davidicus:
Looks ok. Will check once deploy preview is udpated.

packages/react/src/components/Table/StatefulTable.jsx
bjornalm: Ahh, I see. If this is the setup needed, can we rename `totalItems` to `initialTotalItems` to clarify the difference?

bjornalm (approved):
Works well now, but check out the last comment regarding renaming totalItems

packages/react/src/components/Table/StatefulTable.jsx
kevinsperrine: done.

davidicus (approved):

PR #2668, opened 2021-08-13
feat(theshold-form-item): allow strings when equal comparison

### LINKED ISSUE #2388: [ThresholdEditFormItem] Users cannot set comparisons for thresholds for string values
### What package is this for?

- [x] React
- [ ] Angular

### Summary

The CardEditFormItem Threshold editor should allow us to use string values if we're using the = comparison.  Currently it only allows numeric values.

The usecase is that someone wants to change the color of their Value Card if one of their sensor statuses changes.

Go to this story
https://next.carbon-addons-iot-react.com/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-cardeditor-cardeditformitems-thresholdformitem--example-with-initial-values

add a new threshold and change the comparison to equals.  Notice you are not allowed to enter a string value

kevinsperrine:
Closes #2388 

**Summary**

- Allow text input when threshold comparison is set to '='. Previously only allowed number inputs for all comparitors.

- check current comparison operator and use a TextInput if it is '='

- https://deploy-preview-2668--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-cardeditor-cardeditformitems-thresholdformitem--example-with-initial-values Try input any text in the third threshold in this story. Change the comparison operator to '>' or '<' and ensure strings are not allowed.

- All other changes to the thresholds should be unaffected.

### DIFF
--- a/packages/react/src/components/CardEditor/CardEditForm/_card-edit-form.scss
+++ b/packages/react/src/components/CardEditor/CardEditForm/_card-edit-form.scss
@@ -173,6 +173,9 @@
       .#{$prefix}--number input[type='number'] {
         padding-right: 0;
       }
+      .#{$prefix}--form-item input[type='text'] {
+        min-width: 9.375rem;
+      }
     }
     &--item-dropdown {
       margin-right: $spacing-05;
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/ThresholdsFormItem.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/ThresholdsFormItem.jsx
@@ -5,6 +5,7 @@ import omit from 'lodash/omit';
 import isEmpty from 'lodash/isEmpty';
 import uuid from 'uuid';
 import { red60 } from '@carbon/colors';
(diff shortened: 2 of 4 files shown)


packages/react/src/components/CardEditor/CardEditForm/_card-edit-form.scss
```diff
@@ -173,6 +173,9 @@
       .#{$prefix}--number input[type='number'] {
         padding-right: 0;
       }
+      .#{$prefix}--form-item input[type='text'] {
+        min-width: 9.375rem;
```
bjornalm, line 177: nit: is this value "hardcoded" for carbon number input elements or where does it come from?

bjornalm (approved):
Nice work

packages/react/src/components/CardEditor/CardEditForm/_card-edit-form.scss
kevinsperrine: Exactly. This is the hard-coded width in the number elements I used here to make them match.

davidicus (approved):

PR #2669, opened 2021-08-16
fix(table): prevent column widths below min width

### LINKED ISSUE #2474: [Table] min-width of column not always respected
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
The min-width is not being respected on initial rendering for columns with undefined width property.

### To Reproduce

1. Go to https://deploy-preview-2473--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-options-to-explore-column-settings and set "table container width" to be 300px.
2. Note that the columns are below the min-width (62px) that is used for column resize.

### Expected behavior

If the columns do not have an explicit width defined they should not be allowed to shrink below the min-width.

### Environment/versions:

- OS: all
- Browser: all

### Additional context

Relates to needed specification for table columns #2471

bjornalm:
Closes #2474

**Summary**
Fix to prevent the table from rendering column widths below the min width on the initial rendering

- Removed old (CSS) style prop containing widths that was not being used
- Added new check on initial rendering that changes widths below the minimum to the default width
- Added unit test

Instead of affected widths being changed to the actual min-width (62px) I used the default width of 150px instead after discussion with @JordanWSmith15. We reasoned that the users would rarely want the 62px width since it doesn't show much of the column header. The 62px min width still stays as the lower boundary for when the user resizes the columns.

Go to https://deploy-preview-2669--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-options-to-explore-column-settings use the default knob coniguration

**- Test initial column widths**
1. Use config: Initial column width: undefined, table container width: 300px, hasResize:true, wrapCellText: always, useAutoTableLayoutForResize:false, preserveColumnWidths:false
2. Make sure the columns get the default width of 150px. Before the fix the widths would have been below min width.
3. Change "table container width" to 800px. Verify that the columns get a width of 100px. This is the unmodified width set by the browser which we don't modify since it is above the min width.
4. Change "table container width" to "auto" (for a wide browser window) and make sure the columns widths are distributed evenly  

**- Test columns widths on window resize since it uses the same code**
1. Use config: Initial column width: undefined, table container width: auto, hasResize:true, useAutoTableLayoutForResize:true, preserveColumnWidths:false
2. Resize browser window and make sure column widths still adjust to the text of the columns headers and that the min-width of 62 pixels is respected.

Check that resize and column add/remove/toggle works as expected


kevinsperrine (approved):

packages/react/src/components/Table/TableHead/columnWidthUtilityFunctions.js
```diff
@@ -245,6 +244,28 @@ export const adjustLastColumnWidth = (ordering, columns, measuredWidths) => {
...
+ * When the browser layout engine sets the widths dynamically we need to to make
+ * some adjustments to the last column width and also to column widths below
+ * the allowed minimum.
+ * @param {*} ordering
+ * @param {*} columns
+ * @param {*} measuredWidths
+ */
+export const adjustInitialColumnWidths = (ordering, columns, measuredWidths) => {
+  // If the table isn't wide enough for all columns that has a defined width
+  // the browser will will shrink the last column instead of keeping its defined width.
```
davidicus, line 259: some typo of the double "will"

davidicus (approved):

PR #2677, opened 2021-08-16
chore(stateful-table): fix table story

### LINKED ISSUE #2291: [Table] Table story seems wrong
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Look at the story https://next.carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-row-nesting-and-fixed-columns
From the name of the story and column header texts it looks like the columns should be fixed width and the column header texts truncated, but that isn't the case.

### To Reproduce

1. Go to the story linked
2. See the result

### Expected behavior

Story intent and result should match if we wanna keep the story.

### Environment/versions:

- OS: [e.g. MacOS, Windows]

kevinsperrine:
Closes #2291 

**Summary**

- the https://deploy-preview-2677--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--stateful-example-with-row-nesting-and-fixed-columns story was pulling props from another location that had been changed for another story. It was also missing props to make it work as described. These have been fixed.

- The bad story should be renamed and new props added to make it match the described and implied behavior.

n/a

### DIFF
--- a/packages/react/src/components/Table/StatefulTable.story.jsx
+++ b/packages/react/src/components/Table/StatefulTable.story.jsx
@@ -64,7 +64,7 @@ export const StatefulTableWithNestedRowItems = (props) => {
         : undefined,
   }));
   return (
-    <div>
+    <div style={{ width: select('table container width', ['auto', '300px', '800px'], 'auto') }}>
       <StatefulTable
         id="table"
         {...initialState}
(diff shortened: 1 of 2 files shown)


davidicus (approved):

bjornalm (approved):
Great!

PR #2684, opened 2021-08-17
chore(table): remove exports for internal table components

### LINKED ISSUE #1649: [Table] remove export of internal table components
### What package is this for?

- [x] React
- [ ] Angular

### Summary
The TableHead and the TableBody are exported even though our table is not composable for these components, i.e. the Table and the StatefulTable create these components internally and they cannot be overridden. The TableHead and the TableBody are not used by any other components in carbon-addons-iot-react and I think it is unlikely that external components would be using them directly.

kevinsperrine:
Closes #1649 

**Summary**

Since our Tables aren't composable, we're removing exports of TableHead, TableBody, TableToolbar, EmptyTable, and TableSkeletonWithHeaders. As best I can tell, none of these are used outside of the table. Both TableSaveViewModal and TableManageViewsModal are still exported as they have stories documenting their external use. This is potentially a breaking change and is being merged against v3.

- see above.

- items listed above are no longer exported

- nothing breaks.


bjornalm (approved):
Nice one!

davidicus (approved):

PR #2687, opened 2021-08-17
chore(testing): increase coverage on PageTitleBar

### LINKED ISSUE #2531: [PageTitleBar] improve unit test coverage

kevinsperrine:
Closes #2531 

**Summary**

- added a cypress test to cover the missing tests for portions related to scrolling

- added a couple jest tests for small items missing
- added cypress test to cover scrolling behavior

- test coverage above 90%

n/a


packages/react/src/components/PageTitleBar/PageTitleBar.test.jsx
```diff
@@ -204,37 +204,24 @@ describe('PageTitleBar', () => {
...
+          upperActions={
+            <button data-testid="upper-action-delete" type="button">
+              Delete
+            </button>
+          }
+        />
+      </div>
+    );
+    expect(screen.getByTestId('test-description')).toBeVisible();
+    expect(screen.getByTestId('upper-action-delete')).toBeVisible();
```
bjornalm, line 225: nit: According to the testing-library [query priority](https://testing-library.com/docs/queries/about#priority) we should use getByRole when possible and resort to `data-testid` if we "can't match by role or text or it doesn't make sense (e.g. the text is dynamic)"

bjornalm (approved):
Looks good, one small comment

packages/react/src/components/PageTitleBar/PageTitleBar.test.jsx
kevinsperrine: Yeah, I didn't figure it mattered in this case since it's entirely controlled by in the props in the test. When testing the internals of a component I completely agree.

davidicus (approved):

PR #2689, opened 2021-08-18
2317 table column resize fails after all columns have been hidden

### LINKED ISSUE #2317: [table] column resize fails after all columns have been hidden
### What package is this for? 
- [x] React
- [ ] Angular

### Describe the bug
Column resizing fails after all columns are first hidden and then (some are) shown again. This is bug has probably been around for a while. 

### To Reproduce

1. Go to  https://next.carbon-addons-iot-react.com/?path=/story/watson-iot-table--stateful-example-with-expansion-max-pages-and-column-resize
2. Click on the column toggle button
3. Hide **all** the columns
4. Make a few columns visible again
5. Try and resize a column
6. Notice it won't work

### Expected behavior

Column resize should still work

### Environment/versions:
- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version:  `2.145.0-next.73`

### Additional context

image

bjornalm:
Closes #2317

**Summary**
Added fixes to make sure column resize always work after _all_ columns have been hidden. I identified two different scenarios with different solutions that we should verify. Resorting to use min-width as fallback when column widths are missing is not perfect but since this is bit of an edge case I think it is okay.

I kept the code changes isolated to the old resize code since the new behaviour `preserveColumnWidths` did not have these problems. 

* Added fix to `TableHead` and `columnWidthUtilityFunctions.js`
* Added unit tests corresponding two the scenarios
* Added small fix to the story so that knob value "Stateful Table" works as expected

**Scenario 1**
Go to https://deploy-preview-2689--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-options-to-explore-column-settings use the following knob configuration:

```
Type of Table: StatefulTable
Initial column width: undefined
table container width: auto (doesn't matter)
options.hasColumnSelection: true
options.hasResize: true
wrapCellText: auto (doesn't matter)
useAutoTableLayoutForResize: false
preserveColumnWidths: false
```

1. Hide all columns
2. show a few columns
3. Make sure resize works

**Scenario 2**
Go to https://deploy-preview-2689--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-options-to-explore-column-settings use the following knob configuration:

```
Type of Table: Table
Initial column width: undefined
table container width: 800px (doesn't matter)
options.hasColumnSelection: true
options.hasResize: true
wrapCellText: always (doesn't matter)
useAutoTableLayoutForResize: false
preserveColumnWidths: false
```

1. Hide all columns one by one
2. Show the columns "String", "Date", "Select" & "Secret information"
3. Make sure the column "Secret information" can be resized

- tests here

### DIFF
--- a/packages/react/src/components/Table/TableHead/columnWidthUtilityFunctions.js
+++ b/packages/react/src/components/Table/TableHead/columnWidthUtilityFunctions.js
@@ -121,6 +121,14 @@ export const getIDsOfAddedVisibleColumns = (ordering, currentColumnWidths) => {
 };
 
 /**
+ * Returns true if there are visible columns
+ * @param {array} ordering the table ordering prop that specifies the order and visibility of columns
+ */
+export const hasVisibleColumns = (ordering) => {
+  return ordering.some((col) => !col.isHidden);
+};
+
+/**
  * Returns true if all visible columns have a width.
  * @param {array} ordering the table ordering prop that specifies the order and visibility of columns
  * @param {array} columns The table column props
@@ -179,7 +187,8 @@ export const calculateWidthOnShow = (currentColumnWidths, ordering, colToShowIDs
   const newColumnsToShow = colToShowIDs.reduce((accumulator, colToShowId) => {
     const widthOfColumnToShow =
       getExistingColumnWidth(currentColumnWidths, columns, colToShowId) ||
-      getAverageVisibleColumnWidth(visibleColumns);
+      getAverageVisibleColumnWidth(visibleColumns) ||
+      DEFAULT_COLUMN_WIDTH;
     return [...accumulator, { id: colToShowId, width: Math.round(widthOfColumnToShow) }];
   }, []);
   const totalWidthNeeded = newColumnsToShow.reduce((acc, col) => acc + col.width, 0);
@@ -234,6 +243,10 @@ export const calculateWidthOnHide = (currentColumnWidths, ordering, colToHideIDs
  */
 export const adjustLastColumnWidth = (ordering, columns, measuredWidths) => {
   const visibleCols = ordering.filter((col) => !col.isHidden);
+
+  // If there are no visible columns there is nothing to adjust
+  if (!visibleCols.length) return measuredWidths;
+
   const lastIndex = visibleCols.length - 1;
   const lastColumn = columns.find((col) => col.id === visibleCols[lastIndex].columnId);
   const fixedWidth = lastColumn.width ? parseInt(lastColumn.width, 10) : 0;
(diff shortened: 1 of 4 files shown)


kevinsperrine (commented):
Code looks great. just the two very minor comments. 

One thing I see in interaction testing scenario 1 is that b/c the last column is using all the space, the first columns added aren't resizable when preserveWidths is false. Only the last column is resizable. The same is true for scenario 2. When secret information is the last column it's resizable, but it's not otherwise. 

resize-after-all-removed

bjornalm:
> One thing I see in interaction testing scenario 1 is that b/c the last column is using all the space, the first columns added aren't resizable when preserveWidths is false. Only the last column is resizable. The same is true for scenario 2. When secret information is the last column it's resizable, but it's not otherwise.
> 

They are actually resizable, but there is no space for them to shrink since they all are min-width. They also can't grow because that would require the sibling column to shrink, which it can't since it is also at min-width. So you have to resize the last column in order to free up space and then resize the others. It is not ideal, but I hope acceptable for this edge case of the old resize behaviour.

kevinsperrine:
> > One thing I see in interaction testing scenario 1 is that b/c the last column is using all the space, the first columns added aren't resizable when preserveWidths is false. Only the last column is resizable. The same is true for scenario 2. When secret information is the last column it's resizable, but it's not otherwise.
> 
> They are actually resizable, but there is no space for them to shrink since they all are min-width. They also can't grow because that would require the sibling column to shrink, which it can't since it is also at min-width. So you have to resize the last column in order to free up space and then resize the others. It is not ideal, but I hope acceptable for this edge case of the old resize behaviour.

Got it. Yeah, that's fine. Especially, since the new preserveWidths will become the default in v3. That'll make everything so much better.

kevinsperrine (approved):

packages/react/src/components/Table/TableHead/columnWidthUtilityFunctions.js
```diff
@@ -182,18 +190,31 @@ export const addMissingColumnWidths = ({ ordering, columns, currentColumnWidths
...
+
   const totalWidthNeeded = newColumnsToShow.reduce((acc, col) => acc + col.width, 0);
   const shrinkableColumns = [...newColumnsToShow, ...visibleColumns].filter(
     (col) => col.width > MIN_COLUMN_WIDTH
   );
 
-  const adjustedCols = shrinkColumns(shrinkableColumns, totalWidthNeeded);
+  const adjustedCols = [
+    // There are some scenarios where the new columns don't have an existing width
+    // and in that case they are adjusted to get a min with assigned.
```
davidicus, line 213: Typo?
```suggestion
    // and in that case they are adjusted to get a min width assigned.
```

davidicus (approved):
Looks good to me. I think there may be a typo. If so we can merge the suggested change and add the ready to merge tag. 👍

packages/react/src/components/Table/TableHead/columnWidthUtilityFunctions.js
bjornalm: Yes, thanks. I've made that mistake a few times.

PR #2694, opened 2021-08-19
chore(testing): increase coverage on SuiteHeaderAppSwitcher

### LINKED ISSUE #2547: [SuiteHeaderAppSwitcher] improve unit test coverage

kevinsperrine:
Closes #2547 

**Summary**

- more tests

- more tests

- 90% coverage

- n/a


davidicus (approved):

packages/react/src/components/SuiteHeader/SuiteHeaderAppSwitcher/SuiteHeaderAppSwitcher.test.jsx
```diff
@@ -109,13 +108,57 @@ describe('SuiteHeaderAppSwitcher', () => {
...
         noAccessLink="https://www.ibm.com"
       />
     );
     // Expect skeletons
-    expect(wrapper.find('[data-testid="suite-header-app-switcher--loading"]')).toHaveLength(1);
+    expect(screen.getByTestId('suite-header-app-switcher--loading')).toBeVisible();
+  });
+
+  it('calls the default onRouteChange', async () => {
+    jest.spyOn(SuiteHeaderAppSwitcher.defaultProps, 'onRouteChange');
```
bjornalm, line 122: nit: just a minor reflection. I think that it is not actually the call to  `SuiteHeaderAppSwitcher.defaultProps.onRouteChange` that we are interested in since it always returns true no matter what we pass in. Instead, we probably want to make sure that we get a certain functionality even if we don't supply a `onRouteChange` prop.

bjornalm (approved):
Nice work. Just a thought below on the testing of default functions (since we do that a lot).

packages/react/src/components/SuiteHeader/SuiteHeaderAppSwitcher/SuiteHeaderAppSwitcher.test.jsx
kevinsperrine: fair point. added a expect on the href for this.

PR #2697, opened 2021-08-19
fix(datetimepicker): handle null select

### LINKED ISSUE #2653: [DateTimePicker] cannot deselect the default selection with 'null' value
### What package is this for?

- [ ] React
- [x] Angular

### Describe the bug

not able to deselect the default selected value with `[selected] = "null"`.

Screen Shot 2021-08-11 at 1 56 23 PM

Screen Shot 2021-08-11 at 1 13 14 PM

### To Reproduce

### Expected behavior

Should not select any date options when pass `null` to property `selected`.

Screen Shot 2021-08-11 at 1 13 14 PM

### Environment/versions:

- OS: MacOS
- Browser: chrome, firefox
- `carbon-addons-iot-react` version: 2.146.0-next.70

### Additional context

I think this issue is caused by
Screen Shot 2021-08-11 at 1 53 41 PM

upcoming release?

date, the better!_

sophiiae:
Closes #2653 

**Summary**

- summary_here

- items_here

- tests here

- tests here


packages/angular/src/date-time-picker/date-time-picker.component.ts
```diff
@@ -233,13 +234,15 @@ export class DateTimePickerComponent implements OnChanges, OnInit {
...
-      this.selected = [this.dateTimeRanges[0].key];
+      this.selected = [null];
     }
   }
 
   formatCurrentRangeTitle() {
     const [rangeOrType] = this.selected;
-    if (rangeOrType === 'RELATIVE' || rangeOrType === 'ABSOLUTE') {
+    if (!rangeOrType) {
+      return '';
```
cal-smith, line 244: We should probably add a `placeholder` input to return in this case (I'm assuming if you're passing `[null]` you don't want a specific selection)

packages/angular/src/date-time-picker/date-time-picker.component.ts
```diff
@@ -48,6 +48,7 @@ export type DateRange = [Date, Date];
         (keydown.space)="togglePicker()"
         [ibmTooltip]="formatCurrentRange()"
         [offset]="tooltipOffset"
+        [disabled]="!formatCurrentRangeTitle()"
```
cal-smith, line 51: Good thinking, though if we need to disable this we should probably have a separate `disabled` input we can pass down

cal-smith (changes requested):
A couple suggestions, but otherwise thanks for taking this on!

cal-smith (approved):
LGTM! Thanks!

PR #2701, opened 2021-08-20
chore(testing): increase TableCard coverage, fix bugs found

### LINKED ISSUE #2557: [TableCard] improve unit test coverage

kevinsperrine:
Closes #2557 

**Summary**

- Increases the testing coverage on TableCards to 90%, but also fixes a few bugs found along the way.

- upgrade `@testing-library/dom` and `@testing-library/userEvent` to newest version without breaking changes. This was to fix issues userEvent.type not working correctly in some instances.
- spread the linkTemplate properties when determining expanded columns, so that `target` doesn't get lost anymore
- Fix the truthy check on expandedColumns to display `--` when the label is missing
- **Fix row actions passed as an array in table data to be able to render node as icons instead of just strings.** This one doesn't show any adverse reactions in testing--all tests still pass--after the addition of `PropTypes.array` as a valid type passed to the `TableBodyRow` and `TableCellRenderer`. Previously, this wasn't working. Firstly, because the fallback check wasn't taking the array into account. ie: `typeof actionList[0].icon === 'string' ? icons[actionList[0].icon] : actionList.icon` instead of `typeof actionList[0].icon === 'string' ? icons[actionList[0].icon] : actionList[0].icon`. Secondly, because the actions array passed in the data was being run through `JSON.stringify` and then later `JSON.parse` so the React functions for elements was being lost in that conversion. I have no idea why that stringification was happening. There are no comments or reasons. However, I removed it and now allow nodes to be passed as action icons. It's possible this could have performance impact and maybe was the original reason for stringification, I don't know. I've included this change it's own commit so as to be easy to revert if needed. 

- testing coverage is 90%+

- Tables and TableCards still function as expected and at the same speed.


packages/react/src/components/TableCard/TableCard.jsx
```diff
@@ -453,9 +452,7 @@ const TableCard = ({
         : hasActionColumn || filteredPrecisionColumns.length || thresholds
         ? tableData.map((i) => {
             // if has custom action
-            const action = hasActionColumn
-              ? { actionColumn: JSON.stringify(i.actions || []) }
```
kevinsperrine, line 457: The source of the original stringify of action items. Anyone know why this was done?

packages/react/src/components/TableCard/tableCardUtils.jsx
```diff
@@ -104,6 +104,7 @@ export const handleExpandedItemLinks = (row, expandedRow, cardVariables) => {
       updatedExpandedRow.push({
         ...item,
         linkTemplate: {
+          ...linkTemplate,
```
kevinsperrine, line 107: Previously, without this spread, the target on expandedColumn links was being lost.

packages/react/src/components/TableCard/TableCard.jsx
```diff
@@ -200,18 +200,17 @@ const TableCard = ({
     id,
   ]);
 
-  const renderActionCell = (cellItem) => {
-    const actionList = JSON.parse(cellItem.value);
```
kevinsperrine, line 204: Where the original actions array was being re-parsed after being stringified below. Since I removed that I destructured the needed elements in the function.

packages/react/src/components/TableCard/TableCard.jsx
```diff
@@ -200,18 +200,17 @@ const TableCard = ({
...
+  const renderActionCell = ({ value: actionList, rowId }) => {
     return actionList && actionList.length === 1 ? (
       React.createElement(
-        typeof actionList[0].icon === 'string' ? icons[actionList[0].icon] : actionList.icon,
+        typeof actionList[0].icon === 'string' ? icons[actionList[0].icon] : actionList[0].icon,
```
kevinsperrine, line 206: The icon list is always an array, even with only one element, so I assume this was meant to work originally or did at some point. Getting coverage over this part is what opened up the whole can of worms, but seemingly it was a simple issue to fix.

packages/react/src/components/TableCard/TableCard.jsx
bjornalm: I don't know. Perhaps to create a deep clone? Are they modified at some point later?

bjornalm (approved):
Nice work

davidicus (approved):

PR #2745, opened 2021-08-25
Support load more in Table and List component

cgirani:
Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2773

**Summary**

- Support load more for Table and List

image
image

**Updates**
- Loading data support

image

image

- table component update
- list component update
- test update

- Try storybook `load more` for Table and List

- Make sure all the existing functionality still works.


packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,97 @@
...
+  ordering,
+  tableActions: { onRowLoadMore },
+  loadMoreText,
+  nestingLevel,
+}) => {
+  const tableCells = (
+    <>
+      <StyledTableCell
+        key={`${tableId}-${id}-row-load-more-cell`}
+        colSpan={ordering.length + nestingLevel + 1}
```
bjornalm, line 76: Look at the TableBodyRow component. It receives a prop called `totalColumns` that is being passed down from the Table. It contains the real number of columns that a row like this must span. You should use the same prop here to make sure it works for all scenarios/table configurations,

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,97 @@
...
+       *    row: PropTypes.object like this {col: value, col2: value}
+       * }, you should return the node that should render within that cell */
+      renderDataFunction: PropTypes.func,
+    })
+  ).isRequired,
+  /** offset level if row is nested */
+  nestingLevel: PropTypes.number,
+  /** tableActions */
+  tableActions: PropTypes.shape({
+    onRowLoadMore: PropTypes.func,
```
bjornalm, line 36: You can probably skip the wrapping tableActions object and just accept the onRowLoadMore

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+const { TableRow, TableCell } = DataTable;
+
+const propTypes = {
+  /** The unique row id */
+  id: PropTypes.string.isRequired,
+  /** The unique id for the table */
+  tableId: PropTypes.string.isRequired,
+  /** What column ordering is currently applied to the table */
+  ordering: PropTypes.arrayOf(
```
bjornalm, line 17: You don't need the ordering prop, see comment about `totalColumns `

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,97 @@
...
+       *    value: PropTypes.any (current cell value),
+       *    columnId: PropTypes.string,
+       *    rowId: PropTypes.string,
+       *    row: PropTypes.object like this {col: value, col2: value}
+       * }, you should return the node that should render within that cell */
+      renderDataFunction: PropTypes.func,
+    })
+  ).isRequired,
+  /** offset level if row is nested */
+  nestingLevel: PropTypes.number,
```
bjornalm, line 33: You probably don't need to care about nestingLevel either

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,97 @@
...
+const TableBodyLoadMoreRow = ({
+  id,
+  tableId,
+  ordering,
+  tableActions: { onRowLoadMore },
+  loadMoreText,
+  nestingLevel,
+}) => {
+  const tableCells = (
+    <>
```
bjornalm, line 73: No need to wrap using <>

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,97 @@
...
+      <StyledTableCell
+        key={`${tableId}-${id}-row-load-more-cell`}
+        colSpan={ordering.length + nestingLevel + 1}
+        onClick={() => onRowLoadMore(id)}
```
bjornalm, line 77: How about placing an actual button in the table cell instead? It can still span the entire row. We should also assign the button a test-id.

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,97 @@
...
+      >
+        <StyledSpan>
+          {loadMoreText} &nbsp;
```
bjornalm, line 80: no &nbsp;

bjornalm (changes requested):
Nice work! We are on the right track but we need a few changes as suggested by the comments. 

I also wonder about the UX of actually loading more rows. The stories don't show how it looks, but it seems we have to make use of the existing "isLoading" prop which turns the entire list/table into a loading SkeletonText. This means that all the items above (and possible scrollbars) disappears until the loading has finished and that is probably not the smoothest experience. An inline spinner that is added to the load more row until we have the additional data would be nicer. For that we might need an "isLoadingMore" prop.

@dianatran18 do you have any input on this.

packages/react/src/components/List/List.jsx
```diff
@@ -99,6 +102,7 @@ const defaultProps = {
     searchPlaceHolderText: 'Enter a value',
     expand: 'Expand',
     close: 'Close',
+    loadMore: 'Load more',
```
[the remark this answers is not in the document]
cgirani: @CynZhang what is your input about it?

packages/react/src/components/List/List.jsx
```diff
@@ -208,14 +215,34 @@ const List = forwardRef((props, ref) => {
...
+                item.id,
+                getAdjustedNestingLevel(item?.children, level)
+              );
+            })
+            .concat(
+              item.hasLoadMore
+                ? [
+                    <div
+                      key={`${item.id}-list-item-parent-loading`}
+                      className={`${iotPrefix}--list-item load-more-row`}
```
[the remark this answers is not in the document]
cgirani: it's supposed to be just one row, so maybe keep with just in singular?

packages/react/src/components/List/List.jsx
```diff
@@ -113,6 +118,7 @@ const defaultProps = {
   },
   emptyState: 'No list items to show',
   testId: 'list',
+  handleLoadMore: () => {},
```
stuckless, line 121: are they any params to the loadMore?

packages/react/src/components/List/List.test.jsx
```diff
@@ -157,4 +158,41 @@ describe('List', () => {
...
+            children: [
+              { id: 'site-01', content: { value: 'Site 1' } },
+              {
+                id: 'site-02',
+                content: { value: 'Site 2' },
+                children: [
+                  { id: 'system-01', content: { value: 'System 1' } },
+                  { id: 'system-02', content: { value: 'System 2' } },
+                ],
+                hasLoadMore: true,
```
stuckless, line 179: so each child element has to have a property that identifies that the children can be loaded?  So I could set hasLoadMore on every item?

packages/react/src/components/Table/Table.jsx
```diff
@@ -271,6 +271,9 @@ const propTypes = {
       /* (index) => {} */
       onRemoveMultiSortColumn: PropTypes.func,
       onTableErrorStateAction: PropTypes.func,
+
+      /** */
+      onRowLoadMore: PropTypes.func,
```
stuckless, line 276: what are the parameters for the load more?

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,58 @@
...
+  /** The unique row id */
+  id: PropTypes.string.isRequired,
+  /** The unique id for the table */
+  tableId: PropTypes.string.isRequired,
+  /** The unique id for the table test */
+  testId: PropTypes.string.isRequired,
+  /**  total columns has the overall total */
+  totalColumns: PropTypes.number.isRequired,
+  /** call back function for when load more row is clicked */
+  onRowLoadMore: PropTypes.func.isRequired,
```
stuckless, line 21: what are the params?

stuckless (changes requested):
I'm trying to understand how this works for a table, especially.  I mean a table has pages, so is used without paging? with paging?  What happens when I click on loadMore button? Do you go into a loading state waiting to get the items?  Am I expected to just keep adding new data to data structure for the table and just keep passing this down with every load more?

@JordanWSmith15 I'm assuming this is coming from a PAL design.  have you see the design, does it make sense?

packages/react/src/components/List/List.test.jsx
cgirani: yep, I would say mostly for the hierarchy table because the idea behind it is that I would not load all children of that, so I could have "paging"  for that child, where the load of it happens during hierarchy expansion, and I don't have all the data prior loaded.

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
cgirani: updated the comment with `call back function for when load more row is clicked  (rowId) => {} `

packages/react/src/components/List/List.jsx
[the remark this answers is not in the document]
bjornalm: Ah, yes, it is for the actual row with the button

JordanWSmith15:
> @JordanWSmith15 I'm assuming this is coming from a PAL design. have you see the design, does it make sense?

I have not seen this in PAL. @dianatran18 is out on leave, but I don't have a problem with it in the case of hierarchical data tables that expand to show child rows. Lazy loading as you scroll would not really work in that scenario, nor having multiple pagination bars. All I would ask is that this feature should be hidden/unavailable UNLESS hierarchy/child rows are present.

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -0,0 +1,60 @@
...
+  onRowLoadMore,
+  loadMoreText,
+  totalColumns,
+  isLoadingMore,
+}) => {
+  const tableCells = (
+    <TableCell
+      key={`${tableId}-${id}-row-load-more-cell`}
+      colSpan={totalColumns}
+      onClick={() => onRowLoadMore(id)}
```
bjornalm, line 40: With just the onClick it is not possible to use keyboard navigation for this feature. Why don't you put a Button inside the TableCell? Buttons are great for this :-)

(Keyboard navigation works great in the load more for the List component where you used a button)

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.test.jsx
```diff
@@ -0,0 +1,35 @@
...
+  id: 'tableRow01',
+  tableId: 'tableId',
+  testId: 'table-test',
+  totalColumns: 5,
+  loadMoreText: 'Load more...',
+  isLoadingMore: false,
+};
+
+describe('TableBodyLoadMoreRow', () => {
+  it('calls onRowLoadMore when load more id clicked', () => {
```
bjornalm, line 17: nit: spelling id = is

bjornalm (changes requested):
Nice job, we are definitely getting there. 

I think the StatefulTable should listen to the `onRowLoadMore` and set the `isLoadingMore` automatically (but also forward the callback to the application to handle the actual loading of data). The stateful stable should then also set `isLoadingMore` to false when more rows have been loaded. It is kind of what the stateful table does, it manages UI states for the application. 

We probably want to show this in a story and that story should not be the same story as for the normal table since we there manage the `isLoadingMore` state externally. And a unit test for the reducer would be nice, making sure the stateful table does what it should.

JordanWSmith15:
I'd like to see the stateful table example doing a bit more... loading rows, showing those rows and the effect it has on scrolling, and showing the loading row on more than just one level of hiearchy

bjornalm (changes requested):
Nice work, we are almost there.  I've made a few comments to the code, but we also need:

* The full scenario of actually loading data implemented for the Table stories as well, not just the List.
* Update the prop documentation in Table.mdx and List.mdx
* Tests on the TableComponent and the reducer

JordanWSmith15:
>     * The full scenario of actually loading data implemented for the Table stories as well, not just the List.

This. With stateful table selected on "Type" prop, I want to see the full behavior.  in https://deploy-preview-2745--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--row-expansion-and-load-more

packages/react/src/components/Table/TableBody/TableBody.jsx
```diff
@@ -273,7 +282,7 @@ const TableBody = ({
     const rowHasSingleRowEditMode = !!(myRowActionState && myRowActionState.isEditMode);
     const isSelectable = rowEditMode || someRowHasSingleRowEditMode ? false : row.isSelectable;
 
-    const rowElement = (
+    const rowElement = !row.isLoadMoreRow ? (
```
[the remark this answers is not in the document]
cgirani: To add the load more row, I do a check on the prop `hasLoadMore` exist and added the check to make sure the row also have `children` prop

packages/react/src/components/Table/tableReducer.js
```diff
@@ -554,6 +555,9 @@ export const tableReducer = (state = {}, action) => {
             isSelectAllSelected: {
               $set: view ? view.table.isSelectAllSelected : false,
             },
+            loadingMoreIds: {
```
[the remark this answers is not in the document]
bjornalm: So, to follow up on our chat about the TABLE_REGISTER not being triggered when you change the data prop. That is because we are not doing a deep compare of the data array. But if you after appending the loaded data create a new array, e.g. just by spreading the modified one you will trigger the TABLE_REGISTER:

```
const [myData, setMyData] = useState(originalData);

onLoadMore={() => {
  setMyData((currentData) => {
    // Load and append your children to the currentData
    
    // Return a new array
    return [...currentData];
  });
}}

<StatefulTable
  data={myData}
  ...
/>
```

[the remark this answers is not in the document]
cgirani: I do have it in the stories, but it doesn't seems to be triggering it: 
image

It seems that since I do the update of the data it just became to be the `initialData` and update directly the `state reducer` so it seems to have actually changed since it doesn't have a previous anymore.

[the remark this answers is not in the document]
bjornalm: Ok, can you push what you have so far and I will take a look?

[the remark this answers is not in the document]
bjornalm: You are right. We need to dispatch a new action from the StatefulTable when the data has changed. After line 125 you can add something like
```
  useEffect(() => {    
    if (view.table.loadingMoreIds?.length) {
      // Dispatch action to figure out which loading more ID that should be removed 
      dispatch(moreRowsLoaded({data: initialData}))
    }
  }, [initialData]);
```
You also need to create the new action in the reducer :-)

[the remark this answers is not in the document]
bjornalm: UPDATE: We were able to use the TABLE_REGISTER action

packages/react/src/components/Table/Table.jsx
```diff
@@ -961,11 +972,13 @@ const Table = (props) => {
...
                   'onClearRowError',
                   'onRowExpanded',
-                  'onRowClicked'
+                  'onRowClicked',
+                  'onRowLoadMore'
                 )}
                 // TODO: remove 'id' in v3.
                 testId={`${id || testId}-table-body`}
                 showExpanderColumn={showExpanderColumn}
+                isLoadingMore={view.table.loadingState.isLoadingMore}
```
bjornalm, line 981: This is no longer needed right?

packages/react/src/components/Table/tableReducer.test.jsx
```diff
@@ -473,6 +473,50 @@ describe('table reducer', () => {
       // is Loading should be set false and rowCount should be correct
       expect(tableWithSortedData.view.table.loadingState.isLoading).toEqual(false);
       expect(tableWithSortedData.view.table.loadingState.rowCount).toEqual(0);
+
+      // load more
+      const initialStateLoadMore = {
```
bjornalm, line 478: Wouldn't we wanna see a test where the the REGISTER_TABLE is called with e.g. 3 ids in view.table.loadingMoreIds and with data that indicates that 2 rows have gotten more children added. Then have the `expect` verify that the loadingMoreIds only contains the id of the row that didn't get additional children added?

bjornalm (changes requested):

packages/react/src/components/Table/Table.jsx
cgirani: nice catch, removing it.

bjornalm (approved):
Great work with all this. You are now an expert on our table. Thanks for contributing :-)

packages/react/src/components/List/List.story.jsx
```diff
@@ -620,3 +620,100 @@ export const WithPagination = () => (
...
+      },
+    ]);
+    return (
+      <div style={{ width: 400 }}>
+        <List
+          title="Sports Teams"
+          iconPosition="left"
+          items={listItems}
+          expandedIds={expandedIds}
+          toggleExpansion={(id) => setExpandedIds((prev) => [...prev, id])}
```
davidicus, line 663: In order to have the storybook work appropriately we should probably add a way to close parent.
```suggestion
          toggleExpansion={(id) => setExpandedIds((prev) => prev.includes(id) ? prev.filter((prevId) => prevId !== id) : [...prev, id])}
```

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/_table-body-load-more-row.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../../../globals/vars';
+
+.#{$iot-prefix}--load-more-cell {
```
davidicus, line 3: I would rather see us use a more qualified selector rather than the `!important` here. Something like `.#{$iot-prefix}--table-container .#{$iot-prefix}--load-more-cell`. That way we don't have issues with specificity in the future.

davidicus (approved):
Other than the comment about the use of `!important` I think this looks good to go. Thanks @cgirani and @bjornalm for all the work you both put in on this.

davidicus:
@stuckless Are you ok with these changes? I think @JordanWSmith15 has signed off on the design and @bjornalm and @cgirani have worked out a lot of details.

stuckless (approved):

stuckless:
@davidicus if @JordanWSmith15 is good, then I'm good.

PR #2768, opened 2021-08-27
fix(header): Add isActionItemVisible function, account for hidden icons in overflow checking

### LINKED ISSUE #2739: [SuiteHeader] Truncation to three-dot logic breaks when any icons are hidden
In the Configuration app in Maximo, the user is working locally and is not connected to MAS (or logged in). We need to hide things that are not connected up to the MAS APIs, such as the 9-dot suite navigation. When we hide these icons, the three-dot truncation occurs early.

Need to add some props to https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-suiteheader--header-with-custom-action-items that allow you to remove icons as well as add more.

kevinsperrine:
Closes #2739 

**Summary**

The overflow menu wasn't triggering at the prop widths if the header contained hidden icons. This PR accounts for hidden icons when determining if the overflow menu should be triggered, and also adds a new `isActionItemVisible` prop that allows you to choose (by item label--since label is already required) if that item should be shown or not. It also updates the logic for changing the icon into an `x` by only causing that to happen when the panel or menu is opened from within the overflow menu. 

-added cypress retries to config
- added hidden icon to header overflow story
- added isActionItemVisible callback function
- only change button icon to close when triggered from the overflow menu
- increase testing coverage and move overflow tests to e2e in cypress.

- The overflow menu should not trigger until the Announce button reaches the edge of the threshold
- Clicking buttons with menus (when not in overflow mode) should not change their icon to an 'x'
- Buttons with labels that cause isActionItemVisible to return false should not be shown--not even in the DOM.
- 9-dot menu should be hidden: https://deploy-preview-2768--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-suiteheader--header-with-custom-action-items 
- Overflow menu should not trigger until the "Announcements" button reaches edge--even though there's another element that's hidden by a css class: https://deploy-preview-2768--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-header--small-breakpoint
- all tests pass

- backwards compatibility is maintained for all action items


packages/react/src/components/Header/Header.jsx
```diff
@@ -77,8 +77,9 @@ const propTypes = {
     openMenu: PropTypes.string,
     closeMenu: PropTypes.string,
   }),
-
   testId: PropTypes.string,
+  /** Returns true, if the icon should be shown. (itemLabel) => {} */
```
stuckless, line 81: Pass the entire item action to the isItemVisible and then allow action items to have an .id field.  We cannot use the label since this would be too hard to manage from a localization point of view.  ie, we if want to hide the 'logout' action we can test for item.id === 'logout' which will be the same no matter what language we are in.

stuckless (changes requested):

packages/react/src/components/Header/Header.jsx
kevinsperrine: @stuckless, Good call. I've made these adjustments and added ids to the default action items in SuiteHeader.  https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2768/commits/34550b6eac2181711029b982f512b41ac53c2283

stuckless (approved):

davidicus (approved):

PR #2775, opened 2021-08-27
fix(repo): use es build instead of lib in browser

### LINKED ISSUE #2260: Tree shaking is broken due to wrong "browser" value in package.json
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

https://github.com/carbon-design-system/carbon-addons-iot-react/blob/next/packages/react/package.json

Currently the "browser" property is pointing to a CommonJS module folder, which breaks tree shaking in webpack. The correct value should be  "es/src/index.js". 

```
  "browser": "lib/src/index.js",
  "main": "lib/src/index.js",
  "module": "es/src/index.js",
```

### To Reproduce

This is carbon-addons-iot-react size built from graphite, which is 623KB (after gzip) - all the components not being used in my application are bundled in it. 
image

I manually updated "browser" value to "es/packages/react/src/index.js" in the package.json of carbon-addons-iot-react in my node_modules, then re-run graphite build. Now carbon-addons-iot-react is only 40KB. 
image

Here is the webpack tree shaking documentation https://webpack.js.org/guides/tree-shaking. Refer to the first two bullets. 

image

If possible, please fix this ASAP. Since our application is a mobile application, this issue significantly impact the performance.

kevinsperrine:
Closes #2260 

**Summary**

- We were using the `lib` build instead of the `es` build in the browser property of `package.json`. This was forcing everything to be included and breaking tree shaking. The test app was 8.08MB when built and analyzed in its current state, but after the es switch it dropped to 2.1MB. 

- change browser package to use es build

1. Pull this branch down locally
2. create a new directory to hold the built project ie: `mkdir -p ~/tmp/pal-build`
3. run `yarn build`
4. copy built files to project directory created above: `cp -rf package.json umd scss lib es css ~/tmp/pal-build`
5. move to a fresh directory
6. create a test app: `npx create-iot-react-app pal-test`
7. cd into the test app directory:  `cd pal-test`
8. install the locally linked version of carbon-addons. This must be either a relative path or an absolute path (the `~` will not be expanded), so continuing with the `~/tmp/pal-build` example above it would become `yarn add "carbon-addons-iot-react@link:$HOME/tmp/pal-build"` (on mac)
9. build the test application. `yarn build`
10. use serve to serve the built project and confirm it works. `serve -s build`

Optional:
You can compare the build differences by installing source-map-explorer (`yarn add source-map-explorer` and running it on the built application: `yarn source-map-explorer 'build/static/js/*.js'`.

- consuming applications still work as expected

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": "12.x || 14.x"
   },
-  "browser": "lib/index.js",
+  "browser": "es/index.js",
   "main": "lib/index.js",
   "module": "es/index.js",
   "unpkg": "umd/carbon-addons-iot-react.js",


stuckless (dismissed):
I have a vague memory of PAL components changing this value a few months back and it caused build issues for us.  We use a CRA based builds and after consuming the update it broke when did production builds.  I don't recall if this was because of changing the browser field or the main field, but, maybe you can review this history of those groups of lines and see if you can find the issue.

I found the issue I was referring to... https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2026

Can you confirm that a CRA based app both build in dev and production mode with this change.

kevinsperrine:
> I have a vague memory of PAL components changing this value a few months back and it caused build issues for us. We use a CRA based builds and after consuming the update it broke when did production builds. I don't recall if this was because of changing the browser field or the main field, but, maybe you can review this history of those groups of lines and see if you can find the issue.
> 
> I found the issue I was referring to... #2026
> 
> Can you confirm that a CRA based app both build in dev and production mode with this change.

That's correct, @stuckless. The fix in #2026 was to add the browser property to fix your build error. This change points that browser property to use the `es` build of the package instead of lib so that tree-shaking works properly and the bundle size can be reduced greatly for consumers. The detailed steps for approval listed above tests this change against a CRA based app. It built and worked with a BarChartCard, and that's why I included such detailed steps so others can recreate it as a consumer app. Once @bjornalm or @scottdickerson confirms this fix, I'll re-request a review from you.

FYI, we also have an issue slated for September to add this type of build into our CI to catch those types of errors in the future.

davidicus:
I was able to confirm that we could build and start up app for both development and production

davidicus (approved):

bjornalm (approved):
App successfully installed and running. Thanks for the clear instructions!

PR #2777, opened 2021-08-30
testing(list/tile-catalog-new): improve quality of tests for these components

### LINKED ISSUE #1966: [List / TileCatalogNew] rewrite tests
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Tests across these 2 components are very weak and some of the tests are written to simply make the jest coverage pass such as testing that default props are defined or called instead of using react-testing-library methods.

Even worse, our test suite is already getting large and taking longer and longer to run so wasting time on useless tests only hinders us more.

### Expected behavior

Write tests from the user's perspective using react-testing-library. For example, to test searching in a component, fire user events such as click or change and see that the component reacts correctly

kevinsperrine:
Closes #1966 

**Summary**

- previously these components just had test written to get coverage, but weren't actually real functionality. That's fixed now and coverage is also still over 90%

### DIFF
--- a/packages/react/src/components/List/List.test.jsx
+++ b/packages/react/src/components/List/List.test.jsx
@@ -44,21 +44,123 @@ describe('List', () => {
 
   // TODO: write tests that actually test selection
   it('List to have default handleSelect', () => {
-    expect(UnconnectedList.defaultProps.handleSelect).toBeDefined();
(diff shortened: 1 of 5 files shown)


bjornalm (approved):
LGTM, just some old comments we might wanna remove?

stuckless (approved):

PR #2778, opened 2021-08-30
fix(date-time-picker): wait until finished typing to parse dates

### LINKED ISSUE #2771: [DateTimePicker] When I try and create an absolute range both on the same day with different hours, clicking apply resets it to 1969 and misses data
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/2790

reproducible in the storybook too: https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-datetime-picker--default

kevinsperrine:
Closes #2771 

**Summary**

- DateTimePicker wasn't waiting until the user had finished typing to parse times, so it was throwing invalid dates.

- don't parse absolute times until 5 characters have been typed.

- Open DateTimePicker default story
- Click Custom Range
- click Absolute
- Click two dates (same and different)
- Click the Start time input
- Backspace to clear the box and type a time
- tab to the End time input
- type an end time
- click apply
- The correct chosen date should be visible

- no adverse side-effects for absolute dates.

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -398,14 +398,16 @@ const DateTimePicker = ({
       }
       case PICKER_KINDS.ABSOLUTE: {
         let startDate = dayjs(value.absolute.start);
-        if (value.absolute.startTime) {
+        // wait to parse it until fully typed
+        if (value.absolute.startTime && value.absolute.startTime.length === 5) {
           startDate = startDate.hours(value.absolute.startTime.split(':')[0]);
           startDate = startDate.minutes(value.absolute.startTime.split(':')[1]);
         }
         returnValue.absolute.start = new Date(startDate.valueOf());
         if (value.absolute.end) {
           let endDate = dayjs(value.absolute.end);
-          if (value.absolute.endTime) {
+          // wait to parse it until fully typed
+          if (value.absolute.endTime && value.absolute.endTime.length === 5) {
             endDate = endDate.hours(value.absolute.endTime.split(':')[0]);
             endDate = endDate.minutes(value.absolute.endTime.split(':')[1]);
           }


bjornalm (approved):
Nice

scottdickerson (approved):
LGTM

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -398,14 +398,16 @@ const DateTimePicker = ({
       }
       case PICKER_KINDS.ABSOLUTE: {
         let startDate = dayjs(value.absolute.start);
-        if (value.absolute.startTime) {
+        // wait to parse it until fully typed
+        if (value.absolute.startTime && value.absolute.startTime.length === 5) {
```
stuckless, line 402: Not sure this is correct.  Are we saying that all time inpute MUST be HH:MM, if I have to enter 01:30 vs 1:30, that doest seem right.  @JordanWSmith15

stuckless (changes requested):

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
kevinsperrine: @stuckless There's no AM/PM entry on this, so even if we were to allow 4 characters, it's likely to end in confusion, because 1:30 will become 1:30 AM.

scottdickerson:
@sls-ca I believe there was a reason we had to wait for 5 chars, @kevinsperrine try typing 11:30 into one of the fields, I think the widget choked on strings like that because it will try and set the time to 11:03.

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
stuckless: For things like dates and times it would make sense to me that we don't try to validate them until we blur, OR, you move to some additional logic like... 
if the second char is a ':' validate after 4 chars otherwise 5 chars.  I think it's perfectly likely that people are going to enter `7:00' for 7am.

Are you saying that there has been an org wide change to MAS and Maximo where all date inputs are now 24 hour time without AM/PM? Do we also format/display that way as well?  When was this decision made? Does Maximo know?  It's reasonable to me that any time input control allow for AM/PM as part of the entry (even if they use a separate am/pm button).  @JordanWSmith15 any thoughts on this.

I don't get why we are trying to validate dates and times when we type when there are so many variants of how that would work for as you type validation and it eliminates using am/pm for no other reason than we are validating as you are typing.

kevinsperrine: >For things like dates and times it would make sense to me that we don't try to validate them until we blur, OR, you move to some additional logic like...
if the second char is a ':' validate after 4 chars otherwise 5 chars. I think it's perfectly likely that people are going to enter `7:00' for 7am.

I agree. I think onBlur makes perfect sense, but the TimePickerSpinner component doesn't currently have an onBlur callback hence why it's being checked onChange. 

>Are you saying that there has been an org wide change to MAS and Maximo where all date inputs are now 24 hour time without AM/PM? Do we also format/display that way as well? When was this decision made? Does Maximo know? It's reasonable to me that any time input control allow for AM/PM as part of the entry (even if they use a separate am/pm button). @JordanWSmith15 any thoughts on this.

No, that's not what I'm saying at all. I'm simply stating that for many US users, 24-hr time is not typical. A user entering 7:00 and expecting 7:00pm is not out of the question. International users who use 24-hr time regularly would intuitively know to enter 19:00. I don't know the industries of our customers, so maybe that's not a valid concern. The placeholder already shows it in `hh:mm` format, so I don't think it unreasonable for a 0 to be entered to ensure the correct time is entered.

Your suggestion of checking the position of the colon and adjusting the logic to 4 chars instead of 5 is the best solution here if we're going to allow 4.

stuckless: I guess for Graphite we ended re-creating our own date/time components, since, the Carbon ones and the PAL ones are far too restrictive.  Our 100k users certainly would not like the idea of being forced into 24hr format, but since we don't use this, I'll go ahead and approve it anyway, but, I really do think there needs to be more design and UX put into these components.  Even the data validation should not be entirely managed in the component but rather you should be able to provide your own onValidate(string) and provide custom logic for how this component validates.  But since we don't use it, I'm fine with just moving forward with this fix, since my guess is that Monitor is requiring it.

stuckless (approved):

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
kevinsperrine: I completely agree. There are definitely many improvements that can and should be made. We're working on it! 🙂  I'll open an issue with your concerns as feature improvements. Thanks, Sean.

PR #2784, opened 2021-08-31
refactor(table): refactor wrap truncate code

### LINKED ISSUE #1652: [Table] refactor wrap-truncate code
### What package is this for?

- [x] React
- [ ] Angular

### Summary
The wrap and truncate code and unit tests were refactored as part of PR 1645 but we decided to move the refactoring into a separate issue and add it to the next major release in order to mitigate the risk of regressions. Should be part of #956

**Additional context**
See PR https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1645 

We should probably fix issue #1650 before fixing this one. 

TODO:
- Refactor the Table code according to PR 1645
- Merge the internal props 'wrapCellText' and 'truncateCellText' into a new prop 'cellTextOverflow'. Affected files are  TableBody, TableBodyRow, TableHead and TableCellRenderer.
- Remove old unit tests that have been replaced with new one in Table.test 
- Refactor the unit tests in that rely on props in TableCellRenderer.test

bjornalm:
Closes #1652 

**Summary**
The wrap and truncate code and unit tests were refactored as part of PR 1645 but we decided to move the refactoring into a separate issue and add it to the next major release in order to mitigate the risk of regressions. 

* Removed internal props  `wrapCellText` and `truncateCellText` and replaced with `cellTextOverflow`
* Lifted logic from the subcomponents to the table 
* Removed superfluous and old unit tests that were based on checking the presence of prop values

- Verify that the external `wrapCellText` prop still work as expected for the tables

- Verify that the external `wrapCellText` prop still work as expected for the tables

### DIFF
--- a/packages/react/src/components/Table/Table.test.jsx
+++ b/packages/react/src/components/Table/Table.test.jsx
@@ -12,8 +12,6 @@ import { Modal } from '../Modal';
 import { getTableColumns, mockActions, getNestedRows, getNestedRowIds } from './Table.test.helpers';
 import Table, { defaultProps } from './Table';
 import TableToolbar from './TableToolbar/TableToolbar';
-import TableBodyRow from './TableBody/TableBodyRow/TableBodyRow';
-import TableHead from './TableHead/TableHead';
 import { initialState } from './Table.story';
 
 const { iotPrefix, prefix } = settings;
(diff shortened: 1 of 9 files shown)


packages/react/src/components/Table/Table.test.jsx
```diff
@@ -565,81 +563,6 @@ describe('Table', () => {
     expect(wrapper.find('.bx--search-input').prop('value')).toEqual('');
   });
 
-  it('cells should always wrap by default', () => {
```
stuckless, line 568: It scares me when I see people refactoring code AND removing unit tests.  The unit tests are there to ensure that you preserve the same behaviour before and after the refactor.  Now we don't know and I have to believe that your new tests effectively capture the same behaviour.  I would recommend leaving in the existing unit tests until such time that we are confident that your new code and old code behave the same.

stuckless (dismissed):

packages/react/src/components/Table/Table.test.jsx
bjornalm: We had this discussion for these tests almost a year ago :-) 
See https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1645#discussion_r499643701
This is for v3, just to be extra safe.

packages/react/src/components/Table/TableCellRenderer/TableCellRenderer.test.jsx
```diff
@@ -34,104 +38,81 @@ describe('TableCellRenderer', () => {
...
+    expect(screen.getByText('true')).toHaveClass(textTruncateClassName);
+    expect(screen.getByText('127')).toHaveClass(textTruncateClassName);
+    expect(screen.getByText('a div').parentNode).not.toHaveClass(textTruncateClassName);
   });
 
   it('only shows tooltip if text is actually truncated', () => {
     setOffsetWidth(10);
     setScrollWidth(20);
 
     const wrapper = mount(
```
kevinsperrine, line 103: switch these to testing-library, too? or are there a ton of tests in there and we'll do it later?

bjornalm: Nope, just this old bad boy that somehow managed to survive. I'll rewrite it.

packages/react/src/components/Table/Table.test.jsx
stuckless: Ok, but this never did address my original concerns :(  ie, you concluded that the risk is acceptable to you.

My question still remains... did the unit tests HAVE to be removed at this point, or could they have remained and be removed AFTER the refactor has be validated.  The one thing unit tests are really got at, is catching regressions during a refactor operation.  I'm all for refactoring code, I do it all the time.  But if I need to rewrite tests as a result of the refactor then my confidence drops, since, I can't say that my changes, under the hood, still have the same behaviours before and after my changes.

bjornalm: Short answer yes. 
Because these unit tests were based on checking props (not functionality) of _internal sub components_, e.g.
```
// Make sure subcomponent TableBodyRow has prop wrapCellText='always'
`expect(wrapper2.find(TableBodyRow).prop('options').wrapCellText).toEqual('always');`
```
Not a great test to begin with, I know. Now, since this refactoring removes that wrapCellText prop, it can't be used for testing anymore. The actual functionality is covered by other tests added in the previous PR.

packages/react/src/components/Table/TableCellRenderer/TableCellRenderer.test.jsx
bjornalm: Corrections, there are many. I'll fix them all.

davidicus (approved):
While I agree with Sean's sentiment it does not seem to be possible to keep these particular test.

kevinsperrine (approved):
Nice work!

PR #2785, opened 2021-08-31
fix(date-time-picker): date time picker styles and add keyboard navigation/tests

### LINKED ISSUE #2433: [DateTimePicker]: Adjust styles and a11y after design review
There are still a few items in the design review that we need to tackle on the DateTimePicker component in react.

1. Tooltip and styles
2. keyboard navigation
3. inconsistence use of menu controls

PDF: [UI audit - Date Time Range Selector - Angular (v1).pdf](https://github.com/carbon-design-system/carbon-addons-iot-react/files/6668937/UI.audit.-.Date.Time.Range.Selector.-.Angular.v1.pdf)

UI audit - Date Time Range Selector - Angular (v1)  @2x

_Originally posted by @Najil-K in https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2365#issuecomment-863073400_

kevinsperrine:
Closes #2433 

Depends on #2761 since the e2e tests already existed in that coverage PR.

**Summary**

- This begins the process of bring the DateTimePicker into design alignment. There are some elements that haven't been created yet, so this only handles basic items like spacing from a design standpoint. However, it also adds keyboard navigation for accessibility.

- shows tooltip on hover/focus
- match tooltip styles to designs
- add keyboard navigation to presets list (tab, arrow keys)
- fix keyboard navigation between Relative, Absolute radios (arrow keys)
- adds cypress-plugin-tab plugin for tab testing
- fix cypress viewport sizes that mismatched between local and CI
- add e2e tests for keyboard navigation

- DateTimePicker stories should better match provided designs
- The DateTimePicker can be navigated and used via the keyboard
- still meets 90%+ testing threshold

- No averse effects in any stories
- All old stories should be unchanged.

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -336,6 +336,12 @@ const DateTimePicker = ({
     },
   };
 
+  /**
+   * A callback ref to capture the DateTime node. When a user changes from Relative to Absolute
+   * the calendar would capture focus and move the users position adding confusion to where they
+   * are on the page. This also checks if they're currently focused on the Absolute radio button
+   * and captures it so focus can be restored after the calendar has been reparented below.
+   */
   const handleDatePickerRef = useCallback((node) => {
     if (document.activeElement?.getAttribute('value') === PICKER_KINDS.ABSOLUTE) {
       previousActiveElement.current = document.activeElement;
@@ -358,6 +364,7 @@ const DateTimePicker = ({
           dp.appendChild(datePickerElem.cal.calendarContainer);
         }
 
+        // if we were focused on the Absolute radio button previously, restore focus to it.
         if (previousActiveElement.current) {
           previousActiveElement.current.focus();
           previousActiveElement.current = null;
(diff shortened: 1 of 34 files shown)


stuckless (approved):

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -477,8 +498,96 @@ const DateTimePicker = ({
...
+        setIsExpanded(!isExpanded);
+        break;
+    }
+  };
+
+  /**
+   * Moves up the preset list to the previous focusable element or wraps around to the bottom
+   * if already at the top.
+   */
+  const moveToPreviousElement = () => {
```
davidicus, line 537: Could these be helper utilities that we can pull out so they can be used elsewhere? Or are these specific to this component?

kevinsperrine: theoretically, I suppose they could. I don't know exactly where we would use it, but it's only looking for siblings with `tabindex != -1` and wrapping around the ends of the array to move through them...

davidicus (approved):
nice work!

PR #2789, opened 2021-09-01
fix(simple-list): update state properly when external props change

### LINKED ISSUE #2316: [SimpleList] Cannot change Icon component without throwing away and regenning the list
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Because the simple list keeps the items property in local state it doesn't update if the items list changes.  Specifically I change the icon based on an interaction with the list, but the list never updates.  I am forced to change the key of the list so React throws it away

### To Reproduce

Pass the simple list a set of items with icons.  Now change the icons based on some interaction with the page (maybe a button at the top).  The icons will never change in the simple list

kevinsperrine:
Closes #2316 

**Summary**

Previously, the whole list needed to be thrown out when any props changed, but now I've added useEffects to update the state as new items are passed or as pageSize change, for example. It should maintain the previous state, so if you were on Page 2 and new items are added you'll stay on Page 2, but if items are removed you'll reset back to page 1. If you previously had a search value and you add more items the search value will be maintained.

- added useEffects to update state when items or pagination props change
- added new tests to cover all this functionality and increase coverage even more

- Open the Basic SimpleList story
- change the number of items rendering to 10
- confirm only 10 items shown
- change pageSize to sm 
- confirm only 5 items shown
- move to page 2
- change the number of items rendering to 20
- confirm still on page 2
- move to page 3
- confirm 20 items are being rendered
- Repeat similar steps with search and confirm items shown and pagination is maintained when changing other external props

- All previous functionality works as expected.

### DIFF
--- a/packages/react/src/components/List/SimpleList/SimpleList.jsx
+++ b/packages/react/src/components/List/SimpleList/SimpleList.jsx
@@ -129,9 +129,8 @@ const SimpleList = ({
       const searchFilteredItems = items.filter((item) => {
         if (item.content.value !== '' && item.content.value !== undefined) {
           if (
-            item.content.secondaryValue !== '' &&
-            item.content.secondaryValue !== undefined &&
-            typeof item.content.secondaryValue === 'string'
+            typeof item.content.secondaryValue === 'string' &&
+            item.content.secondaryValue !== ''
           ) {
             return (
               item.content.value.toLowerCase().search(searchTerm.toLowerCase()) !== -1 ||
@@ -164,7 +163,7 @@ const SimpleList = ({
   const maxPage = Math.ceil(numberOfItems / rowPerPage);
 
   /**
-   * Then the items array updates, make sure to update the filteredItems, but also re-apply
+   * When the items array updates, make sure to update the filteredItems, but also re-apply
    * search values if there were any.
    */
   useEffect(() => {
--- a/packages/react/src/components/List/SimpleList/SimpleList.jsx
+++ b/packages/react/src/components/List/SimpleList/SimpleList.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo } from 'react';
+import React, { useState, useMemo, useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
 
 import {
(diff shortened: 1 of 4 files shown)


packages/react/src/components/List/SimpleList/SimpleList.jsx
```diff
@@ -102,21 +102,97 @@ const SimpleList = ({
...
+      setCurrentPageNumber(page);
+      setItemsToShow(currentItemsOnPage);
+    },
+    [filteredItems, rowPerPage]
+  );
+
+  const maxPage = Math.ceil(numberOfItems / rowPerPage);
+
+  /**
+   * Then the items array updates, make sure to update the filteredItems, but also re-apply
```
bjornalm, line 167: nit: spelling?

packages/react/src/components/List/SimpleList/SimpleList.test.jsx
```diff
@@ -266,4 +267,171 @@ describe('SimpleList', () => {
...
+    expect(screen.getByText('Item 5')).toBeVisible();
+    expect(screen.getByText('Item 15')).toBeVisible();
+    expect(screen.getByText('Item 25')).toBeVisible();
+    expect(screen.getByText('Item 35')).toBeVisible();
+    expect(screen.getByText('Item 45')).toBeVisible();
+    expect(screen.getByText('Page 1')).toBeVisible();
+    userEvent.click(screen.getByRole('button', { name: 'Next page' }));
+    expect(screen.getByText('Item 50')).toBeVisible();
+    expect(screen.getByText('Page 2')).toBeVisible();
+    expect(screen.getByRole('button', { name: 'Next page' })).toHaveClass(
```
bjornalm, line 354: nit: I guess .toBeDisabled() isn't working on these buttons?

bjornalm (approved):
Nice job, just questions and minor things

packages/react/src/components/List/SimpleList/SimpleList.jsx
```diff
@@ -102,21 +102,97 @@ const SimpleList = ({
...
+
+  const handleSearch = useCallback(
+    (searchTerm) => {
+      setSearchValue(searchTerm);
+      const searchFilteredItems = items.filter((item) => {
+        if (item.content.value !== '' && item.content.value !== undefined) {
+          if (
+            item.content.secondaryValue !== '' &&
+            item.content.secondaryValue !== undefined &&
+            typeof item.content.secondaryValue === 'string'
```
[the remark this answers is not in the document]
kevinsperrine: yeah, I can remove it. It was just copy/pasted from the inlined onChange function below.

davidicus (approved):

PR #2790, opened 2021-09-01
feat(dashboardeditor): add new render callbacks to allow custom renders of the edit form (and card render) without forcing the Card JSON to change

scottdickerson:
Closes #

**Summary**
While it was nice to have custom renderEditContent props on the Card and renderCardPreview prop on the DashboardEditor, each had a couple limitations that kept it from being useful. 

The DashboardEditor:renderCardPreview prop was useful because we could dynamically change the rendered card in the DashboardEditor without changing the card JSON. However, it forced us to reproduce all the current DashboardEditorCardRenderer code (with our minor customizations) because DashboardEditorCardRenderer did not expose its normal render behavior in a way that could be overridden.

The Card.renderEditContent prop was nice because we could dynamically add additional form elements to the CardEditor for that Card.  The problem is that we had to do this in the JSON that was passed to the DashboardEditor and then keep removing it from the onChange methods of the Card.  We needed a way to get these form elements to the CardEditForm WITHOUT changing the underlying Card JON.

- feat(CardEditor): add new custom render callback to the cardeditor allowing additional changes to be made to the card props (to add custom editor elements) WITHOUT requiring it to be passed in the main card JSON.  This allows us to keep the Card JSON focused on the DashboardGrid version of the props, and not add any additional editor props full time to the Card JSON.
- feat(DashboardEditor): add new custom render callback to allow you to change a card JSON dynamically (see above)
- refactor(DashboardEditorCardRenderer): move all the render methods to the DashboardEditorDefaultCardRenderer so it can be reused
- - fix(DashboardEditorCardRenderer): remove the getValidDataItems and dataItems props (and memoization), notice by searching that they are never being used in the Cards downstream, they're only used in the CardEditor
- feat(DashboardEditorDefaultCardRenderer): expose the new DashboardEditorDefaultCardRenderer so downstream components can use it but choose to override it in certain cases. See the new storybook story for an example

- Verify the new story here: http://127.0.0.1:3000/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-cardeditor--dynamically-add-special-content-to-card-edit-form
 It should add the same kind of EditForm extension that this story adds (http://127.0.0.1:3000/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-cardeditor--table-card-with-custom) but it's doing it dynamically
- Verify that the http://127.0.0.1:3000/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-dashboardeditor--custom-card-preview-renderer story still allows you to render custom card types, but it ALSO allows you to extend the render of existing Card Types by reusing the new component DashboardEditorDefaultCardRenderer

- Verify the existing DashboardEditor/CardEditForm stories still work correctly


packages/react/src/components/DashboardEditor/DashboardEditor.jsx
```diff
@@ -754,8 +762,6 @@ const DashboardEditor = ({
                         isResizable={isCardResizable}
                         i18n={mergedI18n}
                         isSelected={isSelected}
-                        getValidDataItems={getValidDataItems}
-                        dataItems={dataItems}
```
kevinsperrine, line 758: This doesn't seem to be a backwards compatible change. Shouldn't these be left and deprecated with a warning in PropTypes?

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.test.jsx
```diff
@@ -231,19 +231,6 @@ describe('DashboardEditorCardRenderer', () => {
     expect(screen.getByText(/defaultCard/)).toBeInTheDocument();
   });
 
-  it('should call getValidDataItems when passed', () => {
```
kevinsperrine, line 234: Why is this test being removed?

packages/react/src/constants/CardPropTypes.js
```diff
@@ -768,7 +768,9 @@ export const CardPropTypes = {
   resizeHandles: PropTypes.array,
   /** Optional callback function that is passed an onChange function and the original cardConfig object.
    * This allows additional information to be passed to be used in the Card Editor for this type.
-   * You need to return an array of child objects with a header: {title, tooltip: {tooltipText: PropTypes.string}} and content element to render * */
+   * You need to return an array of child objects with a header: {title, tooltip: {tooltipText: PropTypes.string}} and content element to render
+   * DEPRECATED: we recommend doing this dynamically with the CardEditor onRenderCardEditForm property
+   * * */
   renderEditContent: PropTypes.func,
```
kevinsperrine, line 774: Can we deprecate this with the warning function in addition to the comment? That way people get the console error.

kevinsperrine (changes requested):
There seem to be a lot of breaking changes in here. We should deprecate instead and leave the removal/cleanup for the v3 branch.

packages/react/src/components/DashboardEditor/DashboardEditor.jsx
scottdickerson: hmm, I'll check the external props interface, I didn't think I changed anything there (at least I didn't intend to).  The DashboardEditor @kevinsperrine

packages/react/src/constants/CardPropTypes.js
```diff
@@ -768,7 +768,9 @@ export const CardPropTypes = {
   resizeHandles: PropTypes.array,
   /** Optional callback function that is passed an onChange function and the original cardConfig object.
    * This allows additional information to be passed to be used in the Card Editor for this type.
-   * You need to return an array of child objects with a header: {title, tooltip: {tooltipText: PropTypes.string}} and content element to render * */
+   * You need to return an array of child objects with a header: {title, tooltip: {tooltipText: PropTypes.string}} and content element to render
+   * DEPRECATED: we recommend doing this dynamically with the CardEditor onRenderCardEditForm property
```
kevinsperrine:
```suggestion
   * we recommend doing this dynamically with the CardEditor onRenderCardEditForm property
```

Let's remove DEPRECATED since this method can still be used within the new `onRenderCardEditForm`

packages/react/src/components/DashboardEditor/DashboardEditor.jsx
kevinsperrine: Got it. So they were being passed to the `BarChartCard` via the `EditorBarChartCard` (https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2790/files#diff-2ff08df7c609521e34782529133308227cf907b56069ea653c8b6ef6eeafed77L328) but the BarChartCard doesn't actually use them. They are only used by the Form side of the DashboardEditor, not anywhere in the Renderer... 👍

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.jsx
```diff
@@ -288,9 +168,6 @@ const DashboardEditorCardRenderer = React.memo(
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [cardProps.values?.hotspots?.length, cardProps.onFetchDynamicDemoHotspots]);
 
-    if (!isCardJsonValid(cardProps)) {
-      return renderDefaultCard(cardProps);
-    }
```
kevinsperrine, line 293: This is being moved into the new `DashboardEditorDefaultCardRender` component below: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2790/files#diff-1cb56abaf3f9f980a20b247c77f76852c3d99620198c40d494aed49883c3924fR45

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.jsx
```diff
@@ -301,7 +178,8 @@ const DashboardEditorCardRenderer = React.memo(
         onDuplicate,
         onRemove,
         isSelected,
-        onShowImageGallery
+        onShowImageGallery,
+        availableDimensions
```
kevinsperrine, line 182: available dimensions are added, so they can be passed to the `DashboardEditorDefaultCardRender` and be used by the `BarChartCard`.

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.jsx
```diff
@@ -311,42 +189,30 @@ const DashboardEditorCardRenderer = React.memo(
...
+    // If it's an image card we need to update its dynamic hotspots
+    const card = useMemo(
+      () =>
+        cardProps.type !== CARD_TYPES.IMAGE
+          ? cardProps
+          : {
+              ...cardProps,
+              values: {
+                ...cardProps.values,
+                hotspots: cardProps.values?.hotspots?.concat(dynamicHotspots),
```
kevinsperrine, line 207: This seems to be a little different than the old renderImageCard function. Previously, we were concatting the dynamicHotspots before passing to the renderImageCard, but then, the renderImageCard would filter the hotspots. Is this an intentional change?

see: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2790/files#diff-2ff08df7c609521e34782529133308227cf907b56069ea653c8b6ef6eeafed77L102
```
hotspots: props.values?.hotspots?.filter((hotspot) => hotspot.type !== 'dynamic') || [],
```

kevinsperrine (commented):
I think we're all good. It's a lot to follow, but the only thing that seems actually different is the filtering of hotspots on the ImageCard.

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.jsx
scottdickerson: no, this is an error, thanks for catching @kevinsperrine .  I'll make sure to filter those again

scottdickerson: here you go

packages/react/src/constants/CardPropTypes.js
bjornalm: Not sure I understand this. So are we going to keep two different approaches for doing the same thing? Is it not possible for you to use onRenderCardEditForm and still add the modifications somehow?

bjornalm (commented):
Nice work, it is a lot to take in for those not too familiar with the dashboard, but the code looks ok. Just a couple of questions.

packages/react/src/constants/CardPropTypes.js
```diff
@@ -768,7 +768,9 @@ export const CardPropTypes = {
   resizeHandles: PropTypes.array,
   /** Optional callback function that is passed an onChange function and the original cardConfig object.
    * This allows additional information to be passed to be used in the Card Editor for this type.
-   * You need to return an array of child objects with a header: {title, tooltip: {tooltipText: PropTypes.string}} and content element to render * */
+   * You need to return an array of child objects with a header: {title, tooltip: {tooltipText: PropTypes.string}} and content element to render
+   * We recommend doing this dynamically with the CardEditor onRenderCardEditForm property rather than hardcoding it in the Card JSON
```
bjornalm, line 772: Not sure I understand this. So are we going to keep two different approaches for doing the same thing? Is it not possible for you to use onRenderCardEditForm and still add the modifications somehow?

packages/react/src/components/DashboardEditor/DashboardEditorCardRenderer.jsx
kevinsperrine: I think we might be missing a push. I still don't see the filter.

packages/react/src/constants/CardPropTypes.js
scottdickerson: Yes there are two ways, we basically we preserve the old static prop (for non-breaking changes) but recommend using the dynamic prop, was the intention with the comment.  We don't want to deprecate the old prop (because we need it to actually use the dynamic prop) but we would probably want to hide it from downstream consumers in a future release.

packages/react/src/components/DashboardEditor/DashboardEditorDefaultCardRenderer.jsx
```diff
@@ -0,0 +1,103 @@
...
+const DashboardEditorDefaultCardRenderer = ({ card: cardProps, availableDimensions }) => {
+  // For time based cards, we need to deduce the correct interval
+  const timeRangeJSON = find(timeRangeToJSON, ({ range }) =>
+    isEqual(range, cardProps?.dataSource?.range)
+  );
+  const interval = useMemo(() => timeRangeJSON?.interval || 'day', [timeRangeJSON]);
+
+  // if our card isn't valid render the default
+  if (!isCardJsonValid(cardProps)) {
+    return renderDefaultCard(cardProps);
```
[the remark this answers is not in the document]
scottdickerson: sure, sounds good, I'll pop an warning in the console

packages/react/src/components/DashboardEditor/DashboardEditor.story.jsx
```diff
@@ -833,35 +834,38 @@ export const CustomCardPreviewRenderer = () => (
...
           {
             // if you want the resizable handles you need to render the children
             cardProps.children
           }
         </Card>
-      ) : undefined;
+      ) : (
+        // if not a custom card you can use the default card renderer, but you can translate and add your own changes to the card JSON
+        <DashboardEditorDefaultCardRenderer
+          card={{ ...cardProps, title: `my custom title: ${cardProps.title}` }}
```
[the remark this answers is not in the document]
scottdickerson: this is just the story, not the real thing

packages/react/src/constants/CardPropTypes.js
bjornalm: I see, thanks for the clarification.

bjornalm (approved):
I think this looks good, but see @kevinsperrine comment about the missing filter still.

kevinsperrine (approved):
Nice work!

davidicus (approved):

PR #2797, opened 2021-09-02
2620 table components in table cell headers are not readable

### LINKED ISSUE #2620: [Table] [A11Y] Table - components in table cell headers are not readable
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

===== Accessibility checkpoint =====
Test Tool Type: IBM Equal Access Accessibility Checker
Checkpoint: IBM 2.1.1 Keyboard (https://www.ibm.com/able/requirements/requirements/#2_1_1)

=====STEPS TO REPRODUCE =====

Open file storage (https://cloud.ibm.com/cloud-storage/file) or block storage (https://cloud.ibm.com/cloud-storage/block)
Login with ID & password
Use Tab key to navigate on the Filter button, tap Enter or Space to display the filter row
Continue use Tab key to navigate on location combobox
Tap Down arrow key to display the menu options
===== Problem Summary =====
When we use Tab key to navigate on location combobox and use Down arrow key to display the menu options, we cannot see the completed option strings or completed placeholder strings.

It's not friendly to keyboard users, they can not see the completed option strings or placeholder strings.
image

bjornalm:
Closes #2620 

**Summary**
Menu items in the combobox for column filters were not visible in and keyboard users could not take advantage of the hover tooltip. Another problem, that has gotten a separate issue is the truncation of the placeholder text that also leaves the keyboard user without the full text. 

The first issue is solved by allowing the menu items expand beyond the width of the parent combobox. Decided that max-width for the menu should be set by the application if needed.

The second issue is only partly mitigated by removing most of the padding around the controls, hence freeing up more space. See https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2620#issuecomment-910751860

- Added `menuFitContent` & `horizontalDirection` props &  and functionality to the PAL combobox
- Updated the FilterHeaderRow to us the PAL combobox instead of Carbons
- Removed old CSS that wasn't working and got rid of some complexity
- Wrote unit tests

1. Go to https://deploy-preview-2797--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table-statefultable--stateful-example-with-create-save-views
2. Resize the Select column to be really wide and verify that the combobox menu isn't smaller than the combobox itself
3. Resize the Select column to be really thin and verify that the combobox menu expands expands enough to show the menu items without ellipses.
5. Move the Select column to the end of the table and verify that it now expands to the left
6. Check that the above also works in RTL
7. Repeat for story https://deploy-preview-2797--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table-table--table-example-with-create-save-views
8. Go to story https://deploy-preview-2797--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table-statefultable--simple-stateful-example and enable knobs "hasFilter", "hasResize" and "hasColumnSelection" and repeat step 2-7.

Test the other table stories that have filters and verify that they still work the same except for the above changes.


packages/react/src/components/ComboBox/_combo-box.scss
```diff
@@ -15,6 +17,17 @@
     }
   }
 
+  &.#{$iot-prefix}--combobox__menu--fit-content {
+    @include list-box-menu-fits-content();
+  }
+
+  // Default for when fit-content is active and the menu is wider than the input
+  // is that the menu expands to the right. This style makes it expand tp the left
```
bjornalm, line 25:
```suggestion
  // is that the menu expands to the right. This style makes it expand to the left
```

packages/react/src/components/Table/TableHead/FilterHeaderRow/_filter-header-row.scss
```diff
@@ -13,22 +15,12 @@
   }
 }
 
-.#{$iot-prefix}--filter-header-row--header-width {
```
bjornalm, line 16: This class, that probably at some point should have been active if there was a table-header-width, was actually active if there was NO table-header-width. And if even if it would have been the other way around, the `var('--table-header-width')` stuff doesn't work since the variables are within singe quotes. I tested all the affected stories and couldn't find any changes breaking when removing this.

packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
```diff
@@ -191,170 +193,169 @@ class FilterHeaderRow extends Component {
...
-                        }
-                      ).text, // eslint-disable-line react/destructuring-assignment
-                    }}
-                    placeholder={column.placeholderText || 'Choose an option'}
-                    onChange={(evt) => {
-                      this.setState(
-                        (state) => ({
-                          filterValues: {
-                            ...state.filterValues,
-                            [column.id]: evt.selectedItem === null ? '' : evt.selectedItem.id,
```
kevinsperrine, line 270: Was this a bug previously or the result of switching from carbon's combobox to ours?

packages/react/src/globals/_mixins.scss
```diff
@@ -41,3 +42,21 @@
     border-bottom-#{$last-button-side}-radius: $border-radius;
   }
 }
+
+@mixin list-box-menu-fits-content() {
+  .#{$prefix}--list-box__menu {
```
kevinsperrine, line 47: Is this `prefix` b/c you're overriding carbon styles or should it be `iotPrefix`?

kevinsperrine (approved):
Nice work! I just have a couple clarifying questions.

packages/react/src/globals/_mixins.scss
bjornalm: I needed to override the Carbon width

packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
bjornalm: It was a result of switching from carbon's combobox to ours. Not really a drop-in replacement it seemed :-)

bjornalm:
@stuckless I would appreciate your input on this PR. I'm replacing the Carbon Combobox in the filters with our PAL version (which wraps the Carbon Combobox).

packages/react/src/components/ComboBox/ComboBox.story.jsx
```diff
@@ -54,6 +54,12 @@ const props = () => ({
   invalidText: text('Invalid text (invalidText)', 'A valid value is required'),
   size: select('Field size (size)', sizes, undefined) || undefined,
   onChange: action('fired onChange'),
+  menuFitContent: boolean('Menu fits content (menuFitContent)', false),
+  horizontalDirection: select(
+    'Horizontal direction (horizontalDirection)',
+    ['left', 'right'],
```
stuckless, line 60: I would avoid "right" and "left" or else it'll be hard to implement RTL support.  ie is "right" still right in RTL or does right mean left in RTL which is confusing..  Normally we try to use "start" and "end" so that when we enter RTL things that were LEFT (start) now become RIGHT(start) in RTL.

stuckless (changes requested):

packages/react/src/components/ComboBox/ComboBox.story.jsx
bjornalm: Good point, I'll fix

davidicus (approved):

PR #2805, opened 2021-09-03
feat(table): allow lazy-loading of columns

### LINKED ISSUE #2436: [Table] Allow loading state that doesn't understand which columns are configured
### What package is this for?

- [x] React
- [ ] Angular

### Summary

When the useViewManagement feature is configured the table doesn't know which columns will be available while it's loading data.  This means that we should not show columns, but instead show loading states in the column header too.  Imagine the normal table loading state, but each column header also shows a SkeletonText.  The number of columns should possible be configurable to span the whole screen, just so there's not a bump in the table width once the number of columns load, but this is negotiable.

https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--with-no-data

request here.

upcoming release?

date, the better!_

### Available extra resources

kevinsperrine:
Closes #2436 

**Summary**

Previously the isLoading prop only effected the rows, but the header would still display and show columns. This PR allows columns to be an empty array and the isLoading prop will loading skeleton text in place of the column headers. It's customizable via a prop to load X number of columns just like rows.

- added new PropTypes to have loadable columns
- changed `<TableHead>` to only display when the columns array isn't empty.
- changed the `<TableSkeletonWithHeaders>` to match the styles of the `<TableHead>` if no columns are passed and show skeleton text as a placeholder while loading
- add tests cases for both Table/StatefulTable
- added new story to simulate async loading of both columns and data
- updated snapshots

- The new `with async column loading` story should show loading indicators for both columns and rows
- after 3 seconds normal table data should appear
- when the knob for StatefulTable is set all function like filters, sorting, etc, should function as expected after loading has completed.

- prior loading states are maintained. ie. if columns are known and passed to the table, the loading header should not appear, but loading rows will
- No adverse side-effects for features not working after columns/data has been async loaded.

### DIFF
--- a/packages/react/src/components/Table/TableSkeletonWithHeaders/TableSkeletonWithHeaders.jsx
+++ b/packages/react/src/components/Table/TableSkeletonWithHeaders/TableSkeletonWithHeaders.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { DataTable, SkeletonText } from 'carbon-components-react';
+import classnames from 'classnames';
 
 import { settings } from '../../../constants/Settings';
 import deprecate from '../../../internal/deprecate';
@@ -13,6 +14,8 @@ const propTypes = {
   hasRowExpansion: PropTypes.bool,
   hasRowActions: PropTypes.bool,
   rowCount: PropTypes.number,
+  /* used to show X number of loading skeletons when columns are unknown */
+  columnCount: PropTypes.number,
   columns: PropTypes.arrayOf(PropTypes.shape({ id: PropTypes.string.isRequired })),
   // TODO: remove deprecated 'testID' in v3
   // eslint-disable-next-line react/require-default-props
@@ -30,6 +33,7 @@ const defaultProps = {
   hasRowExpansion: false,
   hasRowActions: false,
   rowCount: 10,
+  columnCount: 5,
   columns: [],
   testId: '',
   showExpanderColumn: false,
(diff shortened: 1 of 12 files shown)


packages/react/src/components/Table/TableSkeletonWithHeaders/TableSkeletonWithHeaders.jsx
```diff
@@ -42,42 +46,57 @@ const TableSkeletonWithHeaders = ({
...
-          <SkeletonText />
-        </TableCell>
-      ))}
-      {showExpanderColumn ? (
-        <TableCell data-testid={`${testID || testId}-expander-column`} />
-      ) : null}
-      {hasRowActions ? <TableCell /> : null}
-    </TableRow>
-    {[...Array(rowCount > 0 ? rowCount - 1 : 0)].map((row, index) => (
+}) => {
```
kevinsperrine, line 53: Note this indentation change will be much easier to review if ignore whitespace changes are turned on.

packages/react/src/components/Table/TableSkeletonWithHeaders/_table-skeleton-with-headers.scss
```diff
@@ -5,4 +5,22 @@
...
+    td {
+      background-color: $ui-03;
+    }
+
+    .#{$prefix}--skeleton__text {
+      background: $ui-04;
+
+      &:before {
+        background: $text-03;
+      }
```
kevinsperrine, line 23: These are my made-up styles for a darker version of the skeleton text. This may need some design advice.

packages/react/src/components/Table/TableSkeletonWithHeaders/TableSkeletonWithHeaders.jsx
bjornalm: Thanks, I wouldn't mind having that on by default if possible

bjornalm (approved):
Nice work, it looks good.

davidicus (approved):

davidicus (approved):

PR #2806, opened 2021-09-04
Add angular list builder and table column customization modal, refactor entry points

Donisius:
This should be merged AFTER #2922
Needs to be updated to use the new moveColumn function introduced in #3088


Donisius:
Changes moved to #3265

PR #2807, opened 2021-09-06
fix(table): removed css that did not and should not work

### LINKED ISSUE #1683: [StatefulTable] the column header is truncated with "..." when "width" is set in columns parameter
### What package is this for?

- [x ] React
- [ ] Angular

### Describe the bug

Today I updated to the latest Iot carbon version, I found an issue. As you can see below, the first column (should be "Name") and 3rd column (should be "Description") headers are incomplete. There is not such issue in previous versions, so I am sure this is kind of regression issue. I am using StatefulTable component, and in columns array, I set "width:20%" for first column, and "width:30%" for the 3rd one.  I checked the styles of table header and found that a strange style makes the header label occupies 30% of the column header.

image

### Environment/versions:

- `carbon-addons-iot-react` version: v2.133.0

bjornalm:
Closes #1683

**Summary**
The issue was actually already solved by #2669 so it was no longer reproducible, but there were some leftover css using variables no longer present so I clean that up here.

- Remove CSS that was using custom properties (variables) no longer available

- Verify that there is no CSS max-width on the column header elements with class `bx-table-header-label` in the table stories

- Table stories unchanged

### DIFF
--- a/packages/react/src/components/Table/TableHead/_table-head.scss
+++ b/packages/react/src/components/Table/TableHead/_table-head.scss
@@ -138,8 +138,6 @@
 
     .#{$prefix}--table-header-label {
       height: 100%;
-      min-width: var(--table-header-width);
-      max-width: var(--table-header-width);
       white-space: nowrap;
       overflow: hidden;
       text-overflow: ellipsis;


kevinsperrine (approved):
LGTM

davidicus (approved):

PR #2811, opened 2021-09-07
chore(table): cleanup table stories

### LINKED ISSUE #2034: [Table] Clean up table stories
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

There are a few console warnings in multiple stories. There are also too many. We need to segment up to have stateful, non-stateful, and more playground type stories so we can consolidate the number we have. We should also look to see how we can rename the stories so they are more digestible. 

Putting a high impact label because I think this will help all devs looking to get information on the data table and how to use it.

kevinsperrine:
Closes #2034 

**Summary**

- In an effort to make the Table/StatefulTable stories more helpful, this PR add more knobs, fixes prop errors in console, and adds descriptions to many knobs used by `basic dumb table` and `simple stateful table` stories

- fixed prop-type errors in console
- add descriptions to story knobs in [Basic dumb table story](https://deploy-preview-2811--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--basic-dumb-table), and [Simple Stateful Table story](https://deploy-preview-2811--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--simple-stateful-example).
- added Table type knob to all stories to easily switch between Table/StatefulTable on each story.
- remove a withResize story and merged the knobs for `hasColumnSelectionConfig` into the basic dumb table story

- The stories should be more clear and helpful now
- there should be no prop-type warnings (except testID deprecation)

- n/a


bjornalm:
These changes are good, but I feel like we are missing the opportunity to really take the table stories to a higher professional level. Here is a brain dump on what I think we should do, and then I'll comment on the current stories.

We should take a hard look at each story and ask if the purpose of the story is clear and if the name matches that purpose. Most of the stories have been built incrementally and combined over time which makes the "book" itself incoherent. The two story categories "Table" and "StatefulTable" don't make sense when looking at the stories inside and considering the fact that almost all stories in both categories have a knob to change the table type. 

I think each story, except for the main playground, should tell a coherent story around some settings and data. A combination of settings that was turned into a story in order to fix a bug or prevent regression should be moved to unit/e2e testing instead. The documentation sections could be a good starting point when deciding which stories we should focus on. It might be out of scope here, but it would be nice to have some kind of taxonomy to help us in writing and maintaining the stories. Per story type we set some requirements and that could be useful for other components as well. I've found the following:

A. MAIN PLAYGROUND 
Story is the main playground with as many knobs as possible to cover the largest amount of functionality. The documentation should show the full table documentation.

B. FUNCTIONALITY CONFIGURATION
Story shows a for the consumer **relevant and coherent** combination of properties and data. A story could be about showing "how row expansion works" and all the different combinations using knobs, or a story showing "pagination" with all relevant configuration in knobs. We might have to write more than one story to tell the full tale of some complicated functionality but we can group them like carbon does in their table stories. Clicking the docs should open the relevant section. 

C. EXTERNAL FUNCTIONALITY COMBINATION 
Story code itself has custom code/components showing how to extend the table to get a certain functionality in place, e.g. the "create and save view management" functionality that partly resides outside the table. For these stories might want custom documentation.

D. FOCUSED PLAYGROUND 
Story is a playground to highlight special scenario, limitation or workaround that require interactions outside the table, e.g. 1-watson-iot-table-statefultable--with-pre-filled-search or 1-watson-iot-table-table--with-options-to-explore-column-settings. The scenario must be of interest to the common consumer if we are going to export the story, otherwise we might wanna restrict it to the dev environment.


Below is some more specific feedback from before started thinking about the above:

1. Should we rename story "simple stateful example" to "Stateful table playground" or something since it is becoming the goto story for prop exploration? The same for the non stateful "basic dumb table".
2. Story 1-watson-iot-table-statefultable--simple-stateful-example is using sort function even though `options.hasSort` is not checked
3. Combining `useZebraStyles`  and `options.hasRowExpansion` does not play nice
image
4. Rename story  "with single nested hierarchy" to "with simple nested hierarchy" to better math the prop `hasSimpleNestedHierarchy`
5. In screen resolution 1366 (which is the 2nd most common for "desktop/laptops") a lot of our table stories look too crowded, often due to cell texts wrapping 3-4 lines. The only time we actually would like to show that in our table is probably to demonstrate the different `options.wrapCellText`. We always use 8 or more columns + expander and action columns etc while carbon rarely use more than 6 columns (sometimes 3) in their demos. Perhaps we could reduce the default number of default columns and have a story specifically showing all the column types. A quicker fix is to use `options.wrapCellText` = alwaysTruncate but would look less clean.
image
Story 1-watson-iot-table-statefultable--stateful-example-with-expansion-max-pages-and-column-resize looks particulary bad:
image
6. Use a different and shorter date format than `1973-03-03T09:46:40.000Z`
7. The story 1-watson-iot-table-statefultable--stateful-example-with-expansion-max-pages-and-column-resize could probably be renamed. That is has maxPages doesn't seem like a huge deal.

kevinsperrine:
@davidicus FYI, This one was accidentally merged and closed, because the `size` prop PR was dependent on this one b/c of all the changed stories.

bjornalm:
@kevinsperrine Was any of the feedback addressed?

kevinsperrine:
> @kevinsperrine Was any of the feedback addressed?

No, didn't have the opportunity yet.

PR #2818, opened 2021-09-09
feat(table): add support for size prop to effect header/aggregation footer

### LINKED ISSUE #1986: [Table] `size` prop does not make header any smaller
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

There a Carbon/PAL table `size` prop. I can be used to make the rows and header larger, but doesn't work with "compact" and "short" sizes. Those will shrink the rows, but not the header.

Additionally, if rows have row actions, those buttons enlarge the rows, so they can't shrink either.

This works with the Carbon DataTable directly.

This is not a critical problem, but does make it harder to integrate this table into some "legacy app" pages where other tables/lists are much more compact.

### To Reproduce

Use prop `size: 'small'` or `'compact'`.

Normal size:
image

Small size (notice header doesn't shrink):
image

### Expected behavior

Rows *and* headers shrink in size. Including rows with row actions.

### Environment/versions:

- `carbon-addons-iot-react` version: [e.g. `v2.98`]

kevinsperrine:
Closes #1986 

**Depends on: #2811**

Since this addition basically touched all the stories, I based it on the table cleanup stories PR to prevent a ton of merge conflicts.

**Summary**
- Previously the carbon `size` prop would only effect rows within the table, but now it will also change the size of the aggregation row, or the header. The header will only shrink as small as it's able, so if you have an overflow menu, or sortable columns then it will remain roughly the same default size; however, it will grow larger if you pass the `xl` option.

- Added `size` prop to table
- updated styles to allow various sizes to adjust the header/aggregation footer height properly
- added `size` knob to all stories

- check all the table/stateful table stories changing the size knob on each and confirm the header/rows/aggregation footer all update accordingly. Note the header may not shrink smaller if it has sorting or an overflow menu.

- Confirm no adverse visual side-affects in the stories


bjornalm:
When using the **Table - basic dumb table** story I get an error when checking knob `options.hasMultiSort`:
image

This does not happen in PR https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2811

kevinsperrine:
> When using the **Table - basic dumb table** story I get an error when checking knob `options.hasMultiSort`:
> image
> 
> This does not happen in PR #2811

Nice catch. This is b/c I removed `isSortable` on the object column, so that the header would shrink with lower sizes; however, that caused the bug in the multisort modal to rear its head--it wasn't anticipating multi-sort being turned on with no sortable columns. Fixed the bug in the modal, and adjust the story to add some sortable columns when hasMultiSort is enabled.

bjornalm:
I noticed that the XS rows grow larger to fit the checkbox for multi select rows. In the carbon implementation they do not. Perhaps we wanna address that now that we have the size knob in all the stories

bjornalm (approved):

kevinsperrine:
> I noticed that the XS rows grow larger to fit the checkbox for multi select rows. In the carbon implementation they do not. Perhaps we wanna address that now that we have the size knob in all the stories

It's odd to me that they don't, honestly. I'd assume for "tappable" area requirements for mobile they would _want_ the extra padding in there... I can follow suit with them and remove that padding though.

davidicus (approved):

PR #2824, opened 2021-09-10
Simple fix to hotspot content to check for null values

### LINKED ISSUE #2823: [HotspotContent] If no values are passed it throws an exception
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Bug in the Hotspot Content code where if values is null, it throws a NPE

I made a new story to reproduce and fixed it in my PR

scottdickerson:
Closes #2823 

**Summary**

- If values were null, the HotspotContent was throwing a NPE

- Wrote a story and added a null check

- Verify the new story

- Verify the other Hotspot Content stories haven't changed

### DIFF
--- a/packages/react/src/components/ImageCard/HotspotContent.story.jsx
+++ b/packages/react/src/components/ImageCard/HotspotContent.story.jsx
@@ -169,3 +169,21 @@ Locale.parameters = {
     escapeHTML: false, // needed for greater than less than
   },
 };
+
+export const EmptyValues = () => {
+  return (
+    <Tooltip open direction="right" triggerId="tooltipTrigger" id="tooltip" tooltipId="tooltip">
+      <HotspotContent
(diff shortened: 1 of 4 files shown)


packages/react/src/components/ImageCard/HotspotContent.story.jsx
```diff
@@ -169,3 +169,21 @@ Locale.parameters = {
     escapeHTML: false, // needed for greater than less than
   },
 };
+
+export const EmptyValues = () => {
```
bjornalm, line 173: Do we need a story for this, or is a unit test sufficient?

kevinsperrine (approved):
LGTM. I agree with Bjorn on the story though. A new story doesn't really seem necessary and the test suffices. We're about to start working on creating visual regression tests using Cypress. The empty story seems more suited to that if anything.

PR #2836, opened 2021-09-14
fix(datetimepicker): disable apply for invalid values

### LINKED ISSUE #2832: [DateTimePicker] accepts invalid date ranges
### What package is this for?

- [ x] React
- [ ] Angular

### Describe the bug

Component allows user to apply an invalid date range

### To Reproduce

1. Expand DateTimePicker component
2. Click Custom Range
3. Make sure Relative Radio is checked, put invalid number in the number-input
4. number-input component will display "Number is not valid", but user can still hit apply

### Expected behavior

User should not be able to click apply

### Environment/versions:

- OS: MacOs
- Browser: Chrome
- `carbon-addons-iot-react` version: 2.146.0-next.151

### Additional context

image

upcoming release?

date, the better!_

bjornalm:
Closes #2832

**Summary**
Fix to disabled the apply button when the "last" NumberInput in relative dates is invalid.

- Added check for invalid date and a state to use it 
- Added unit tests

1. Got story ?path=/story/1-watson-iot-datetime-picker--default
2. Expand DateTimePicker component
3. Click Custom Range
4. Make sure Relative Radio is checked
5. Input an invalid number in the number-input, e.g. empty or minus, plus
6. Verify that the apply button is disabled
7. Change to Custom range: Absolute and verify that the apply button is enabled 
8. Click "Back" and verify that the button is enabled when not in custom range mode

Verify that the other DatePicker functionality is unaffected

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -308,6 +308,7 @@ const DateTimePicker = ({
   const [absoluteValue, setAbsoluteValue] = useState(null);
   const [focusOnFirstField, setFocusOnFirstField] = useState(true);
   const [isTooltipOpen, setIsTooltipOpen] = useState(false);
+  const [relativeLastNumberInvalid, setRelativeLastNumberInvalid] = useState(false);
 
   // Refs
   const [datePickerElem, setDatePickerElem] = useState(null);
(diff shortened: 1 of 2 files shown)


kevinsperrine (commented):
Should we also do it for other fields anytime the date is invalid?
Screen Shot 2021-09-14 at 2 11 35 PM

Screen Shot 2021-09-14 at 2 13 13 PM

bjornalm:
> Should we also do it for other fields anytime the date is invalid?
> 

Good catch. TimePickerSpinner is a PAL component so we should try and prevent the user from being able to add `kevin` as a time. Not sure why that isn't already in place, if it turns out to be problematic to guarantee valid input only it should show some validation error. I will create a new issue for that.

kevinsperrine (approved):
Nice work!

davidicus (approved):

PR #2840, opened 2021-09-16
chore(repo): change workflows so test run in parallel & build test app

### LINKED ISSUE #1730: [Repo]Build test app that can spin up in CI to test build
We need to have a way to test down stream consumers build using our package. We need to leverage our test CLI to build an app in our continuous integration tool.

- [ ] It should be able to create the project
- [ ] install our package as dependency
- [ ] run a build without errors
- [ ] Try selecting component on page?

### LINKED ISSUE #2844: [Repo] Break up testing into separate jobs
In order to reduce the time needed to run our checks we need to separate `jest` and `cypress` test runs into their own job inside our Github test action. 

This change will require that we store the coverage artifacts from both runs so we can combine and report to coveralls in its own separate action.

davidicus:
Closes #1730 
Closes #2844 

**Summary**

This PR attempts to set up a test app that will use the built PAL project to build a CRA app in order to test our packages deliverables. 

- Split up the cypress and jest test into separate jobs in order to speed up the testing workflow. 
- Have the result of the coverage report downloaded so our combine action job has access to both and put in its own job
- created a test app workflow that will build the PAL project, yarn link it and use it to try to build a CRA app using the cli tool.


packages/react/src/components/Card/_card.scss
```diff
@@ -5,6 +5,7 @@
 @import './data-state-renderer';
 @import '../../globals/mixins';
 
+// delcare variables
```
davidicus, line 8: Added to trigger github actions for react test

PR #2860, opened 2021-09-20
fix(tabletoolbar): provide aria-label to CarbonTableToolbar

### LINKED ISSUE #2607: [Table] TableToolbar not forwarding aria props
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The Carbon component [accepts aria-label](https://github.com/carbon-design-system/carbon/blob/2a752847635e0fb01ba968e4b899e76aa95447f3/packages/react/src/components/DataTable/TableToolbar.js#L47), which must be unique, but the value is [not forwarded by the Table component](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/a597a20bb477b1378d4bacd1a21c361d5a9b7d8d/packages/react/src/components/Table/TableToolbar/TableToolbar.jsx#L211-L214).
When there are multiple tables on a page, this becomes an a11y issue (as well as when there is a single table, there is no way to customize this aria value).

===== Problem Summary =====
Multiple elements with "region" role do not have unique labels.
Each element with a "region" role must have a unique label.
When there are multiple elements with a "region" role, they must have unique labels that clearly describe the region's content. This includes labelled <section> elements. This label helps people using assistive technologies to quickly find and navigate to each region.

For each element with a "region" role, either set the value of the aria-labelledby attribute of each region to the id value of any visible text that appropriately and uniquely describes that region's content;
OR, use the aria-label attribute to add an invisible, unique, non-empty label to each element with the "region" role.
image

TheSharpieOne:
Closes #2607

**Summary**

- Allows TableToolbar to send aria-label to the underlying carbon-component.

- Add `toolbarLabelAria` to `Table`'s `i18n` prop and forwards it to `TableToolbar`
- `TableToolbar` now pull `toolbarLabelAria` out of `i18n` to pass to the underlying carbon-component

- snapshots

- Table still Renders
- TableToolbar still Renders

### DIFF
--- a/packages/react/src/components/Table/Table.jsx
+++ b/packages/react/src/components/Table/Table.jsx
@@ -426,6 +426,7 @@ export const defaultProps = (baseProps) => ({
     itemSelected: (selectedCount) => `${selectedCount} item selected`,
     rowCountInHeader: (totalRowCount) => `Results: ${totalRowCount}`,
     toggleAggregations: 'Toggle aggregations',
+    toolbarLabelAria: undefined,
     /** empty state */
     emptyMessage: 'There is no data',
     emptyMessageBody: '',
@@ -727,6 +728,7 @@ const Table = (props) => {
               downloadIconDescription: i18n.downloadIconDescription,
               rowCountInHeader: i18n.rowCountInHeader,
               toggleAggregations: i18n.toggleAggregations,
+              toolbarLabelAria: i18n.toolbarLabelAria,
             }}
             actions={{
               ...pick(
--- a/packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
+++ b/packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
@@ -75,6 +75,7 @@ const propTypes = {
     filterAscending: PropTypes.string,
     filterDescending: PropTypes.string,
     toggleAggregations: PropTypes.string,
+    toolbarLabelAria: PropTypes.string,
   }),
   /**
    * Action callbacks to update tableState
@@ -215,6 +216,7 @@ const TableToolbar = ({
       // TODO: remove deprecated 'testID' in v3
       data-testid={testID || testId}
       className={classnames(`${iotPrefix}--table-toolbar`, className)}
+      aria-label={i18n.toolbarLabelAria}
     >
       <TableBatchActions
         // TODO: remove deprecated 'testID' in v3
--- a/packages/react/src/components/Table/TablePropTypes.js
+++ b/packages/react/src/components/Table/TablePropTypes.js
@@ -170,6 +170,7 @@ export const I18NPropTypes = PropTypes.shape({
   closeMenuAria: PropTypes.string,
   clearSelectionAria: PropTypes.string,
   batchCancel: PropTypes.string,
+  toolbarLabelAria: PropTypes.string,
   /** String 'items selected' or function receiving the selectedCount as param:
    * (selectedCount) => `${selectedCount} items selected` */
   itemsSelected: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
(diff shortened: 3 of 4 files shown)


packages/react/src/components/Table/Table.jsx
```diff
@@ -426,6 +426,7 @@ export const defaultProps = (baseProps) => ({
     itemSelected: (selectedCount) => `${selectedCount} item selected`,
     rowCountInHeader: (totalRowCount) => `Results: ${totalRowCount}`,
     toggleAggregations: 'Toggle aggregations',
+    toolbarLabelAria: '',
```
bjornalm, line 429: Why default to empty string? Do you need a default at all here?

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -215,6 +216,7 @@ const TableToolbar = ({
       // TODO: remove deprecated 'testID' in v3
       data-testid={testID || testId}
       className={classnames(`${iotPrefix}--table-toolbar`, className)}
+      aria-label={i18n.toolbarLabelAria || secondaryTitle ? `${secondaryTitle} Toolbar` : undefined}
```
davidicus, line 219: Hi @TheSharpieOne. Thanks for the PR! Since this could change aria-lables that are out in the wild I think we should just use undefined if `toolbarlabelAria` is not present.

packages/react/src/components/Table/Table.jsx
TheSharpieOne: Following the established patterned, See L432 (3 lines below)

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
TheSharpieOne: Since aria-label could _not_ be specified, the _only_ value in the wild would be `"data table toolbar"`, which is not a sane default. So it's not like the default truly is `undefined`. Default values should be useful so that users don't need to specify every prop every time (only needing to use props such as this for specials cases (e.g. i18n)).
Look at the changed snapshots and tell me this was not an improvement in a11y for the default case providing `secondaryTitle`.
I can mark this as a breaking change if you would like (carbon react doesn't seem to believe changes like this are considered breaking.)

bjornalm: It is an improvement, and I'll let David answer your question, but it is also a known problem for some of our clients that changes to default labels and roles etc are not considered breaking since they are sometimes used in automated tests that fail after the change.

packages/react/src/components/Table/Table.jsx
bjornalm: I see, good that you noticed the default for `emptyMessageBody` but that is not a pattern we want to use here (we should change `emptyMessageBody` to use `undefined`). The best option, if you are going with the breaking change, is to use `undefined` as default value here as well.

TheSharpieOne: I updated `toolbarLabelAria` to default to `undefined`. I also changed the commit message to indicate it is a breaking change. Depending on how breaking changes are handled in this repo / how often major release are released, I may create an additional PR that doesn't change the default (just to get that out more quickly. I still believe this change to the default is better and should be used.)

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
TheSharpieOne: _rant_
Yes, I believe it is a breaking change as well. It will definitely break snapshots (as seen by snapshots changing in the PR.) I just wasn't sure if this repo was more aligned with carbon react in it's definition of a breaking change since it's also within the carbon-design-system org.
We have had carbon react break us many times. Here are the most recent:
[changes that prevent onChange callbacks from be called](https://github.com/carbon-design-system/carbon/pull/8018) which [break validation](https://github.com/carbon-design-system/carbon/issues/9337).
[changes that add internal validation which affect component rendered output](https://github.com/carbon-design-system/carbon/pull/9132) that cause visual changes.
_/rant_

I updated the commit message to indicate it is a breaking change. I included mitigation steps for users to preserve the current functionality to avoid changes.

davidicus:
@TheSharpieOne I would be more than happy to include this in our next major release but that release date is still quite a ways off in the future. If this is something you will want to consume in the near term we should leave the defaults as undefined. If you are ok to wait for the next major release then we will want to open this against the v3 branch.

davidicus (approved):

PR #2876, opened 2021-09-23
feat(empty-state): add angular empty state component

Donisius:


packages/angular/src/empty-state/empty-state.component.ts
```diff
@@ -0,0 +1,142 @@
...
+
+@Component({
+  selector: 'ai-empty-state',
+  template: `
+    <div class="iot--empty-state">
+      <div class="iot--empty-state--content">
+        <ng-container *ngIf="icon !== 'no-icon'">
+          <ng-container *ngIf="isTemplate(icon)" [ngTemplateOutlet]="$any(icon)"></ng-container>
+          <ng-container *ngIf="!isTemplate(icon)" [ngSwitch]="icon">
+            <empty-state-no-results-icon
```
zvonimirfras, line 11: Can we just (somehow) import and use svgs directly instead?

packages/angular/src/table/head/table-head.component.ts
```diff
@@ -43,8 +43,13 @@ import { TableHead } from 'carbon-components-angular';
             [attr.rowspan]="column.rowSpan"
             [filterTitle]="getFilterTitle()"
             (sort)="sort.emit(i)"
-            [class]="column.className"
+            [class]="column.className + ' iot--table-head--table-header'"
```
zvonimirfras, line 46: would one of these make more sense in ngClass?

packages/angular/src/table/table-model.class.ts
```diff
@@ -3,6 +3,13 @@ import { Subject } from 'rxjs';
 
 export type HeaderType = number | 'select' | 'expand';
 
+export class AITableHeaderItem extends TableHeaderItem {
+  /**
+   * Defines the alignment of the the header item and the column below it.
+   */
+  alignment: 'start' | 'center' | 'end';
```
zvonimirfras, line 10: and defaults to?

packages/angular/src/table/table.stories.ts
```diff
@@ -38,10 +42,27 @@ simpleModel.setData([
   ],
 ]);
 
+emptyModel.setHeader([
```
zvonimirfras, line 45: empty model is not really empty? 😛

zvonimirfras (changes requested):
Seems like a bunch of changes doesn't belong here. Some PR still needs to be merged before this?

Would it make sense to make empty state declarative?

Table looks good. Can this go in after #3023 since scss is moved out to react package and then built into styles package?

packages/angular/src/empty-state/empty-state.component.ts
```diff
@@ -0,0 +1,142 @@
...
+  @Input() containerClass: string;
+
+  /**
+   * Title of empty state
+   */
+  @Input() title: string;
+  /**
+   * Description of empty state
+   */
+  @Input() body: string;
```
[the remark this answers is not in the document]
Donisius: body and title can be in ng-content, but then

packages/angular/src/table/head/table-head.component.ts
Donisius: not sure, neither are conditional

packages/angular/src/empty-state/empty-state.component.ts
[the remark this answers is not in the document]
Donisius: nvm I agree

davidicus:
@zvonimirfras have your requested changes been addressed. I would like to get these PRs either merged or closed out as they are getting a bit old.

packages/angular/src/empty-state/empty-state.stories.ts
```diff
@@ -0,0 +1,100 @@
...
+        </ai-empty-state-action>
+      </ai-empty-state>
+		`,
+    props: {
+      icon: select(
+        'icon',
+        ['error', 'error404', 'not-authorized', 'no-results', 'success', 'default', 'no-icon'],
+        'default'
+      ),
+      title: text('title', 'You don’t have any [variable] yet'),
```
zvonimirfras, line 43: That apostrophe 👀

packages/angular/src/table/table-model.class.ts
```diff
@@ -98,7 +116,7 @@ export class AITableModel implements PaginationModel {
   /**
    * Contains information about the header cells of the table.
    */
-  protected header: TableHeaderItem[][] = [[]];
+  protected header: AITableHeaderItem[][] = [[]];
```
zvonimirfras, line 119: Hold on, this looks a bit breaky... Could we keep it as is?

If they don't need alignment, they just use the cca default.

zvonimirfras (changes requested):

packages/angular/src/empty-state/empty-state-action-wrapper.component.ts
```diff
@@ -0,0 +1,11 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'ai-empty-state-action',
+  template: `
+    <div class="iot--empty-state--action">
```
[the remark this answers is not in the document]
Donisius: It isn't able to apply the margins without the div

packages/angular/src/empty-state/empty-state-secondary-action-wrapper.component.ts
```diff
@@ -0,0 +1,11 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'ai-empty-state-secondary-action',
+  template: `
+    <div class="iot--empty-state--link">
```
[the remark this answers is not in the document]
Donisius: It isn't able to apply the margins without the div

packages/angular/src/table/table-model.class.ts
Donisius: It would result in a type error when you check for alignment in the template if it was left as is. Can we maybe do `TableHeaderItem[][] | AITableHeaderItem[][]` and do some kind of type checking in the template?

zvonimirfras: Perhaps... It wouldn't be a breaking change in that case :)

zvonimirfras (approved):

davidicus (approved):

PR #2898, opened 2021-09-24
fix(table): aggregation column placement

### LINKED ISSUE #2859: [table] aggregation data placed in wrong columns
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The aggregation column label ends up in the first column which is not wide enough to display any text and the actual sum of data ends up in the incorrect column (to the left of where it suppose to be). 

### To Reproduce

Use the following settings:
```
const tableData = addRowActions(getTableData(20, words, selectData));
    tableData[0].children = [
      { ...tableData[0], id: 'row-0-A' },
      { ...tableData[0], id: 'row-0-B' },
    ];
    tableData[1].children = [
      { ...tableData[1], id: 'row-1-A', children: [{ ...tableData[1], id: 'row-1-A-1' }] },
    ];

    const columnsOverflowMenu = getTableColumns(selectData).map((col) => ({
      ...col,
      align: 'center',
    }));
    columnsOverflowMenu[0].width = '300px';
    columnsOverflowMenu[1].width = `${MIN_COLUMN_WIDTH}px`;
    columnsOverflowMenu[1].isSortable = false;
    columnsOverflowMenu[2].width = `${DEFAULT_COLUMN_WIDTH}px`;

<Table
        testId="snapshot-2"
        columns={columnsOverflowMenu}
        data={tableData}
        useZebraStyles={false}
        lightweight
        options={{
          hasAggregations: true,
          hasColumnSelection: true,
          hasColumnSelectionConfig: false, 
          hasFilter: true,
          hasPagination: true,
          hasResize: true,
          hasRowEdit: false,
          hasRowExpansion: false,
          hasRowActions: false,
          hasRowNesting: true,
          hasRowSelection: 'single',
          hasSearch: true,
          hasSort: true,
          hasSingleRowEdit: true,
          useAutoTableLayoutForResize: false,
          wrapCellText: 'always',
          size: 'md',
          shouldExpandOnRowClick: false,
        }}
        view={{
          aggregations: {
            label: 'Total:',
            columns: [
              {
                id: 'number',
                align: 'center', 
                isSortable: false,
              },
            ],
          },
          pagination: { isItemPerPageHidden: false },
          table: {
            isSelectAllSelected: false,
            selectedIds: ['row-4'],
            expandedIds: ['row-0', 'row-1', 'row-1-A'],
            sort: {
              columnId: 'select',
              direction: 'ASC',
            },
          },
          toolbar: {
            activeBar: 'column',
            search: { defaultExpanded: true },
          },
        }}
      />
```

### Expected behavior

The label "total" should appear in the second column (the string column) and not be wrapped.

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context

image

upcoming release?

date, the better!_

kevinsperrine:
Closes #2859 

**Summary**

- When using multiselect and nested rows at the same time, the aggregation label was in the wrong place. This fixes that by using colspan to position the label based on various prop combinations.

- use colspan instead of cells to position label
- add tests to confirm

- Go to [`basic dumb table` story](https://deploy-preview-2898--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--basic-dumb-table)
- turn on pagination
- experiment with all combinations of `hasRowNesting`, `hasRowExpansion`, and `hasRowSelection`. The 'Total' label should also be aligned with the pagination below.

- aggregation placement is still correct on other aggregation stories/settings

### DIFF
--- a/packages/react/src/components/Table/TableFoot/TableFoot.jsx
+++ b/packages/react/src/components/Table/TableFoot/TableFoot.jsx
@@ -59,7 +59,8 @@ const TableFoot = ({
 
   const hasMultiSelect = hasRowSelection === 'multi';
   const hasExpandOrNest = hasRowExpansion || hasRowNesting;
-  const colSpan = hasExpandOrNest || hasMultiSelect ? 2 : 1;
+  const labelColSpan =
+    hasMultiSelect && hasExpandOrNest ? 3 : hasExpandOrNest || hasMultiSelect ? 2 : 1;
 
   return (
     <tfoot className={`${iotPrefix}-table-foot`} data-testid={testId}>
@@ -76,13 +77,10 @@ const TableFoot = ({
                 className={`${iotPrefix}-table-foot--label`}
                 data-testid={cellTestId}
                 key={cellKey}
-                colSpan={colSpan}
+                colSpan={labelColSpan}
               >
                 {aggregations.label}
               </TableCell>
-              {colSpan === 2 && hasMultiSelect && hasExpandOrNest ? (
-                <TableCell key="spacer-cell" />
-              ) : null}
             </Fragment>
           ) : aggregated ? (
             <Fragment key={`aggregated-cell-fragment-${index}`}>
--- a/packages/react/src/components/Table/TableFoot/TableFoot.jsx
+++ b/packages/react/src/components/Table/TableFoot/TableFoot.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { TableCell, TableRow } from 'carbon-components-react';
 import classnames from 'classnames';
@@ -12,6 +12,7 @@ const propTypes = {
     hasRowExpansion: PropTypes.bool,
     hasRowSelection: PropTypes.oneOf(['multi', 'single', false]), // TODO: don't duplicate this one
     hasRowActions: PropTypes.bool,
+    hasRowNesting: PropTypes.bool,
   }),
   tableState: PropTypes.shape({
     aggregations: PropTypes.shape({
@@ -50,17 +51,19 @@ const defaultProps = {
 
 const TableFoot = ({
   testId,
-  options: { hasRowExpansion, hasRowSelection, hasRowActions },
+  options: { hasRowExpansion, hasRowSelection, hasRowActions, hasRowNesting },
   tableState: { aggregations, ordering },
   showExpanderColumn,
 }) => {
   const visibleColumns = ordering.filter((col) => !col.isHidden);
 
+  const hasMultiSelect = hasRowSelection === 'multi';
+  const hasExpandOrNest = hasRowExpansion || hasRowNesting;
+  const colSpan = hasExpandOrNest || hasMultiSelect ? 2 : 1;
+
   return (
     <tfoot className={`${iotPrefix}-table-foot`} data-testid={testId}>
-      <TableRow>
-        {hasRowExpansion ? <TableCell key="row-expansion" /> : null}
-        {hasRowSelection === 'multi' ? <TableCell key="row-selection" /> : null}
+      <TableRow key="aggregate-row">
         {visibleColumns.map((orderedCol, index) => {
           const aggregated = aggregations.columns.find((col) => orderedCol.columnId === col.id);
           const isLabelCell = !aggregated && index === 0;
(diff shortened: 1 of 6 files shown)


bjornalm:
image

To reproduce:
1. Go to basic dumb table story
2. Set table type to StatefulTable
3. enable hasPagination & hasColumnSelection
4. move the number column all the way to the left

bjornalm (changes requested):
See screenshot

kevinsperrine:
> See screenshot

got it. adjusted so it matches next. deploying now.

packages/react/src/components/Table/TableFoot/TableFoot.jsx
```diff
@@ -50,46 +51,59 @@ const defaultProps = {
...
+            <Fragment key="label-cell-fragment">
+              <TableCell
+                className={`${iotPrefix}-table-foot--label`}
+                data-testid={cellTestId}
+                key={cellKey}
+                colSpan={colSpan}
+              >
+                {aggregations.label}
+              </TableCell>
+              {colSpan === 2 && hasMultiSelect && hasExpandOrNest ? (
```
bjornalm, line 83: Hmm, if either hasMultiSelect or hasExpandOrNest is true, then I don't think we don't need to test for colSpan === 2. But it is also a bit confusing that we calculate the colspan and at the same time introduce a spacer cell that we just merge.  Is that needed?

packages/react/src/components/Table/TableFoot/TableFoot.jsx
```diff
@@ -50,46 +51,59 @@ const defaultProps = {
...
-  options: { hasRowExpansion, hasRowSelection, hasRowActions },
+  options: { hasRowExpansion, hasRowSelection, hasRowActions, hasRowNesting },
   tableState: { aggregations, ordering },
   showExpanderColumn,
 }) => {
   const visibleColumns = ordering.filter((col) => !col.isHidden);
 
+  const hasMultiSelect = hasRowSelection === 'multi';
+  const hasExpandOrNest = hasRowExpansion || hasRowNesting;
+  const colSpan = hasExpandOrNest || hasMultiSelect ? 2 : 1;
```
bjornalm, line 62: nit: call it labelCellColspan to clarify where we gonna use it?

packages/react/src/components/Table/TableFoot/TableFoot.jsx
kevinsperrine: sure, I can merge it into the colSpan.

packages/react/src/components/Table/TableFoot/TableFoot.jsx
```diff
@@ -50,46 +51,57 @@ const defaultProps = {
...
 
           return isLabelCell ? (
-            <TableCell
-              className={`${iotPrefix}-table-foot--label`}
-              data-testid={cellTestId}
-              key={cellKey}
-            >
-              {aggregations.label}
-            </TableCell>
+            <Fragment key="label-cell-fragment">
```
bjornalm, line 75: We can remove the fragment now

packages/react/src/components/Table/TableFoot/TableFoot.jsx
```diff
@@ -50,46 +51,57 @@ const defaultProps = {
...
-              })}
-              align={aggregated.align ? aggregated.align : undefined}
-              data-testid={cellTestId}
-              key={cellKey}
-            >
-              {aggregated.value}
-            </TableCell>
+            <Fragment key={`aggregated-cell-fragment-${index}`}>
+              {index === 0 && (hasMultiSelect || hasExpandOrNest) ? (
+                <TableCell colSpan={hasMultiSelect && hasExpandOrNest ? 2 : 1} />
```
bjornalm, line 88: nit: this could also be simplified into just to having a single condition to add the extra table cell or not.

bjornalm (approved):
Nice work, I left two small comments.

packages/react/src/components/Table/TableFoot/TableFoot.jsx
kevinsperrine: removed.

packages/react/src/components/Table/TableFoot/TableFoot.jsx
kevinsperrine: given that all of this is going to be reworked to account for aggregations in the first column or potentially multiple labels, I don't really think that's important right now.

davidicus (approved):

PR #2910, opened 2021-09-29
2827 table create visual regression test fixes

### LINKED ISSUE #2827: [Table] Create visual regression test
Visual regression test are needed on the table component. In order to limit the amount of test we run we will want to create a Table instance that has as many permutations as possible. If two different instances are needed to capture all the functionality let's include them both in one test that snapshots both at once. 

Since there could be differences between headless and headed instance of the test runner you the `onlyOn` method from `@cypress/skip-test` package to skip visual regression test  during headed runs. CardCodeEditor e2e test have example of this.

bjornalm:
Closes #2827

Summary
Adds visual regression testing to the table. I used 3 table instances in separate visual snapshots to cover as much UI as possible. Separate snapshots was needed since I needed to capture UI elements that are only visible on click/hover.

- refactored table test helper code so that it could be share between tests
- added tests (See attached excel file for what is covered in each table instance)
[IBM PAL table visual regression testing.xlsx](https://github.com/carbon-design-system/carbon-addons-iot-react/files/7205615/IBM.PAL.table.visual.regression.testing.xlsx)
- FIX where I had to set the environment HEIGHT to the value of the width in order to get the viewport working. This also affected the CardCodeEditor snap which actually got the width defined by the test (minus 1px which seems to be happening in some tests). This is still very confusing, but at least it is working now.

Do we capture enough and the right stuff?

Regression Test (how to make sure this PR doesn't break old functionality)
Old unit tests still working as expected


packages/react/src/components/Table/Table.test.jsx
```diff
@@ -1369,7 +1298,7 @@ describe('Table', () => {
       />
     );
 
-    expect(screen.queryAllByTestId('table-head--overflow').length).toBe(5);
+    expect(screen.queryAllByTestId('table-head--overflow').length).toBe(8);
```
davidicus, line 1301: Should we have any changes to existing unit test like this?

davidicus (approved):
Looks good other than the changed test case.

packages/react/src/components/Table/Table.test.jsx
bjornalm: In general no, but this stems from that I wanted to share the test data between the unit and the e2e tests and for the e2e I needed 3 additional columns so I added those to the `getColumns` function in the Table.test.helpers.jsx. Hence we went from 5 to 8 columns in our tests and I had to adjust this one unit test.

packages/react/src/components/Table/Table.test.jsx
```diff
@@ -173,7 +102,7 @@ describe('Table', () => {
   const expandedData = [
     {
       rowId: 'row-1',
-      content: <RowExpansionContent rowId="row-1" />,
+      content: <div style={{ padding: '8px' }}>Expnded content</div>,
```
kevinsperrine, line 105:
```suggestion
      content: <div style={{ padding: '8px' }}>Expanded content</div>,
```

kevinsperrine (approved):
Looks good. I wonder why snapshot 1 failed and the others didn't?

packages/react/src/components/Table/Table.test.jsx
bjornalm: Thanks, good catch. I need to work on my spelling.  

Regarding the failed test, I'm looking into it but I think it is related to the "load more" PR that we just merged.

bjornalm: Ahh.. no of course. It was your fix to the Aggregation column. In the old screenshot it was in the wrong column, that is how I discovered it in the first place.

PR #2920, opened 2021-10-01
fix(hierarchy-list): reset page (as necessary) when pageSize changes

### LINKED ISSUE #2918: [HierarchyList] pageSize doesn't reset when not enough items to fill current page
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

If the pageSize prop changes while you're on a page higher than the total pages available you get stock on a page and cannot get back to the items.

### Specific timeline issues / request
list-reset-pagination
s

kevinsperrine:
Closes #2918 

**Summary**

- Previously, if you paged through the results to a higher page, but then changed the pagination (or removed it) You could get stuck on the higher page and be unable to get back to the list items.

- add useEffect to check current page and max page and reset to page 1 if necessary.

- Go the HierarchyList with large number of items story
- change pageSize to small
- change to page 2
- change pageSize to undefined
- you should return to page 1 with all results

- no adverse effects on other stories


davidicus:
If you change the knob to page size `xl` then go to page 33 and then change the prop to `lg` the results go back to 0-9 but you remain on page 33. you are still able to go to page 32 and that seems to reset to the right number but def unexpected.

bjornalm (approved):
Nice work. Perhaps you want to add a unit test for this?

kevinsperrine:
> Nice work. Perhaps you want to add a unit test for this?

Yeah, I'm going to. These are just small issues I've come across adding the virtual list stuff. I'll add the unit tests in that PR when I open it to cover all the basis with both regular and virtual list content.

kevinsperrine:
> If you change the knob to page size `xl` then go to page 33 and then change the prop to `lg` the results go back to 0-9 but you remain on page 33. you are still able to go to page 32 and that seems to reset to the right number but def unexpected.

Adjusted this so that it stays on page 33 and updates the list to show the correct items for page 33.

davidicus (approved):

PR #2921, opened 2021-10-01
fix(hierarchy-list): maintain list search state on drag and drop

### LINKED ISSUE #2917: [HierarchyList] list resets when dragging-and-dropping with active search
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

When you drag and drop in a HierarchyList with an active search the list will reset and the search is no longer reflected.

list-reset-search

kevinsperrine:
Closes #2917 

**Summary**

- Previously, if you searched the list, and then drag-and-drop the list would reset

- call search handler from handleMove

- Go to the HierarchyList with large number of items story
- change pageSize to small
- search for 5
- drag 45 to the top of the list
- the search state should be maintain and 45 is now at the top of the list.

- no adverse effects to other stories


packages/react/src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -359,9 +359,12 @@ const HierarchyList = ({
    * the total filtered array. The next category's children then needs to
    * be searched in the same fashion.
    * @param {String} text keyed values from search input
+   * @param {Array} currentItems the current state of items.
+   *                             Used to maintain state when the list is updated (drag and drop)
+   *                             and there's a current search value
    */
-  const handleSearch = (text) => {
-    const tempFilteredItems = searchForNestedItemValues(items, text);
+  const handleSearch = (text, currentItems = null) => {
```
bjornalm, line 366: Nit: Wouldn't it make more sense that the signature had a required  "haystack" parameter? Now it can be either implicit or explicit.

bjornalm (approved):
Nice work!

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
kevinsperrine: sure. updated.

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -359,9 +359,12 @@ const HierarchyList = ({
    * the total filtered array. The next category's children then needs to
    * be searched in the same fashion.
    * @param {String} text keyed values from search input
+   * @param {Array} currentItems the current state of items.
+   *                             Used to maintain state when the list is updated (drag and drop)
```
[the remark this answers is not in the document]
kevinsperrine: shortened.

PR #2927, opened 2021-10-05
2891 hierarchy list drag performance

### LINKED ISSUE #2891: [HierarchyList] poor reorder performance with lots of items
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

If there are a lot of items in the list, dragging & dropping them becomes very laggy.

### To Reproduce

1. Create a list with lots of items (> 1000) 
2. Try to reorder an item

kevinsperrine:
Closes #2891 

**Summary**

- With a large list the drag and drop performance is very slow. This PR adds the ability to virtualize the content of the list either via an override pattern using the new VirtualListContent component, or by simply using the `isVirtualList` content to use the default virtual list.

- use react-window to create new VirtualListContent component
- export ListContent and VirtualListContent components for use in override props
- add `isVirtualList` props to HierarchyList and List
- copied jest/cypress tests and set `isVirtualList` list prop to confirm backwards compatibility even when using a virtual list
- minor style adjustments to handle absolutely positioned list elements when using virtual list

- HierarchyList and List stories should function the same with the `isVirtualList` knob turned on.

- There should be no backwards compatibility issues as all the functionality is encapsulated in the new prop, but double check the stories just to make sure.


bjornalm:
The selected item is scrolled out of view. Go to story https://deploy-preview-2927--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-hierarchylist--with-default-selected-id and enable `hasVirtualList`. 
1. Scroll down and expand 'Washington Nationals'
2. Scroll to the bottom and select 'Max Scherzer'
3. The list gets scrolled up and we don't se the selected item

bjornalm:
In this story the overflow menu floats on top of list when hasVirtualList is true https://deploy-preview-2927--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-list--with-row-actions-multiple

That is not the case for the old version in next BUT in my opinion it is actually a real improvement.

PR #2932, opened 2021-10-06
2929 upgrade carbon

### LINKED ISSUE #2437: [Card][A11Y] Info tooltips remain open when tabbing through
### What package is this for?
React

### Describe the bug

When using tab to navigate through cards, the info icon tooltip remains open. 
https://carbon-addons-iot-react.com/?path=/story/watson-iot-card--with-ellipsed-title-tooltip-external-tooltip
Result on a large dashboard can have impact for a11y users:
Screen Shot 2021-06-29 at 10 31 23 AM

Note that this does not happen with the tooltip component itself at https://carbon-addons-iot-react.com/?path=/story/watson-iot-tooltip--default-bottom

Graphite issue:
https://jsw.ibm.com/browse/GRAPHITE-35449
Health issue:
https://github.ibm.com/maximo/Asset-Health-Insight/issues/5787

### LINKED ISSUE #2929: [Repo] upgrade Carbon to 10.47/7.47

kevinsperrine:
Closes #2929 
Closes #2437

**Summary**

## 10.40
- MenuButtons are now opened in a portal, so tests using `container` to query for classes needed to be switched to `document.body`

## 10.41
- Pagination buttons were reduced in size causing style breaks
- - `.bx--pagination__button` was reduced from `3rem` to `2.5rem`
- - added `size` parameter to `SimplePagination` and defaulted it to `lg`

## 10.42
- No new test failures

## 10.43
- `title` attributes were added to the selected item in Dropdowns, so tests using `screen.getByRole('button', { name: 'Size Small (4x1)' })` have now changed, b/c the title attribute doesn't match the text content that it used to `screen.getByRole('button', { name: 'Size Small (4x1) Open menu' })` in one case
- Custom dropdown-based components (SimpleIconDropdown, ColorDropdown, etc) will now have a `title` attribute that is `[Object object]`, because carbon is now using `itemToString` to add a title to the dropdown. This doesn't prevent the dropdowns from working, but they become less accessible. An issue has been opened requesting the ability to render the selected item with `itemToElement` instead of always using `itemToString` 

## 10.44
- The prop `wrapperClassName` for Checkbox will be deprecated in V11 in favor of `className`. `className` will then be placed on the outer wrapper.
- The handleClick prop for ClickableTile has been deprecated in favor of onClick. It will be removed in the next major release.

## 10.45
- no new test failures

## 10.46
- no new test failures

## 10.47
- The `iconDescription` prop is being deprecated in favor of other props. On the `Modal`, it has been replaced with `closeIconLabel`, and on the `TimePickerSelect` it has been replaced with `aria-label`. https://github.com/carbon-design-system/carbon/issues/9624

- as with all carbon upgrades we check all stories
- ~~Confirm expanded [SideNav item collapse when in isRail mode](https://deploy-preview-2932--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-sidenav--side-nav-component)~~
- - ~~Actually, this one doesn't seem to be working in the preview, but it definitely works locally. Their fix is using `child.type.name` and I'm betting that's getting mangled by minification. Posted a comment on the [original issue](https://github.com/carbon-design-system/carbon/issues/9725#issuecomment-965164158)~~.
- Confirm tabbing through [cards with tooltips closes the tooltip](https://deploy-preview-2932--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card--with-ellipsed-title-tooltip-external-tooltip)

- as with all carbon upgrades we check all stories
- with specific attention paid to:
- - any items using pagination to make sure the buttons are the correct size
- - SimpleIconDropdown, IconDropdown, ColorDropdown (any dropdowns hijacking `itemToString`


packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/__snapshots__/ThresholdsFormItem.story.storyshot
```diff
@@ -89,6 +89,22 @@ exports[`Storybook Snapshot tests and console checks Storyshots 2 - Watson IoT E
...
+                      }
+                    }
+                  >
+                    <div
+                      className="iot--icon-dropdown__item-border"
+                    >
+                      <ForwardRef(Warning24) />
+                    </div>
+                  </div>
+                }
```
kevinsperrine, line 107: Carbon started using `itemToString` to add a title to the currently selectedItem in a dropdown; however, this breaks our current implementation of IconDropdowns. I've opened an issue with them to add another prop to allow us to render the selectedItem as an element instead of always calling the `itemToString` https://github.com/carbon-design-system/carbon/issues/10038

bjornalm: FYI. I think we have had that conversation with them before and they didn't want to support that functionality because they didn't agree with the design use case, but let's hope they change their minds.

packages/react/src/components/List/ListContent/ListContent.jsx
```diff
@@ -157,17 +157,7 @@ const ListContent = ({
...
-                  if (isSelectable && isCheckboxMultiSelect) {
-                    // When combing checkboxes with selectable rows (isSelectable) we are
-                    // getting click events from the ListItemWrapper before this handler is called.
-                    // Therefor we stop this one and rely on the ListItem onSelect handler to handle
-                    // the selection. If we don't the handleSelect callback will be called multiple times.
-                    evt.stopPropagation();
-                  } else {
-                    handleSelect(item.id, parentId);
-                  }
-                }}
```
kevinsperrine, line 170: carbon seems to have fixed the double-event firing that was happening when clicking the checkbox, so this stopProp isn't needed anymore. You can confirm with: 

1. Go to story path=/story/1-watson-iot-list--with-checkbox-multi-selection
2. Verify that the multi selection works by clicking on the row, clicking the checkbox and using enter key when row is focused.
3. Verify the same but with knob isVirtualList checked
4. Go to story path=/story/1-watson-iot-list--with-checkbox-multi-selection-and-hierarchy
5. Repeat step 2-3 but also verify that child / parent selections works in BOTH directions including the indeterminate state
```

File: packages/react/src/components/ImageGalleryModal/ImageGalleryModal.jsx [hunk 3]
```diff
@@ -171,7 +171,7 @@ const ImageGalleryModal = ({
           secondaryButtonText={modalSecondaryButtonLabelText}
           modalHeading={deleteModalTitleText(selectedImage?.id)}
           size="xs"
-          iconDescription={modalCloseIconDescriptionText}
```
kevinsperrine, line 174: iconDescription is being deprecated in favor of other attributes. On the Modal it's the aria-label.

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -30558,9 +30856,6 @@ Map {
       "labelText": Object {
         "type": "node",
       },
-      "listFiles": Object {
-        "type": "bool",
-      },
```
kevinsperrine, line 30563: Unused. https://github.com/carbon-design-system/carbon/pull/9915

kevinsperrine:
> Not looking forward to having to deal with this in Graphite :(

Yeah, that's why we're trying to be thorough on documenting what changes and why, so you're not flying completely blind. 😕

davidicus:
There are some issue with [PageTitlebar](https://deploy-preview-2932--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-pagetitlebar--with-select).

| Bread crumbs are cut off | Sizing on sticky header is messed up |
|--------------------------|--------------------------------------|
| image vs `next`. image | image vs `next` image |

davidicus:
There is an issue with the menu button menu placement

image

davidicus (changes requested):
whew! That was a doozy. I think this last one will be it and it should be ok to 🚢

davidicus (approved):

PR #2938, opened 2021-10-08
2630 date time picker spec alignment

### LINKED ISSUE #2630: [Card] Align experimental DateTimePicker feature with spec
### What package is this for?

- [X] React
- [ ] Angular

### Summary

For date time range picker in the Card component we need to align with the suggestions made from design in [this](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2455#issuecomment-871529449) thread. Requires that the custom range picker Card action be an icon only with the text that normally populates the input piped to the subtitle of the card. 

- [ ] Modify DateTimePicker component to surface the values in a callback
- [ ] Modify Card component to use the custom range picker as an icon only card action with value being added to subtitle
- [ ] Document new feature in Card markdown
- [ ] Modify test to verify value is in new location.

kevinsperrine:
Closes #2630 

**Summary**

- Adds readableValue to the onApply callback and catches that variable in the onCardAction callback to display it in the subtitle, if one isn't already given.

- add `humanValue`, `tooltipValue` to onApply callback for DateTimePickerV2
- add state management for subtitle to display the currently selected DateTimeValue in the subtitle of a card if one isn't already given
- add docs for Card
- refactor parseValue into a utility and use in both DateTimePicker/DateTimePickerV2
- update tests as necessary and add new ones

- in the Card with datetime range selector story
- open the DateTimePicker and choose a date
- click onApply
- the chose date range should display in the subtitle

- make sure existing CardRangePicker works as expected
- make sure the DateTimePicker (V1 and V2) continue to work as expected without changes

### DIFF
--- a/packages/react/src/components/Card/Card.story.jsx
+++ b/packages/react/src/components/Card/Card.story.jsx
@@ -245,7 +245,7 @@ export const WithCustomRangeSelector = () => {
 WithCustomRangeSelector.storyName = 'with custom range selector';
 
 export const WithDateTimePickerRangeSelector = () => {
-  const dateTimePickerSetting = select('range', [true, false, 'iconOnly', 'full'], 'iconOnly');
+  const dateTimePickerSetting = select('range', [true, false, 'iconOnly'], 'iconOnly');
   return (
     <div style={{ width: `300px`, margin: 20 }}>
       <Card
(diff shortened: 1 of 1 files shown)


davidicus:
Looks good. Few items I see in the preview:
- [This](https://deploy-preview-2938--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card--with-date-time-picker-range-selector) story throws an error when you choose `true` as value for `range` knob and then click the calendar icon
- The knobs still reference the option to be `full` or `icon-only`. We should update these knobs?

PR #2945, opened 2021-10-11
2794 table group column headers

### LINKED ISSUE #2794: [Table] Group Column Headers
### What package is this for?

- [X] React
- [ ] Angular

### Summary

Support table header grouping labels
Design spec:
https://pages.github.ibm.com/ai-applications/design/components/datatable/groupedcolumns

Need to ensure that text ellipsis occurs on strings that do not fit, and provide either a Carbon or browser tooltip to display the full label
We will need to disable column resize and column pickers when grouping is enabled

bjornalm:
Closes #2794

**Summary**
New feature to add grouping to the columns.

- API. Added a new columnGroups prop to the table that create groups. Added a columnGroupId to the view.table.ordering objects that lets the developer place a column in a group.
- Added a new Component to handle the column group row. Since the first row of column headers will define the column width of all following rows and since rowspan only works downward I kept the normal column headers row at the top and added the group row as the second one. Using position relative to swap places of the individual headers to give the appearance the group row comes first. 
- Had to write plenty of css to handle all the combinations of table props that would influence the table header cells and their borders. I wanted to keep this new functionality as isolated as possible so none of this css is used unless the table has column groups. 
- Added prop to TableHeader to handle headers that should span two rows.
- Wrote more tests and updated the Visual regression snapshot1 to contain column groups. This snapshot catches most of the styling, but not all. We could also add e2e tests to verify the styling.
- Updated documentation, added more knobs to the simple-stateful-example story (as playground and testing) and added one more story specific for this feature to serve as an example for our library consumers.

_A note on the design_
I've implemented an additional design of the groups to fit the ligthweight prop. I also decided to show vertical borders on the edges of normal columns if there are "special" columns before or after. This is mostly because of consistency since there has to be a border at the edge by the resize spacer column.

- Read the documentation first about the sequential ordering of groups and columns 
- Test in multiple browsers
- Use https://deploy-preview-2945--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table-table--table-with-column-grouping and try to "break" the table head by combining the props in different ways. The props in this story mainly affects how the borders are rendered. 

- Use https://deploy-preview-2945--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table-statefultable--simple-stateful-example and try to "break" the table head by combining different props. Test all the knobs but extra hot candidates that have been particularly naughty are:
* resize (and don't forget to actually resize)
* ligthweight
* size 
* multisort
* hasFilter
* hasMultiSort
* hasResize
* hasRowActions
* hasRowExpansion
* hasRowNesting
* hasRowSelection = multi
* hasSort
* Demo column tooltips
* and of course, do it all again using the RTL knob

- There should be no side effects for the table when this feature isn't used. So to make sure we should re-test all table stories and compare with next.
- No Snapshot diffs for the table except id and the `modified simple-stateful-example` and the new `table-with-column-grouping`. 

### DIFF
--- a/packages/react/src/components/Table/TableHead/ColumnGrouping/_column-grouping.scss
+++ b/packages/react/src/components/Table/TableHead/ColumnGrouping/_column-grouping.scss
@@ -46,22 +46,6 @@
         top: calc(#{$col-height} * -1);
       }
     }
-
-    &.lightweight {
-      // lightweight adds extra 2 borders at top and bottom
-      $extra-borders-width: calc(2px * 2);
-      // Subtract width for extra borders from the overflow button height.
-      .#{$iot-prefix}--table-head--overflow {
-        max-height: calc(#{$col-height} - #{$extra-borders-width}) !important;
-      }
-      // Subtract width for extra borders from the table headers
-      th.#{$iot-prefix}--table-header--span-group-row {
-        button.#{$iot-prefix}--table-head--table-header,
-        a.#{$iot-prefix}--table-head--table-header {
-          height: calc(calc(#{$group-height} + #{$col-height}) - #{$extra-borders-width});
-        }
-      }
-    }
   }
 }
 


kevinsperrine:
The groupings disappear in the simple stateful example when using `hasColumnSelection`.

1. Go to https://deploy-preview-2945--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--simple-stateful-example
2. Turn on `hasColumnSelection`
3. Open column selection
4. Turn off a column

kevinsperrine:
> > The groupings disappear in the simple stateful example when using `hasColumnSelection`.
> > 
> > 1. Go to https://deploy-preview-2945--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--simple-stateful-example
> > 2. Turn on `hasColumnSelection`
> > 3. Open column selection
> > 4. Turn off a column
> 
> Correct. This is what we discussed in the meeting last week, these props can't be combined. The docs state this and there is also a console warning when that is done anyway.

Yeah, I read that and then completely forgot it when I start running through the options. 🥴 My bad.

PR #2956, opened 2021-10-13
fix(table): table expand icon consistency

JoelArmendariz:
**Summary**
When the table option "shouldExpandOnRowClick" is passed as true, the expand icon shows up in the table when the row is clicked whether the row has children or not. Need to only handle expand if there are children defined for any given row.

 - Add `&& nestingChildCount` condition to the `onClick` that fires the `onRowExpanded` function so that it doesn't fire unless there are children (nestingChildCount is passed by the parent `TableBody` like this `nestingChildCount={row.children ? row.children.length : 0}`)

Go to the `stateful table with row nesting and resizable columns with initial width` story where I have added `shouldExpandOnRowClick`. Make sure that rows without children do not show the expand icon when you click on them.

Make sure normal functionality for an expandable row is still good.

### DIFF
--- a/packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
+++ b/packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
@@ -543,7 +543,10 @@ const TableBodyRow = ({
           data-nesting-offset={nestingOffset}
           onExpand={(evt) => stopPropagationAndCallback(evt, onRowExpanded, id, false)}
           onClick={() => {
-            if (shouldExpandOnRowClick && nestingChildCount) {
+            if (
+              shouldExpandOnRowClick &&
+              ((hasRowNesting && nestingChildCount) || hasRowExpansion)
+            ) {
               onRowExpanded(id, false);
             }
             if (hasRowSelection === 'single' && isSelectable !== false) {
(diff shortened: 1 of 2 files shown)


kevinsperrine (dismissed):
Thank you!

davidicus:
@JoelArmendariz bump to address Bjorn's comments.

davidicus:
@bjornalm can we get a re review here?

packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -543,7 +543,10 @@ const TableBodyRow = ({
           data-nesting-offset={nestingOffset}
           onExpand={(evt) => stopPropagationAndCallback(evt, onRowExpanded, id, false)}
           onClick={() => {
-            if (shouldExpandOnRowClick) {
+            if (
```
bjornalm, line 546: Nice, this should work. 
We have plenty of code duplication for the 2 `onClick` functions now so I would refactor and reference a single function from the JSX code. The only change to the function body would be the  `onRowSelected(id, !isExpanded)`;

```
const onRowClick = () => {
    if (shouldExpandOnRowClick && ((hasRowNesting && nestingChildCount) || hasRowExpansion)) {
      onRowExpanded(id, !isExpanded);
    }
    
    if (hasRowSelection === 'single' && isSelectable !== false) {
      onRowSelected(id, true);
    }
    if (isSelectable !== false) {
      onRowClicked(id);
    }
}
```

bjornalm (approved):

PR #2965, opened 2021-10-14
fix(table): improve dropdown experience with filter dropdown

### LINKED ISSUE #2833: [Table] Table is stretched when filter contains large number of items.
### What package is this for?

- [X] React

### Describe the bug
See below example in our product. The tables contains only 3 items, but there is more items in the filter drop down list, which causes the table has been stretched when we open filter drop down list.
image

### Environment/versions:
- Browser: [All browsers]
- `carbon-addons-iot-react` version: 2.146.0-next.182

kevinsperrine:
Closes #2833 

**Summary**

- The dropdown menus overflowing the container and causing a scroll was a weird experience. While we can't make the dropdown appear outside of the container, this solutions attempts to find a middle-ground by adding a small min-height to the table based on the size prop given, and then dynamically calculates the max-height of the dropdown to fit into the available area.

- add size classes for container
- set min-height based on size
- calculate max-height of dropdown list

- The dropdown menus should not overflow the container
- The min-height should exist on tables with fewer than 3 rows.

- confirm the dropdowns are still accessible and min-height doesn't break any other table functions.


bjornalm:
We still get a horizontal scrollbar on the table when we have hasResize and pagination active, I think we can avoid that by subtracting the horizontal scrollbar from the max height.

image

davidicus:
Looks great. We may want to add a background color to the table for instances where the table is larger than the rows it has. 
image

PR #2972, opened 2021-10-18
feat(table): allow drag and drop sorting of multisort columns

### LINKED ISSUE #2399: [Table] multi-sort drag and drop ordering
### What package is this for?

- [X] React
- [ ] Angular

### Summary

Allow dragging and dropping to change the order of multi-sort columns as mentioned here: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2336#pullrequestreview-677710194

kevinsperrine:
Closes #2399 

**Summary**

- adds a drag handle to the TableMultiSortModal to allow drag and drop re-ordering of the MultiSort columns

- adds drag and drop hooks to TableMultiSortRow
- adds tests to confirm functionality

- in the StatefulTable with multi-sort
- open the multi-sort modal
- drag and drop the columns into a different order
- save
- new order should be reflected on the table

- confirm existing multi-sort behavior wasn't effected.

### DIFF
--- a/packages/react/src/components/Table/StatefulTable.story.jsx
+++ b/packages/react/src/components/Table/StatefulTable.story.jsx
@@ -11,6 +11,7 @@ import StoryNotice from '../../internal/StoryNotice';
 import FlyoutMenu, { FlyoutMenuDirection } from '../FlyoutMenu/FlyoutMenu';
 import { csvDownloadHandler } from '../../utils/componentUtilityFunctions';
 import Button from '../Button/Button';
+import { DragAndDrop } from '../../utils/DragAndDropUtils';
 
 import StatefulTable from './StatefulTable';
 import {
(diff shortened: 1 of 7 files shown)


kevinsperrine:
> Nice work, and thanks for commenting the more complex reorder code.
> 
> Using story /story/1-watson-iot-table-statefultable--with-multi-sorting I get a few issues.
> 
> 1. click on multisort in the numbers column
> 2. click the bottom "add columns" 2 times
> 3. drag the bottom row  up and down the list without releasing
> 4. issue A: notice how the rows are somehow duplicated
> 5. now select another story and then select the multisort story again
>    6: issue B: we get the error "Cannot have two HTML5 backends at the same time"
> 
> I did also find some strange behaviour that isn't in next:
> 
> 1. click on multisort in the numbers column
> 2. click the bottom "add columns" a few times
> 3. click sort
> 4. repoen multisort dialog
> 5. click clear sorting
> 6. not all rows are cleared

Nice catch. These are both caused by bad keys on the elements that were just added. Fixed. I added a DragAndDrop decorator on the story, I think that'll take care of that, too. (At least I haven't ran into it again yet).

PR #2986, opened 2021-10-21
2879 dashboard grid visual regression tests

### LINKED ISSUE #2879: [DashboardGrid] visual regression tests

kevinsperrine:
Closes #2879 

**Summary**

- Adds visual regression testing to all variations of breakpoints and cards.

- Adds VRT to DashboardGrid

- Just double check we're not missing variations on features used by the cards.

- n/a


davidicus:
> Wow, how many MB are all the images? I thought we would only use a few screenshots in total for our e2e testing considering binary data cannot not be diffed by git (i.e. repo size will keep increasing when we update the images).

Yes, I was thinking we would have one pic with all the cards in one shot. We can discuss if this is not possible.

kevinsperrine:
> > Wow, how many MB are all the images? I thought we would only use a few screenshots in total for our e2e testing considering binary data cannot not be diffed by git (i.e. repo size will keep increasing when we update the images).
> 
> Yes, I was thinking we would have one pic with all the cards in one shot. We can discuss if this is not possible.

One pic with everything isn't possible, because it causes out-of-memory errors when running, so I split them into an image for each card, BUT I'm also including every breakpoint size here, too, so that would drop it a lot if we only included one breakpoint.

kevinsperrine:
> > > Wow, how many MB are all the images? I thought we would only use a few screenshots in total for our e2e testing considering binary data cannot not be diffed by git (i.e. repo size will keep increasing when we update the images).
> > 
> > 
> > Yes, I was thinking we would have one pic with all the cards in one shot. We can discuss if this is not possible.
> 
> One pic with everything isn't possible, because it causes out-of-memory errors when running, so I split them into an image for each card type, BUT I'm also including every breakpoint size here, too, so that would drop it a lot if we only included one breakpoint.

@davidicus What do you want to do here?

davidicus:
I

> > > > Wow, how many MB are all the images? I thought we would only use a few screenshots in total for our e2e testing considering binary data cannot not be diffed by git (i.e. repo size will keep increasing when we update the images).
> > > 
> > > 
> > > Yes, I was thinking we would have one pic with all the cards in one shot. We can discuss if this is not possible.
> > 
> > 
> > One pic with everything isn't possible, because it causes out-of-memory errors when running, so I split them into an image for each card type, BUT I'm also including every breakpoint size here, too, so that would drop it a lot if we only included one breakpoint.
> 
> @davidicus What do you want to do here?

I think we need these for the dashboard. If any can be combined it would be great.

kevinsperrine:
> I
> 
> > > > > Wow, how many MB are all the images? I thought we would only use a few screenshots in total for our e2e testing considering binary data cannot not be diffed by git (i.e. repo size will keep increasing when we update the images).
> > > > 
> > > > 
> > > > Yes, I was thinking we would have one pic with all the cards in one shot. We can discuss if this is not possible.
> > > 
> > > 
> > > One pic with everything isn't possible, because it causes out-of-memory errors when running, so I split them into an image for each card type, BUT I'm also including every breakpoint size here, too, so that would drop it a lot if we only included one breakpoint.
> > 
> > 
> > @davidicus What do you want to do here?
> 
> I think we need these for the dashboard. If any can be combined it would be great.

@davidicus @bjornalm I stripped this down to just the `lg` breakpoint. That makes it 9 files (one for each card type with all sizes) and < 2MB.

davidicus (approved):

bjornalm (approved):

PR #3009, opened 2021-10-29
fix(table): remove state management of aggregations from table component

### LINKED ISSUE #2434: [Table]: remove state management from table for aggregations
We need to remove the state management of showing/hiding aggregations from the `Table` component and force the consumer to use the `onToggleAggregations` callback to manage that state themselves. More info is available here: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2389/files#diff-950233c53355828ab38327e2f495c3465ec1b87c3ea11a6635099bf1668cda3dR583

- [x] The `onToggleAggregations` function needs to be added to the defaultProps for the table.
- [x] The `onToggleAggregations` function needs to be added back to the actions for table stories
- [x] The useEffect, useState, useCallbacks all need to be removed to allow the consumer to manage the `isHidden` aggregation prop, and only rely on the state of the `view.aggregations.isHidden` prop for showing/hiding the aggregation footer.

kevinsperrine:
Closes #2434

### Merging into V3

**Summary**

- cleans up the state management of aggregations that was in the basic table.

- add onToggleAggregations callback to actions/defaults.
- remove all the useEffects, useStates, etc

- table should no longer automatically manage state for aggregations, but the user should supply the `isHidden` prop to toggle it on/off.

- this is a breaking change for v3.

### DIFF
--- a/packages/react/src/components/Table/Table.jsx
+++ b/packages/react/src/components/Table/Table.jsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState, useRef, useCallback, useEffect } from 'react';
+import React, { useMemo, useState, useRef } from 'react';
 import PropTypes from 'prop-types';
 import merge from 'lodash/merge';
 import pick from 'lodash/pick';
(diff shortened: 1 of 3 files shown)


davidicus (approved):
Looks good. One thing I was thinking is that we should probably start a migration guide doc that we can continue to add to for these v3 changes. Just so we can keep track of all of them when we finally do publish v3

PR #3015, opened 2021-11-01
fix(date-time-picker): allow selecting ranges across months

### LINKED ISSUE #3007: [DateTimePicker] Date time picker doesn't allow selecting on a previous day in absolute mode
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Is this the same bug I'm aware of where if I select a previous Month in the absolute date time picker and click on a date it defaults to the current month on that date instead of the previous Month?

For instance:

image

Now click the back button to go to September
image

Now click on the 14th of September
image

It switches the Month to October and starts trying to select a range but in the wrong month
image

(We need this in our 8.7 release too)

### To Reproduce

You can see this in the latest next storybook online if you follow those instructions/scenario

We need this pretty badly, its causing lots of issues for our demos

kevinsperrine:
Closes #3007 

**Summary**

- Fixes an issue where changing the month in the DateTimePicker when picking an absolute range would cause it to not show the correct month and lead to a confusing experience.

- added checks to call flatpickr's `jumpToDate` function when selecting dates
- added tests to confirm for both DateTimePicker and DateTimePickerV2

- go to https://deploy-preview-3015--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-datetime-picker--default
- Click "Last 30 minutes"
- Click "Custom Range"
- Click "Absolute"
- Click "Back arrow" to go back to a previous month
- Choose a date -- The month should still be the same month you were in...
- Click "Forward arrow" to go forward to a more recent month
- Choose a date -- The month should still be the same month you navigated to.
- Click "Apply"
- The chosen range should match what you entered
- Repeat with [DateTimePickerV2](https://deploy-preview-3015--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--default)

- double check DateTimePicker and DateTimePickerV2 for adverse affects related to absolute ranges

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -539,20 +539,33 @@ const DateTimePicker = ({
     }
   }, [datePickerElem, focusOnFirstField]);
 
-  const onDatePickerChange = ([start, end]) => {
+  const onDatePickerChange = ([start, end], _, flatpickr) => {
+    if (
+      dayjs(absoluteValue.start).isSame(dayjs(start)) &&
+      dayjs(absoluteValue.end).isSame(dayjs(end))
+    ) {
+      return;
+    }
+
     const newAbsolute = { ...absoluteValue };
+    newAbsolute.start = start;
+    newAbsolute.startDate = dayjs(newAbsolute.start).format('MM/DD/YYYY');
+    const prevFocusOnFirstField = focusOnFirstField;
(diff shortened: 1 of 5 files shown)


scottdickerson (approved):
this fixes my issues

bjornalm (approved):
Great work, and thanks for adding the comments explaining that the code is needed to workaround issues in flatpicker

PR #3021, opened 2021-11-02
chore(table): remove isSortable prop where inappropriate

### LINKED ISSUE #2962: [table] remove isSortable knobs and references
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The table has no `options.isSortable` prop but we have such knobs in the stories and also in some tests.
We should remove that.

### To Reproduce

1. Go to https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-statefultable--simple-stateful-example and https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--basic-dumb-table
2. See the isSortable knob.

### Expected behavior

No knobs or test props that don't exist

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

kevinsperrine:
Closes #2962 

**Summary**

- removes extraneous uses of isSortable in stories/tests where not appropriate.

- updated stories, tests, snapshots

- isSortable should not appear as knobs in stories
- or in tests except where necessary

- n/a

### DIFF
--- a/packages/react/src/components/Table/StatefulTable.test.jsx
+++ b/packages/react/src/components/Table/StatefulTable.test.jsx
@@ -190,7 +190,6 @@ describe('stateful table with real reducer', () => {
         }}
         secondaryTitle={`Row count: ${initialState.data.length}`}
         actions={mockActions}
-        isSortable
         options={{
           ...initialState.options,
           hasFilter: 'onKeyPress',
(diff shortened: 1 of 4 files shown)


bjornalm (changes requested):
Nice, I think you missed one though. `isSortable` is also used in packages/react/src/components/Table/StatefulTable.test.jsx

kevinsperrine:
> Nice, I think you missed one though. `isSortable` is also used in packages/react/src/components/Table/StatefulTable.test.jsx

Good catch. tough to sort through and catch all of them. fixed.

bjornalm (approved):

davidicus (approved):

PR #3022, opened 2021-11-02
fix(stateful-table): handle resetting pagination when necessary

### LINKED ISSUE #2971: [StatefulTable] current page doesn't reset to initial value when new data is passed to the table
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
If the user is on the page higher that 1 when data changes and there is not enough data entries to fill all pages before the current one and be displayed on the current page - the empty state is shown;

### To Reproduce

1. Go to 'https://stackblitz.com/edit/carbon-table-example-3zk62j?file=components%2FStatefulTable.jsx'
2. Choose `Table`
2. Go to page 2 of the table
3. Click the `short` button above the table (it will change table data)
4. Empty state is shown because there is not enough data to fill the first page and be shown on the second page

test case:
https://stackblitz.com/edit/carbon-table-example-3zk62j?file=components%2FStatefulTable.jsx
please note that in the test case the version is v2.141.0-next.36 just because it crashes every time you try to use a newer version, but the bug is reproducible with newer versions.

### Expected behavior

Current page number should be reset to the initial value

Related to release at the end of this month

kevinsperrine:
Closes #2971 

**Summary**

- Fixes the StatefulTable reducer to update the current page in situations where the number of data items is less than what the currently selected page is able to display

- add knob to stateful table example story to display `n` number of items
- update reducer to check current page against number of items and pageSize
- add tests to confirm

- on the [StatefulTable example story](https://deploy-preview-3022--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--simple-stateful-example)
- turn on pagination
- change to page 3
- change number of items to 5
- check that page reset to page 1
- repeat with various combinations of pageSize, page, and number of items

- check other pagination related StatefulTable stories for errors.

### DIFF
--- a/packages/react/src/components/Table/StatefulTable.story.jsx
+++ b/packages/react/src/components/Table/StatefulTable.story.jsx
@@ -157,6 +157,14 @@ export const SimpleStatefulExample = () => {
       id="table"
       key={`table${demoInitialColumnSizes}`}
       {...initialState}
+      data={initialState.data.slice(
+        0,
+        select(
+          'number of data items in table',
+          [initialState.data.length, 50, 20, 5],
+          initialState.data.length
+        )
+      )}
       actions={tableActions}
       columns={initialState.columns
         .map((column) => {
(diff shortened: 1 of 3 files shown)


bjornalm (approved):
LGTM

davidicus (approved):

PR #3024, opened 2021-11-02
fix(list): fix styles to keep secondaryValue aligned when overflowing

### LINKED ISSUE #2939: [List] secondary value out of alignment with long primary value
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When the primary value is too long, the secondary value starts going out of alignment. I have found & reproduced this on HierarchyList but I suspect it's on all the list components.

### To Reproduce

Component to reproduce is below with a screenshot of the outcome attached
```

import React from "react";
import {HierarchyList} from "carbon-addons-iot-react";


export default class ReorderListTest extends React.Component<any, any> {

    handleItemWillMove(): boolean {
        return true;
    }

    render() {

        const items: any = [];
        for (let i = 0; i < 10; i++) {
            items.push({
                id: i + '',
                content: {
                    value: "Item ".repeat(i * 10),
                    secondaryValue: "Jo"
                }
            });
        }

        return (
            <HierarchyList
                title={"Joanne"}
                items={items}
                editingStyle={"single"}
                itemWillMove={this.handleItemWillMove}
            >

            </HierarchyList>

        );

    }
}
```

misalignment

### Expected behavior

The second value should always be aligned

### Environment/versions:

- OS: [Windows]
- Browser: [chrome]
- `carbon-addons-iot-react` version: [v2.147.0-next.14]

kevinsperrine:
Closes #2939 

**Summary**

- small style fix to prevent the flexbox from creeping up in size and causing the secondaryValue to be out of alignment

- add min-width to css

- in the [HierarchyList with large number of items story](https://deploy-preview-3024--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-hierarchylist--with-large-number-of-items)
- item 10 secondary value should be aligned and overflow with ellipsis
- item 20 primary value should be aligned and overflow with ellipsis

- scan List stories to make sure it doesn't cause adverse effects

### DIFF
--- a/packages/react/src/components/List/HierarchyList/HierarchyList.story.jsx
+++ b/packages/react/src/components/List/HierarchyList/HierarchyList.story.jsx
@@ -471,9 +471,14 @@ export const WithLargeNumberOfItems = () => (
       items={[...Array(1000)].map((_, i) => ({
         id: `item-${i}`,
         content: {
-          value: i === 20 ? `Item ${i} `.repeat(i + 1 * 10) : `Item ${i}`,
+          value:
+            i === 20
+              ? `Item ${i} that has an extra long label that will definitely be truncated`
+              : `Item ${i}`,
           secondaryValue:
-            i === 10 ? `Item ${i} Subvalue `.repeat(i + 1 * 10) : `Item ${i} Subvalue`,
+            i === 10
+              ? `Item ${i} that has an extra long label that will definitely be truncated`
+              : `Item ${i} Subvalue`,
         },
       }))}
       editingStyle={EditingStyle.Single}
--- a/packages/react/src/components/List/HierarchyList/HierarchyList.story.jsx
+++ b/packages/react/src/components/List/HierarchyList/HierarchyList.story.jsx
@@ -471,7 +471,9 @@ export const WithLargeNumberOfItems = () => (
       items={[...Array(1000)].map((_, i) => ({
         id: `item-${i}`,
         content: {
-          value: `Item ${i}`,
+          value: i === 20 ? `Item ${i} `.repeat(i + 1 * 10) : `Item ${i}`,
+          secondaryValue:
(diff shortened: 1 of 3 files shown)


packages/react/src/components/List/HierarchyList/HierarchyList.story.jsx
```diff
@@ -471,7 +471,9 @@ export const WithLargeNumberOfItems = () => (
       items={[...Array(1000)].map((_, i) => ({
         id: `item-${i}`,
         content: {
-          value: `Item ${i}`,
+          value: i === 20 ? `Item ${i} `.repeat(i + 1 * 10) : `Item ${i}`,
```
bjornalm, line 474: When I was testing the story and saw the repeated value in the drag preview I was sure it was a bug. 

```suggestion
          value: i === 20 ? `Item ${i} that has an extra long label` : `Item ${i}`,
```

packages/react/src/components/List/HierarchyList/HierarchyList.story.jsx
```diff
@@ -471,7 +471,9 @@ export const WithLargeNumberOfItems = () => (
...
+          secondaryValue:
+            i === 10 ? `Item ${i} Subvalue `.repeat(i + 1 * 10) : `Item ${i} Subvalue`,
```
bjornalm, line 476:
```suggestion
            i === 10 ? `Item ${i} Subvalue that has an extra long label` : `Item ${i} Subvalue`,
```

bjornalm (approved):
Clean and simple, just the suggestion about the confusing label

davidicus (approved):

packages/react/src/components/List/HierarchyList/HierarchyList.story.jsx
kevinsperrine: Good point. Updated and merged.

PR #3026, opened 2021-11-03
fix(table): hide multisort options in rowEdit to match current sort

### LINKED ISSUE #2863: [table] multisort rendering incorrectly when activeBar = rowEdit
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The column headers that has multisort are missing the sort order arrows when activeBar = rowEdit. See attached screenshot.

It should look like this:
image

But it looks like this
image

There could be more recent header functionality that are affected by the activeBar = rowEdit so we should expand the search for that.

### To Reproduce
Requires code changes to a story.

Use a story to enable both activeBar = rowEdit and multisort headers

### Expected behavior

Arrows showing directions should still be shown (but the sorting should be disabled)

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

kevinsperrine:
Closes #2863 

**Summary**

- When in rowEdit mode the sort options are hidden, but those for multi-sort were still visible. This removes them when in rowEdit mode to match behavior of singular sort.

- add checks for `isDisabled` prop and hide multi-sort options/classes when necessary
- added MultiSort knob to rowEdit table story to verify

- open the Table with full rowEdit example
- turn on MultiSort
- click the edit button in the toolbar to turn on rowEdit mode
- confirm the sort icons/numbers/multisort overflow menu is hidden

- check other MultiSort stories for inconsistencies


bjornalm (approved):
Nice work!

davidicus (approved):

PR #3029, opened 2021-11-03
feat(table): focus on first filterable column when filter row opened

### LINKED ISSUE #1340: [table] UX, expanding the Filter Rows should automatically place focus into the first filterable column
### Describe the bug

This is a accessibility/UX improvement.  After clicking the Filter Columns button where the Filter row expands, we should automatically put focus into the first filterable column so they can easily tab between them to add

kevinsperrine:
Closes #1340 

**Summary**

- adds refs to the FilterHeaderRow to apply focus to the first column when the FilterHeaderRow opens

- adds forwardRef to ComboBox
- adds refs to FilterHeaderRow items
- checks type of first filterable column and focuses on it

- open StatefulTable example
- turn on hasColumnSelection and hasFilter
- open filter row
- String column should be focused
- open column selection and move "Status" column to first position
- open Filters
- String column should still be focused
- Repeat moving Select column to first position
- Repeat moving Select column to first position and toggle "force MultiSelect filter" knob to true

- ensure focus doesn't mess with any other FilterHeaderRow stories

### DIFF
--- a/packages/react/src/components/ComboBox/ComboBox.jsx
+++ b/packages/react/src/components/ComboBox/ComboBox.jsx
@@ -52,232 +52,237 @@ const defaultProps = {
   horizontalDirection: 'end',
 };
 
-const ComboBox = ({
-  loading,
-  wrapperClassName,
-  closeButtonText,
-  editOptionText,
-  hasMultiValue,
-  onChange,
-  items,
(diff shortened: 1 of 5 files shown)


bjornalm (approved):
Nice work

Issue #1885, opened 2020-12-16
[table] Not support to set overflow menu direction


zhaozhuo813:
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When the table has many rows, I want to set the overflow menu direction (top/bottom) to control how to pop up my menu depends on if the row is on top of the list or bottom of the list.
Currently in each row in the table data, I set row action like this :
row.rowActions.push({
          id: 'metadata',
          isOverflow: true,
          labelText: 'xxx'
        }); 
There is no parameter supported to control the overflow menu direction.

### To Reproduce

1. Go to story book
2. No parameter to control the overflow menu direction

### Environment/versions:

- OS: macos
- Browser: chrome
- `carbon-addons-iot-react` version: `2.140.6`

katedraper:
@davidicus please estimate this item

davidicus:
We have recently added a smart detection placement of drop downs for the table that will place the dropdown on screen if too close to the edge. @kevinsperrine can you confirm this? If so does that solve the use case here @zhaozhuo813 ?

jlrgore:
+1 that we are running into a similar issue with overflow menus causing issues in modals. The overflow menu can flow outside of the viewport at the bottom of a list: https://github.ibm.com/tools-for-aps/zh/issues/26714#issuecomment-39202143

kevinsperrine:
Hi @jlrgore, We merged an auto-positioning version of this in `v2.147.0-next.55`, so the overflow should switch to top when near the bottom of the screen as seen here: https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--table-example-with-create-save-views. 

<img width="1399" alt="Screen Shot 2022-01-06 at 4 25 13 PM" src="https://user-images.githubusercontent.com/609466/148454459-5b61c2ce-8134-4017-a2be-853f94798083.png">

This was left open to add a prop to allow it to be controlled by the developer.

jlrgore:
Nice! Will this be available in base Carbon, as well? I don't believe our dev team is using the IOT variant.

kevinsperrine:
Unfortunately, no. The auto-positioning is only available in the IOT variant. We wrote a wrapper around the OverflowMenu and Tooltip to automatically re-position them based on screen position.

Issue #2951, opened 2021-10-12
[Table] toggle aggregation off-screen in RTL mode


kevinsperrine:
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

![Screen Shot 2021-10-12 at 9 08 21 AM](https://user-images.githubusercontent.com/609466/136962085-229a5d03-dfb8-4b2a-9f67-1230489c1e44.png)

### To Reproduce

1. https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--basic-dumb-table
2. Turn on RTL
3. Open Overflow menu in Toolbar

PR #3031, opened 2021-11-03
fix(overflow-menu): increase width of the spacer to match button width

### LINKED ISSUE #3013: [StatefulTable] with row actions: when hovering over the row actions redundant white space over the row action button appears
### What package is this for?

- [x] React
- [ ] Angular

### To Reproduce

1. Go to [Stateful Table with Row actions](https://carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--with-row-expansion-and-actions)
2. Click on three dots button (any row action)
3. Notice extra white space line appear over the bottom edge of the hovered button (please see screenshot)

image

### Environment/versions:

- OS: [e.g. MacOS, Windows]

kevinsperrine:
Closes #3013 #1885 

**Summary**

- The small spacer tab in the overflow menu wasn't quite long enough to fill the space leaving a small gap. This fixes that.

- increase width of spacer tab to match button width

- Go to [StatefulTable with row action](https://deploy-preview-3031--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--stateful-example-with-expansion-max-pages-and-column-resize) or [OverflowMenu stories](https://deploy-preview-3031--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-overflowmenu--basic)
- open the overflow menu
- the spacer tab should match the width of the button

- Check other stories using overflow menu (in RTL) to confirm the fixes don't break anything
- https://deploy-preview-3031--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--table-example-with-create-save-views
- https://deploy-preview-3031--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--basic-table-with-full-row-edit-example
- https://deploy-preview-3031--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-overflowmenu--basic

### DIFF
--- a/packages/react/src/hooks/usePopoverPositioning.jsx
+++ b/packages/react/src/hooks/usePopoverPositioning.jsx
@@ -1,4 +1,6 @@
 import * as React from 'react';
+import { useLangDirection } from 'use-lang-direction';
+
 /**
  * This is used as the callback for menuOffset on Tooltips, OverflowMenus, and FlyoutMenus.
  * That callback returns the underlying FloatingMenu element, direction, trigger element, and flipped
@@ -70,6 +72,7 @@ export const usePopoverPositioning = ({
   const [adjustedFlipped, setAdjustedFlipped] = React.useState();
   const [{ flyoutAlignment }, setDirections] = React.useState({});
   const previousDirection = React.useRef();
+  const langDir = useLangDirection();
 
   React.useEffect(() => {
     previousDirection.current = direction;
@@ -105,7 +108,6 @@ export const usePopoverPositioning = ({
         : buttonElement.getBoundingClientRect();
       const windowWidth = window.innerWidth || document.documentElement.clientWidth;
       const directionChange = `${previousDirection.current}->${adjustedDirection}`;
-
       if (previousDirection.current !== adjustedDirection && flyoutAlignment) {
         switch (directionChange) {
           default:
@@ -114,6 +116,12 @@ export const usePopoverPositioning = ({
               left: tooltipRect.x - buttonRect.x,
             };
         }
+      }
+
+      let left = 0;
+
+      if (langDir === 'rtl' && isOverflowMenu) {
+        left = -16;
       }
 
       if (previousDirection.current !== adjustedDirection) {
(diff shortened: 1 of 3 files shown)


bjornalm (approved):
Nice, work. This is not from your code changes, but it is related so perhaps we can fix it with this PR as well (or a new one of you want):

image

kevinsperrine:
> Nice, work. This is not from your code changes, but it is related so perhaps we can fix it with this PR as well (or a new one of you want):
> 
> image

I actually think this is a carbon bug in this version, but I added a fix to my auto-positioning hook to account for it, and replaced the overflow menu in TableActionRow with the auto-positioning version (as requested in #1885), so I've re-requested a review for those changes.

bjornalm:
Code looks good but the  build failed so I the regression test links point to the previous version

packages/react/src/hooks/usePopoverPositioning.jsx
```diff
@@ -116,6 +118,12 @@ export const usePopoverPositioning = ({
         }
       }
 
+      let left = 0;
+
+      if (langDir === 'rtl' && isOverflowMenu) {
+        left = -15;
```
bjornalm, line 124: Put this value 15 in a well named const that explains the origin?

bjornalm (approved):

bjornalm:
I approved this, but I also found another related issue that also exists in next. 
image

packages/react/src/hooks/usePopoverPositioning.jsx
kevinsperrine: It's not necessary. Carbon has fixed it and I need to take it out in the carbon upgrade PR...

kevinsperrine:
> I approved this, but I also found another related issue that also exists in next. image

Yeah, there's already an issue for this. https://github.com/carbon-design-system/carbon-addons-iot-react/issues/2951

packages/react/src/hooks/usePopoverPositioning.jsx
kevinsperrine: Actually, nvm. It's not. I'll add a constant.

davidicus (approved):

PR #3034, opened 2021-11-03
fix(table): preserve whitespace in cells with multiple spaces

### LINKED ISSUE #3011: [statefulTable] Extra Spaces are truncated in a text input
### What package is this for?

- [ y] React
- [ ] Angular

### Describe the bug
within the InitialView of statefulTable, extra spaces are truncated (even between the characters). with the editMode, it looks good.
https://carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--basic-table-with-full-row-edit-example
provided the screenshots. 1. Edit mode.. 2. View mode.
Tried rendering at cell level, it still happens. 
image
image

### To Reproduce: 
create a statefulTable or use the link : https://carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--basic-table-with-full-row-edit-example
Edit a text Input field and provide the value "1  3     6"  click on DONE, and in view mode, the text Input displays "1 3 6".
When you click on Edit again, it restores the correct value to "1  3     6"

1. Go to 'https://carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--basic-table-with-full-row-edit-example'
2. Click on 'Edit'
3. Enter any value with extra spaces 
4. Click Done. 
4. See display

### Expected behavior

Extra spaces needs to be restored even in the view mode. When click Done, text field should display "1  3     6"

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?
yes 
date, the better!_

kevinsperrine:
Closes #3011 

**Summary**

- adds a class setting `white-space: pre` in table cells that have multiple continuous spaces, so that whitespace is preserved in those cells.

- add `iot--table__cell-text--preserve` class
- set class in TableCellRenderer when preserveCellWhiteSpace prop is true
- add tests to confirm

- Go to the table with full rowEdit example
- turn on preserveCellWhiteSpace knob
- edit the first row and add numbers with multiple spaces in the number column
- save
- confirm your multiple spaces are still visible in the output and whitespace is preserved.

- no adverse effects on other strings


bjornalm (commented):
Code looks good, can't test the preview though. Do you think everyone wants this behaviour? I can imagine that there is a good reason for `white-space: pre` not being the default setting for white-space in general. Perhaps the best option is to put it behind a prop?

kevinsperrine:
> Code looks good, can't test the preview though. Do you think everyone wants this behaviour? I can imagine that there is a good reason for `white-space: pre` not being the default setting for white-space in general. Perhaps the best option is to put it behind a prop?

That's a good thought. That's why I used double spaces as the boolean check, but a prop probably makes more sense. I'll add that.

kevinsperrine:
> Code looks good, can't test the preview though. Do you think everyone wants this behaviour? I can imagine that there is a good reason for `white-space: pre` not being the default setting for white-space in general. Perhaps the best option is to put it behind a prop?

@bjornalm Just pushed the code with the prop added. Hopefully, netlify completes this time. 🤞

bjornalm (approved):
Nice work

davidicus (approved):

PR #3039, opened 2021-11-05
3025 list add allowed drop target

### LINKED ISSUE #3025: [list] add option to explicitly specify which list item that can be drop targets during a drag
### What package is this for?

- [x] React
- [ ] Angular

### Summary
For #2603 we need to be able to limit reordering to within the group only. A clean and generic way to support that is to allow consumers of the List and HierarchyList to specify a set of row ids that should get drop targets based on which row is being dragged.

**Additional context** 

upcoming release?

date, the better!_

### Available extra resources

bjornalm:
Closes #3025

**Summary**
Adds the option to limit which rows can act as drop targets  using a callback getAllowedDropIds that receives the id of the item about to be dragged and should return an array of ids of the rows that are allowed as drop targets. I decided on using  a callback like this because providing some kind of map of AllowedDropIds up front would be too costly and using the existing `itemWillMove` wouldn't work since it was called after the drop targets were rendered.

- Added prop `getAllowedDropIds` from component ListItemWrapper all the way up to the List and HierarchyList.
- Memoized `getAllowedDropIds` in the ListComponent to prevent a possibly expensive function to be executed for each row.
- Refactored some poorly written dnd code
- Created new stories for List and HierarchyList in order to have meaningful examples
- Added new unit tests
- Added new e2e tests
- Updated docs

1. Go to https://deploy-preview-3039--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-hierarchylist--with-nested-reordering-restricted
2. Make sure players can only be reordered within their team and not dropped in other teams
3. Make sure teams can only be reordered among other teams and not dropped in other teams 
4. Enable knob hasVirtualList and repeat tests

1. Go to https://deploy-preview-3039--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-list--with-reorder-and-drop-target-restrictions
2. Make sure drops are only allowed on rows defined in knob `getAllowedDropIds return value`
3. Enable knob hasVirtualList and repeat tests

- Make sure all other stories for List and HierarchyList works as before


kevinsperrine (changes requested):
Code looks great! But i did find one issue we'll want to address re: dragging parents into other parents.

(ignore the artifacting on the gif)
drop-target-test

bjornalm:
> Code looks great! But i did find one issue we'll want to address re: dragging parents into other parents.
> 
Good catch, I've updated the story to prevent nesting using the `itemWillMove`.

davidicus (approved):

packages/react/src/components/List/List.jsx
```diff
@@ -155,6 +163,9 @@ const List = forwardRef((props, ref) => {
   const ListHeader = overrides?.header?.component || DefaultListHeader;
   const ListContent =
     overrides?.content?.component || isVirtualList ? VirtualListContent : DefaultListContent;
+  // getAllowedDropIds will be called by all list items when a drag is initiated and the
+  // paramater (id of the dragged item) will be the same until a new drag starts.
+  const memoizedGetAllowedDropIds = getAllowedDropIds ? memoize(getAllowedDropIds) : null;
```
stuckless, line 168: I wonder how well this scales.  If I have a list of 10k items, will it call this for 10k items?

stuckless: also, what is a virtual list and how is different than a normal list

bjornalm: Since the `getAllowedDropIds` function is memoized and will receive the same param it will only be evaluated once per drag no matter how many items you have. 

For the old non virtual list with 10k items that doesn't use pagination the function `memoizedGetAllowedDropIds` will indeed be called 10k times but for such a list there are a lot of things that will happen 10k times and the calling of a memoized function would likely not be the straw that broke the camel's back.

The potential bottleneck lies in the number of items returned by the `getAllowedDropIds` since it is an array and not a map there will be a search for each rendered item's id. For large lists we now recommend using `isVirtualList` which means that only the items visible on the screen (I assume no more than around 30) will be calling and searching the result of  `memoizedGetAllowedDropIds` at the "same" time. Returning an array of 1000 ids does work but for more than that we probably need to support returning a map.

But even with `isVirtualList` and without the `getAllowedDropIds` our list component will be slow when using drag and drop in a list of 10k items. I also don't know what our UX designers would say about drag and drop in such a huge list even if it was fast though.

bjornalm: The virtual list only renders the items currently visible, see https://github.com/bvaughn/react-window

PR #3055, opened 2021-11-11
V3 upgrade

kevinsperrine:
Closes #

## This is being merged into V3

**Summary**
It looks like a lot--and it is--but the majority of it is the eslinting upgrade changes and changing `bx--` and `iot--` prefixes to use constants. The items that specifically had merge conflicts are listed here and should be the ones checked more thoroughly. There were merge conflicts with:
- src/index.js - linting change for `export {default ` versus items that are not being exported in v3. TableHead, TableBody, EmptyTable, TableWithSkeletonHeader
- src/components/SuiteHeader.jsx - href as `#` instead of `javascript:void(0)`
- src/components/Table/Table.jsx - adding shouldExpandOnRowClick prop removal of wrapText/truncateText props
- src/components/Table/Table.test.jsx - removing tests related to wrapText/truncateText
- src/components/Table/TablePropTypes.js - adding new props to end of file
- src/components/Table/TableBody.jsx - adding preserveCellWhiteSpace prop
- - src/components/Table/TableBodyRow.jsx - adding preserveCellWhiteSpace prop
- src/components/Table/TableCellRenderer.test.jsx - adding preserveCellWhiteSpace prop / removing wrapText/truncateText tests
- src/components/Table/TableHead/TableHead.jsx - adding hasRowNesting prop, columnGrouping code

- check public api snaphots for inadvertent removals.
- check table stories (specifically related to these features: preservingWhiteSpace, columnGrouping, and cell truncation)

- see above


packages/react/src/components/FlyoutMenu/FlyoutMenu.jsx
```diff
@@ -398,8 +400,8 @@ const propTypes = {
 
   /** Set of internationalized button names */
   i18n: PropTypes.shape({
-    cancelButton: PropTypes.string,
```
davidicus, line 401: Is this breaking change?

packages/react/src/components/Header/Header.jsx
```diff
@@ -11,42 +11,11 @@ import { Switcher20 } from '@carbon/icons-react';
 import { settings } from '../../constants/Settings';
 
 import HeaderActionGroup from './HeaderActionGroup';
+import { HeaderActionItemPropTypes, HeaderPanelPropTypes } from './HeaderPropTypes';
```
davidicus, line 14: I see we are deleting three objects but only importing two here. Is this on purpose?

packages/react/src/components/FlyoutMenu/FlyoutMenu.jsx
kevinsperrine: No, it was never correct to begin with. https://github.com/carbon-design-system/carbon-addons-iot-react/blame/0019171e71520aa92357a3b059120747e58cca08/packages/react/src/components/FlyoutMenu/FlyoutMenu.jsx

packages/react/src/components/Header/Header.jsx
kevinsperrine: Yeah, ChildContentPropTypes is only used within HeaderActionItemPropTypes, so it didn't need to be re-imported.

packages/react/src/components/ImageCard/ImageCard.test.jsx
```diff
@@ -42,12 +46,6 @@ describe('ImageCard', () => {
     expect(screen.getByTestId(`${testId}-empty`)).toBeDefined();
   });
 
-  it("should show 'size not supported' when an invalid size is given", () => {
```
davidicus, line 45: How come we remove this test?

kevinsperrine: We changed the behavior on the ImageCards when we added all the prop-type warning when using an invalid size. The preferred method was to attempt to render the content of the card as best we can, but throw a prop-type warning for using an invalid size. So, we removed the test because this message will never be shown anymore and replaced it with a test checking that the prop-type warning is shown.

davidicus (approved):
I think this should be good to go once Bjorn gives the thumbs up,

bjornalm (approved):
Great work. What a massive PR.
I focused on the table related changes code and functionality looks ok-

PR #3058, opened 2021-11-12
 fix (table) : filtering for timestamps

EdmondCShaffer:
Closes #

**Summary**

- Fix issue for filtering on table columns with dates 

- items_here

- tests here

**Regression Test (Test all filterable columns in table to insure filterFunction only runs on timestamp column)**

- tests here

### DIFF
--- a/packages/react/src/components/TableCard/TableCard.jsx
+++ b/packages/react/src/components/TableCard/TableCard.jsx
@@ -386,7 +386,14 @@ const TableCard = ({
           isSortable: true,
           width: i.width ? `${i.width}px` : newSize === CARD_SIZES.LARGETHIN ? '150px' : '', // force the text wrap
           filter: i.filter
-            ? i.filter
+            ? i.type === 'TIMESTAMP'
+              ? {
+                  filterFunction: (cellValue, filterValue) => {
+                    const dateString = dayjs(cellValue).format('L HH:mm');
+                    return dateString.includes(filterValue);
+                  },
+                }
+              : i.filter
             : { placeholderText: mergedI18n.defaultFilterStringPlaceholdText }, // if filter not send we send empty object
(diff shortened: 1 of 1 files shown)


packages/react/src/components/TableCard/TableCard.jsx
```diff
@@ -386,7 +386,14 @@ const TableCard = ({
           isSortable: true,
           width: i.width ? `${i.width}px` : newSize === CARD_SIZES.LARGETHIN ? '150px' : '', // force the text wrap
           filter: i.filter
-            ? i.filter
+            ? i.type === 'TIMESTAMP'
+              ? {
+                  filterFunction: (cellValue, filterValue) => {
```
kevinsperrine, line 391: We should spread the other filter options into this object. Additionally, we need to check if a filterFunction was supplied. We wouldn't want to replace it with this default if the devs want to filter differently.

kevinsperrine (changes requested):
Thanks for the contribution, @EdmondCShaffer. 

Can you also add a test case confirming these new behaviors?

scottdickerson (approved):
I'm good once @kevinsperrine changes are incorporated

bjornalm:
This PR is not for `next`, it is for branch `carbon-design-system:test-tablecard-filter`. Is that on purpose?

kevinsperrine (approved):
Looks good.

bjornalm:
@EdmondCShaffer We also need an issue to link this PR to

bjornalm (changes requested):
Nice work, just some console.log statements that we can skip + all PRs need to be linked to an issue.

scottdickerson:
I fixed the failing testcases now @EdmondCShaffer

bjornalm (approved):

PR #3064, opened 2021-11-16
fix(hierarchy-list): uncheck children when parent unchecked

### LINKED ISSUE #3005: [HierarchyList] unchecking parent checkbox does not uncheck children
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

Unchecking a parent in the HierarchyList does not uncheck all the children.

### To Reproduce

1. Go to https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-hierarchylist--with-nested-reorder
2. Change "Editing Style" knob to "multiple-nesting"
3. Check checkbox on "Atlanta Braves"
4. Click chevron to open "Atlanta Braves"
5. Uncheck "Atlanta Braves"
6. All Children of "Atlanta Braves" should now be unchecked, but only the parent was unchecked.

kevinsperrine:
Closes #3005 

**Summary**

- previously when a parent was selected and then unselected, all the children would remain checked. This fixes that.

- fix onClick being called twice on label and input
- fix locked rows being selected when parent is checked
- fix children not being unselected when parent is unselected
- move the handleSelect call on the checkbox to onChange instead of onClick
- add tests to confirm

1. https://deploy-preview-3064--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-hierarchylist--with-nested-reorder
2. check a parent
3. uncheck a parent
4. confirm children unselected

1. check a parent with a locked child
2. confirm locked child is not selected

- no adverse effects on other hierarchylist stories

### DIFF
--- a/packages/react/src/components/List/ListContent/ListContent.jsx
+++ b/packages/react/src/components/List/ListContent/ListContent.jsx
@@ -167,7 +167,15 @@ const ListContent = ({
                 name={item.value}
                 data-testid={`${item.id}-checkbox`}
                 labelText=""
-                onClick={() => handleSelect(item.id, parentId)}
+                onClick={(evt) => {
+                  // after the carbon upgrade to 10.47 this event
+                  // is being fired twice, once on the label and once on the
+                  // input, so filtering out the LABEL event here and only firing on
+                  // the input.
+                  if (evt.target.tagName === 'INPUT' && !isLocked) {
+                    handleSelect(item.id, parentId);
+                  }
+                }}
                 checked={isSelected}
                 disabled={isLocked}
                 indeterminate={isIndeterminate}
--- a/packages/react/src/components/List/ListContent/ListContent.jsx
+++ b/packages/react/src/components/List/ListContent/ListContent.jsx
@@ -167,15 +167,7 @@ const ListContent = ({
                 name={item.value}
                 data-testid={`${item.id}-checkbox`}
                 labelText=""
-                onClick={(evt) => {
-                  // after the carbon upgrade to 10.47 this event
-                  // is being fired twice, once on the label and once on the
-                  // input, so filtering out the LABEL event here and only firing on
-                  // the input.
-                  if (evt.target.tagName === 'INPUT' && !isLocked) {
-                    handleSelect(item.id, parentId);
-                  }
-                }}
+                onChange={() => handleSelect(item.id, parentId)}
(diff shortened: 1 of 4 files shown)


packages/react/src/components/List/ListContent/ListContent.jsx
```diff
@@ -167,7 +167,15 @@ const ListContent = ({
                 name={item.value}
                 data-testid={`${item.id}-checkbox`}
                 labelText=""
-                onClick={() => handleSelect(item.id, parentId)}
+                onClick={(evt) => {
+                  // after the carbon upgrade to 10.47 this event
+                  // is being fired twice, once on the label and once on the
+                  // input, so filtering out the LABEL event here and only firing on
+                  // the input.
+                  if (evt.target.tagName === 'INPUT' && !isLocked) {
```
kevinsperrine, line 175: The argument could be made this should fire on label and not input. I chose input in this case, because many of our tests are using testIds tied to the checkbox element. Setting this to fire on the label breaks those tests. This is a carbon bug. I've filed an issue. It's re-creatable in their own storybook for the checkbox playground. https://github.com/carbon-design-system/carbon/issues/10122

kevinsperrine: switching this to an onChange seems to be fine and tests all still pass. We need to double check the stories though.

bjornalm:
Code looks good, but I think we should fix the parent child selection relationship properly. For that we also need:
1. When all children are selected, the parent also gets selected.
2. When some, but not all children are selected, set the indeterminate state on the parent.

Checkout the list story at https://deploy-preview-3064--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-list--with-checkbox-multi-selection-and-hierarchy

kevinsperrine:
> Code looks good, but I think we should fix the parent child selection relationship properly. For that we also need:
> 
> 1. When all children are selected, the parent also gets selected.
> 2. When some, but not all children are selected, set the indeterminate state on the parent.
> 
> Checkout the list story at https://deploy-preview-3064--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-list--with-checkbox-multi-selection-and-hierarchy

That's a separate issue: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3066

davidicus (approved):

PR #1462, opened 2020-08-04
[WIP] chore(Carbon): update to v10.17.0

dakahn:
Closes #1323 
Closes #1469 

Smoke testing our v10.17 release candidate (with a bonus v10.16 bump coming along for the ride :surfer:)

### To test 
I'll include a list of bug fixes per relevant packages to check for bugs and breaks:
#### carbon-components-react
- fix(dropdown): add ...other prop (#6479) ([`810168bf8`](https://github.com/carbon-design-system/carbon/commit/810168bf8))
- fix(Copy): avoid label text mismatch (#6428) ([`177a6bd77`](https://github.com/carbon-design-system/carbon/commit/177a6bd77))
- fix(SideNav): a11y fixes for SideNavMenu, MenuItem (#6468) ([`efa7a9bb1`](https://github.com/carbon-design-system/carbon/commit/efa7a9bb1))
- fix(dropdown,multiselect): add support for refs on buttons (#6441) ([`2485371ed`](https://github.com/carbon-design-system/carbon/commit/2485371ed))
- fix(react): update downshift to 5.2.1 (#6418) ([`69151cc8c`](https://github.com/carbon-design-system/carbon/commit/69151cc8c))
- fix(multiselect): remove aria-selected prop (#6440) ([`2692813cc`](https://github.com/carbon-design-system/carbon/commit/2692813cc))
- fix(pagination-nav): add data-page attributes (#6437) ([`8c84f349b`](https://github.com/carbon-design-system/carbon/commit/8c84f349b))
- fix(react): add prop adapter for downshift changes (#6427) ([`0037736f7`](https://github.com/carbon-design-system/carbon/commit/0037736f7))
- fix(react): update useId to match id on server and client (#6369) ([`6d3e3dc6c`](https://github.com/carbon-design-system/carbon/commit/6d3e3dc6c))
- fix(data-table): sortable, boolean column, and overflow menu row fixes (#6268) ([`5d7986990`](https://github.com/carbon-design-system/carbon/commit/5d7986990))
- feat(CodeSnippet): allow copy button to be optional (#6505) ([`ad4d9a298`](https://github.com/carbon-design-system/carbon/commit/ad4d9a298))
- feat(grid): add narrow grid mode (#6520) ([`2c7f9aab5`](https://github.com/carbon-design-system/carbon/commit/2c7f9aab5))
- feat(modal): support alertdialog aria role (#5955) ([`825f03ab7`](https://github.com/carbon-design-system/carbon/commit/825f03ab7))
[...]

PR #3071, opened 2021-11-18
refactor(repo): remove vulnerable polished library, clean up other un-used deps

### LINKED ISSUE #3050: Use of a vulnerable version of polished npm package
The package `polished@^2.3.3` is used as the dependency. This version is vulnerable to the ReDoS described here: https://security.snyk.io/vuln/SNYK-JS-POLISHED-1298071

Vulnerability is fixed in versions 3.7.2, 4.1.3 or higher.

kevinsperrine:
Closes #3050 

**Summary**

`polished` is a helper library for styled-components that wasn't used anymore, so I thought I'd check if there were other things that aren't used anymore or other ways we could clean up the repo a bit.

- `mockdate` is only used in tests -- switched to devDep
- `polished` isn't used at all -- removed. It was originally added as a helper to styled-components years ago, but isn't used anywhere now.
- `react-transition-group` isn't used at all -- removed. It was originally added years ago but has since been removed.
- update all references to `/lib/` to `/es/` in carbon packages (icons mainly)
- `fbjs` removed -- I completely removed it, and didn't have any issues with linking it locally with using `"carbon-addons-iot-react": "link:../../pal-build"` in the package.json (pal-build is a directory containing the built files copied into it), but it was only added back in April https://github.com/carbon-design-system/carbon-addons-iot-react/pull/2276, so I'd like for @scottdickerson to smoke-test this for us, but it's passing in the CI build test.
- `downshift` isn't used. removed. It was originally added when smoke-testing an old carbon upgrade, but it's a dependency of carbon, so we don't need to explicitly include it or pin the version. https://github.com/carbon-design-system/carbon-addons-iot-react/pull/1462
- `lodash` - replace `lodash` with `lodash-es`. @carbon/charts uses `lodash-es`, so instead of having consumed applications have both, this makes them only build with one. This was a pretty automated replacement using a regex search and replace, and then running `eslint --fix` and `yarn format` to pretty it up. Added the custom eslint rule to point people toward `lodash-es` instead.

| Current with `lodash` | After Switch to `lodash-es` |
| ----- | ----- |
| Screen Shot 2021-11-18 at 8 32 53 AM | Screen Shot 2021-11-18 at 8 07 14 AM |

- confirm removed packages have no affect

- `fbjs` could potentially cause issues with local linking, so we need to test that thoroughly


davidicus (approved):
This is great work! I would say if the build and deploy test works in our CI then we are able to successfully link the package even with the removal of the `fbjs` but since @scottdickerson added and was having issues we should wait for some confirmation or clarification from him.

bjornalm (approved):
Looks good ot me, nice clean up!

davidicus:
hey @scottdickerson. Do you have a moment to check this PR?

scottdickerson:
if a local `yarn link` from a downstream project worked for you @davidicus then we're good I think

davidicus:
> if a local `yarn link` from a downstream project worked for you @davidicus then we're good I think

Yes, we `yarn link` in our test app that we build in our github actions so we should be good.

PR #3081, opened 2021-11-20
feat(hierarchy-list): add controlled expandedIds prop

### LINKED ISSUE #3037: [HierarchyList] Expand/Collapse all or some rows
### What package is this for?

- [x] React
- [ ] Angular

### Summary

I want to be able to expand/collapse all levels of a HierarchyList (basically pass an array of row Ids as a prop to allow expanding or collapsing of rows).
Using `defaultExpandedIds` only expands specified rows on the first render. I can change the value of that prop by changing a state, but it doesn’t trigger a change in the collapse/expanded state of the rows.

When clicking the Open All button
Screen Shot 2021-11-04 at 9 17 13 AM

All rows specified in the prop are expanded
Screen Shot 2021-11-04 at 9 17 43 AM

The sooner the better. We have a product release coming up in January, so would be great to have the feature in before then.

### Available extra resources

I am available to test the feature

kevinsperrine:
Closes #3037 

**Summary**

- Adds the ability to explicitly control which ids are expanded or collapsed in the HierarchyList. 
- Also adds an `onExpandedChange` callback that passes the array to the user, so they can know what ids are currently expanded

- adds `expandedIds` prop to control which ids in the list are expanded or collapsed
- adds `onExpandedChange` callback to let the consumer know the current internal state of the expansions.
- added tests to confirm

- Check https://deploy-preview-3081--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-hierarchylist--with-overflow-menu
- confirm Chicago White Sox is expanded
- Add another team to the expandedIds array knob (comma separated no space--ie. Chicago White Sox,New York Mets)
- Check https://deploy-preview-3081--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-hierarchylist--with-default-expanded-ids
- confirm the white sox and yankees are expanded by default
- Add a different team in the expandedIds array knob and confirm that team is expanded and the white sox and yankees close

- check for errors with defaultExpandedIds story
- check expanding and collapsing on click still works.
- expanded state should only change when the `expandedIds` prop is updated


bjornalm (approved):
Nice work as always. I also tested the deep nesting in https://deploy-preview-3081--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-hierarchylist--with-mixed-hierarchies and it works fine.

Since we are doing this I think it also makes sense to expose an `onExpandedChange` so that consumers can be informed if the "controlled" state is changed by the user in the HierarchyList itself. Could be useful to determine the enabled state of an external "Expand all" button like in the issue example.

kevinsperrine:
> Nice work as always. I also tested the deep nesting in https://deploy-preview-3081--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-hierarchylist--with-mixed-hierarchies and it works fine.
> 
> Since we are doing this I think it also makes sense to expose an `onExpandedChange` so that consumers can be informed if the "controlled" state is changed by the user in the HierarchyList itself. Could be useful to determine the enabled state of an external "Expand all" button like in the issue example.

Good idea. I'll add that.

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -276,6 +286,33 @@ const HierarchyList = ({
...
+    }
+    isFirstRender.current = false;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [expandedIdsProp]);
+
+  /**
+   * effect to trigger calling the onExpandedChange callback. Ignore it on the initial render
+   * when previousExpandedIds is undefined, but fire it on every change afterward
+   */
+  useEffect(() => {
```
bjornalm, line 305: Would it not be easier to use the `toggleExpansion` callback from the list, and also keep the APIs more aligned?

kevinsperrine: I started to, but we call setExpandedIds in lots of other places (searches or when the items change, for example), so I thought the effect was cleaner than calling it in multiple places any time it changed.

bjornalm: Ok, makes sense

davidicus (approved):

PR #3083, opened 2021-11-20
fix(table): properly round maxPages to prevent decimals

### LINKED ISSUE #2156: [Table] Pagination display incorrect when set number per page is 50 and total count is 201
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When using <Table> component,  If set item per page to 50, and if the total count is 201, then the pagination is incorrect.
image

### Environment/versions:
- `carbon-addons-iot-react` version: [e.g. `v2.144.0`]

kevinsperrine:
Closes #2156 

**Summary**

- in certain cases if the number of items in the table was _less_ than the `pageSize` multiplied by the `maxPages`, this could result in un-rounded situations. In this case, for example, the number of items was 201. The page size was 50, but I believe `maxPages` was being passed in as `4.02` which when multiplied in javascript actually gives you `200.99999999999997`. I've updated the documentation and prop-types to explain the reasoning for `maxPages` and throw a warning if it isn't a positive integer.

- updated documentation for `maxPages` in pagination portion
- added prop-type warnings for maxPages that aren't positive integers
- added tests to confirm

- I don't know how to verify this one in a story without adding a bunch of different knobs and props for a one-off confirmation. Hopefully, the test passing suffices.

- n/a


bjornalm (changes requested):
The whole `maxPages` is a bit confusing, and I'm afraid we are making it even more confusing now. From my understanding it is there to serve as a `limit` when dealing with huge data sets so that the select page dropdown doesn't crash on rendering thousands of items.

It doesn't really make sense for the consuming application to just divide their total number of items with the page size and pass it as the `maxPages` which is what I assume is happening here. The application should only use that prop to set an actual limit and and let the `totalItems` prop do its work instead.

So I don't think the reported issue is a bug that we should solve, instead I propose that we:
1. Update the documentation/comments to really clarify the usage of `maxPages`
2. Warn if `maxPages` receives something other than a positive integer.

kevinsperrine:
> The whole `maxPages` is a bit confusing, and I'm afraid we are making it even more confusing now. From my understanding it is there to serve as a `limit` when dealing with huge data sets so that the select page dropdown doesn't crash on rendering thousands of items.
> 
> It doesn't really make sense for the consuming application to just divide their total number of items with the page size and pass it as the `maxPages` which is what I assume is happening here. The application should only use that prop to set an actual limit and and let the `totalItems` prop do its work instead.
> 
> So I don't think the reported issue is a bug that we should solve, instead I propose that we:
> 
> 1. Update the documentation/comments to really clarify the usage of `maxPages`
> 2. Warn if `maxPages` receives something other than a positive integer.

Updated.

davidicus:
This change would not have prevented the original issue author's floating number total pages. It seems like we are doing a calculation somewhere that needs a `Math.ceil()` to round up this number and prevent it from happening.

kevinsperrine:
> This change would not have prevented the original issue author's floating number total pages. It seems like we are doing a calculation somewhere that needs a `Math.ceil()` to round up this number and prevent it from happening.

That's what I did originally, but @bjornalm disagreed. See: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3083#pullrequestreview-813462449

davidicus:
> > This change would not have prevented the original issue author's floating number total pages. It seems like we are doing a calculation somewhere that needs a `Math.ceil()` to round up this number and prevent it from happening.
> 
> That's what I did originally, but @bjornalm disagreed. See: [#3083 (review)](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3083#pullrequestreview-813462449)

Yes, but I would have to disagree. If a user is entering a positive integer as a prop and we end up having a floating point number in the UI then that is a bug. I am fine with keeping the documentation changes but we should still  fix the issue.

kevinsperrine:
> > > This change would not have prevented the original issue author's floating number total pages. It seems like we are doing a calculation somewhere that needs a `Math.ceil()` to round up this number and prevent it from happening.
> > 
> > 
> > That's what I did originally, but @bjornalm disagreed. See: [#3083 (review)](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3083#pullrequestreview-813462449)
> 
> Yes, but I would have to disagree. If a user is entering a positive integer as a prop and we end up having a floating point number in the UI then that is a bug. I am fine with keeping the documentation changes but we should still fix the issue.

Re-added the rounding and included it in the test with the prop-type warning to ensure the correct output is shown in the UI.

bjornalm:
> > > This change would not have prevented the original issue author's floating number total pages. It seems like we are doing a calculation somewhere that needs a `Math.ceil()` to round up this number and prevent it from happening.
> > 
> > 
> > That's what I did originally, but @bjornalm disagreed. See: [#3083 (review)](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3083#pullrequestreview-813462449)
> 
> Yes, but I would have to disagree. If a user is entering a positive integer as a prop and we end up having a floating point number in the UI then that is a bug. I am fine with keeping the documentation changes but we should still fix the issue.

@davidicus Just to clarify, they were **actually not** entering a positive integer, they were incorrectly entering a floating point. What we do is just multiplying the `maxPages` with the `pageSize` and none of those props should allow floating points.

davidicus:
> > > > This change would not have prevented the original issue author's floating number total pages. It seems like we are doing a calculation somewhere that needs a `Math.ceil()` to round up this number and prevent it from happening.
> > > 
> > > 
> > > That's what I did originally, but @bjornalm disagreed. See: [#3083 (review)](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3083#pullrequestreview-813462449)
> > 
> > 
> > Yes, but I would have to disagree. If a user is entering a positive integer as a prop and we end up having a floating point number in the UI then that is a bug. I am fine with keeping the documentation changes but we should still fix the issue.
> 
> @davidicus Just to clarify, they were **actually not** entering a positive integer, they were incorrectly entering a floating point because they probably didn't understand the purpose of the `maxPages` prop. What we do in the Table is just multiplying the `maxPages` with the `pageSize` and none of those props should allow floating points in my opinion.

That makes more sense. I should have thought of that.

kevinsperrine:
> > > > > This change would not have prevented the original issue author's floating number total pages. It seems like we are doing a calculation somewhere that needs a `Math.ceil()` to round up this number and prevent it from happening.
> > > > 
> > > > 
> > > > That's what I did originally, but @bjornalm disagreed. See: [#3083 (review)](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3083#pullrequestreview-813462449)
> > > 
> > > 
> > > Yes, but I would have to disagree. If a user is entering a positive integer as a prop and we end up having a floating point number in the UI then that is a bug. I am fine with keeping the documentation changes but we should still fix the issue.
> > 
> > 
> > @davidicus Just to clarify, they were **actually not** entering a positive integer, they were incorrectly entering a floating point because they probably didn't understand the purpose of the `maxPages` prop. What we do in the Table is just multiplying the `maxPages` with the `pageSize` and none of those props should allow floating points in my opinion.
> 
> That makes more sense. I should have thought of that. I do think preventing the break is still the right thing to do but perhaps in the warning we tell them the floating point number will be rounded up.

@bjornalm @davidicus Added the rounding message to the prop type warning.

bjornalm (approved):
LGTM

davidicus (approved):

PR #3085, opened 2021-11-20
fix(table): fix a11y issues on column selection button and toolbar tooltip

### LINKED ISSUE #2593: [Table] Header button a11y issue
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

===== Problem Summary =====
The WAI-ARIA role 'presentation' is not valid for the element <button> and may be ignored by the browser since the element is focusable.
WAI-ARIA roles must be valid for the element to which they are assigned.
Some HTML elements or attributes have native semantics that map, by default, to implied WAI-ARIA semantics. For predictable interaction and behavior of UI controls, these implicit semantics should only be overridden as defined in the W3C ARIA in HTML Document conformance requirements for use of ARIA attributes in HTML specification.

If a native HTML element fits the intended behavior, use that instead of adding ARIA markup;
OR, update the role on the element using the guidance in the Document conformance requirements for use of "ARIA" attributes in HTML section of the W3C ARIA in HTML specification. This reference provides details of valid ARIA markup on native HTML elements.
image

kevinsperrine:
Closes #2593 

- changes button role from "presentation" to "button" in column selection row
- adds `toolbarTooltipLabel` to i18n to add aria-label to table toolbar tooltips
- fixes incorrect "options" parameter passed to some tests

- go to basic dumb table test
- turn on hasColumnSelection knob
- open in new window
- scan with IBM Equal Access Accessibility Checker
- There should only be 5 violations about Bypass blocks and content not being within a landmark element

- column selection drag and drop and hiding still works as expected


bjornalm:
I only get 2, not 5. I assume that is even better...
image

bjornalm (approved):

davidicus (approved):

PR #3088, opened 2021-11-23
feat(table): account for rowSpans in moveColumn

Donisius:


packages/angular/src/table/table-model.class.ts
```diff
@@ -996,8 +984,135 @@ export class AITableModel implements PaginationModel {
...
       list.splice(blockStart, blockEnd - blockStart + 1);
     }
   }
+
+  protected tabularToNested(
+    header: TableHeaderItem[][],
+    data: TableItem[][],
+    headerRow: TableHeaderItem[] = [],
+    availableHeaderItems: TableHeaderItem[][] = [],
+    leafIndex = { index: 0 }, // So we can pass by reference
```
Donisius, line 997: change to `leafIndexRef`, with property `current`

Donisius, line 997:
```suggestion
    // This allows us to walk the leaves as if they were in a list from left to right.
    // We need to pass by reference so that we can update this value from within the recursion.
    leafIndex = { index: 0 },
```

packages/angular/src/table/table-model.class.ts
```diff
@@ -996,8 +984,135 @@ export class AITableModel implements PaginationModel {
...
       const block = list.slice(blockStart, blockEnd + 1);
-      list.splice.apply(list, [index, 0].concat(block));
+      list.splice.apply(list, [index + 1, 0].concat(block));
       list.splice(blockStart, blockEnd - blockStart + 1);
     }
   }
+
+  protected tabularToNested(
+    header: TableHeaderItem[][],
+    data: TableItem[][],
```
Donisius, line 994: Don't pass these in use this.header and this._data instead

davidicus:
@zvonimirfras bump.

zvonimirfras (approved):

davidicus (approved):

PR #3092, opened 2021-11-29
V3 upgrade

kevinsperrine:
Closes #

**Summary**

- next round of merge changes from next into v3
- This one is _very_ straightforward and everything has already been approved in the previous PRs. The only merge conflicts were in the publicAPI snapshot and in table.test.jsx where I added a test next to tests for `wrapCellText` and `truncateCellText` that have been removed in v3. This review should be just a formality.

**Change List (commits, features, bugs, etc)**
- `packages/react/src/components/Table/Table.test.jsx`: merge conflict for removal of old tests for props that are removed in V3.
- `packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap`: conflicts between the addition of a new prop to the table and removal of those mentioned above. Re-ran tests to update snapshots.

**Acceptance Test (how to verify the PR)**

- all tests pass

**Regression Test (how to make sure this PR doesn't break old functionality)**

- nothing new here really--and no merge conflicts of note--so everything in this has already been approved, just moving it into v3.

**Things to look for during review**

- [ ] Make sure all references to `iot` or `bx` class prefix is using the prefix variable
- [ ] (React) All major areas have a `data-testid` attribute. New test ids should have test written to ensure they are not changed or removed.
- [ ] UI should be checked in RTL mode to ensure the proper handling of layout and text.
- [ ] All strings should be translatable.
- [ ] The code should pass a11y scans (The storybook a11y knob should show no violations). New components should have a11y test written.
- [ ] Unit test should be written and should have a coverage of 90% or higher in all areas.
- [ ] All components should be passing visual regression test. For new styles or components either a visual regression test should be written for all permutations or the base image updated.
- [ ] Changes or new components should either write new or update existing documentation.
- [ ] PR should link and close out an existing issue

PR #3093, opened 2021-11-29
refactor(table): make preserveColumnWidths:true the default, deprecate false

### LINKED ISSUE #2812: [table] make preserveColumnWidths default
### What package is this for?

- [x] React
- [ ] Angular

### Summary

When `preserveColumnWidths` is true then adjacent column widths are preserved during resize and column toggle/add/remove operations. This behaviour should become default and the prop `preserveColumnWidths`, that when set to be false can be used to switch to the old behaviour, should be deprecated.

upcoming release?

### Available extra resources

kevinsperrine:
Closes #2812 

### Depends On: #3092

**Summary**

- makes the preserveColumnWidth:true behavior the default and deprecates the usage of preserveColumnWidth:false

- set preserveColumnWidth:true as default
- deprecate preserveColumnWidth:false
- update a couple of tests to explicitly set preserveColumnWidth:false to maintain their checks
- update all story knobs to use preserveColumnWidth:true by default

- deprecation warning is shown in stories when using preserveColumnWidth:false
- tests pass

- n/a


bjornalm (approved):
Seems like all the v3 changes from PR 3092 made it into this one as well, but I filtered on the commit `refactor(table): make preserveColumnWidths:true the default` which I assume covers this PR?

kevinsperrine:
> Seems like all the v3 changes from PR 3092 made it into this one as well, but I filtered on the commit `refactor(table): make preserveColumnWidths:true the default` which I assume covers this PR?

Correct. I based it off the other stuff, just so that the v3 changes were the latest in case this change caused any conflicts.

davidicus (approved):

PR #3096, opened 2021-11-30
chore(repo): upgrade cypress and related plugins/images

### LINKED ISSUE #3072: [Repo] upgrade Cypress
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

We're seeing weird failures on previously working tests (CardCodeEditor). The hunch is that Chrome upgrades are causing issues, so we might explore upgrading Cypress to see if they're addressed in newer versions.

kevinsperrine:
Closes #3072 

**Summary**

- upgrades cypress to v9.1.0 and the docker images along with it.

- upgrade cypress and all cypress related plugins
- all tests are run in `--headless` mode by default, so removed `--headless` where unnecessary and added `--headed` where needed
- removed `nodeVersion="system"` as it's deprecated and all runs will use the version of node used to launch cypress
- updated docker image and github action containers to newer Node and Chrome versions
- updated visual regression test images for slight rendering differences in new version
- added viewport calls to TimeSeriesCard tests to maintain proper aspect ratio, so mouse position calls are correct

- tests pass

- n/a


bjornalm (approved):
LGTM

davidicus (approved):

PR #3105, opened 2021-12-03
3100 list inconsistent checkbox behaviour

### LINKED ISSUE #3100: [List] inconsistent checkbox behaviour
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The callback behaviour when clicking on a checkbox in the List differs depending on if  `VirtualListContent` or the older `ListContent` is used.  

### To Reproduce

1. Go to [list--with-checkbox-multi-selection](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-list--with-checkbox-multi-selection)
2. Click on a list item's checkbox and notice that the Item is selected
3. Check the `isVirtual` knob 
4. Click on another list item's checkbox and notice that the Item is NOT selected

### Expected behavior

The behaviour should be the same for both  `VirtualListContent` and `ListContent` and that is that only one callback is made when clicking the checkbox

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

bjornalm:
Closes #3100

**Summary**
It was hard to tell why this bug occurred but it turned out it was related to the carbon checkbox bug that we thought we already had a workaround for. 

The checkbox emits a double click event, which will trigger double click events for our wrapping list rows. But, the actual workaround is simple, we need to stop propagation on the checkbox click even though we weren't using it. 

1. Go to https://deploy-preview-3105--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-list--with-checkbox-multi-selection
2. Click on a list item's checkbox and notice that the Item is selected
3. Check the isVirtual knob
4. Click on another list item's checkbox and notice that the Item is also selected

Verify that the other List and HierarchyList stories that use checkboxes works as expected. For the hierarchyList you can use this story knob config
https://deploy-preview-3105--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-hierarchylist--with-nested-reorder&knob-Allow%20Item%20Movement=true&knob-Title=MLB%20Expanded%20List&knob-Default%20Selected%20Id=New%20York%20Mets_Pete%20Alonso&knob-Editing%20Style=multiple-nesting&knob-Page%20Size=lg&knob-isLoading=&knob-isLargeRow=&knob-hasSearch=true&knob-hasDeselection=true&knob-hasMultiSelect=&knob-hasVirtualList=&knob-lockedIds=["New%20York%20Mets_Jeff%20McNeil"]


packages/react/src/components/List/List.story.jsx
```diff
@@ -493,6 +493,7 @@ export const WithCheckboxMultiSelectionAndHierarchy = () => {
     };
 
     const handleSelection = (items, id) => {
+      action('handleSelect');
```
kevinsperrine, line 496:
```suggestion
      action('handleSelect')();
```
This action isn't firing in the actions panel.

kevinsperrine (approved):
Good catch. Just the one note on the action not firing.

davidicus (approved):

PR #3110, opened 2021-12-06
feat(repo):  move text strings into i18n prop

### LINKED ISSUE #1811: [repo] flatten i18n prop objects
### What package is this for?

- [x] React
- [ ] Angular

### Summary

Would like to propose that we flatten all i18n prop objects out to individual properties no longer within an object.

- [ ] All translatable strings should be standardized to be suffixed with `Text`. If an i18n prop does not have `Text` as it's suffix, we add it.
- [ ] Write and provide a codemod script to help consumers migrate from the old props to the new ones (stretch goal)

```diff
const propTypes = {
-  i18n: PropTypes.shape({
-    openEditorButton: PropTypes.string,
-    cardTitle: PropTypes.string,
-    description: PropTypes.string,
-  }),
+  openEditorButtonText: PropTypes.string,
+  cardTitleText: PropTypes.string,
+  descriptionText: PropTypes.string,
};
```

### Why?

It adds additional complexity to the codebase with no quantifiable value. Currently we have to merge defaultProps.i18n with passed i18n props to ensure defaults and passed parameters are all together.

https://github.com/carbon-design-system/carbon-addons-iot-react/blob/a6444bd4ee64aa7bb4c551784175f8adebdb39ea/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItem.jsx#L142

It's also something that we do differently than Carbon and causes our API to be inconsistent between components we modify from Carbon and ones that we re-export with no modifications.

This would be a breaking change and could be included with v3

kevinsperrine:
Closes #1811 

**Summary**

- moves all "non-falsy" strings into an `i18n` prop to bring consistency to all components

- deprecates the existing prop
- sets the existing prop to `undefined`
- add the prop to an `i18n` prop

- check that all non-falsey strings that should be translated are in an i18n and their existing prop is deprecated

- confirm no snapshot changes
- confirm no warning in test output


bjornalm (changes requested):
Nice work, 

just the small modification of adding the deprecated props to the merge. 

Did you check if any mdx files needed updating? I had a quick look but didn't find any.

packages/react/src/components/ColorDropdown/ColorDropdown.jsx
```diff
@@ -107,6 +119,8 @@ const ColorDropdown = ({
       );
     }
   }, []);
+  const mergedI18n = useMerged(defaultProps.i18n, i18n);
```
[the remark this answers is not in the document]
kevinsperrine: > `const mergedI18n = useMerged(defaultProps.i18n, {label, titleText}, i18n)`
> Then we can skip the additional "or-expressions" (label || mergedI18n.label) and only used the mergedI18n which can remain unchanged once we remove the old props.

Brilliant! I didn't even think of that (obviously). Yeah, I checked through the MDX files and didn't see anything, but I'll give them another check.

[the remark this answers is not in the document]
kevinsperrine: Updated.

bjornalm (approved):
Nice job!

packages/react/src/components/TileCatalog/TileCatalog.jsx
```diff
@@ -23,9 +23,17 @@ export const propTypes = {
...
-    nextPageText: PropTypes.string,
-    prevPageText: PropTypes.string,
+    pageOfPagesText: deprecate(
+      PropTypes.func,
+      `The prop \`pageOfPagesText\` has been deprecated for the \`SimplePagination\` component. It will be removed in the next major release. Please use \`i18n.pageOfPagesText\` instead.`
+    ),
+    nextPageText: deprecateString(),
+    prevPageText: deprecateString(),
+    i18n: PropTypes.shape({
+      pageOfPagesText: PropTypes.func,
```
davidicus, line 33: For our new strings are we removing these functions? I have not really thought out the repercussions of that but I do know it's something that we have been requested to do.

packages/react/src/hooks/useMerged.js
```diff
@@ -0,0 +1,8 @@
+import { useMemo } from 'react';
+import { merge } from 'lodash-es';
```
davidicus, line 2: I am not necessarily opposed to keeping this but is there something `merge` does that we can't achieve with just spreading the args onto an object literal?

packages/react/src/components/TileCatalog/TileCatalog.jsx
kevinsperrine: I assumed that would be tackled in a separate issue.

packages/react/src/internal/deprecate.js
```diff
@@ -26,8 +27,37 @@ export default function deprecate(propType, message) {
       );
     }
 
-    return propType(props, propName, componentName, ...rest); // eslint-disable-line consistent-return
+    return propType(props, propName, componentName, ...rest);
```
davidicus, line 30: This change and the `useMerge` helper are nice. Should we bring them into next as well in order to use them until the major transition.

packages/react/src/hooks/useMerged.js
kevinsperrine: The main thing is that `merge` will only replace the existing item if it's undefined whereas spreading will replace anything that's falsy. Using merge allows a parent to set a string as `""` (when we don't want to show a label, for example) without a child replacing it with it's default.

packages/react/src/internal/deprecate.js
kevinsperrine: Sure, I can easily cherry-pick them into next.

davidicus (approved):
This looks good. Just the one question about needing the `merge` function from lodash. Either way I think this is fine to merge.

packages/react/src/hooks/useMerged.js
```diff
@@ -0,0 +1,8 @@
...
+
+const useMerged = (...args) => {
+  return useMemo(() => merge({}, ...args), [args]);
```
stuckless, line 5: I could be wrong, but I don't think this doing what you think it's doing.

you are passing in an arg like `useMeged(obj1, obj2)` and you are spreading them for the merge, which is fine, but the useMemo(, [args]) is getting a new array instance every time, and likely it means that every place we are using useMerge we doing extra work to merge and memorize it but nothing is ever memorized.   (ie, if I pass in (obj1, obj2) then what I'm passing to the hook in the array is [[obj1, obj2]] and not [obj1, obj2].

it could be that `return useMemo(() => merge({}, ...args), args);` might work, but you'd need to unit test it.  Or maybe... `return useMemo(() => merge({}, ...args), [...args]);` but again, you'd have to create a unit test to verify this.

As I said, I could be wrong, and maybe the hook is smart enough to handle this, but I'm skeptical.  Could you confirm.

davidicus: @stuckless You are absolutely correct. Good catch. Because of the de-structuring here we are creating a new variable each time this is called. @kevinsperrine We can fix this by using a regular function and spreading the `arguments` variable. I have demo'd [here](https://codesandbox.io/s/fast-hill-jtu61).

bjornalm: Nice touch using the "old" arguments object, haven't seen that one in a while

PR #3112, opened 2021-12-07
[DateTimePicker]: Added validation for start/end time

### LINKED ISSUE #3134: [DateTimePicker] User should not be allowed to select a start date after end date
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

You can select a start time that's later than the end time. 

### To Reproduce

1. Go to https://carbon-addons-iot-react.com/?path=/story/1-watson-iot-datetime-picker--default
2. Click on DateTimePicker
3. Click on Custom Range
4. Click on the Absolute radio button
5. Change the start date to be after the end date

image

### Expected behavior

Apply button should be disabled until start time and end time for absolute time range is valid

### Environment/versions:

- OS: MacOS
- Browser: Chrome
- `carbon-addons-iot-react` version: 2.147.0-next.79

upcoming release?

date, the better!_

eosineye1:
Closes #3134
**Summary**

- When setting a custom time range for DateTimePicker one can select a start time that's later than the end time.

- Added validation to prevent a start time after the end time to be selected. An error message is displayed.

- Try to select a start time after the end time.
- Try to select an end time before the start time.

- Run unit tests 


bjornalm (changes requested):
Nice work and thanks for contributing!

Code looks good. Could you also write a unit test for this and create an issue in our  carbon-addons-iot-react repo before I approve it?

eosineye1:
@bjornalm Removed some code from old tests as the apply button is disabled for absolute range initially when start and end time are 00:00

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -229,10 +229,6 @@ describe('DateTimePicker', () => {
     expect(wrapper.find(`.${iotPrefix}--date-time-picker__field`).first().text()).toEqual(
       '2020-04-01 13:34 to 2020-04-06 11:49'
     );
-
-    wrapper.find(`.${iotPrefix}--date-time-picker__menu-btn-apply`).first().simulate('click');
-    jest.runAllTimers();
-    expect(dateTimePickerProps.onApply).toHaveBeenCalled();
```
kevinsperrine, line 235: Why was this removed? The portion above is clicking the increment button setting a valid date and time range, so apply should be enabled and the callback fired.

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -685,6 +680,40 @@ describe('DateTimePicker', () => {
...
+
+    // Change start time to 02:00 - start time is after end time
+    fireEvent.change(startTime, { target: { value: '02:00' } });
+    expect(applyBytton).toBeDisabled();
+
+    // Change end time to 02:00 - start time is after end time
+    fireEvent.change(endTime, { target: { value: '00:00' } });
+    expect(applyBytton).toBeDisabled();
+  });
+
```
kevinsperrine, line 716: Can you add a little more testing that changes the date and confirms situations where the start time before an end time is valid? ie. start: `02:00` and end: `01:00` is valid if startDate is `2021/10/01` and endDate is `2021/11/01`

kevinsperrine (changes requested):
Thanks for the PR. The code looks good so far. Just a few more minor changes around testing different time ranges and it'll be ready.

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -494,7 +490,6 @@ describe('DateTimePicker', () => {
     expect(screen.queryByText(i18nDefault.endTimeLabel)).not.toBeInTheDocument();
     // click apply
     fireEvent.click(screen.getByText(i18nTest.applyBtnLabel));
-    expect(screen.getAllByTitle(new RegExp(`.*${i18nTest.toLabel}.*`))[0]).toBeInTheDocument();
```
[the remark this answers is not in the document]
eosineye1: Added the test back and took your comments into consideration and updated the test.

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
eosineye1: Re-added the test to reflect changed made.

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
eosineye1: The date wouldn't matter because the date picker will always set the start date as the first date selected and set the last date as the second selected date. So the test mentioned is already covered in the current test cases.

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -787,13 +787,35 @@ const DateTimePicker = ({
...
   };
 
+  // Validate that the start time if less than the end time
+  const validateStartEndTime = (startTime, endTime) => {
+    const timefrom = new Date();
+    let temp = startTime.split(':');
+    timefrom.setHours(Number(temp[0]));
+    timefrom.setMinutes(Number(temp[1]));
+
+    const timeto = new Date();
```
kevinsperrine, line 797: You'll need to pass the start/end dates from the absolute range picker into this method, too. By comparing only against the current date with `new Date()` you're showing any `startTime` that's after `endTime` as invalid, but that's not always true. If I have chosen `11/1/2021` as my start date, and `11/3/2021` as my end date. I can have whatever startTime/endTime I want and all of them are valid.

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
kevinsperrine: Thanks, @eosineye1, but unfortunately, this isn't correct. The DateTimePicker can be passed a selected date range, so these tests aren't covering the validation correctly. See this [existing story in your preview](https://deploy-preview-3112--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-datetime-picker--selected-absolute). It marks the startTime as invalid, but it _is_ a valid time.

kevinsperrine (changes requested):

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
eosineye1: Thanks for letting me know to include the start and end date. I've taken that into consideration and refactored the code.

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -787,13 +797,54 @@ const DateTimePicker = ({
...
+      timefrom.setHours(Number(temp[0]));
+      timefrom.setMinutes(Number(temp[1]));
+
+      // Include end time in end date
+      const timeto = new Date(absoluteValues.end);
+      temp = endTime.split(':');
+      timeto.setHours(Number(temp[0]));
+      timeto.setMinutes(Number(temp[1]));
+
+      console.log(timefrom, timeto);
```
kevinsperrine, line 816: Let's remove the console.log.

```suggestion

```

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -685,6 +694,86 @@ describe('DateTimePicker', () => {
...
+    const defaultValues = {
+      timeRangeKind: PICKER_KINDS.ABSOLUTE,
+      timeRangeValue: {
+        startDate: '2020-04-01',
+        startTime: '12:34',
+        endDate: '2020-04-01',
+        endTime: '11:49',
+      },
+    };
+    const wrapper = mount(<DateTimePicker {...dateTimePickerProps} defaultValue={defaultValues} />);
```
kevinsperrine, line 707: Looking great. Thanks, @eosineye1. I should have mentioned this yesterday, but didn't think of it. Can you switch these tests over to using testing-library instead of enzyme. We're working on replacing all the existing enzyme tests and write all new tests in testing library.

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -685,6 +694,86 @@ describe('DateTimePicker', () => {
...
+    jest.runAllTimers();
+    expect(wrapper.find(`.${iotPrefix}--date-time-picker__field`)).toHaveLength(1);
+    expect(wrapper.find(`.${iotPrefix}--date-time-picker__field`).first().text()).toEqual(
+      '2020-04-01 12:34 to 2020-04-01 11:49'
+    );
+
```
kevinsperrine, line 713: We can make these test more concrete by checking for the existence of the "Invalid time format." message, so here, for example, we could check:
```jsx
expect(screen.getAllByText('Invalid time format.')).toHaveLength(2);
expect(screen.getByRole('button', {name: 'Apply'})).toBeDisabled();
```

packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
```diff
@@ -685,6 +694,86 @@ describe('DateTimePicker', () => {
...
+    expect(wrapper.find(`.${iotPrefix}--date-time-picker__field`).first().text()).toEqual(
+      '2020-04-01 13:34 to 2020-04-01 12:49'
+    );
+
+    wrapper.find(`.${iotPrefix}--time-picker__controls--btn.up-icon`).at(1).simulate('click');
+    jest.runAllTimers();
+    expect(wrapper.find(`.${iotPrefix}--date-time-picker__field`).first().text()).toEqual(
+      '2020-04-01 13:34 to 2020-04-01 13:49'
+    );
+
```
kevinsperrine, line 731: And then here the opposite of above:

```jsx
expect(screen.queryByText('Invalid time format.')).toBeNull();
expect(screen.getByRole('button', {name: 'Apply'})).not.toBeDisabled();

File: packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx [hunk 9]
```diff
@@ -685,6 +694,86 @@ describe('DateTimePicker', () => {
...
+        endTime: '11:49',
+      },
+    };
+    const wrapper = mount(<DateTimePicker {...dateTimePickerProps} defaultValue={defaultValues} />);
+    jest.runAllTimers();
+    expect(wrapper.find(`.${iotPrefix}--date-time-picker__field`)).toHaveLength(1);
+    expect(wrapper.find(`.${iotPrefix}--date-time-picker__field`).first().text()).toEqual(
+      '2020-04-01 12:34 to 2020-04-06 11:49'
+    );
+
```
kevinsperrine, line 753: same checks as comments in test above to check the invalid message does not appear with valid times and button is enabled (and vice-versa).

kevinsperrine (changes requested):
That's awesome, @eosineye1. Looks great while testing it in storybook now. Just a few comments to make the jest tests more concrete and this is ready to go. Thanks for this addition, It's a good improvement on the component.

eosineye1:
Hi @kevinsperrine, I am having issues adding default dates and times to the DateTimePicker using the testing library. I am currently going through the docs, but can't find resources about passing in default values. If you know a way to do so using the testing library could you point me in the right direction?

kevinsperrine:
> 

Sure thing. You can pass them in the `deafultValue` prop when render is called. You can look at some of the other tests. For example: 

https://github.com/carbon-design-system/carbon-addons-iot-react/blob/1f3e2513117d4d98fdc811e14a285a5f50f0b110/packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx#L531

https://github.com/carbon-design-system/carbon-addons-iot-react/blob/1f3e2513117d4d98fdc811e14a285a5f50f0b110/packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx#L552

Or an example of dynamically typing in the time boxes: 

https://github.com/carbon-design-system/carbon-addons-iot-react/blob/1f3e2513117d4d98fdc811e14a285a5f50f0b110/packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx#L669

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -787,13 +797,52 @@ const DateTimePicker = ({
...
   // on change functions that trigger a absolute value update
   const onAbsoluteStartTimeChange = (pickerValue, evt, meta) => {
     setAbsoluteStartTimeInvalid(meta.invalid);
+    const invalidStartEndTime = invalidAbsoluteDate(
+      pickerValue,
+      absoluteValue.endTime,
+      absoluteValue
+    );
+    setAbsoluteStartTimeInvalid(meta.invalid || invalidStartEndTime);
+    setAbsoluteEndTimeInvalid(absoluteEndTimeInvalid && invalidStartEndTime);
```
bjornalm, line 832: This will cause a bug if you:
1. Select an absolute date range 
2. Input an invalid number like `9999` in the "End time" input
3. Enter a valid number in the "Start time" input.
4. The "End time" input is still invalid but will no longer be shown as invalid because of the `absoluteEndTimeInvalid && invalidStartEndTime` where `invalidStartEndTime` is false

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -787,13 +797,52 @@ const DateTimePicker = ({
...
   const onAbsoluteEndTimeChange = (pickerValue, evt, meta) => {
     setAbsoluteEndTimeInvalid(meta.invalid);
+    const invalidStartEndTime = invalidAbsoluteDate(
+      absoluteValue.startTime,
+      pickerValue,
+      absoluteValue
+    );
+    setAbsoluteEndTimeInvalid(meta.invalid || invalidStartEndTime);
+
+    setAbsoluteStartTimeInvalid(absoluteStartTimeInvalid && invalidStartEndTime);
```
bjornalm, line 844: Same bug here

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -787,13 +797,52 @@ const DateTimePicker = ({
...
+      pickerValue,
+      absoluteValue.endTime,
+      absoluteValue
+    );
+    setAbsoluteStartTimeInvalid(meta.invalid || invalidStartEndTime);
+    setAbsoluteEndTimeInvalid(absoluteEndTimeInvalid && invalidStartEndTime);
     changeAbsolutePropertyValue('startTime', pickerValue);
   };
   const onAbsoluteEndTimeChange = (pickerValue, evt, meta) => {
     setAbsoluteEndTimeInvalid(meta.invalid);
```
bjornalm, line 836: Can be removed

bjornalm (changes requested):

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
eosineye1: Logic has been updated to account for this and to validate each TimrSpinner individually. Also, a test has been added to test this

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -787,13 +795,60 @@ const DateTimePicker = ({
     setAbsoluteValue(newAbsolute);
   };
 
+  const isValidDate = (d, time) => {
```
bjornalm, line 798: We are getting there :-) This check will however incorrectly return true for any time value following `nn:nn`, e.g. the value `99:99`. If you instead of splitting the time before and and pass in the full time string you could use the same regex used in the TimePickerSpinner.jsx:

```
const isValid24HoursRegex = /^([01][0-9]|2[0-3]):([0-5][0-9])$/;
const isValidTime = isValid24HoursRegex.test(time);
```

Ideally this would not be duplicated but in a separate file and imported int both components..

bjornalm (changes requested):

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
eosineye1: Utilizing the regrex in the isValidDate function not. I was not sure where to put the regrex and was not sure where to put a new file if I made one for me to import it for both components, so I copied it into the DateTimePicker component file.

bjornalm (approved):
Nice job on this one!
There are still some edge cases not covered by the tests but I think we can merge this now since the functionality is ok. Just commit the suggested spelling fix.

davidicus (approved):
Should be good once Kevin approves

kevinsperrine (commented):
I swear, I thought I did this yesterday, but I'm not seeing the comment now. Nice work, @eosineye1. Just a few linting errors and this is ready to merge!

```carbon-addons-iot-react: /home/runner/work/carbon-addons-iot-react/carbon-addons-iot-react/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
carbon-addons-iot-react:   598:7   error  'invalidEndDate' was used before it was defined    no-use-before-define
carbon-addons-iot-react:   601:7   error  'invalidStartDate' was used before it was defined  no-use-before-define
carbon-addons-iot-react:   800:37  error  Unexpected use of 'isNaN'                          no-restricted-globals```

**Review by kevinsperrine — APPROVED** [review]

File: packages/react/src/components/DateTimePicker/DateTimePicker.jsx [hunk 14]
```diff
@@ -701,13 +701,15 @@ const DateTimePicker = ({
...
         if (!absolute.hasOwnProperty('start')) {
-          absolute.start = dayjs(absolute.startDate).valueOf();
+          absolute.start = dayjs(`${absolute.startDate} ${absolute.startTime}`).valueOf();
         }
         if (!absolute.hasOwnProperty('end')) {
-          absolute.end = dayjs(absolute.endDate).valueOf();
+          absolute.end = dayjs(`${absolute.endDate} ${absolute.endTime}`).valueOf();
         }
         absolute.startDate = dayjs(absolute.start).format('MM/DD/YYYY');
+        absolute.startTime = dayjs(absolute.start).format('HH:mm');
```
stuckless, line 710: Seem wrong to me that we are hard coding the date/time formats.  We should be looking at a way to externize this.  I don't think it is simple enough to expose a "pattern" since, what is the format of the pattern?  we can't have different components using different kinds of patterns... some using moment, some using dayjs, etc.

stuckless: I think PAL needs to have a standardize set of of API around globalization and that object should be made available to components that need to parse/format data, get access to masks, etc.

In graphite when we use "formats" we use named formats like "short_date", "long_date", "date_fully_month", etc, we translate those into the appropriate masks based on user locale.

kevinsperrine: Thanks, Sean. We've talked about this internally and have started creating a document defining data formatting, i18n, data loading, etc--all the main items we've talked about over the last few months. We're moving that direction. This particular component was written 2 years ago, and this PR is to add time validation to the component as it already exists. The refactors based on new technical design will come later.

davidicus (approved):

PR #3113, opened 2021-12-07
feat(usescrollintoview): hook for infinite loading functionality

### LINKED ISSUE #3054: [Repo] Create infinite loading helper
We have several components that render a list of sub-items. A user needs to have a mechanism to call for additional items as to not have to call for all data up front. 

This helper function would need to implement something like an intersection observer that we place on the last (close to last) item in a list that would call a callback when that item is scrolled into viewport.

I think a hook that returns a callback ref should serve this purpose but whoever takes this on can determine the best approach.

- [ ] create custom hook
- [ ] create issues to track implementation of hook in each list/ table/ dropdown? component

davidicus:
Closes #3054

**Summary**

This PR adds a custom hook for creating an intersection observer that can be placed on items in a list in order to implement infinite scrolling. 

- created useScrollIntoViewCallback.jsx and associated test

- tests should pass

### DIFF
--- a/packages/react/src/hooks/useScrollIntoViewCallback.jsx
+++ b/packages/react/src/hooks/useScrollIntoViewCallback.jsx
@@ -1,0 +1,48 @@
+import { useRef, useCallback, useEffect } from 'react';
+
+/**
+ *
+ * @param {Function} callback - required callback function to call when intersecting
+ * @param {Boolean} loading - optional loading boolean to return without setting up observer
+ * @returns {Object} scrollItemRef
+ * @example const { scrollItemRef } = useScrollIntoViewCallback(callback, loading);
+ */
+export default function useScrollIntoViewCallback(callback, loading) {
+  const observer = useRef();
+  useEffect(
+    () => () => {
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/useScrollIntoViewCallback.jsx
```diff
@@ -0,0 +1,48 @@
...
+      });
+      // Once component mounts tell our observer to observe it
+      if (node && observer.current) {
+        observer.current.observe(node);
+      }
+    },
+    [loading, callback]
+  );
+
+  return { scrollItemRef };
```
bjornalm, line 47: Why is the function wrapped in an object?

bjornalm (approved):
This looks good to me. I'm not familiar with the IntersectionObserver but I trust you are now :-) 

Do you wanna use this one or the one from https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3114 and pick the best ideas from both?

davidicus:
closing this PR in favor of the hook created for #3114

PR #3114, opened 2021-12-07
feat(repo): replace react-visibility-sensor and react-sizeme with hooks

### LINKED ISSUE #2364: [Card] Getting intermittent error when unmounting cards from dashboards
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

It's hard to say why this is occurring, but in our logs occasionally while unmounting a dashboard we see this:
image

It seems like something to do with the Size measurement of the card

### To Reproduce

It's pretty hard to reproduce, hopefully this is an error in an underlying library that we could search for.

kevinsperrine:
Closes #2364 

**Summary**

The errors on unmount are stemming from a dependency of react-sizeme, so to fix those and clean up the code, I've replaced react-sizeme with a ResizeObserver hook, and react-visibility-sensor with a IntersectionObserver hook. Many of these changes are whitespace based, so it's easier to read this PR while [ignore whitespace changes](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3114/files?diff=split&w=1). Additionally, I did split the commits to make it easier to follow, so doing [one commit at a time](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3114/commits/29c6219a11a5cc37775736c52ae8281911582c8c?diff=split&w=1) will be easier, too. 

I used this opportunity to explore how a IntersectionObserver hook could be used to build an infinite loader. The new list story demonstrates how this might work, but we still need to combine it with another components that would handle data loading, pagination, etc. 

- remove all usages of react-sizeme
- remove all usages of react-visibility-sensor
- replace those with custom IntersectionObserver and ResizeObserver hooks
- refactor importing of { DataTable} to import components directly from carbon-components-react
- - `import { TableBody } from 'carbon-components-react';` instead of `import { DataTable } from 'carbon-components-react`
- - Is tree shaking smart enough to do this automatically from previous `const { TableBody } = DataTable;` usages?

- The [card with render prop story](https://deploy-preview-3114--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card--basic-with-render-prop) shows the correct size

- The [HotspotEditorModal with existing hotspots story](https://deploy-preview-3114--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-hotspoteditormodal--with-existing-hotspots)
- - resize the window and confirm the image resizes and hotspots stay in the correct place

- The [basic table story](https://deploy-preview-3114--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--basic-dumb-table)
- - turn on `hasPagination` knob
- - make the window very narrow (less than 500px)
- - confirm the pagination text disappears and only the buttons remain

- The [basic table story with shouldLazyLoad knob on](https://deploy-preview-3114--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--basic-dumb-table&knob-title=Munich%20Building&knob-isEditable=&knob-isLoading=&knob-Trend%20color=&knob-Trend=up&knob-Threshold%20label%20(%3E%2080%)=Good&knob-values={%22temperature%22:35.35,%22humidity%22:99}&knob-Choose%20how%20text%20should%20wrap%20witin%20columns%20(options.wrapCellText)=always&knob-Select%20with/without%20section=Title&knob-Type%20of%20Table=Table&knob-Title%20shown%20in%20bar%20above%20header%20row%20(secondaryTitle)=Basic%20`dumb`%20table&knob-Alternate%20colors%20in%20table%20rows%20(useZebraStyles)=&knob-Enables%20choosing%20which%20columns%20are%20visible%20or%20drag-and-drop%20reorder%20them%20(options.hasColumnSelection)=&knob-Enables%20choosing%20which%20columns%20are%20visible%20or%20drag-and-drop%20reorder%20them%20and%20adds%20callback%20to%20manage%20which%20columns%20are%20available%20to%20the%20table%20(options.hasColumnSelectionConfig)=&knob-Enables%20sorting%20the%20table%20by%20multiple%20dimensions%20(options.hasMultiSort)=&knob-Sets%20the%20height%20of%20the%20table%20rows%20(size)=lg&knob-Aggregates%20column%20values%20and%20displays%20in%20a%20footer%20row%20(options.hasAggregations)=true&knob-Enables%20filtering%20columns%20by%20value%20(options.hasFilter)=true&knob-Enables%20only%20loading%20table%20rows%20as%20they%20become%20visible%20(options.shouldLazyRender)=true&knob-Enables%20pagination%20for%20the%20table%20(options.hasPagination)=&knob-Enables%20resizing%20of%20column%20widths%20(options.hasResize)=&knob-Enables%20expanding%20rows%20to%20show%20additional%20content%20(options.hasRowExpansion)=&knob-Enables%20rows%20to%20have%20nested%20rows%20within%20(options.hasRowNesting)=&knob-Enable%20or%20Disable%20selecting%20single,%20multiple,%20or%20no%20rows%20(options.hasRowSelection)=multi&knob-Enable%20search%20as%20typing%20(default)%20or%20only%20on%20%27Enter%27%20(options.hasFastSearch).=true&knob-Enable%20searching%20on%20the%20table%20values%20(options.hasSearch)=&knob-Enable%20sorting%20columns%20by%20a%20single%20dimension%20(options.hasSort)=&knob-Preserve%20column%20widths%20when%20resizing%20(options.preserveColumnWidths)=&knob-Removes%20table-layout:fixed%20to%20allow%20resizable%20tables%20(options.useAutoTableLayoutForResize)=&knob-Disable%20the%20table%20toolbar%20(view.toolbar.isDisabled)=&knob-Show%20the%20loading%20state%20for%20the%20table%20(view.table.loadingState.isLoading)=&knob-The%20number%20of%20skeleton%20rows%20to%20be%20included%20in%20the%20loading%20state%20(view.table.loadingState.rowCount)=7&knob-The%20number%20of%20skeleton%20columns%20to%20be%20included%20in%20the%20loading%20state%20(view.table.loadingState.columnCount)=6&knob-i18n.columnSelectionConfig=__Manage%20columns__)
- - confirm new rows are loaded dynamically as you scroll down the page

- The [table with create & save view management ](https://deploy-preview-3114--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--table-example-with-create-save-views)
- - click "View: Filters and search"
- - click "Search view"
- - note the `View: ` prefix should be hidden--this is because the search box takes up extra space, so the resize observer is checking that and setting `isCompact` true to hide that prefix.

- The [List with infinite scrolling story](https://deploy-preview-3114--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-list--with-infinite-scroll) should load new items as you scroll down the list
- - repeat this check with isVirtualList:true

- regressions should be covered by all the above tests, but double check I didn't miss a usage of react-visibility-sensor or react-sizeme somewhere in the diffs


packages/react/src/hooks/useVisibilityLoader.js
```diff
@@ -0,0 +1,75 @@
+import { useEffect, useLayoutEffect, useState } from 'react';
+import warning from 'warning';
+
+import { browserSupports } from '../utils/componentUtilityFunctions';
+
+const useVisibilityLoader = (ref, { maxToLoad = 1, loadBatchSize = 20 } = {}) => {
+  if (!browserSupports('IntersectionObserver')) {
+    warning(
+      !__DEV__,
+      'The current browser does not support ResizeObserver. You will need to include a ResizeObserver polyfill for this component to function properly.'
```
davidicus, line 10: need to update the message to reflect particular observer or generalize if its all or nothing support wise.

packages/react/src/hooks/useVisibilityLoader.js
```diff
@@ -0,0 +1,75 @@
...
+  const [moreToLoad, setMoreToLoad] = useState(true);
+  const [nextVisibleIndex, setNextVisibleIndex] = useState(0);
+
+  useEffect(() => {
+    setMoreToLoad(nextVisibleIndex < maxToLoad);
+  }, [maxToLoad, nextVisibleIndex]);
+
+  useLayoutEffect(() => {
+    if (ref.current && moreToLoad) {
+      const observer = new IntersectionObserver(
```
davidicus, line 26: I think this will create a new intersection observer with each render. We may want to store the observer in a ref 
in order to clean up on each render.

packages/react/src/hooks/useVisibilityLoader.js
```diff
@@ -0,0 +1,75 @@
...
+          });
+        },
+        {
+          rootMargin: '300px 0px 300px 0px',
+        }
+      );
+
+      observer.observe(ref.current);
+
+      return () => {
```
davidicus, line 63: I think we need to move this out of the current conditional and instead check if observer ref exist. I suspect that this would not clean up if `moreToLoad` is false.

kevinsperrine: yep. good call.

packages/react/src/hooks/useVisibilityLoader.js
kevinsperrine: Done.

packages/react/src/hooks/useVisibilityLoader.js
kevinsperrine: updated.

packages/react/src/hooks/useVisibilityLoader.js
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @param {*} ref html element to check if it's visible before rendering more items
+ * @param {object} options
+ * @returns Array an array containing [isVisible: bool | arrayOf(bool), moreToLoad: bool]
+ */
+const useVisibilityLoader = (ref, options) => {
+  if (!browserSupports('IntersectionObserver')) {
+    warning(
+      !__DEV__,
+      'The current browser does not support IntersectionObserver. You will need to include a IntersectionObserver polyfill for this component to function properly.'
```
stuckless, line 37: Since this is replacing the Visibility Sensor does the visibility sensor require the polyfill?  ie would this be a breaking change when this is applied?

packages/react/src/components/List/List.jsx
```diff
@@ -92,6 +93,12 @@ const propTypes = {
   handleLoadMore: PropTypes.func,
   /** RowIds for rows currently loading more child rows */
   loadingMoreIds: PropTypes.arrayOf(PropTypes.string),
+  /** does this list use infinite scrolling */
+  isInfiniteScroll: experimental(PropTypes.bool),
+  /** callback to fire when the last element in an infinite scroll list is visible */
+  onInfiniteScroll: experimental(PropTypes.func),
```
stuckless, line 99: can you document the function params

packages/react/src/components/List/ListContent/ListContent.jsx
```diff
@@ -68,6 +69,12 @@ const propTypes = {
   /** icon can be left or right side of list row primary value */
   iconPosition: PropTypes.oneOf(['left', 'right']),
   selectedItemRef: HtmlElementRefProp,
+  /** does this list use infinite scrolling */
+  isInfiniteScroll: PropTypes.bool,
+  /** callback to fire when the last element in an infinite scroll list is visible */
+  onInfiniteScroll: PropTypes.func,
```
stuckless, line 75: what are the params

packages/react/src/hooks/useSizeObserver.js
```diff
@@ -0,0 +1,23 @@
+import useResizeObserver from 'use-resize-observer';
+import warning from 'warning';
+
+import { browserSupports } from '../utils/componentUtilityFunctions';
+
+const useSizeObserver = (ref) => {
+  if (!browserSupports('ResizeObserver')) {
```
stuckless, line 7: is the only purpose of this function to add the warning?

packages/react/src/hooks/useVisibilityLoader.js
```diff
@@ -0,0 +1,95 @@
...
+  const { hasMoreToLoad, isLoading, onVisible, intersectionObserverOptions } = merge(
+    {},
+    defaultOptions,
+    options
+  );
+
+  /* eslint-disable react-hooks/rules-of-hooks */
+  const observerRef = useRef(null);
+  const [isVisible, setIsVisible] = useState(false);
+  const intersectionCallback = useCallback(
```
stuckless, line 51: This is a very common "optimization" flaw when using hooks. The use of useCallback here is an attempt to make the code more efficient, but in fact, it makes it less efficient.

The passed function to useCallback() is going to get created every time.  Just now with useCallback it accepts the function and array and figures out if it should return the newly created function or a memorized one, wasting CPU cycles since the end result is that the callback function doesn't need to be memoized.

given that the useEffect is monitoring 2 of the 3 props here, this would actually be better if you didn't use useCallback and just put isVisible in the useEffect.  The end result is the same with less CPU cycles.

If you disagree... Read this.  https://kentcdodds.com/blog/usememo-and-usecallback

I'm curious why you had to disable rules-of-hooks.  Was it warning about this very thing?

stuckless (changes requested):

packages/react/src/hooks/useSizeObserver.js
kevinsperrine: That's the correct. And, that's why I suppressed the "rules-of-hooks" eslint message. There's no concept of a ErrorBoundary within a hook, so I thought it was pretty to at least print the warning rather than throwing an error.

packages/react/src/components/List/ListContent/ListContent.jsx
kevinsperrine: Currently, In this example there aren't any params passed. This was added purely as an example to get your comments and feedback about how we should build this feature. There a more robust example that is built on top of this PR here (https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3140). In that example, I pass `pageSize`, `start`, and `retry` parameters. `pageSize` is the number of items to load, `start` is the starting index to load from, and `retry` is a reference to most recently called attempt to load to handle retry logic.

packages/react/src/hooks/useVisibilityLoader.js
kevinsperrine: Good catch. I can adjust that.

packages/react/src/hooks/useVisibilityLoader.js
kevinsperrine: I thought I commented on this already, but I'm not seeing my response now, so I'm sorry if this is a duplicate. Yes, this will be a breaking change for browsers that don't support IntersectionObserver. I just check the internals of react-visibility-sensor and it's using onScroll detection. We can either lazy-load the polyfill here in this check or push this change to v3 to prevent the potential breaks.

davidicus: IntersectionObserver is supported in all modern browsers so this should just be a safeguard. 
image

packages/react/src/hooks/useVisibilityLoader.js
```diff
@@ -86,7 +79,7 @@ const useVisibilityLoader = (ref, options) => {
         observerRef.current = null;
       }
     };
-  }, [isVisible, ref, intersectionObserverOptions, intersectionCallback, hasMoreToLoad]);
+  }, [isVisible, ref, intersectionObserverOptions, hasMoreToLoad, isLoading, onVisible]);
```
bjornalm, line 82: While we are getting nerdy on the subject, this useEffect will still be called on every render right? Because of the fact that we use lodash merge to create the referenced variables `hasMoreToLoad, isLoading, onVisible, intersectionObserverOptions` and they will be created from a new object each time. Our linting is still gonna complain if we remove or use an empty dependency array so perhaps this is the way to go anyway, but just wanted to clarify.

kevinsperrine: That's a good point. I didn't think about the merge effecting that. Would a `useMemo` handle that?

bjornalm: sorry, removed my comment. Wanted to modify it but got interrupted by the baby. I think only the non primitives like intersectionObserverOptions are problematic...

bjornalm: just skip the lodash merge for those

bjornalm: not sure if the default props are recreated on render...

stuckless: I'm having a sense of Deja Vu :)   Recently there was a need in Graphite for a data grid that supports the potential for infinite rows and columns.  And by that, it should perform well with large datasets both vertically and horizontally.  The developer went away (never talked to me) and came back a couple of weeks later with the finished solution.  I loaded it up with 200 rows and 100 columns and it crawled to a halt.  We finally talked and he went away and created what was needed and it's blazingly fast for both infinite rows and infinite columns.  Part of the issue is that sometimes I don't think devs are fully aware of what they are solving (and I've been there many times myself).

So let's look at this... what are we solving here... because I see the code and I see it trying to solve 2 things and neither of which is going to help the current table perform better. 

When I worked with the Graphite dev on building the data-grid, I sat down and drew up this diagram, so that they could better visualize what they were doing (or trying to do).

image

For grid the biggest performance bottleneck is going to be rendering and it's typically rendering information that is not visible.  So in the case of the graphite data grid it was rendering columns and cells for information that was not visible.

The other thing is that loading sequential data and rendering sequential data shouldn't be enforced.  Today, the Table has a dropdown for "Pages".  Imagine if when the table first loaded, and you click the dropdown it showed 1 and 2, and then later when you click 2, and then after it loaded, you clicked the dropdown and it showed, 1, 2,  and 3, you click 3, it loads, and then it shows 1, 2, 3, and 4.  If it worked like that you'd think that was a silly implementation of data paging.

Loading data via some loadMore really just implies a sequential data model and that limits how you can load the data.  Also, tying it to a visibility trigger might not be the best approach.  Fundamentally you are building an incremental loading interface in such a way because your components are built a certain way.  The data model of the table and datalist is flawed and I think these need to be addressed before implementing a visibility loader.  And I think that if the visibility load is just loading more data and that data, both horizontally and vertically is being rendered into the DOM when its not visible, then I don't think it's solving any problems with the table.  This where a data contract is really needed that all component can adhere to.  In the simplest form... you need...

* totalCount (how many items does it know about)
* isItemLoaded(index) (check if an item is loaded)
* itemsAreLoading (flag indicating that we are busy loading items)
* loadItems(start, size) (async function that loads a range of items)

This is the bare minimum, but, using this information you can...
* totalCount can allow us to create scrollbars that reflect the total amount of items
* isItemLoaded(index) can be used to check if an item is loaded and show some sort of loading indicator if not...
* itemsAreLoading property can be used to signal that the "loader" is busy loading items
* loadItems(start, size) can be used to get a range of items that you need to render

Using a data contract you can get items without actually managing them internally, but, this still doesn't solve the lazy rendering of item data.  ie, that is a completely separate solution and has nothing to do with loading data, it just about how we can only render into the DOM the elements we need, no matter how much data we might have access to.

So, I think that the first step, in my mind is really fixing the data model for the Table so that it is really a data model, and NOT an hybrid data/ui model as it stands today.  Data really needs to be separated from the UI model.

kevinsperrine: Thanks, @stuckless. This is great feedback and I completely agree. That's what I'm working toward in these two PRs. They aren't finished solutions, but are steps toward a solution by building these new hooks while maintaining existing functionality, creating some potential examples, and gathering your feedback. I would appreciate you taking a look at the [useDataLoader hook](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3140/files#diff-20e8db082c0a53fafbec9a5df135593fff9f6f00ebff08c10c0ea200930d0253R9) and [story](https://deploy-preview-3140--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-list--with-infinite-scroll-data-loader) in [this PR](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3140). It's built on this PR, so you'll see many of the same changes, but just takes them a bit farther. It's using the IntersectionObserver to load/unload data as it becomes visible on the screen, doesn't worry about managing data at all, and leaves that up to the consumer in the onLoadData callback. That callback is passed the `pageSize` and `start` parameters you mentioned and needs to return `data`, `loading` and `hasMoreData` flags, but I can easily re-work that to be based off the `totalCount` variable you mention above.

stuckless: @kevinsperrine I think we need to decouple the concepts of lazily loading data and lazily rendering data.  These are 2 different things and they don't depend on each other.  But more importantly, I feel like the data model of the table should be refactored because right now, we are building "interfaces" that sort of work with the current data model, but, these really aren;t going to benefit me, personally.  I will look at the use data hook, but I feel like existing data loading is missing the important concept that data doesn't need to be loaded sequentially... and that UI components should not be managing holding on to references of complete data sets.

bjornalm:
image
style attr is correct but not the actual result

packages/react/src/components/List/ListContent/ListContent.jsx
```diff
@@ -68,6 +69,12 @@ const propTypes = {
   /** icon can be left or right side of list row primary value */
   iconPosition: PropTypes.oneOf(['left', 'right']),
   selectedItemRef: HtmlElementRefProp,
+  /** does this list use infinite scrolling */
+  isInfiniteScroll: PropTypes.bool,
```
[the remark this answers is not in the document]
bjornalm: Found the answer in your story example :-)

packages/react/src/hooks/useVisibilityLoader.js
```diff
@@ -0,0 +1,88 @@
...
+
+const defaultOptions = {
+  /** are there more items available to load */
+  hasMoreToLoad: true,
+  /** is the application already loading new data */
+  isLoading: false,
+  /** a callback that will be triggered when the ref element becomes visible */
+  onVisible: () => {},
+  /* options passed to IntersectionObserver, see: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API */
+  intersectionObserverOptions: {
```
stuckless, line 15: Not a fan of exposing the intersection options.  The fact that we are using intersection observer should be private, and we should craft a set of parameters that we end up rewriting for the intersection observer vs exposing the entire "options" to the developer and forcing them to understand what an intersection object is and how it works, etc.  Sadly even after I read the document for intersection observer it's not clear to me what rootMargin is doing :(

stuckless (approved):

packages/react/src/hooks/useVisibilityLoader.js
kevinsperrine: That's a good point. I can simply that so we're not using all IntersectionObserver options. It may not be necessary to expose any of them, but I included them to give devs more flexibility. By exposing the rootMargin for example, it gives them control over how large the "visible" area is and how many items are being "pre-loaded" off-screen as the user is scrolling.

Yes, the docs are extremely unclear. A positive margin expands the intersection area outside the bounds of the root object, so the callback is trigger sooner where a negative margin shrinks the area inside the bounds of the root element. So, `100px 0px 100px 0px' is expanding the area 100px above and 100px below the root area, so that the element is marked as intersecting 100px sooner than it would if the margin were "0px" (the default).

packages/react/src/hooks/useSizeObserver.js
```diff
@@ -0,0 +1,23 @@
...
+    warning(
+      !__DEV__,
+      'The current browser does not support ResizeObserver. You will need to include a ResizeObserver polyfill for this component to function properly.'
+    );
+    return [{ height: 0, width: 0 }, ref];
+  }
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const { height = 0, width = 0 } = useResizeObserver({
```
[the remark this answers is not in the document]
kevinsperrine: Good point. Adding the ability to pass a default is probably needed.

packages/react/src/hooks/useVisibilityLoader.js
```diff
@@ -0,0 +1,88 @@
...
+    rootMargin: '100px 0px 100px 0px',
+  },
+};
+
+/**
+ * A simple hook to return a bool telling if the given element is visible on the screen. It can be
+ * used to load a single item or multiple items. If maxToLoad is greater than 1, the first item in
```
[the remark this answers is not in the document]
kevinsperrine: It is not. I went through a few iterations of how this should be built before I landed her. I'll adjust.

packages/react/src/components/HotspotEditorModal/HotspotEditorModal.jsx
```diff
@@ -303,6 +303,7 @@ const HotspotEditorModal = ({
 
   const hasNonEditableContent = React.isValidElement(selectedHotspot?.content);
   const mergedI18n = useMemo(() => ({ ...defaultProps.i18n, ...i18n }), [i18n]);
+  const [modalContentSize, modelContentRef] = useSizeObserver(useRef(null));
```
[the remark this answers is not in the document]
kevinsperrine: None, really. Could easily not pass it and use the one returned from useResizeObserver.

packages/react/src/hooks/useSizeObserver.js
```diff
@@ -0,0 +1,23 @@
+import useResizeObserver from 'use-resize-observer';
+import warning from 'warning';
+
+import { browserSupports } from '../utils/componentUtilityFunctions';
+
+const useSizeObserver = (ref) => {
```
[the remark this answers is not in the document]
stuckless: What is the purpose of this hook?  is it to measure the size of an element after it has rendered?

And more importantly, since this is replacing the sizeme with this (I think), what are the benefits of this vs sizeme?  ie what was it about sizeme that made it unsuitable and needed to be replaced?

packages/react/src/hooks/useVisibilityObserver.test.js
```diff
@@ -0,0 +1,141 @@
...
+    expect(result.current).toEqual([false]);
+    expect(onChange).not.toHaveBeenCalled();
+
+    const newRef = React.createRef();
+    newRef.current = document.createElement('div');
+    rerender([newRef, initialProps[1]]);
+
+    expect(observer.unobserve).not.toHaveBeenCalled();
+    expect(observer.observe).toHaveBeenCalledTimes(2);
+    expect(observer.disconnect).toHaveBeenCalledTimes(2);
```
bjornalm, line 139: Why has `observer.disconnect` been called twice?

bjornalm (changes requested):
Nice work, I don't remember much from my review before the holidays so this was almost new code again ;-) 

I only have some questions and smaller change request.

davidicus (commented):
This is good to go once @bjornalm's comments are addressed.

davidicus:
@kevinsperrine bump to address Bjorn's comments

packages/react/src/components/Table/TableViewDropdown/TableViewDropdown.jsx
```diff
@@ -71,121 +71,127 @@ const defaultProps = {
...
-  const MyDropDown = overrides?.dropdown?.component || Dropdown;
-  const MyTableViewDropDownItem = overrides?.dropdownItem?.component || TableViewDropdownItem;
-  const [containerSize, containerRef] = useSizeObserver({ initialWidth: 200 });
-  const [dropdownRef, updateTitle] = useDropdownTitleFixer();
+    const mySelectedItem = allItems.find((item) => item.id === selectedViewId) || viewAllItem;
+    const MyDropDown = overrides?.dropdown?.component || Dropdown;
+    const MyTableViewDropDownItem = overrides?.dropdownItem?.component || TableViewDropdownItem;
+    const [containerSize, containerRef] = useSizeObserver({ initialWidth: 200 });
+    // destructure the array as an object by index
+    const { 1: updateTitle } = useDropdownTitleFixer(ref?.current);
```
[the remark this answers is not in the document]
kevinsperrine: I ended up changing this in a later commit. The useDropdownTitleFixes doesn't take a param. It returns the ref from the hook. I initially added it because I was getting a "forwardRef" error, but I realized that was because of an override component being used in the test/story that was just a function component. So passing the ref isn't needed after I fixed that with forwardRef.

packages/react/src/components/IconDropdown/dropdownHooks.js
```diff
@@ -10,22 +10,29 @@ const { prefix } = settings;
...
  * @returns updateTitle function
  */
-export const useDropdownTitleFixer = () => {
-  const ref = useRef(null);
+export const useDropdownTitleFixer = (ref) => {
+  const realRef = useRef(null);
+
+  useEffect(() => {
+    if (ref?.current) {
+      realRef.current = ref.current;
```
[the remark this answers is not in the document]
kevinsperrine: see comment above: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3114#discussion_r787083812

packages/react/src/hooks/useVisibilityObserver.js
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * If true, stop tracking the element after it has become visible otherwise, it will continue to
+   * track the element and trigger onChange each time it changes visibility
+   */
+  unobserveAfterVisible: false,
+};
+
+/**
+ * A simple hook to return a bool telling if the given element is visible on the screen. It can be
+ * used to load a single item or multiple items. If maxToLoad is greater than 1, the first item in
```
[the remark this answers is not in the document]
kevinsperrine: Yeah, remove the irrelevant portions.

packages/react/src/hooks/usePrevious.jsx
```diff
@@ -6,8 +6,9 @@ import { useEffect, useRef } from 'react';
  * @param {*} value something you want to store the value of between renders to do comparisons against
  * @returns * the value passed to the hook on the previous render
  */
-export function usePrevious(value) {
-  const ref = useRef();
+export function usePrevious(value, initialValue) {
```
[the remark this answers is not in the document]
kevinsperrine: good call. Done.

packages/react/src/components/List/VirtualListContent/VirtualListContent.jsx
```diff
@@ -124,7 +124,7 @@ const defaultProps = {
   selectedIds: [],
   testId: 'list',
   toggleExpansion: () => {},
-  virtualListRef: undefined,
+  virtualListRef: null,
```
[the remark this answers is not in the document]
kevinsperrine: accidental change. Reverted.

packages/react/src/components/Table/TableViewDropdown/TableViewDropdown.jsx
```diff
@@ -130,15 +130,17 @@ const TableViewDropdown = ({
   const mySelectedItem = allItems.find((item) => item.id === selectedViewId) || viewAllItem;
   const MyDropDown = overrides?.dropdown?.component || Dropdown;
   const MyTableViewDropDownItem = overrides?.dropdownItem?.component || TableViewDropdownItem;
-  const [dropdownRef, updateTitle] = useDropdownTitleFixer();
+  const [containerSize, containerRef] = useSizeObserver({ initialWidth: 200 });
+  // destructure the array as an object by index
```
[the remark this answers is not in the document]
kevinsperrine: done.

packages/react/src/components/Table/TableViewDropdown/TableViewDropdown.jsx
```diff
@@ -130,15 +130,17 @@ const TableViewDropdown = ({
...
+  const [ref, updateTitle] = useDropdownTitleFixer();
```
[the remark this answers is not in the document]
kevinsperrine: Yeah, good idea.

packages/react/src/hooks/useVisibilityObserver.js
```diff
@@ -0,0 +1,135 @@
...
+ * loadBatchSize prop determines how many new items to mark as visible each time the reference element
+ * is visible on screen. For example, if the loadBatchSize is 20 and you're scrolling down a table,
+ * when the reference row becomes visible it will update the isVisible array to contain 20 new true
+ * booleans.
+ *
+ * @param {*} elementRef html element to check if it's visible
+ * @param {object} options see defaultOptions above
+ * @returns Array an array containing [isVisible: bool]
+ */
+const useVisibilityObserver = (element, options) => {
```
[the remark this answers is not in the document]
kevinsperrine: changed.

packages/react/src/hooks/useVisibilityObserver.test.js
```diff
@@ -0,0 +1,141 @@
...
+      }
+    );
+
+    // the initial render should observe the ref one time
+    expect(observer.observe).toHaveBeenCalledWith(ref.current);
+    expect(observer.observe).toHaveBeenCalledTimes(1);
+    expect(result.current).toEqual([false]);
+    expect(onChange).not.toHaveBeenCalled();
+
+    // // mock the observer calling the callback when the element is visible,
```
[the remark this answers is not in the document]
kevinsperrine: fixed.

packages/react/src/hooks/useVisibilityObserver.test.js
kevinsperrine: once for changing to the new ref. I'm honestly not certain about the second. I tried to track it down for a long time. I can't figure out where it's coming from unless the useEffect cleanup is happening oddly in jest. I don't have any issues in storybook.

bjornalm (approved):
Great work on this one!

PR #3116, opened 2021-12-08
feat(cardeditor): add maximumDataPoints to data series cards

scottdickerson:
Closes #

image

**Summary**

- Support the ability to capture the maximumDataPoints a DataSeriesCard should show.  This is important because the TimeSeriesCards can reach scale issues once we get to a certain size

- feat(DataSeriesFormSettings): add the property to capture maximumDataPoints for the card
- feat(DashboardEditor): support passing the 18n to upstream default cards 
- feat(index): export the getDefaultCard utility function so downstream consumers can use it in addition to their own getDefaultCard callbacks in the DashboardEditor

- Verify that a TimeSeriesCard in the card editor has a new setting to capture the maximumDataPoints

- Verify the existing getDefaultCard callbacks still work


packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormSettings.jsx
```diff
@@ -30,6 +30,8 @@ const propTypes = {
       includeZeroOnYaxis: PropTypes.bool,
       timeDataSourceId: PropTypes.string,
       showLegend: PropTypes.bool,
+      /** maximum amounts of data points to render in the graph */
+      maximumDataPoints: PropTypes.bool,
```
kevinsperrine, line 34: bool?

kevinsperrine (commented):
Code looks good. Is `bool` the correct prop type? Just for my own clarity, how does this actually get applied to the card? For example, if I edit the [dashboard editor with initial value story](https://deploy-preview-3116--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-dashboardeditor--with-initial-value), and set a maximum number of data points of 3 on the TimeSeriesCard. I still see 14 data points in the chart, and I'm able to add Torque, firmware, and manufacturer as additional data sets, so the chart grows to 35 data points across all five types.

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormSettings.jsx
scottdickerson: nope, thanks for catching.  The data point limitation doesn't actually affect the cards themselves currently, only is stored in the JSON and up to the downstream consumers to use when they fetch their data.  Should I add to the chart itself?

kevinsperrine: Got it. No, I don't think it's necessary to add to the chart itself, but maybe clarify that in the prop-type comment?

bjornalm (approved):
Looks good once Kevin's comments have been addressed

davidicus (approved):

PR #3130, opened 2021-12-09
Nested table max-width prop

### LINKED ISSUE #3129: [Table] Nested table test wrap
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

For table nested table, at some point column can get overridden by one with bigger content
### To Reproduce

1. Go to StatefulTable with row nesting and resizable columns with initial width
2. Click on expand to one with 4 levels
3. See the error

### Expected behavior

Max-width of the span should be based on parent width - current offset

### Environment/versions:

- OS:  MacOS
- Browser: chrome
- `carbon-addons-iot-react` version:  `v2.147.0-next.79`

### Additional context

image

Do you want this work within a specific time period? Yes.
Is it related to an upcoming release? MAS Monitor.

cgirani:
Closes #3129

**Summary**

- Add a max-width in the span, based on the nested offset.

Before:
image

After:
image

- fix(tablerow): max-width for nested table

- Go to StatefulTable - with row nesting and resizable columns with initial width
- Open an item with at least 4 nested levels and see the width is not overwritten the next column

- Make sure the nested table keeps working properly.


kevinsperrine (approved):
Thanks for the help. Looks great!

bjornalm (commented):
Fix looks clean and simple. I just wanna test it properly before approving. I've restarted the failed deployment.

bjornalm (approved):

davidicus (approved):

packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
```diff
@@ -381,6 +381,7 @@ const StyledExpansionTableRow = styled(({ hasRowSelection, ...props }) => <Table
 const StyledNestedSpan = styled.span`
   position: relative;
   left: ${(props) => props.nestingOffset}px;
+  max-width: calc(100% - ${(props) => props.nestingOffset}px);
```
davidicus, line 384: We should not be doing this with styled components. This prop should be added as a css custom property.

davidicus (changes requested):
We do not support styled components and these will be removed in the future. Any new features that need a prop value should be done using css custom properties

PR #3133, opened 2021-12-09
fix(use-popover-positioning): minor style and offset adjustments

### LINKED ISSUE #3108: [FlyoutMenu] slightly incorrect auto positioning
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The menu auto positioning is off by a pixel

### To Reproduce

1. Go to https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-flyoutmenu--auto-positioning-example
2. Click on the buttons to expand the menus

### Expected behavior

Pixelperfection

### Environment/versions:

- OS: MacOS
- Browser: Chrome Version 96.0.4664.55

### Additional context

image

upcoming release?

date, the better!_

kevinsperrine:
Closes #3108 

**Summary**

- fixes minor positioning issues on FlyoutMenu
- fixes minor style issues on FlyoutMenu in RTL
- adds VRT for auto-positioning of FlyoutMenu in RTL/LTR

- check FlyoutMenu auto-positioning story and confirm everything is correct in RTL/LTR

- check Tooltip auto-positioning story and confirm everything is correct in RTL/LTR
- check OverflowMenu auto-positioning story and confirm everything is correct in RTL/LTR


packages/react/src/components/FlyoutMenu/FlyoutMenu.test.e2e.jsx
```diff
@@ -0,0 +1,196 @@
...
+          display: 'flex',
+          alignItems: 'flex-start',
+          justifyContent: 'flex-start',
+          flex: 1,
+        }}
+      >
+        <FlyoutMenu
+          direction={FlyoutMenuDirection.LeftEnd}
+          useAutoPositioning
+          menuOffset={{ top: 2, left: isRTL ? -2 : 0 }}
```
kevinsperrine, line 30: I've had to manually adjust the positioning of these with menuOffset objects. They are off by a few pixel in Cypress, but look perfect in storybook. I don't know if these offsets will need to be adjusted over time, or if I should leave them off and take the snapshots with them in the "wrong" position since they are correct in storybook/chrome.

packages/react/src/hooks/usePopoverPositioning.test.jsx
```diff
@@ -88,7 +88,7 @@ describe('usePopoverPositioning', () => {
     userEvent.click(button);
     const menu = screen.getByTestId(testId);
 
-    expect(menu).toHaveClass(`${iotPrefix}--flyout-menu--body__${FlyoutMenuDirection.LeftEnd}`);
```
kevinsperrine, line 91: This test was incorrect and is changed, because of the flyoutAlignment changes that fixes an edge case above: https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3133/files#diff-c56a39e4778901918956ade07124779426ef2c20782d0ac1bdc7d7981805c08cR205

packages/react/src/components/FlyoutMenu/FlyoutMenu.test.e2e.jsx
bjornalm: I also tested in the Storybook iframe canvas where it looked good, it is strange that they are off here. If you leave the adjustments here we should add a comment like the above to the code that explains why.

bjornalm (approved):
LGTM. It would be better if possible to find the root cause of the required adjustments, but I assume you have investigated that?

packages/react/src/components/FlyoutMenu/FlyoutMenu.test.e2e.jsx
kevinsperrine: Yeah, I spent quite-a-while yesterday trying different padding on the cypress window, different sizes, etc. I don't know. I'll copy my comment into the code, though.

davidicus: That is aggravating!

davidicus: I wonder if these offsets would be necessary if we used grid instead of flexbox (just an example). Wondering if its just a weird bug from Cypress's implementation

davidicus (approved):
I think this should be fine. I wonder if it makes more sense to leave out the offsets but keep in a note of the discrepancy in case we need to investigate in the future.

kevinsperrine:
> 

Yeah, You can make an argument for it either way. I don't really know which is best.

PR #3140, opened 2021-12-13
PROTOTYPE: Infinite loader test using `useDataLoader` hook

kevinsperrine:
Closes #

**Summary**

- A prototype useDataLoader hooks that ties into the useVisibleLoader hook from #3114. It leaves the onInfiniteScroll code from that same PR in place, so these are two different examples of solving the same problem. The basic premise is consumers will pass an onLoadData function that handles getting the data. When called, the onLoadData function will be called with an object containing `pageSize` (how many items to load), `start` the index from which to start loading, and `retry` which is a reference to the current call, so that if it fails a retry UI can be shown. This function must return an object containing `data` (an array of the newly loaded items), `hasMoreData` (boolean if more data is available), and `loading` (boolean to tell if it is still loading. The [example story](https://deploy-preview-3140--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-list--with-infinite-scroll-data-loader) shows how this can be used to load items infinitely as the user scrolls (`isInfiniteScroll` knob true), loading a section of data (`isInfiniteScroll` knob false), or async loading of pages on demand (`demo with pagination` knob true). This currently works with normal lists, but not VirtualLists. All jest tests pass, but there are some issues with drag-and-drop in the HierarchyList cypress tests that need to be addressed. This is not a true PR, but rather a draft prototype to determine actual implementation.

- adds useDataLoader hook


packages/react/src/hooks/useDataLoader.js
```diff
@@ -0,0 +1,117 @@
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+
+const useDataLoader = ({
```
stuckless, line 3: There's an expression, "when all you have is a hammer, everything looks like a nail".  This is what I'm seeing lately as developers try to put EVERYTHING as a hook :(

Data loading is not something that is not well suited for a hook.  It is complex enough that it should well defined and not tied to React, hooks, etc, and doing so, just makes it even more complicated.

For example, it would be much nicer to pass a `loader` into a component and the loader knows how to deal with data, manages the state of the data, and the concurrency around loading, and notifying component when it changes, etc.

Given that you build a clean interface around a `loader` it can easily unit tested, and it has nothing to do with React, since, loading data is not a react problem to solve.

Now, once you have a loader, you can likely create a clean hook, or set of hooks around that for easier integration with React components.

eg, ```
<MyCoolDataComponent loader={customLoader} options={renderingOptions}/>
```
The Loader interface might be something like this... ```
Loader
+ loadRange(start, size, callback)
+ isItemLoaded(index)
+ on(event, callback)
+ off(event, callback)
+ isLoading()
```

You can now create an ArrayLoader that implements this interface and just uses an Array as the source.  You can create a PALTableLoader and accepts another "loader" and delegates the loading but brings back the items and puts it into a PAL Table format for items.  The Loader interface doesn't care about the shape of data, it just manages getting the data and letting you know when it is working or have new data, etc, etc.

Once loaders are in place, you can now create a hook around using them in your components.  And because it's no longer cool to use a 'function' for functional components, we'll use the anonymous declaration instead...

const useLoader(loader) => {
    const [isLoading, setIsLoading] = useState(loader.isLoading());
    useEffect(()=>{
       const setLoading = (b)=>setIsLoading(b);
       loader.on('loading', setLoading);
       return () => {
          loader.off('loading', setLoading);
       }
    }, [loader]);
   ...
};

All the hook is doing is taking care of the tedious business of subscribing and unsubscribing from the loader events.  Now, this loader can be used in hooks, HOCs, whatever, and we don't have to rewrite data loading because something comes along that replaces hooks.  This is the intention of hooks.  ie, to be minimal pieces of code that basically just auto-wires the tedious integration of other systems.

And, if you later decide to add debouncing to the loader, the hooks don't have to be updated.  Multiple components might have the same instance of the loader, and they would all get notified when new data is loaded, etc, and app devs don't have to worry about concurrency.  ie, if multiple components make a request to load the data, then "loader" can manage this and debounce the requests and notify everyone later that data is loaded.  You won't get 2 requests to load the same data from different components.

In graphite, our "loader" is a "Datasource" with pluggable data adapters and it manages everything from loading, searching, filtering, selections, validations, changes, etc.  And we have an HOC that manages how you connect it into components.  We might later decide to create hooks for this, but, nothing changes in how they manage and load data, just how components listen for changes to data, and it has zero impact on downstream consumers.

This is just an example of how to effectively create data interfaces and integrate with hooks.  A real discussion should have around the architecture of this before we run off and build an implementation.

kevinsperrine: Thanks, @stuckless. That's helpful to understand your requirements and expectations. My contractual hours are up for the year, so let's continue this discussion in the new year and we'll work to build something that meets these various use cases.

PR #3159, opened 2022-01-03
3150 list missing props in prop types definition

### LINKED ISSUE #3150: [List] missing props in propTypes definition
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The shape propType describing the `items` prop is missing several props:

1. secondaryValue
2. rowActions
3. isCategory

There might be more missing. These should be defined and added to the documentation in the mdx file .

If needed, also update the documentation for components using the List

### To Reproduce

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

bjornalm:
Closes #3150

**Summary**
Adds missing PropTypes definition for ListItemPropTypes

- Add proptypes
- Update mdx doc
- update snapshot

- List is working and documentation has been updated

- List stories and stories built on the list works as expected


packages/react/src/components/List/List.mdx
```diff
@@ -99,6 +91,22 @@ const yankees = {
...
 
+### ListItemPropTypes
+
+| Name                   | Type                             | Default | Description                                                     |
+| :--------------------- | :------------------------------- | :------ | :-------------------------------------------------------------- |
+| children               | arrayOf(object)                  |         | Array of child items if the item is a parent                    |
+| content                | shape                            |         |                                                                 |
+| content.icon           | node                             |         | The icon to show for each item                                  |
+| content.rowActions[]   | oneOfType([arrayOf(node), func]) |         | The row actions for each item                                   |
```
kevinsperrine, line 101: We should define a deprecated designation for our docs, so it's clear when something is deprecated like the arrayOf(nodes) here. Maybe the † (<kbd>⌥</kbd>+<kbd>t</kbd>) since it won't effect the layout of the markdown in anyway?

bjornalm: Yea, up until now we have just added "Deprecated" in the description or in the type column (which seems a bit weird). I'll add it to description here.

davidicus (approved):

Issue #3177, opened 2022-01-07
[Table] Margin in toolbar is not correct in RTL mode


Ritu-Thakur:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

There is no margin between trash icon and 'Delete' button in the toolbar when RTL mode is selected

RTL:

![image](https://user-images.githubusercontent.com/30444945/148581282-1baa737c-adfd-4c0b-ae4c-2235da157881.png)

LTR:

![image](https://user-images.githubusercontent.com/30444945/148581550-5994ac92-a5fd-434f-8f85-d94db8724ea2.png)

### Steps to reproduce
1) Goto https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--row-selection-and-batch-actions
2) Enable Right-to-left under RTL tab

PR #3178, opened 2022-01-07
fix(table): margin between trash icon and delete on toolbar

Ritu-Thakur:
Closes # [3177](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3177)

**Summary**

Margin in table#toolbar is not correct in RTL mode 

Added margin-right style in element : html[dir="rtl"] .bx--action-list .bx--btn .bx--btn__icon

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/Table/TableToolbar/_table-toolbar.scss
+++ b/packages/react/src/components/Table/TableToolbar/_table-toolbar.scss
@@ -63,6 +63,10 @@ html[dir='rtl'] {
     .#{$prefix}--action-list {
       left: 0;
       right: auto;
+      .#{$prefix}--btn .#{$prefix}--btn__icon {
+        margin-left: 0;
+        margin-right: $spacing-03;
+      }
     }
 
     .#{$prefix}--batch-summary {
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Table/TableToolbar/_table-toolbar.scss
```diff
@@ -77,6 +77,9 @@ html[dir='rtl'] {
       right: 0;
     }
   }
+  .#{$prefix}--action-list .#{$prefix}--btn .#{$prefix}--btn__icon {
```
bjornalm, line 80: We can move this inside the ruleset starting on line 63 to avoid repeating the `.#{$prefix}--action-list` selector.
We also need to "undo" the margin-left from the normal ltr mode otherwise there will be too much space between the icon and the vertical divider:

`margin-left: 0;`

bjornalm (changes requested):
Nice work, just two smaller changes needed.

packages/react/src/components/Table/TableToolbar/_table-toolbar.scss
Ritu-Thakur: Thanks for your review @bjornalm . I have tried to fix it.

bjornalm (approved):
Nice job!

davidicus (approved):

PR #3181, opened 2022-01-10
List secondary value

### LINKED ISSUE #3180: [ListItem] Bad title when mousing over
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When we provide `secondaryValue` as a function the title is being `{value}--secondary-value`

### To Reproduce

1. Go to ListItem
2. Provide a function as a value
4. Hover the item and see error

### Expected behavior

Have support to provide a custom label when using functions

### Environment/versions:

- OS: MocOs
- Browser: Chrome
- `carbon-addons-iot-react` version: latest next
### Additional context

image

Do you want this work within a specific time period?  Yes
Is it related to an
upcoming release?  Yes

date, the better!_

cgirani:
Closes #3180 

**Summary**

- Support secondaryValue prop an object with:
```
      {
         value: PropTypes.func,
         label: PropTypes.string,
      }
```

- fix(listitem): support secondary value obj with custom label when func
- chore(test): snapshots update

- Go to ListItem with object and render should keep working

- N/A no changes to the current behaviour


bjornalm:
Looks like the unit tests are failing

packages/react/src/components/List/List.story.jsx
```diff
@@ -253,6 +253,27 @@ export const WithSecondaryValue = () => (
 
 WithSecondaryValue.storyName = 'with secondaryValue';
 
+export const WithSecondaryValueAsObject = () => (
```
bjornalm, line 256: I don't think we need this story since we you also created the new ListItem story

packages/react/src/components/List/ListItem/ListItem.jsx
```diff
@@ -39,7 +39,14 @@ const ListItemPropTypes = {
   expanded: PropTypes.bool,
   value: PropTypes.string.isRequired,
   /** string value or callback render function */
-  secondaryValue: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  secondaryValue: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.func,
+    PropTypes.shape({
```
bjornalm, line 45: You will also have to update the file `packages/react/src/components/List/ListPropTypes.js` which is used by the List component.

packages/react/src/components/List/ListItem/ListItem.jsx
```diff
@@ -39,7 +39,14 @@ const ListItemPropTypes = {
   expanded: PropTypes.bool,
   value: PropTypes.string.isRequired,
   /** string value or callback render function */
-  secondaryValue: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  secondaryValue: PropTypes.oneOfType([
```
bjornalm, line 42: We should also update the documentation in packages/react/src/components/List/List.mdx

packages/react/src/components/List/ListItem/ListItem.jsx
```diff
@@ -234,13 +241,23 @@ const ListItem = ({
   const renderSecondaryValue = () =>
     secondaryValue ? (
       <div
-        title={typeof secondaryValue === 'function' ? `${value}--secondary-value` : secondaryValue}
+        title={
+          typeof secondaryValue === 'function'
+            ? `${value}--secondary-value`
```
bjornalm, line 246: I think we also need to do something about this, because it never makes sense to show `${value}--secondary-value`. We could check if the function returns a string and if so use that string, otherwise use an empty string.

bjornalm (changes requested):
Nice work, only some smaller changes needed :-)

davidicus:
@bjornalm bump

packages/react/src/components/List/List.mdx
```diff
@@ -45,7 +45,7 @@ const yankees = {
...
-| :------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------- | :---------------------------------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| :------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------- | :---------------------------------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --- | --- | ------------------------------------------------------------------ |
```
bjornalm, line 48: Seems like two additional columns were accidentally added here. You can view/preview your documentation changes by clicking on the docs tab for any List story.

image

bjornalm (changes requested):
One small change needed

bjornalm (approved):
Great job!

davidicus (approved):

PR #3183, opened 2022-01-11
Change file extension for codesandbox test

### LINKED ISSUE #3179: [Repo] Remove codesandbox action
There is an open bug around resolving large dependencies for codesandbox. With Carbon charts and icons package our dependencies are way to large to load in the browser. 

We need to remove this check until we fix the issue that prevents our sandbox from loading.

davidicus:
Closes #3179 

Codesandbox fails to load our package due to the size of our dependencies, namely Carbon icons and charts. This change renames the file so we no longer run this check on PRs until we can get the underlying issue resolved.


kevinsperrine (approved):
Looks good. Why did you need to remove the spaces?

bjornalm (approved):

davidicus:
> Looks good. Why did you need to remove the spaces?

I think the formatter must have changed it. I had only changed the file extension.

PR #3185, opened 2022-01-12
3044 menu button style audit button variations

### LINKED ISSUE #3044: [MenuButton] Style audit + button variations
### What package is this for?

- [x] React
- [ ] Angular

## 1. Menu heights

Similar to other components like [dropdowns](https://react.carbondesignsystem.com/?path=/story/components-dropdown--playground) and [overflow menu](https://react.carbondesignsystem.com/?path=/story/components-overflowmenu--playground), there are multiple size variations for the component, and the menu items should match the height of the component. 

### Current

Storybook: https://next.carbon-addons-iot-react.com/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-menubutton--single-menu-button
Screen Shot 2021-11-08 at 3 58 34 PM

### Proposed
Design spec: https://pages.github.ibm.com/ai-applications/design/components/buttons/style

Screen Shot 2021-11-08 at 3 53 08 PM

## 2. Button variations

The menu button in the design can be used with the primary, secondary, tertiary, and ghost.
The split (combo) button in the design can be used with the primary, secondary, and tertiary.

Note: the secondary button is up for discussion to see if we need to support that, since secondary buttons are generally used in conjunction with primary buttons and acts as the negative action of the set, but it would be good to support the tertiary and ghost (for menu only) as well.

Design spec: https://pages.github.ibm.com/ai-applications/design/components/buttons/style

## 3. Fix z-index on shadow for icon only button

The shadow from the popover shows above the icon button. Also, same note on the button size vs the menu items being the same size.

### Current

Storybook: https://next.carbon-addons-iot-react.com/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-menubutton--icon-only-menu-button

Screen Shot 2021-11-08 at 4 05 43 PM

### Proposed

Screen Shot 2021-11-08 at 4 05 57 PM

bjornalm:
Closes #3044

**Summary**
Fixed the styling remarks from #3044 and also fixed some additional RTL issued. In order to understand the changes it is easiest to go through the commits one by one.

1. Fixed button and menu item sizes according to the spec  + unit tests
2. Added button variations (kind) according to the spec + unit tests
3. Fixed the shadow from the menu that was incorrectly dropping in the button (complex fix needed) + unit tests
4. Fixed additional RTL issues

Start by reading the spec.

1. Go to https://deploy-preview-3185--carbon-addons-iot-react.netlify.app?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-menubutton--single-menu-button and use the size knob to verify that the sizes renders according to spec. Note, we use "default" instead of large until carbon v.11

2.  Repeat step 1 for https://deploy-preview-3185--carbon-addons-iot-react.netlify.app?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-menubutton--split-menu-button and https://deploy-preview-3185--carbon-addons-iot-react.netlify.app?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-menubutton--icon-only-menu-button

3. Verify that https://deploy-preview-3185--carbon-addons-iot-react.netlify.app?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-menubutton--single-menu-button renders the different kinds of buttons available in the kind knob and that the button and the menu looks as expected. Compare to the spec.

4. Verify that https://deploy-preview-3185--carbon-addons-iot-react.netlify.app?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-menubutton--split-menu-button renders the different kinds of buttons available in the kind knob and that the button and the menu looks as expected. Compare to the spec.

5.  Go to https://deploy-preview-3185--carbon-addons-iot-react.netlify.app?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-menubutton--auto-positioning-example and for all COMBINATIONS of positions, sizes and RTL-mode for the IconOnlyButton verify that: 
a) the shadow from the menu does not fall on the button (for any state of interaction)
b) the blue focus border of the menu is intact
c) the blue focus border of the button is intact (to test this click the button to open the menu and then press down on the button with the mouse again but move the cursor outside the button before releasing)

- All stories work as before except for the changes described by the spec.

### DIFF
--- a/packages/react/src/components/MenuButton/SplitMenuButton.jsx
+++ b/packages/react/src/components/MenuButton/SplitMenuButton.jsx
@@ -42,6 +42,11 @@ const propTypes = {
    * The size of the button and the dropdown items
    */
   size: PropTypes.oneOf(['sm', 'md', 'default']).isRequired,
+
+  /**
+   * The kind of button to render
+   */
+  kind: PropTypes.oneOf(['primary', 'secondary', 'tertiary']).isRequired,
 };
 
 export const SplitMenuButton = React.forwardRef(
@@ -54,6 +59,7 @@ export const SplitMenuButton = React.forwardRef(
       onSecondaryActionClick,
       size,
       testId,
+      kind,
     },
     ref
   ) => {
@@ -64,6 +70,7 @@ export const SplitMenuButton = React.forwardRef(
           onClick={onPrimaryActionClick}
           testId={`${testId}-primary`}
           size={size}
+          kind={kind}
         >
           {label}
         </Button>
@@ -79,6 +86,7 @@ export const SplitMenuButton = React.forwardRef(
           onClick={onSecondaryActionClick}
           testId={`${testId}-secondary`}
           size={size}
+          kind={kind}
         />
       </>
     );
--- a/packages/react/src/components/MenuButton/SplitMenuButton.jsx
+++ b/packages/react/src/components/MenuButton/SplitMenuButton.jsx
@@ -37,11 +37,24 @@ const propTypes = {
    * be read by screen readers
    */
   iconDescription: PropTypes.string.isRequired,
+
+  /**
+   * The size of the button and the dropdown items
+   */
+  size: PropTypes.oneOf(['sm', 'md', 'default']).isRequired,
 };
 
 export const SplitMenuButton = React.forwardRef(
   (
-    { label, iconDescription, renderIcon, onPrimaryActionClick, onSecondaryActionClick, testId },
+    {
+      label,
+      iconDescription,
+      renderIcon,
+      onPrimaryActionClick,
+      onSecondaryActionClick,
+      size,
+      testId,
+    },
     ref
   ) => {
     return (
@@ -50,6 +63,7 @@ export const SplitMenuButton = React.forwardRef(
           className={classnames(`${iotPrefix}--menu-button__primary`)}
           onClick={onPrimaryActionClick}
           testId={`${testId}-primary`}
+          size={size}
         >
           {label}
         </Button>
@@ -64,6 +78,7 @@ export const SplitMenuButton = React.forwardRef(
           renderIcon={renderIcon}
           onClick={onSecondaryActionClick}
           testId={`${testId}-secondary`}
+          size={size}
         />
       </>
     );
--- a/packages/react/src/components/MenuButton/_menu-button-shadow-blocker.scss
+++ b/packages/react/src/components/MenuButton/_menu-button-shadow-blocker.scss
@@ -1,0 +1,192 @@
+@import '../../globals/vars';
+
+$size-sm: rem(32px);
+$size-md: rem(40px);
+$size-lg: rem(48px);
+$shadow-blocker-height: rem(8px);
+$shadow-blocker-border-width: rem(1px);
+
+@mixin shadow-blocker-element {
+  position: absolute;
+  display: block;
+  height: calc(#{$shadow-blocker-height} - #{$shadow-blocker-border-width});
+}
+
+// COVER THE SHADOW WITH A NEW DIV
+// We add a new div on a higher z-index in order to cover the shadow of the menu.
+.#{$iot-prefix}--menu-button--icon-only.#{$prefix}--menu--open {
+  .#{$iot-prefix}--menu__shadow-blocker {
+    @include shadow-blocker-element;
+    top: calc((-#{$shadow-blocker-height}) + #{$shadow-blocker-border-width});
+    left: 0;
+    width: $size-sm; // small is the default size
+    background-color: $ui-01;
+    border-bottom: $shadow-blocker-border-width solid $ui-01;
+
+    &--flip-y {
+      left: unset;
+      right: 0;
+    }
+
+    &--flip-x {
+      left: unset;
+      top: var(--menu-height);
+    }
+
+    html[dir='rtl'] & {
+      &:not(.#{$iot-prefix}--menu__shadow-blocker--opens-horizontally) {
+        left: 0;
+      }
+    }
+
+    &--md {
+      width: $size-md;
+    }
+    &--lg {
+      width: $size-lg;
+    }
+
+    // The menu is attached to the side (instead of above or below)
+    &--opens-horizontally {
+      $width: calc(#{$shadow-blocker-height} - #{$shadow-blocker-border-width});
+
+      top: 0;
+      height: $size-sm; // small is the default size
+      width: $width;
+      left: calc((#{$width}) * (-1));
+      border-bottom: none;
+      border-right: $shadow-blocker-border-width solid $ui-01;
+
+      &.#{$iot-prefix}--menu__shadow-blocker--flip-y {
+        right: calc((#{$width}) * (-1));
+        left: unset;
+        border-right: none;
+        border-left: $shadow-blocker-border-width solid $ui-01;
+      }
+
+      &.#{$iot-prefix}--menu__shadow-blocker--md {
+        height: $size-md;
+        width: $width;
+      }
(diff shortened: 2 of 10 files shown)


kevinsperrine (approved):
This looks great. The only question I'm uncertain about is the extra space at the top and bottom of a menu item (It's easily visible when hovering "Delete" or "Publish"). It doesn't _seem_ to be part of the spec, but it's hard to know for certain.

Screen Shot 2022-01-13 at 8 45 45 AM

bjornalm:
> This looks great. The only question I'm uncertain about is the extra space at the top and bottom of a menu item (It's easily visible when hovering "Delete" or "Publish"). It doesn't _seem_ to be part of the spec, but it's hard to know for certain.
> 
> Screen Shot 2022-01-13 at 8 45 45 AM

I don't know either. It was not introduced by these changes, it is the standard carbon padding on the surrounding *bx-menu* element so it is only available above the first item and below the last item. The carbon *bx--menu-divider* also has  top-bottom padding which is what we seen in the screenshot. 
@dianatran18 Should we remove the padding for the *bx--menu* and *bx--menu-divider*?

packages/react/src/components/MenuButton/SplitMenuButton.jsx
```diff
@@ -37,11 +37,30 @@ const propTypes = {
    * be read by screen readers
    */
   iconDescription: PropTypes.string.isRequired,
+
+  /**
+   * The size of the button and the dropdown items
+   */
+  size: PropTypes.oneOf(['sm', 'md', 'default']).isRequired,
```
davidicus, line 44: Do we need to make these required? Can we add a default and make this optional? Same comment with `kind`.

packages/react/src/components/MenuButton/_menu-button-shadow-blocker.scss
```diff
@@ -0,0 +1,192 @@
+@import '../../globals/vars';
```
davidicus, line 1: Should we use the new `@use` syntax for new sass files?

kevinsperrine: Theoretically, I guess we could. I've just been adjusting it in the sass-module PR each time I update it. 🤷‍♂️

dianatran18:
> > This looks great. The only question I'm uncertain about is the extra space at the top and bottom of a menu item (It's easily visible when hovering "Delete" or "Publish"). It doesn't _seem_ to be part of the spec, but it's hard to know for certain.
> > Screen Shot 2022-01-13 at 8 45 45 AM
> 
> I don't know either. It was not introduced by these changes, it is the standard carbon padding on the surrounding _bx-menu_ element so it is only available above the first item and below the last item. The carbon _bx--menu-divider_ also has top-bottom padding which is what we seen in the screenshot. @dianatran18 Should we remove the padding for the _bx--menu_ and _bx--menu-divider_?

Let me confirm with the Carbon folks because that was a change I saw more recently too. I don't know if they're moving towards having padding on the bottom like that, but I know there's padding between divider lines now, which is new from before.

dianatran18:
> > > This looks great. The only question I'm uncertain about is the extra space at the top and bottom of a menu item (It's easily visible when hovering "Delete" or "Publish"). It doesn't _seem_ to be part of the spec, but it's hard to know for certain.
> > > Screen Shot 2022-01-13 at 8 45 45 AM
> > 
> > 
> > I don't know either. It was not introduced by these changes, it is the standard carbon padding on the surrounding _bx-menu_ element so it is only available above the first item and below the last item. The carbon _bx--menu-divider_ also has top-bottom padding which is what we seen in the screenshot. @dianatran18 Should we remove the padding for the _bx--menu_ and _bx--menu-divider_?
> 
> Let me confirm with the Carbon folks because that was a change I saw more recently too. I don't know if they're moving towards having padding on the bottom like that, but I know there's padding between divider lines now, which is new from before.

@bjornalm I confirmed with the Carbon designer that for overflow menus and combo/menu buttons there will be _no_ padding. It is only for the [Carbon context menu](https://react.carbondesignsystem.com/?path=/story/experimental-unstable-menu-contextmenu--context-menu) where there is padding.

>Lauren Rice:rice:  12:19 PM
The context menu will have the padding, but the combo button / menu button / overflow menu will not have the padding.

So it should be like this (let me know if this makes sense and if I've missed any other areas)

Screen Shot 2022-01-13 at 12 25 27 PM

packages/react/src/components/MenuButton/SplitMenuButton.jsx
bjornalm: SplitMenuButton.jsx is an internal component and MenuButton is the only consumer, so I made the size prop required in order to not duplicate the default value. But I can add a default if you think that makes more sense?

packages/react/src/components/MenuButton/_menu-button-shadow-blocker.scss
bjornalm: @davidicus so, should we just take that decision then, that all new files use `@use`

packages/react/src/components/MenuButton/SplitMenuButton.jsx
davidicus: No, if its internal only there is no need.

packages/react/src/components/MenuButton/_menu-button-shadow-blocker.scss
davidicus: Since this old way is going to be deprecated I think it makes sense to avoid having to do so on a timeframe.

davidicus (approved):
Looks good. @dianatran18 can you sign off on this before we merge?

packages/react/src/components/MenuButton/_menu-button-shadow-blocker.scss
bjornalm: Sounds good, I'll update the PR

bjornalm: Nope, It doesn't work, it throws:
`Error: @use rules must be written before any other rules.`

bjornalm:
@dianatran18 
An update on the left menu item indentation: 
I went through the carbon code for the menu to make sure I didn’t write a workaround for something that already existed, and it turns out that only if there are any selectable menuItems (`MenuSelectableItem`) then the indentation element showing the icon will be added. 

image


So this means:
1. a list without any indentation and icons will be automatically created by using only “normal” menu items (`MenuItem`)
image

2. if we want custom icons on the left side like in our stories we also have to include at least one `MenuSelectableItem`
image
 

@JordanWSmith15 is number 2 ok with you

packages/react/src/components/MenuButton/_menu-button-shadow-blocker.scss
davidicus: hmmm. Is it because this is pulled in somewhere else that is using imports? @kevinsperrine thoughts?

packages/react/src/components/MenuButton/_menu-button.scss
```diff
@@ -1,5 +1,6 @@
 @import '~carbon-components/scss/components/menu/menu';
 @import '../../globals/vars';
+@import './menu-button-shadow-blocker';
```
davidicus, line 3: If we change these to `@use` will this fix the sass error you received?

bjornalm: Nope, I did try that too

bjornalm: See [comment here](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3185#discussion_r785830751)

dianatran18:
LGTM!

PR #3192, opened 2022-01-14
3172 tech stack

### LINKED ISSUE #3172: [Repo] Add onboarding documentation - tech stack and libraries
New developers should be able to quickly familiarize themselves with the pertinent technologies and libraries used in this repo. This documentation will possibly be used to upskill PAL and internal contributors. The goal is a crash course overview with links for more learning on each section. Should cover the following:
- [ ] JavaScript
- [ ] NPM
- [ ] Yarn
- [ ] Storybook
- [ ] Jest
- [ ] Cypress
- [ ] Sass
- [ ] React
- [ ] ?

davidicus:
Closes #3172 

**Summary**

- summary_here

- items_here

- tests here

- tests here

### DIFF
--- a/packages/react/docs/guides/tech-stack.md
+++ b/packages/react/docs/guides/tech-stack.md
@@ -2,7 +2,7 @@
 
 ### _last updated:_ 01/05/2022
 
-The following is a brief breakdown into the technolgies used to develop, test, and ship components in the AI Application's Pattern Asset Library. This is not meant to be an exhaustive intro into these technologies but a good "crash course" overview with links to more information.
+The following is a brief breakdown into the technolgies used to develop, test, and ship components in the AI Application's Pattern Asset Library. This isn't meant to be an exhaustive intro into these technologies but a good "crash course" overview with links to more information.
 
 ### Table of Contents
 
@@ -24,8 +24,9 @@ Javscript is used to run our development server, run test, author markup, and ma
 
 I won't go deeply into every aspect of JavaScript but there are a few features of the language that you will use often when building React components.
 
-1. [Destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)
-   You will see destructring used throughout the codebase. Whether using them to destructure our props from the prop object passed to components
+### [Destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)
+
+You will see destructring used throughout the codebase. Whether destructuring our props from the prop object passed to components
 
 ```js
 function Component({ propOne, propTwo }) {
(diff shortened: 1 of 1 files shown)


packages/react/docs/guides/tech-stack.md
```diff
@@ -0,0 +1,283 @@
...
+
+```js
+const dogName = adventurer.dog?.name;
+adventurer.dog?.someNonExistentMethod?.();
+
+// vs
+
+if (adventurer.dog && adventurer.dog.someNonExistentMethod) {
+  const dogName = adventurer.dog.name;
+  adventurer.dog.someNonExistentMethod.();
```
Ritu-Thakur, line 187: At 187 line , do we need .() in the end?

Ritu-Thakur (changes requested):

packages/react/docs/guides/tech-stack.md
```diff
@@ -0,0 +1,283 @@
...
+```js
+function MyComponent(props) {
+  return <div>Hello World</div>;
+}
+```
+
+but you will definitely see them defined using arrow functions both in our library and in the wild.
+
+```js
+const MyComponent = (props) => <div>Hello World</div>;
```
stuckless, line 133: A point to make here, is that while regular function is preferred, cases where you need wrap the function with some other HOC, the arrow function is preferred.
```js
const MyComponent = observer((props) => <div>Hello World</div>);
```

packages/react/docs/guides/tech-stack.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+## [Tertiary conditional](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)
+
+is the only JavaScript operator that takes three operands: a condition followed by a question mark (?), then an expression to execute if the condition is truthy followed by a colon (:), and finally the expression to execute if the condition is falsy. This operator is frequently used as an alternative to an if...else statement.
+
+In react there will be times when you want to conditionally render an element. Like the content of an accordion.
+
+```js
+<AccordionItem>{isOpen ? <AccordionItemContent>Hey</AccordionItemContent> : null}</AccordionItem>
```
stuckless, line 159: While ternaries can make single line expressions very readable... I do find that they quickly become confusing when... 
a) they are nested
b) they include large multi-line responses on both the true and false expression

davidicus: Agreed, and there are cases where a traditional if statement may be necessary but for most cases I think the ternary is best practice over the use of `&&` to avoid the case of false rendering.

The example below would render `0`, unintentionally, if content was an empty array.

```js
{ content.length && <AccordionItemContent>content[i]</AccordionItemContent> }
```
I do think it is worth noting where the ternary can be abused and what to use in those cases.

packages/react/docs/guides/tech-stack.md
davidicus: Do not need the `.`. Thanks

bjornalm (approved):

PR #3203, opened 2022-01-19
fix(table): table toolbar action overflow truncation

### LINKED ISSUE #3198: [Table] Toolbar action's overflow menu shows items wrapped.
### What package is this for?

- [ x] React
- [ ] Angular

### Describe the bug

1. The toolbar action’s overflow menu items wrap by default making the text to overflow. 
2. Also, icon size changes automatically with longer texts.
3. If all items in overflow menu are hidden (not disabled but hidden), then the overflow icon(3 dots) should also be hidden. Currently there's no way to hide the overflow menu.

### To Reproduce

image

https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--with-filters

### Expected behavior

1. Longer menu item texts should not wrap by default, instead should be displayed with ellipsis and show a tooltip on-hover.
2. icon size should remain unchanged.
3. Hide overflow icon(3 dots) if all items in overflow menu are hidden

### Environment/versions:

- OS: [e.g. MacOS, Windows]
- `carbon-addons-iot-react` version: [e.g. `v2.147.0`]

kevinsperrine:
### Merging into `master` to make it into 8.7

Closes #3198 

**Summary**

- forces text to truncate instead of wrap when using icons in the overflow menu
- hides the toolbar action overflow menu when all items are hidden
- hides toolbar actions that aren't in the overflow menu when they are hidden

- change html structure/classes to cause long text to truncate when using icons
- update story to demo long text with and without icons
- update story to add demo all items as 'hidden' knob
- update tests for new structure

- go to [with filtering, toolbarActions, and custom toolbar content](https://deploy-preview-3203--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--with-filters)
- open toolbar overflow menu
- confirm long text is truncated
- turn on "demo all toolbar overflow items hidden" knob
- confirm overflow menu and actions are hidden

- confirm overflow menu works as expected
- check inline overflow menu actions ([this story, for example](https://deploy-preview-3203--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--table-example-with-create-save-views)) to confirm they still function and look as expected


bjornalm:
The item with an icon is not truncated:
image

kevinsperrine:
> The item with an icon is not truncated: image

Yeah, the preview isn't updating after I changed something. Hopefully that clears soon.

kevinsperrine:
> > The item with an icon is not truncated: image
> 
> Yeah, the preview isn't updating after I changed something. Hopefully that clears soon.

@bjornalm This finally updated the preview and is ready now.

davidicus:
This may not be related by I am seeing the overflow menu not scroll  with the table. 
image

davidicus (approved):
Just the one question about the moving overflow but the code looks good and the overflow is working. 👍

kevinsperrine:
> This may not be related by I am seeing the overflow menu not scroll with the table. image

In which story? I can't re-create it in my PR or in next... 🤔

bjornalm:
> > This may not be related by I am seeing the overflow menu not scroll with the table.
> 
> In which story? I can't re-create it in my PR or in next... 🤔

I could replicate it here
https://deploy-preview-3203--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-table--table-example-with-create-save-views

But it is not a regression error by this PR, it also happens in next

packages/react/src/components/Table/TableToolbar/_table-toolbar.scss
```diff
@@ -93,13 +93,15 @@ html[dir='rtl'] {
...
   text-align: left;
+  min-width: 0;
 
   [dir='rtl'] & {
     text-align: right;
   }
 
   svg {
     margin-right: $spacing-03;
+    min-width: $spacing-05;
```
bjornalm, line 104: I think it is shrinking because it is in a flex container, so I would use `flex-shrink: 0;` instead of adding a min-width

bjornalm (approved):
Nice job. Approving since this might be urgent. Non critical feedback to adress if there is time.

packages/react/src/components/Table/TableToolbar/_table-toolbar.scss
kevinsperrine: Ah! Of course. Thanks!

PR #3216, opened 2022-01-21
feat(list): add `hasFastSearch` and example async search story

### LINKED ISSUE #3078: [List] support async search
### What package is this for?

- [ x] React
- [ ] Angular

### Summary

Since the list now as (rudimentary) functionality to load more items in an async matter we should also support async search. This is to some extent already supported but we probably need some UX improvements like triggering search on enter instead of key press. We also need a search loading indicator, not sure if the general list loading is sufficient or not.

request here.

upcoming release?

date, the better!_

### Available extra resources

kevinsperrine:
Closes #3078 

**Summary**

- Adds `hasFastSearch` prop to search object on List to allow search to only happen on Enter or Blur. This gives us the ability to do async searches on a List
- Adds an example story with async loading and search.

- adds `hasFastSearch` to List and ListHeader
- adds logic to call onChange via onBlur/Enter or onClear
- adds tests to confirm
- adds story to demo async loading/search

- go to with async loading and search
- type 8
- confirm loading indicator appears
- after 1.5seconds you should have only items with `8` in them

- go to basic (single column) story
- turn on search
- type `dj`
- confirm search happens immediately
- click clear search button
- turn off check the hasFastSearch knob
- type `dj`
- confirm search does not happen until typing enter

- check list search story and confirm default experience is the same as before


packages/react/src/components/List/List.story.jsx
```diff
@@ -204,6 +213,65 @@ export const BasicSingleColumnWithSearch = () => {
...
+          isVirtualList={boolean('isVirtualList', false)}
+        />
+      </div>
+    );
+  };
+
+  return <ListWithAsyncLoadingAndSearch />;
+};
+
+BasicSingleColumnWithAsyncLoadingAndSearch.storyName = 'with async loading and search';
```
bjornalm, line 273: nit: I think it is best to keep these names synced so that that the URL and the story name matches e.g.
`WithAsyncLoadingAndSearch.storyName = 'with async loading and search';`

packages/react/src/components/List/ListHeader/ListHeader.jsx
```diff
@@ -35,6 +38,17 @@ const defaultProps = {
 };
 
 const ListHeader = ({ title, buttons, search, i18n, testId }) => {
+  const previousSearchValueRef = useRef(search?.value);
+  const hasFastSearch = search?.hasFastSearch !== false;
+  const handleSearch = (e) => {
+    const { value: currentValue } = e.target;
+    const { current: previousValue } = previousSearchValueRef;
+    if (previousValue !== currentValue) {
```
bjornalm, line 46: Why is this needed? Is it to prevent multiple searches if the user hits enter multiple times, or is it a fix for the `handleSearch` being called multiple times due to some other reason?

bjornalm:
Regarding the acceptance test instructions:
1. go to with async loading and search
2. type 8
3. confirm loading indicator appears 

**-- It does not. I have to press Enter for it to happen, but I assume that is correct**

4. go to basic (single column) story
5. turn on search
6. type dj

**-- Got error: `Cannot have two HTML5 backends at the same time.` **
This is not reproducible and might not be related to this PR but it happened a few times during testing.
Perhaps we should include the DND wrapper for all List stories: 
```
WithStoryName.decorators = [
  (Story) => (
    <DragAndDrop>
      <Story />
    </DragAndDrop>
  ),
];

```

packages/react/src/components/List/ListHeader/ListHeader.jsx
kevinsperrine: Exactly. If you hit enter, but then blur the input the handler is called twice. This prevents that if nothing has changed.

packages/react/src/components/List/List.story.jsx
kevinsperrine: works for me! I'll adjust.

kevinsperrine:
> Regarding the acceptance test instructions:
> ...
> -- It does not. I have to press Enter for it to happen, but I assume that is correct

Yeah, you are correct. It was getting late. I'll adjust these steps. And add the drag and drop decorator.

bjornalm (approved):
Nice work!

packages/react/src/components/List/List.test.jsx
```diff
@@ -1013,5 +1041,37 @@ describe('List', () => {
       );
       expect(screen.getByRole('checkbox')).toBeDisabled();
     });
+
+    it('should not trigger search onChange when hasFastSearch:false', () => {
```
davidicus, line 1045: are these duplicate test as line 481?

davidicus (approved):
Looks good just the one question about the test.

packages/react/src/components/List/List.test.jsx
kevinsperrine: These are with `isVirtualList:true`. They're in their own `describe` block.

Issue #3190, opened 2022-01-13
[Table] Load more nested rows using disabled styling


ben-bacon:
### What package is this for?

- [X] React
- [ ] Angular

### Summary

https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--row-expansion-and-load-more

After clicking "load more" under nested rows, the loading spinner styling looks to be disabled. Let's show it in its "full color"/ enabled state instead. 

Proposed styling:
![image](https://user-images.githubusercontent.com/52676199/149410220-64c0e86f-f5dd-4910-af71-0b14846c1395.png)

Shady-Nessim:
This diverges from the standard way we have the buttons in the loading state. Just confirming, do we need to diverge?

ben-bacon:
Good point @Shady-Nessim. Instead, let's follow how the table loads the cells initially with skeleton states, as in this [example](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--async-column-loading), after clicking "load more".

PR #3224, opened 2022-01-24
fix(table): add skeleton column for nesting rows when lazy loading

### LINKED ISSUE #3220: [Table] shouldLazyRender renders incorrect columns
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The prop `shouldLazyRender` renders incorrect columns when combined with `options.hasRowNesting` and `options.hasRowSelection:multi

### To Reproduce

1. Go to a playground story
2. Enable shouldLazyRender
3. Enable options.hasRowNesting 
4. Set options.hasRowSelection to "multi"

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context

image

upcoming release?

date, the better!_

kevinsperrine:
Closes #3220 

**Summary**

- The placeholder skeleton column was missing when lazy loading with `hasNestedRows`. This fixes that.

- add check for `hasRowNesting` when including skeleton columns

- go the the basic table story
- turn on hasRowNesting
- turn on hasRowSelection to 'multi'
- turn on shouldLazyRender
- confirm the columns align with header and Object ID column isn't missing a skeleton

- tests here

### DIFF
--- a/packages/react/src/components/Table/TableBody/TableBodyRowRenderer.jsx
+++ b/packages/react/src/components/Table/TableBody/TableBodyRowRenderer.jsx
@@ -197,11 +197,15 @@ const TableBodyRowRenderer = (props) => {
         ref={rowVisibilityRef}
         data-testid={`${tableId}-lazy-row-${row.id}`}
       >
-        {[...Array(totalColumns)].map((c, colIndex) => (
+        {hasRowSelection === 'multi' ? <TableCell /> : null}
+        {hasRowExpansion || hasRowNesting ? <TableCell /> : null}
+        {columns.map((v, colIndex) => (
           <TableCell key={`empty-cell-${colIndex}`}>
             <SkeletonText />
           </TableCell>
         ))}
+        {showExpanderColumn ? <TableCell /> : null}
+        {hasRowActions ? <TableCell /> : null}
       </tr>
     );
   }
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Table/TableBody/TableBodyRowRenderer.jsx
```diff
@@ -198,7 +198,7 @@ const TableBodyRowRenderer = (props) => {
         data-testid={`${tableId}-lazy-row-${row.id}`}
       >
         {hasRowSelection === 'multi' ? <TableCell /> : null}
-        {hasRowExpansion ? <TableCell /> : null}
+        {hasRowExpansion || hasRowNesting ? <TableCell /> : null}
```
bjornalm, line 201: We can use the `totalColumns` which contains the number of visible columns prop (calculated by the table and passed down to its subcomponents) and simplify this section by only writing:  

```
  {totalColumns.map((v, colIndex) => (
    <TableCell key={`empty-cell-${colIndex}`}>
      <SkeletonText />
    </TableCell>
  ))}
```

bjornalm (changes requested):

packages/react/src/components/Table/TableBody/TableBodyRowRenderer.jsx
kevinsperrine: I had to add a fix to make `hasRowNesting` add an extra column to the totalColumn count, but otherwise, it's much cleaner. Thanks!

bjornalm: Man I'm sorry. But I was probably wrong here. This would work fine if we were just inserting empty cells for the whole row, but I realise now that only the actual data columns should show skeleton texts. Here all extra spacer columns also get the skeleton:

image

Your original solution was the best option here

kevinsperrine: reverted.

bjornalm (approved):
Nice work!

davidicus:
@Shady-Nessim FYI, this may affect your issue #3190.

davidicus (approved):

Issue #3189, opened 2022-01-13
[Table] Filter tag using wrong color on hover


ben-bacon:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-statefultable--stateful-table-with-advanced-filters

The tags used for table filters are using the wrong color token on hover, instead of [Carbon's default styling](https://react.carbondesignsystem.com/?path=/story/components-tag--filter) (select blue tag in knob section).

Proposed styling:
<img width="263" alt="Screen Shot 2022-01-13 at 3 02 56 PM" src="https://user-images.githubusercontent.com/52676199/149408710-d47ab9f2-6a9c-4cbd-afc2-f3b718ae8035.png">

Ritu-Thakur:
@ben-bacon 
Hi Ben,
The hover text which says "Clear Filter" has black background in the above summary screenshot. But the Carbon default styling has grey background:
<img width="758" alt="image" src="https://user-images.githubusercontent.com/30444945/151066827-b3b6f758-8f8d-4aca-a100-032dcf475611.png">

Which background should I use for "Clear Filter" ?

ben-bacon:
@Ritu-Thakur Sorry, my issue was not clear. The issue is the hover state over the close icon is incorrect in the iot-addons library. see below,

**iot-addons (incorrect hover state styling over close icon):**
<img width="284" alt="Screen Shot 2022-01-26 at 9 56 06 AM" src="https://user-images.githubusercontent.com/52676199/151198897-d4520c71-7b51-4f5e-9266-77e32f20b64c.png">

**Carbon (the proposed styling for hover state over close icon):**
<img width="157" alt="Screen Shot 2022-01-26 at 9 56 23 AM" src="https://user-images.githubusercontent.com/52676199/151199018-0039b707-4e94-4cc3-8560-8b4b974dabd0.png">

The hover state for the close icon in the iot-addons library is #c6c6c6, but it should be blue30 (#a6c8ff)

Now that I look at it more closely, the color of the text is incorrect in the iot-addons library as well. The text and icon, including in the hover state, should be blue80 (#002d9c).

Ritu-Thakur:
@ben-bacon Thanks for explaining it more. Yes I noticed the close button's background on hover yesterday and tried to fix it.
I'll fix text/icon too.

PR #3226, opened 2022-01-24
fix(table): filter tag has wrong color on hover

Ritu-Thakur:
Closes : https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3189

**Summary**

Filter tag has wrong color on hover

Added "blue" type in the Table's tag

Check Filter tag and hover over the "x" button:
https://deploy-preview-3226--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table-statefultable--stateful-table-with-advanced-filters

- tests here


bjornalm:
@Ritu-Thakur Is this one ready for review?

kevinsperrine (changes requested):
Thanks, @Ritu-Thakur. These color changes are good, but Ben is also talking about the hover state for the close icon. You'll want to add another selector in that same block for `.#{$prefix}--tag__close-icon:hover` and set the appropriate color for the button, also.

bjornalm:
> > @Ritu-Thakur Is this one ready for review?
> 
> @bjornalm Yes, but need input from you guys if this is the correct fix?

Great, we normally add the label "status: ready for review" to indicate that the PR is ready. I did that. It is not a requirement but it also helps if the section "Acceptance Test (how to verify the PR)" contains a link to the story in the PR preview where the fix can be tested.

Ritu-Thakur:
I tried to find the exact color for close button in hover state, I found that $link-02 is the one we need:


image


But use of  doesn't work for me 👍 


  .#{$prefix}--tag__close-icon:hover {
    background-color: #a6c8ff;
  }

kevinsperrine (approved):
Nice work!

davidicus (approved):

davidicus:
This is good but it is failing our visual regression test for Table. We will want to update these.

PR #3228, opened 2022-01-25
fix(composed-modal): adjust height based on content and available area

### LINKED ISSUE #3077: [ComposedModal] prevent height of modal to change with content
### What package is this for?

- [x] React
- [ ] Angular

### Summary

The height of the modal currently changes with the content, e.g. the height of the modal is decreased if the user reduces the number of items in a list using search. We can set a fixed height of the modal content but doesn't take into account the actual viewport height and if we set the height of the content in `vh` we are still missing information about the space occupied by the surrounding modal (like height of header and footer that could vary).

Not sure what the best solution to this problem is. It could be nice if the content of the modal would be passed the maximum height allowed, which would be the viewport minus all modal paddings, margins, header and footer.

**Additional context** 
image
image

kevinsperrine:
Closes #3077 

**Summary**

- This is a tricky issue, and I'm not certain I love my solution here, but wanted to gather some feedback. Using the refs on the modal elements, I'm measuring the footer/header to know their height to be able to subtract it from the whole modal and know the maxHeight of the content. This is set as a style object on the ModalBody; however, since the Carbon modal's css doesn't have a fixed height (it sets a min-height, and a max-height--and the max-height varies based on screen size)this approach doesn't completely solve the issue of the modal changing height as content grows. It will prevent it from growing too large, but the modal will still change size until it reaches that max-height set on the overall modal. Additionally, I'm setting a css var of the "extrasHeight" (header and footer), so that those can be read by components that use the ComposedModal to set their own height restrictions.

Old summary: ~I'm measuring the height of the window, the modal, the header and footer, and calculating the maximum height the content can be. If the modal is _not_ `isLarge` I set that height as the maxHeight and allow the modal to grow as needed; however, if it _is_ `isLarge` then, I set that as the `height` of the modal body, so that it maintains the same height all the time.~ 

- measure the window and modal components to calculate a maximum height for the content
- set that height as the maxHeight on the modalBody style
- set a css var of the height of the header/footer combined.

- check the SelectUserModal stories or TableColumnCustomizationModal stories
- open and close groups in SelectUserModal and see how modal stays the same height (it's using isLarge)
- check the "with many columns and 'Load more'" column customization story

- confirm no height issues exist at various window sizes
- check the ComposedModal stories for regressions with isLarge knob being used or not.


packages/react/src/components/ComposedModal/ComposedModal.jsx
```diff
@@ -101,90 +122,96 @@ export const ComposedModalPropTypes = {
...
-    } = this.props;
-    const { label, title, helpText } = header;
-    // First check for dataErrors as they are worse than form errors
+  const modalWrapperRef = useRef(null);
 
-    if (__DEV__ && passiveModal && (footer || onSubmit)) {
-      warning(
-        false,
-        'You have set passiveModal to true, but also passed a footer or onSubmit handler.  Your footer will not be rendered.'
+  useEffect(() => {
```
bjornalm, line 177: I wonder if it is better to use `useLayoutEffect` here to prevent the modal from jumping in height on the first render...

packages/react/src/components/ComposedModal/ComposedModal.jsx
```diff
@@ -101,90 +122,96 @@ export const ComposedModalPropTypes = {
...
+    if (modalWrapperRef.current) {
```
bjornalm, line 178: I don't know if  this is working, but perhaps we could use the new useSizeObserver on the `CarbonComposedModal` and on the `ModalBody`. Then subtract the initial height of the ModalBody from the CarbonComposedModal to find out what the modal extras (height including header, footer and paddings etc) is. Then content max height could be set as CSS `calc(100vh - ${extrasHeight}`

kevinsperrine: Yeah, that was my initial plan, but none of the Carbon ComposedModal parts forward ref, hence the need for the wrapper and selectors. We can certainly do it that way, but it'll just be blocked for a while waiting on Carbon to merge a PR.

packages/react/src/components/ComposedModal/ComposedModal.jsx
kevinsperrine: Yes, I'll make that change.

packages/react/src/components/ComposedModal/ComposedModal.jsx
bjornalm: Yea I don't know, but I don't think this is time critical. @davidicus What is your take on this, wait a while and get a cleaner solution or merge this now?

packages/react/src/components/ComposedModal/ComposedModal.jsx
```diff
@@ -210,6 +237,8 @@ class ComposedModal extends React.Component {
               // Prevent double scrollbars
               [`${iotPrefix}--composed-modal__body--small-margin-bottom`]: error,
             })}
+            // subtracting 4rem for padding/margin
+            style={{ [isLarge ? 'height' : 'maxHeight']: `calc(${maxContentHeight}px - 3rem)` }}
```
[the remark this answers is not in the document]
bjornalm: On the other hand, if we only set `maxHeight` it will fix the double scrollbar problem, but we still need a CSS sizes solution and add those to the individual specialised components (SelectUsersModal, TableColumnCustomizationModal etc)

packages/react/src/components/ComposedModal/ComposedModal.jsx
kevinsperrine: I'm fine to wait. I can open the PRs on the carbon side to add the forwardRefs. I think that'll be better. This is pretty hacky.

kevinsperrine:
@davidicus Added the carbon issue in the summary. It was merged last week, but haven't been released yet.

davidicus (approved):

Issue #2820, opened 2021-09-10
[Table] Create coherent table stories


bjornalm:
### What package is this for?

- [x] React
- [ ] Angular

### Summary
Most of the stories have been built incrementally and combined over time which makes both the individual stories and the "book" itself incoherent. We lack a clear goal and plan for how the stories of this complex component should evolve. There is also no direct link from the story to the parts of the documentation covering the functionality displayed.

**Additional context** 
We should take a hard look at each story and ask if the purpose of the story is clear and if the name matches that purpose.  The two story categories "Table" and "StatefulTable" don't make sense when looking at the stories inside and considering the fact that almost all stories in both categories have a knob to change the table type. 

I think each story, except for the main playground, should tell a coherent story around some settings and data. A combination of settings that was turned into a story in order to fix a bug or prevent regression should be moved to unit/e2e testing instead. The documentation sections could be a good starting point when deciding which stories we should focus on. It would be nice to have some kind of taxonomy to help us in writing and maintaining the stories. Per story type we set some requirements and that could be useful for other components as well. I've thought of the following:

A. MAIN PLAYGROUND
Story is the main playground with as many knobs as possible to cover the largest amount of functionality. The documentation should show the full table documentation.

B. FUNCTIONALITY CONFIGURATION
The most common. Story shows a for the consumer relevant and coherent combination of properties and data. A story could be about showing "how row expansion works" and all the different variants and combinations using knobs, or a story showing "pagination" with all relevant configuration in knobs. We might have to write more than one story to tell the full tale of some complicated functionality but we can group them like carbon does in their table stories. Clicking the docs should open the relevant section.

C. EXTERNAL FUNCTIONALITY COMBINATION
Story code itself has custom code/components showing how to extend the table to get a certain functionality in place, e.g. the "create and save view management" functionality that partly resides outside the table. For these stories we might need custom documentation.

D. FOCUSED PLAYGROUND
Story is a playground to highlight special scenario, limitation or workaround that require interactions outside the table, e.g. 1-watson-iot-table-statefultable--with-pre-filled-search or 1-watson-iot-table-table--with-options-to-explore-column-settings. The scenario must be of interest to the common consumer if we are going to export the story, otherwise we might wanna restrict it to the dev environment.

### Specific timeline issues / requests
The work can be divided into several issues, one per story/story group, and old stories can be removed once they are covered by the new ones. That way this can be turned into an epic instead of a single giant issue.

[...]

PR #3238, opened 2022-01-27
fix(table): add hidden prop to row action

### LINKED ISSUE #3208: [Table]  Hidden property is not available for rowActions.
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

rowAction(list of actions visible on row hover) currently has only disabled prop, but no hidden prop is exposed

### To Reproduce

Following are the rowAction props. Since 'hidden' prop is missing in the list, a row-action cannot be hidden.

image

### Expected behavior

Expecting 'hidden' prop be made available so that buttons/overflowMenu items defined in rowAction can be hidden when needed.

### Environment/versions:

- OS: [e.g. MacOS, Windows]
- `carbon-addons-iot-react` version: [e.g. `v2.147.0`]

bjornalm:
Closes #3208

**Summary**
Adds hidden prop to Table rowActions

- Adds hidden prop and proptype to Table rowActions
- Update Table story example data to include a hidden action and a hidden overflow action
- Update `SimpleStatefulExample` to include a row where all actions are hidden
- Write unit tests

1. Go to https://deploy-preview-3238--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table-statefultable--simple-stateful-example
2. Enable knob hasRowActions
3. Verify that the second row is not showing any row actions at all
4. Verify that there are no actions named "Hidden" visible
5. Verify that there are no overflow actions named "Hidden overflow" visible

- Make sure the other Table stories with actions work as expected

### DIFF
--- a/packages/react/src/components/Table/StatefulTable.story.jsx
+++ b/packages/react/src/components/Table/StatefulTable.story.jsx
@@ -156,14 +156,31 @@ export const SimpleStatefulExample = () => {
       id="table"
       key={`table${demoInitialColumnSizes}`}
       {...initialState}
-      data={initialState.data.slice(
-        0,
-        select(
-          'number of data items in table',
-          [initialState.data.length, 50, 20, 5],
-          initialState.data.length
+      data={initialState.data
+        .slice(
+          0,
+          select(
+            'number of data items in table',
+            [initialState.data.length, 50, 20, 5],
+            initialState.data.length
+          )
         )
-      )}
+        .map((row, index) => ({
+          ...row,
+          rowActions:
+            index === 1
+              ? [
+                  {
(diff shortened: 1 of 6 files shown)


kevinsperrine (approved):
Looks good

davidicus (approved):

davidicus:
@bjornalm Did we not add the hidden row actions to the docs? I think we may have missed that.

bjornalm:
> @bjornalm Did we not add the hidden row actions to the docs? I think we may have missed that.

@davidicus Oh true, the prop table docs was not updated. We are also completely missing all normal documentation about the row actions but I will write such a section as part of #2820 and I'll make sure to update the prop table as well.

PR #3239, opened 2022-01-27
 fix(table): make table rows load more in skeleton state

### LINKED ISSUE #3190: [Table] Load more nested rows using disabled styling

Shady-Nessim:
Closes #3190

**Summary**

- In the [table row expansion](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table-table--row-expansion-and-load-more) story, changed the row state while loading more so that it shows skeleton states rather than the loading button.

- Add check for loading more state in TableBodyLoadMoreRow
- Add SkeletonText to each cell in the loading row
- Add unit test and a set of params/properties for it

- Go to https://deploy-preview-3239--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table-table--row-expansion-and-load-more
- Click load more on table row
- Verify skeleton loading state

- Make sure the other Table stories with actions work as expected


packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.test.jsx
```diff
@@ -13,12 +13,21 @@ const tableRowProps = {
   isLoadingMore: false,
 };
 
+const tableRowPropsLoading = {
```
bjornalm, line 16: Since these are not "real constants" in the sense that the objects can be modified and since they are declared in the global scope there is a risk that one test could modify the test data and have it affect the other tests. So, even though this was already here before this PR I would suggest a small refactor to use functions that return data objects instead e.g.

```
const getTableRowPropsNotLoading = () => ({
  id: 'tableRow01',
  tableId: 'tableId',
  testId: 'table-test',
  totalColumns: 5,
  loadMoreText: 'Load more...',
  isLoadingMore: false,
});
```

You can also shallow copy the `tableRowPropsNotLoading` object and just change the attribute `isLoadingMore`:

`const getTableRowPropsLoading = () => ({...getTableRowPropsNotLoading(), isLoadingMore: true});`

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.jsx
```diff
@@ -35,21 +35,29 @@ const TableBodyLoadMoreRow = ({
...
-        className={`${iotPrefix}--load-more-cell`}
-        data-testid={`${testId}-${id}-load-more`}
-      >
-        <Button
-          className={`${iotPrefix}--load-more-cell--content`}
-          onClick={() => onRowLoadMore(id)}
-          kind="ghost"
-          loading={isLoadingMore}
+      {isLoadingMore ? (
+        Array.from(Array(totalColumns)).map((_, index) => (
```
bjornalm, line 39: I don't this is gonna work as expected. TotalColumns was great for creating one big column using colspan but now we are going to have "skeletons" in all the cells, not only the actual data cells. 

You need to know how many "extra" columns the row has before and after the actual data columns. You can have a look to see how it is handled in the `TableBodyRowRenderer.jsx`

bjornalm (changes requested):
Nice work, the table is a complicated component with many edge cases so we need to adjust this code a bit.

packages/react/src/components/Table/TableBody/TableBodyRowRenderer.jsx
```diff
@@ -192,21 +191,18 @@ const TableBodyRowRenderer = (props) => {
 
   if (shouldLazyRender && !isVisible) {
     return (
-      <tr
-        key={`lazy-row-${row.id}`}
-        ref={rowVisibilityRef}
-        data-testid={`${tableId}-lazy-row-${row.id}`}
```
kevinsperrine, line 198: @Shady-Nessim This is the source of your `[data-testid="/lazy-row/i"]` failure. By replacing this with the new `SkeletonRow` component, it uses a different `data-testid`, so maintain consistency, you could change the `testId` prop on the the SkeletonRow component so that it overrides the testId entirely instead of inserting it into the string. So by default the SkeletonRow's testId can be simply[`${tableId}-${testId}`](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3239/files#diff-cf9d26532604dbcc58c7c16672f1f12aae701eb962096455f325bf5d79e26fe9R56) with a default `testId` of `skeleton`. But with the shouldLazyRender check we could pass 
```testId={`${tableId}-lazy-row-${row.id}`}``` instead.

File: packages/react/src/components/Table/TableBody/SkeletonRow.jsx [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+  rowVisibilityRef,
+  hasRowActions,
+  hasRowExpansion,
+  hasRowNesting,
+  hasRowSelection,
+  showExpanderColumn,
+  columns,
+}) => {
+  return (
+    <tr key={`lazy-row-${id}`} ref={rowVisibilityRef} data-testid={`${tableId}-${testId}-skeleton`}>
```
kevinsperrine, line 56:
```suggestion
    <tr key={`lazy-row-${id}`} ref={rowVisibilityRef} data-testid={`${tableId}-${testId}`}>
```

Based on my [comment here](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3239#discussion_r797053340), you can change this, and then set the default `testId` to be `skeleton`.

packages/react/src/components/Table/TableBody/SkeletonRow.test.jsx
```diff
@@ -0,0 +1,102 @@
...
+});
+
+describe('SkeletonRow', () => {
+  it('creates standard data skeleton row', () => {
+    const { container } = render(<SkeletonRow {...skeletonRowProps()} />, {
+      container: document.body.appendChild(document.createElement('tbody')),
+    });
+    expect(container).toBeDefined();
+    expect(container.rows[0].cells.length).toBe(3);
+    expect(container.getElementsByClassName(`${prefix}--skeleton__text`).length).toBe(3);
```
kevinsperrine, line 46: just a note, you can also use `toHaveLength(3)` and skip the `.length`.

packages/react/src/components/Table/TableBody/SkeletonRow.test.jsx
```diff
@@ -0,0 +1,102 @@
...
+  ],
+});
+
+describe('SkeletonRow', () => {
+  it('creates standard data skeleton row', () => {
+    const { container } = render(<SkeletonRow {...skeletonRowProps()} />, {
+      container: document.body.appendChild(document.createElement('tbody')),
+    });
+    expect(container).toBeDefined();
+    expect(container.rows[0].cells.length).toBe(3);
```
kevinsperrine, line 45: just a note, you could also do `expect(container.querySelectorAll('td')).toHaveLength(3)`

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.test.jsx
```diff
@@ -2,23 +2,55 @@ import React from 'react';
...
 import userEvent from '@testing-library/user-event';
 
+import { settings } from '../../../../constants/Settings';
+
 import TableBodyLoadMoreRow from './TableBodyLoadMoreRow';
 
-const tableRowProps = {
+const { prefix } = settings;
+
+const getTableRowPropsNotLoading = () => ({
```
kevinsperrine, line 11: again, just a note, since you're already using a function, you could just make it take a parameter to override any of the defaults. ie:

```suggestion
const getTableRowPropsNotLoading = (overrides) => ({
```

and then add the overrides with a spread operator at the bottom of the object:

```
...overides,
});
```

packages/react/src/components/Table/TableBody/TableBodyLoadMoreRow/TableBodyLoadMoreRow.test.jsx
```diff
@@ -2,23 +2,55 @@ import React from 'react';
...
+      name: 'Number',
+      isSortable: true,
+    },
+  ],
+});
+
+const getTableRowPropsLoading = () => ({
+  ...getTableRowPropsNotLoading(),
+  isLoadingMore: true,
+});
```
kevinsperrine, line 46: and then using the `overrides` comment above, you can simply call `getTableRowProps({isLoadingMore: true})` to get the Loading version instead of calling two functions.

kevinsperrine (approved):
Nice work, @Shady-Nessim!

bjornalm (approved):
Really nice work!

davidicus (approved):

PR #3241, opened 2022-01-28
V3 upgrade

kevinsperrine:
Closes #

**Summary**

- merges next into v3

- there was one merge conflict in Table.story.jsx where Bjorn's last PR moved all the table actions into a function, so I add to also re-enable the `onToggleAggregations: action('onToggleAggregations')` instead of it being commented out as it currently is in next.

- confirm the Table playground story works with aggregations and the `onToggleAggregations` callback action is called.

- tests here


bjornalm (approved):
Nice job, it seems to be working. Thanks for taking ownership of this reoccurring merge task

davidicus (approved):

PR #3257, opened 2022-02-03
docs(repo): initial onboarding documentation

### LINKED ISSUE #3173: [Repo] Add onboarding documentation - coding styles and best practices
New developers should be able to quickly familiarize themselves with the coding styles used in this repo. They should also be able to see general best practices used to develop in this project Should cover the following:
- [ ] Adding data-test ids
- [ ] using prefixes for class names
- [ ] using classname package for dynamic classnames
- [ ] defining functions not pertaining to state outside of our components
- [ ] properly memoizing functions
- [ ] using i18n object for translatable strings
- [ ] BEM naming convention
- [ ] Using Carbon tokens in place of hard coded spacing or color values
- [ ] prop naming conventions
- [ ] creating and naming feature branches
- [ ] data handling 
- [ ] ?

kevinsperrine:
Closes #3173 

**Summary**

- A list of common styles and issues to help new developers be on-boarded to this codebase.

- added on-boarding docs 

- check for typos or missing information you think is relevant

- n/a

### DIFF
--- a/packages/react/docs/guides/onboarding.md
+++ b/packages/react/docs/guides/onboarding.md
@@ -44,6 +44,45 @@ Input.defaultProps = defaultProps;
 <Input i18n={{ labelText: 'A fancy input' }} />;
 ```
 
+If you have data that needs to be embedded in that string, we're moving toward using index-based placeholders
+to make it easy for translators.
+
+```jsx
+// example needing i18n string interpolation
+const List = ({ items, title }) => {
+  <div>
+    <h3>{title}</h3>
+    <ul>
+      {items.map(({ text }) => {
+        return <li>{text}</li>;
+      })}
+    </ul>
+    <div>Item Count: {items.length}</div>
+  </div>;
+};
+
+// after refactoring
+const defaultProps = {
+  i18n: {
+    itemCountText: 'List count: {0}',
+  },
+};
+
+const List = ({ items, title, i18n }) => {
+  const mergedI18n = getMerged(defaultProps.i18n, i18n);
+
+  <div>
+    <h3>{title}</h3>
+    <ul>
+      {items.map(({ text }) => {
+        return <li>{text}</li>;
+      })}
+    </ul>
+    <div>{mergedI18n.itemCountText.replace('{0}', items.length)}</div>
+  </div>;
+};
+```
+
 ### Adding data-testid
 
 We want to make sure our components have dependable way to test all major functionality, so we provide
@@ -77,6 +116,33 @@ or Enzyme could find by test id:
 
 - `find('button[data-testid="new-window-button"]')`
 
+As components become more complex with multiple items the consumers are expected to interact with we
+need to make sure testIds are also applied to those elements.
+
+```jsx
+const Button = ({ testId = 'Button', i18n, children }) => {
+  const mergedI18n = useMerged(defaultProps.i18n, i18n);
+  return (
+    <button data-testid={testId} aria-label={mergedI18n.buttonLabel}>
+      {children}
+    </button>
+  );
+};
+
+// a simple example with multiple buttons the user will want to test for their availability based
+// on props. This allows the consumer to test for `form-buttons-reset-button` or `form-buttons-submit-button`.
+const FormButtons = ({ testId = 'form-buttons', showReset = true }) => {
+  return (
+    <div>
+      {showReset ? <Button testId={`${testId}-reset-button`}>Reset</Button> : null}
+      <Button testId={`${testId}-submit-button`}>Submit</Button>
+    </div>
+  );
+};
+
+// usage
+```
+
 ### Using prefixes for component classes
 
 Carbon uses a prefix `bx` (soon to be `cds`) to ensure unique classnames. We also follow suite in
@@ -99,10 +165,10 @@ This same structure is available within the scss files, too.
 ```scss
 // this is using @use from newer versions of sass, but isn't supported until v3 of this library
 // is released in 2022. The method in v2 is the usual @import
-@use '<react-package>/src/globals/vars' as *;
+// @use '<react-package>/src/globals/vars' as *;
 
 // for v2
-// @import '<react-package>/src/globals/vars';
+@import '<react-package>/src/globals/vars';
 
 .#{$prefix}--button {
   // styles to override the carbon button
@@ -134,12 +200,12 @@ button {
 }
 
 // Good example, v3 @use
-@use '<react-package>/src/globals/spacing' as *;
-@use '<react-package>/src/globals/colors' as *;
+// @use '<react-package>/src/globals/spacing' as *;
+// @use '<react-package>/src/globals/colors' as *;
 
 // v2 @import
-// @import '<react-package>/src/globals/vars';
-// @import '<react-package>/src/globals/spacing';
+@import '<react-package>/src/globals/vars';
+@import '<react-package>/src/globals/spacing';
 
 button {
   padding: $spacing-05;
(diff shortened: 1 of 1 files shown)


packages/react/docs/guides/onboarding.md
```diff
@@ -0,0 +1,259 @@
...
+// component
+const defaultProps = {
+  onClick: undefined,
+  i18n: {
+    labelText: 'button label',
+  },
+};
+
+const Button = ({ i18n, onClick, children }) => {
+  const mergedI18n = useMerged(defaultProps.i18n, i18n);
```
Shady-Nessim, line 31: Just wanna make sure I get this, here `mergedI18n` will have the default if `i18n` is empty/null, otherwise it will use `i18n`'s value?
If `i18n` contained more than one `labelText`, some empty and some with values, would the `useMerged` use the default only for the empty ones?

packages/react/docs/guides/onboarding.md
```diff
@@ -0,0 +1,259 @@
...
+.#{$iot-prefix}--menu {
+  padding: $spacing-02;
+}
+
+// prefix       block  element
+.#{$iot-prefix}--menu__item {
+  color: $text-01;
+}
+
+// prefix       block  element
```
Shady-Nessim, line 189: Is `active` here the modifier?

Shady-Nessim (approved):
Thanks for putting this together Kevin, very useful information and good summary to have handy.

packages/react/docs/guides/onboarding.md
kevinsperrine: Yes, I added missing comment.

packages/react/docs/guides/onboarding.md
kevinsperrine: I've replaced this with a better example and more comments to describe. Let me know if that clarifies it better.

kevinsperrine: See line 39 and following

Shady-Nessim: Much clearer, thanks for elaborating

packages/react/docs/guides/onboarding.md
```diff
@@ -0,0 +1,259 @@
...
+Below you will find major items we strive to provide in all of our code and try to check for in all
+code reviews.
+
+### Prop naming conventions
+
+When naming props all names should be in camelCase, events should be prefixed with `on` (onClick), and booleans should be prefixed with `is` or `has` (isActive, hasLabel).
+
+### Localization
+
+All non-falsy translatable strings should be passed in an `i18n` prop and the defaultProps should include
```
bjornalm, line 17: Should we say something about variables and substring substitution? Not sure if we have a standard there yet, but at least I think we can agree that we are moving away from using functions as i18n values.

For graphite they wanted to use indexes rather than variable names right?

packages/react/docs/guides/onboarding.md
```diff
@@ -0,0 +1,259 @@
...
+
+// usage
+<Button i18n={{ labelText: 'A fancy button' }}>Fancy button</Button>;
+```
+
+### Adding data-testid
+
+We want to make sure our components have dependable way to test all major functionality, so we provide
+this to our consumers with `data-testid` props on important parts of the component. For example,
+a button component could look like this:
```
bjornalm, line 51: Something about also adding test-id to the subcomponents and how that is done?

packages/react/docs/guides/onboarding.md
```diff
@@ -0,0 +1,259 @@
...
+- Block: A standalone entity that is meaningful on its own.
+- Element: An element is a part of a block that has no standalone meaning and is semantically tied to its block.
+- Modifier: A flag on a block or element. Use them to change appearance or behavior.
+
+You can learn more about BEM in this [documentation](http://getbem.com/introduction/).
+
+```scss
+// v3 @use
+@use '<react-package>/src/globals/vars' as *;
+@use '<react-package>/src/globals/spacing' as *;
```
bjornalm, line 167: comment this out for now instead of the v2 import since we are still mainly developing on v2?

bjornalm (commented):
Nice work, from the issue I'm missing the section on 

- defining functions not pertaining to state outside of our components
(by which I think that David means that functions that can be "pure", i.e. they only rely on  their input parameters to return a result and never have any side effects) should be defined in the same file but outside the component functions itself.

and I would personally like to see something about the size of the JSX and the amount of logic we should allow in there. We don't have a rule for this but it would be nice to have some guidance. I personally think it is much easier to read when the JSX is kept clean and as short as possible so complex logic can be placed in the component but outside the return-function. If the JSX is too long I think we should abstract parts of it... but what is the best way to do that? Functions within the component that returns sections of JSX or should we aim at creating additional components like the TableBodyRowRenderer. I would lean towards additional and smaller components...

packages/react/docs/guides/onboarding.md
kevinsperrine: Good call. I've added the example we discussed.

packages/react/docs/guides/onboarding.md
kevinsperrine: Yep. I've changed everything to v2 being uncommented.

packages/react/docs/guides/onboarding.md
kevinsperrine: Good idea. Added an example.

bjornalm (commented):
Reviewing this again made me think of additional info we might wanna add, see comments.

Two big things we aren't touching here are the stories and mdx files. We have developed (or are in the process of) some best practices here and I think it makes sense for that info be added here as well. From the top of my head:

Stories:
- Naming convention and placement, experimental, deprecated etc
- Playground story (all knobs, active in dev, inactive in production)
- Stories should focus on a functionality or a group of related functionality 
- Stories link to mdx docs

Mdx-files:
- Getting started section with import etc?
- Demo simple code example
- If complex, use table of content
- Structure of mdx files and folders for complex components 
- Always include a prop table and make sure new props are always added to the docs

packages/react/docs/guides/onboarding.md
```diff
@@ -0,0 +1,445 @@
...
+## Coding practices
+
+Below you will find major items we strive to provide in all of our code and try to check for in all
+code reviews.
+
+### Prop naming conventions
+
+When naming props all names should be in camelCase, events should be prefixed with `on` (onClick), and booleans should be prefixed with `is` or `has` (isActive, hasLabel).
+
+### Localization
```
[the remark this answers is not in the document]
kevinsperrine: Good points. Added.

packages/react/docs/guides/onboarding.md
```diff
@@ -0,0 +1,445 @@
...
+pick up the styles and practices we use to make this a coherent code base.
+
+## Coding practices
+
+Below you will find major items we strive to provide in all of our code and try to check for in all
+code reviews.
+
+### Prop naming conventions
+
+When naming props all names should be in camelCase, events should be prefixed with `on` (onClick), and booleans should be prefixed with `is` or `has` (isActive, hasLabel).
```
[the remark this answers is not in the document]
kevinsperrine: Yeah, I didn't think about that! Addressed.

kevinsperrine:
@bjornalm Added docs on stories and docs. See [lines](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3257/files#diff-168be1fa7e1861d6cf3a03c764f9f163123e6bde161a7f65cf9b2144147b97c5R550) and following.

packages/react/docs/guides/onboarding.md
```diff
@@ -0,0 +1,720 @@
...
+Other stories for a component should be structured around a single behavior or a group of related behaviors.
+See the [Table stories](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table--playground)
+for examples. The [playground](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table--playground) gives a broad overview of everything the table offers and other stories focus on a single function such as [sorting](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table--with-sorting)
+or [row expansion](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table--with-row-expansion).
+
+#### Naming stories
+
+The storyName should match the name of the story component with spaces and the name should clearly
+indicate the behavior being demonstrated. The goal is to make it easy for consumers to quickly scan
+and find example and documentation related to the specific functionality they're implementing.
```
kevinsperrine, line 572:
```suggestion
and find examples and documentation related to the specific functionality they're implementing.
```

davidicus (approved):
This looks great! Thanks for writing this.

bjornalm (approved):
This is really great!

Issue #3147, opened 2021-12-15
[List] Drag preview not covering full width for large rows


bjornalm:
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
The drag preview is not covering full width for large rows. This is true for all components using the List.

### To Reproduce

1. Go to https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-hierarchylist--with-nested-reorder
2. Enable knob isLargeRow
3. Drag an item 
4. Notice how the Drag preview (the blue line) and Drop target indications (blue border around other items that accepts drop) are not full width

> Please create a reduced test case in CodeSandbox:
> https://codesandbox.io/s/github/carbon-design-system/carbon-addons-iot-react/tree/next/packages/react/examples/codesandbox

### Expected behavior

Blue line is full width

### Environment/versions:

- OS: [e.g. MacOS, Windows]
- Browser: [e.g. chrome, safari]
- `carbon-addons-iot-react` version: [e.g. `v2.60.0`]

### Additional context

Screenshot from TableColumnCustomizationModal showing the same problem:
![image](https://user-images.githubusercontent.com/5167091/146200737-f29e11f8-43b0-4581-b2eb-9c995190c398.png)

### Specific timeline issues / requests

Do you want this work within a specific time period? Is it related to an
upcoming release?

_NB: The core contributors will try to work with your timeline, but it's not
guaranteed. The earlier you make a request in advance of a desired delivery
date, the better!_

bjornalm:
@dianatran18 Just verifying with you that this is a bug right?

dianatran18:
Yes that looks like a bug. It should span the full width like the other single row examples.

PR #3264, opened 2022-02-04
fix width of drag preview on large rows

Shady-Nessim:
#3147

Closes #

**Summary**

- The drag preview was not covering full width for large rows. This is now fixed and accounted for with RTL.

- `_list-item.scss` added sub styles for `--drop-targets`

- Verify on watson-iot-hierarchylist--with-nested-reorder story that the drag preview borders on large rows are correct

- Ensure regular rows dragging looks fine as well as other list actions

### DIFF
--- a/packages/react/src/components/List/ListItem/_list-item.scss
+++ b/packages/react/src/components/List/ListItem/_list-item.scss
@@ -31,6 +31,14 @@
     flex-direction: column;
     position: absolute;
     align-items: stretch;
+
+    &.#{$iot-prefix}--list-item__large {
+      margin-left: -($spacing-05);
+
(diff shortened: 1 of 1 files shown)


kevinsperrine (commented):
You're close, but it looks like the "above" rules might need some extra width in both LTR and RTL.
Screen Shot 2022-02-07 at 8 04 15 AM
Screen Shot 2022-02-07 at 8 04 04 AM

bjornalm (changes requested):
I'm still seeing the issue described by Kevin in the preview build if my drag starts by grabbing the "six dots"

bjornalm (approved):
Nice work!

PR #3267, opened 2022-02-07
Card title resize

### LINKED ISSUE #3266: [Card] Title and Subtitle do not handle tooltip overflows properly.
### What package is this for?

- [X] React

### Describe the bug

Internal JIRA item [here](https://jsw.ibm.com/browse/GRAPHITE-43003).

The hook used to determine if text overflows its container has a flaw that prevents it from running on updates to the card size and to changes in the text. 

The hook and the card component should be more dynamic to handle these edge cases. 

### To Reproduce

1. Go to [this story](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-card--with-ellipsed-title-tooltip-external-tooltip)
2. Click on expand icon to go full screen
3. Click on subtitle text
4. See that a tooltip is still applied even though the text is no longer bigger than the containing space.

davidicus:
Closes #3266

**Summary**

This PR fixes a bug in the logic for `useHasTextOverflow` hook and updates the card to handle the changes from this fix.

- Updated the `useHasTextOverflow.jsx` to make more dynamic
- Updated `Card.jsx` to handle changes need for text overlfow
- Wrote some jest unit test to test if it would change via prop update
- Wrote cypress test since our jest test manually update the values for `clientWidth` and `scrollWidth` to ensure it behaves well in actual browser. 

- Go to 
[this story](https://deploy-preview-1824--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card--with-ellipsed-title-tooltip-external-tooltip)
- click on subtitle and see that there is a tooltip
- click on isExpanded knob
- Check that tooltip is off for subtitle
then:
- Click on title and see tooltip
- Update the title in knobs to something short
- click on title and see tooltip has gone away

### DIFF
--- a/packages/react/src/components/Card/Card.test.jsx
+++ b/packages/react/src/components/Card/Card.test.jsx
@@ -417,6 +417,31 @@ describe('Card', () => {
       expect(tooltipButton).toHaveAttribute('aria-expanded', 'true');
     });
 
+    it('should remove the tooltip if the title changes to a shorter string', async () => {
+      const aLongTitle =
+        'A very very long title which will almost certainly overflow and require a tooltip and we must test these things, you know.';
+
+      const aShortTitle = 'A Title';
+      const { rerender } = render(<Card {...cardProps} title={aLongTitle} />);
+      const tooltipButton = screen.getByRole('button', {
+        name: aLongTitle,
+      });
+      expect(tooltipButton).toBeVisible();
+      expect(tooltipButton).toHaveClass(`${iotPrefix}--card--title--text__overflow`);
+      Object.defineProperty(HTMLElement.prototype, 'clientWidth', {
+        writable: true,
+        configurable: true,
+        value: 500,
+      });
+      Object.defineProperty(HTMLElement.prototype, 'scrollWidth', {
+        writable: true,
+        configurable: true,
+        value: 500,
+      });
+      rerender(<Card {...cardProps} title={aShortTitle} subtitle="This is subtitle" />);
+      expect(screen.getByTestId('Card-title-notip')).toBeVisible();
+    });
+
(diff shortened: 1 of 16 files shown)


bjornalm:
Regarding the Acceptance Test:
1. Story link does not work. Assume you meant this https://deploy-preview-1824--carbon-addons-iot-react.netlify.app/?path=/story/watson-iot-card--with-ellipsed-title-tooltip-external-tooltip
2. I don't see a subtitle, there is only the title

packages/react/src/components/Card/Card.test.jsx
```diff
@@ -417,6 +417,31 @@ describe('Card', () => {
...
+        'A very very long title which will almost certainly overflow and require a tooltip and we must test these things, you know.';
+
+      const aShortTitle = 'A Title';
+      const { rerender } = render(<Card {...cardProps} title={aLongTitle} />);
+      const tooltipButton = screen.getByRole('button', {
+        name: aLongTitle,
+      });
+      expect(tooltipButton).toBeVisible();
+      expect(tooltipButton).toHaveClass(`${iotPrefix}--card--title--text__overflow`);
+      Object.defineProperty(HTMLElement.prototype, 'clientWidth', {
```
bjornalm, line 431: Do we need to reset these prototypes to prevent them from possibly spilling over on other tests?

davidicus: We do reset them in an `afterAll` on line 401

bjornalm: I don't see `clientWidth` being reset there

bjornalm (changes requested):

davidicus:
@bjornalm added.

bjornalm (approved):
Nice work!

PR #3268, opened 2022-02-07
fix(stateful-table): fix multi-sort and advanced filters cooperatively

### LINKED ISSUE #3254: [table] multisort not working with advanced filters
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The multisort doesn't work when hasAdvancedFilter is true

### To Reproduce

2. Click on knobs tab "Sort & filter"
3. Enable `isSortable`, `hasMultiSort`, and `hasAdvancedFilter`
4. Click the overflow button for the Select column and select "Multi-sort"
5. Click the "Sort" button in the dialog
6. See that the result is empty, which is incorrect

### Expected behavior

Result should be sorted rows instead of empty

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context

image

upcoming release?

date, the better!_

kevinsperrine:
Closes #3254 

**Summary**

- fixes a couple of bugs preventing MultiSort and AdvancedFilters to be used on a stateful table at the same time.

- pass currently selected advanced filters when saving multi-sort
- only add anticipated multi-sort column if it doesn't already exist
- add e2e tests to confirm

- go to the [playground story](https://deploy-preview-3268--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--playground)
- turn on MultiSort and Advanced Filters knobs
- set advanced filters
- add multi-sort columns
- confirm table shows data
- repeat same process, but add multi-sort first and then add a filter

- confirm multi-sort and advanced filters still function as expected


bjornalm (approved):
Nice work, clean code

kevinsperrine:
> It may be when I invert the sort that the filter is blown away. Steps to reproduce:
> 
> * turn on both knobs to enable multiSort and AdvancedFilters
> * Choose second option for advanced filter and hit save
> * Click on Status multiSort dropdown and open dialogue
> * Click Save
> * Click on the Sort toggle to make it descending
> * See now that we have true values for boolean on the table.
> 
> My expectation would be that the sorting would be within the realm of the filtered data

@davidicus Very nice catch! I've fixed it. Just waiting for the deploy to update.

kevinsperrine:
@davidicus Preview is updated.

PR #3277, opened 2022-02-10
fix(table): update table pagination to include an initialWidth

kevinsperrine:
Closes #

**Summary**

- simple addition of a initialWidth on the table pagination 
In graphite one of the table tests has it wrapped in a stretch flex container that doesn't resize correctly to trigger the ResizeHandler and cause the pagination text to show as it does in our stories. This is a simple addition of the initialWidth to force it to show it on initial render.

- add initialWidth to Table pagination size hook

- check the playground story
- turn on pagination
- resize the window to ensure the pagination text still disappears when less than 500 
- resize to ensure it re-appears when greater than 500
- check the statefultable with pre-filled search story
- confirm text is hidden when less than 500px

- see above

### DIFF
--- a/packages/react/src/components/Table/Pagination.jsx
+++ b/packages/react/src/components/Table/Pagination.jsx
@@ -21,7 +21,7 @@ const SizedPagination = ({
   size,
   ...rest
 }) => {
-  const [{ width }, paginationRef] = useSizeObserver();
+  const [{ width }, paginationRef] = useSizeObserver({ initialWidth: 500 });
   return (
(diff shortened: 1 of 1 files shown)


kevinsperrine:
> I did find an issue related to the pagination size. It seems that the breakpoint for when to activate `--pagination-text-display:none;` should be larger:
> 
> image
> 
> Since this is not related to your change perhaps we create a new issue for this?

Yeah, I'll create an issue and we can explore it.

PR #3281, opened 2022-02-11
feat(hierarchy-list): call onExpandedChange after rows are expanded

### LINKED ISSUE #3191: [HierarchyList] onExpandedChange function to be called once all of the rows are opened
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
Copy from https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3037#issuecomment-1008551773 :

I am currently having a slight issue with this at the moment, as the onExpandedChange seems to be called as soon as the array of ids to open are passed in, and not when the actual rows themselves are opened.

This is an issue for us as we have a table with a couple thousand entries, and there is a good second or two delay between hitting our "Open All" button, and actually having all of the rows opened.

We'd like for the onExpandedChange function to be called once all of the rows are opened (and not immediately after getting the array of row ids passed in), as we need to display a loading spinner while this delay occur

### To Reproduce

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

kevinsperrine:
Closes #3191 

**Summary**

- I'm using an `onAfterExpand` callback to allow the HierarchyList to know when the items have all been expanded. I've added a new story to demonstrate an "Open all" behavior. I don't _love_ it, but I can't come up with a better solution. Additionally, when using the VirtualList, because of how the virtual list only renders what is currently visible, we can't (easily) know what the last element is, so I've left the previous code in place to fire the onExpandedChange event when the prop is updated when using VirtualList, because with the virtual list it happens almost instantly anyway.

- pass onAfterExpand to list components
- pass isVirtualList to ListItem
- add new story to demonstrate an "Open all" behavior
- update tests to mock timer/requestAnimationFrame
- 

- check the new [with a large number of expandable items story](https://deploy-preview-3281--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-hierarchylist--with-large-number-of-expandable-items)
- click expand all button
- change to Actions tab
- click "Open all" button
- confirm the "onExpandedChange" function isn't call until the dom is updated

- repeat with `isVirtualList` knob

- tests here


bjornalm (commented):
Nice work, this is a tricky problem. 

Perhaps one alternative (but very much related) solution that is slightly less complex could be to call the `onExpandedChange` already in the loop creating the ListItems in ListContent.jsx.  Timing wise it would be more or less the same.
```
const notifyOnLastExpansionChange = (itemId, index) => {
    const isLastItem = index === items.length - 1;
    if (isLastItem && expandedIds.includes(itemId) !== previousExpandedIds.includes(itemId)) {
    setTimeout(() => {
      window.requestAnimationFrame(() => {
        onExpandedChange(expandedIds);
      });
    }, 0);
  }
};
  
const listItems = items.map((item, index) => {
  notifyOnLastExpansionChange(item.id, index);
  return renderItemAndChildren(item, index, null, getAdjustedNestingLevel(items, 0));
});
```

packages/react/src/components/List/ListItem/ListItem.jsx
```diff
@@ -144,8 +150,17 @@ const ListItem = ({
...
   dragPreviewText,
   preventRowFocus,
+  onAfterExpand,
+  isVirtualList,
 }) => {
   const mergedI18n = useMemo(() => ({ ...ListItemDefaultProps.i18n, ...i18n }), [i18n]);
+  const previousExpansion = usePrevious(expanded, expanded);
+
+  useEffect(() => {
+    if (!isVirtualList && onAfterExpand && previousExpansion !== expanded && isExpandable) {
```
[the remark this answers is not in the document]
kevinsperrine: Yeah, that was my plan to call it on both. I'll rename it with toggle.

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -496,6 +503,25 @@ const HierarchyList = ({
...
   };
 
+  /**
+   * Used to trigger the onExpandedChange callback after the _last_ item in the expandedIds
+   * array has official expanded. Only used on regular lists, not virtual, because the virtual
+   * list only renders the rows shown, so we can't know what the _last_ item actually is.
+   *
+   * @param {string} id The ID of the item that was just expanded
+   */
+  const handleAfterExpand = (id) => {
```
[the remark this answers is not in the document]
kevinsperrine: 🤔 Hmm, yeah, you're right; however, even with a useCallback, it's still going to change quite frequently because of the dependencies on `expandedIds` and `previousExpandedIds`. Does that cancel out the benefits of useCallback?

packages/react/src/components/List/HierarchyList/HierarchyList.jsx
```diff
@@ -496,6 +503,25 @@ const HierarchyList = ({
...
+    const [lastId] = expandedIds.slice(-1);
+    const [previousLastId] = previousExpandedIds.slice(-1);
+    if (id === lastId || id === previousLastId) {
+      setTimeout(() => {
```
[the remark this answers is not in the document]
kevinsperrine: Yeah, I actually have it backwards. The setTimeout should be inside the requestAnimationFrame. I'll add a note. https://firefox-source-docs.mozilla.org/performance/bestpractices.html#how-do-i-avoid-triggering-uninterruptible-reflow

kevinsperrine:
> Perhaps one alternative (but very much related) solution that is slightly less complex could be to call the `onExpandedChange` already in the loop creating the ListItems in ListContent.jsx. Timing wise it would be more or less the same.
> 
> ```
> const notifyOnLastExpansionChange = (itemId, index) => {
>     const isLastItem = index === items.length - 1;
>     if (isLastItem && expandedIds.includes(itemId) !== previousExpandedIds.includes(itemId)) {
>     setTimeout(() => {
>       window.requestAnimationFrame(() => {
>         onExpandedChange(expandedIds);
>       });
>     }, 0);
>   }
> };
>   
> const listItems = items.map((item, index) => {
>   notifyOnLastExpansionChange(item.id, index);
>   return renderItemAndChildren(item, index, null, getAdjustedNestingLevel(items, 0));
> });
> ```

This works great for triggering it only after the last item, but will fail to trigger the `onExpandedChange` for all other changes. Then, we'd have to still use the useEffect, except _block_ it from firing a second time if the item that changed was the last item... Not sure which is better.

kevinsperrine:
> > Perhaps one alternative (but very much related) solution that is slightly less complex could be to call the `onExpandedChange` already in the loop creating the ListItems in ListContent.jsx. Timing wise it would be more or less the same.
> > ```
> > const notifyOnLastExpansionChange = (itemId, index) => {
> >     const isLastItem = index === items.length - 1;
> >     if (isLastItem && expandedIds.includes(itemId) !== previousExpandedIds.includes(itemId)) {
> >     setTimeout(() => {
> >       window.requestAnimationFrame(() => {
> >         onExpandedChange(expandedIds);
> >       });
> >     }, 0);
> >   }
> > };
> >   
> > const listItems = items.map((item, index) => {
> >   notifyOnLastExpansionChange(item.id, index);
> >   return renderItemAndChildren(item, index, null, getAdjustedNestingLevel(items, 0));
> > });
> > ```
> 
> This works great for triggering it only after the last item, but will fail to trigger the `onExpandedChange` for all other changes. Then, we'd have to still use the useEffect, except _block_ it from firing a second time if the item that changed was the last item... Not sure which is better.

Actually with a minor adjustment this works great. Pushing this up now as I think it's cleaner.

packages/react/src/components/List/ListContent/ListContent.jsx
```diff
@@ -257,10 +262,24 @@ const ListContent = ({
       ...(!hasChildren && item.hasLoadMore ? [renderLoadMore(item, isLoadingMore, level)] : []),
     ];
   };
+  const previousExpandedIds = usePrevious(expandedIds, expandedIds);
+  const notifyOnLastExpansionChange = (itemId) => {
```
bjornalm, line 266: Hmm... there is a problem if we wanna support "toggle" and not only "onExpand" changes since this does not work if you expand row 1, then expand row 2 and then collapse row 1.

bjornalm: But perhaps we can take the last item from the diff between expandedIds and previousExpandedIds?

kevinsperrine: Yep! The diff works. I have to do it both ways (ie. the diff of current to previous and then previous to current) to handle both opens and closes, but it seems to work regardless of order.

packages/react/src/components/List/ListContent/ListContent.jsx
```diff
@@ -257,10 +262,33 @@ const ListContent = ({
...
+    // get the lastId of the diff between previous and current, this handles closes
+    const [previousLastId] = previousExpandedIds
```
bjornalm, line 268: We should probably create the diffs outside the loop in order to not upset Big O

kevinsperrine: 🤦 duh. corrected. Thanks!

packages/react/src/components/List/ListContent/ListContent.jsx
```diff
@@ -257,10 +262,30 @@ const ListContent = ({
       ...(!hasChildren && item.hasLoadMore ? [renderLoadMore(item, isLoadingMore, level)] : []),
     ];
   };
+  const previousExpandedIds = usePrevious(expandedIds, expandedIds);
+  // get the lastId of the diff between previous and current, this handles closes
+  const [previousLastId] = previousExpandedIds.filter((id) => !expandedIds.includes(id)).slice(-1);
```
bjornalm, line 267: Sorry, just realised that we can perhaps do a little more for performance here. We can't use `useMemo` but I assume lodash `isEqual` is still more efficient than comparing every item in X with every item in Y.

```
const selectionChanged = !isEqual(expandedIds, previousExpandedIds);
const previousLastId = selectionChanged && previousExpandedIds.filter((id) => !expandedIds.includes(id)).slice(-1)[0];
...
```

it might require some additional changes to get this to work

kevinsperrine: Done.

bjornalm: Sweet, I tested it, it works fine and you already have the approval.

PR #3282, opened 2022-02-11
Some changes and fixes to the idle user detection logic

blechner:
**Summary**

- IdleTimer logic uses a cookie to control its current state (before inactivity timeout expired and after inactivity timeout expired). However, after a user logs out in one tab, the cookie is cleared by the logout process and other active tabs running IdleTimer will end up in an invalid state where the cookie does not exist anymore, which may cause the Logout confirmation dialog to popup at random times. The fix consists of considering a deleted cookie to be the same as entering the state where inactivity timeout had expired, because the only reason why the cookie would be deleted is after a logout event in one of the active tabs, so other tabs would then be redirected to the logout page when they detect that the cookie has been deleted.

- Another fix made in this PR is to disable the "Stay Logged in" button after the user has clicked "Log out" in the Inactivity Log Out Confirmation dialog, because the logout action is a redirect action and it may not happen instantly upon clicking the Log out button.

- Handling the scenario where the cookie is deleted in IdleTimer, adding unit tests and replacing plain numbers with a variable in the tests
- Disabling the Stay Logged in button after Logout button is clicked

Acceptance tests for the IdleTimer fix:

- Include this line inside the `onSecondarySubmit` handler function in `IdleLogoutConfirmationModal` to simulate the cookie deletion that happens during an actual logout process:
```
document.cookie = `${idleTimeoutData.cookieName}=; Max-Age=0; path=/; domain=${routes.domain};`
```
- Then open the IdleLogoutConfirmationModal story in PAL storybook in 2 separate tabs. When the confirmation dialog shows up in both tabs, click "Log out" in one of them and notice that the countdown will end and the tab will be redirected to the same page as the one where you clicked logout. (Before the fix, the other tab would intermittently show and then hide the logout confirmation dialog due to bad handling of the "cookie deleted" scenario.)

Acceptance tests for the "disable Stay Logged in" fix:

- In the IdleLogoutConfirmationModal story, click "Log out" and notice that the "Stay Logged in" button is disabled just before the redirect happens.

- All tests are still passing. A new test for IdleTimer has been added to cover the scenario where the cookie is deleted by another browser tab.


kevinsperrine (commented):
Thanks, @blechner. This all looks good. The only question that came up is that when you click "Log out" in one tab, should it be possible to click "Stay logged in" in the other? In the same vein, if I manually delete the cookie in dev tools should it be possible to click "Stay logged in"?

packages/react/src/components/SuiteHeader/util/IdleTimer.js
```diff
@@ -39,8 +39,9 @@ class IdleTimer {
     }
     // Initialize the setInterval thread
     this.intervalHandler = setInterval(() => {
-      // Check if user is idle
-      if (this.getUserInactivityTimeout() < Date.now()) {
+      // Check if user is idle or if userInactivityTimeout is not a number, which indicates that a logout has already happened in another tab
+      const userInactivityTimeout = this.getUserInactivityTimeout();
+      if (Number.isNaN(userInactivityTimeout) || userInactivityTimeout < Date.now()) {
```
bjornalm, line 44: nit: we could use well named variables to explain what is going on, they normally stand the test of time better than comments.

```
const userInactivityTimeoutValue = this.getUserInactivityTimeout();
const userIsIdle = userInactivityTimeoutValue < Date.now();
const userLoggedOutInDifferentTab = Number.isNaN(userInactivityTimeoutValue);
if (userIsIdle || userLoggedOutInDifferentTab) {
...
```

packages/react/src/components/SuiteHeader/util/IdleTimer.js
```diff
@@ -77,6 +78,10 @@ class IdleTimer {
     )};expires=${expires};path=/;domain=${this.COOKIE_DOMAIN};`;
   }
 
+  deleteUserInactivityTimeout() {
```
bjornalm, line 81: Perhaps call it `deleteUserInactivityTimeoutCookie`?

packages/react/src/components/SuiteHeader/util/IdleTimer.js
```diff
@@ -77,6 +78,10 @@ class IdleTimer {
...
+    document.cookie = `${this.COOKIE_NAME}=;Max-Age=0;path=/;domain=${this.COOKIE_DOMAIN};`;
```
bjornalm, line 82: Do we need to set the "expires" to a date in the past?

blechner: This is an example I found on stackoverflow to programmatically delete a cookie. If there is a better way, I don't mind changing it.

packages/react/src/components/SuiteHeader/util/IdleTimer.js
blechner: Good idea, I'll change the method name. I didn't include `Cookie` in the name because we already have `setUserInactivityTimeout` and `updateUserInactivityTimeout`. So, perhaps I'll add the `Cookie` suffix to those as well.

packages/react/src/components/SuiteHeader/util/IdleTimer.js
blechner: Thanks, good idea, I'll make the changes (but `userLoggedOutInDifferentTab` comes first in the `||` logic)

packages/react/src/components/SuiteHeader/util/IdleTimer.js
```diff
@@ -65,7 +66,7 @@ class IdleTimer {
       document.cookie.split('; ').reduce((r, v) => {
         const parts = v.split('=');
         return parts[0] === this.COOKIE_NAME ? decodeURIComponent(parts[1]) : r;
-      }, Date.now()),
+      }, NaN),
```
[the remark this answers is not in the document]
blechner: Thanks, I replaced the existing code with your version of the `getUserInactivityTimeout` method. However, I needed to add the `?` optional chaining to `cookie.split` to avoid throwing an error when the cookie has been deleted.
So, the before the `return` statement, the line is now like this:
```
const cookieValue = cookie?.split('=')[1];
```

packages/react/src/components/SuiteHeader/util/IdleTimer.js
bjornalm: no, I double checked and the `Max-Age=0` apparently does the same trick as setting the expires

blechner:
> Thanks, @blechner. This all looks good. The only question that came up is that when you click "Log out" in one tab, should it be possible to click "Stay logged in" in the other? In the same vein, if I manually delete the cookie in dev tools should it be possible to click "Stay logged in"?
> 
> logout

yes, in other tabs, the user would still be able to click "Stay Logged In", however, if a logout happen in any other tab, all API calls in a tab would redirect the user to the login page.

bjornalm (approved):
Nice work!

blechner:
> > Thanks, @blechner. This all looks good. The only question that came up is that when you click "Log out" in one tab, should it be possible to click "Stay logged in" in the other? In the same vein, if I manually delete the cookie in dev tools should it be possible to click "Stay logged in"?
> > logout
> 
> yes, in other tabs, the user would still be able to click "Stay Logged In", however, if a logout happen in any other tab, all API calls in a tab would redirect the user to the login page.

@kevinsperrine I've changed this behavior so now when the user logs out in one tab, the others will detect that the cookie has been cleared and also redirect to the logout page.

kevinsperrine (approved):
Thanks, @blechner!

davidicus (approved):

PR #3290, opened 2022-02-15
fix(IdleTimer): Adding the 'Secure;' property to the IdleTimer cookie and fixing the strategy to update the countdown in the logout confirmation dialog

blechner:
**Summary**

- Making the `_user_inactivity_timeout` cookie `Secure;` only when running on `https`.
- Calculating the countdown using the actual time elapsed since the expired timeout cookie instead of just decreasing countdown by 1 on every `setInterval` execution. This is to account for the browser lowering the priority of javascript execution in minimized and inactive tabs, which causes the actual interval of `setInterval` to be higher than the specified interval in this scenario. The effect of the slowdown of `setInterval` in inactive tabs is more noticeable in Safari.

Secure cookie:
- When running the storybook in a netlify build (which is https), open the browser dev tools and confirm that the `_user_inactivity_timeout` cookie has the `secure` flag set. When running the storybook locally (`yarn storybook`), check that the cookie does not have the `secure` flag set.

Fixed countdown in logout confirmation dialog:
- In Safari, open 2 tabs with the `IdleLogoutConfirmationModal` story (not side by side, because one tab needs to be considered "inactive" by Safari). Wait for the active tab to reach a low countdown value. Then click the other tab and notice that, although the countdown value is higher, it will immediately update to the same value that was in the first tab. (To make it easier to test this, change the countdown from 10 to 30 in the `IdleLogoutConfirmationModal` story code.)

- Some unit tests have been fixed to account for this change

### DIFF
--- a/packages/react/src/components/SuiteHeader/util/IdleTimer.test.js
+++ b/packages/react/src/components/SuiteHeader/util/IdleTimer.test.js
@@ -1,3 +1,6 @@
+import { act, waitFor } from '@testing-library/react';
+import MockDate from 'mockdate';
+
 import IdleTimer from './IdleTimer';
 
 let timer;
@@ -17,6 +20,7 @@ describe('IdleTimer', () => {
     timer.cleanUp();
     window.document.cookie = originalWindowDocumentCookie;
     jest.useRealTimers();
+    MockDate.reset();
   });
   it('starts the timer', () => {
     // Make sure that there are default values for all IdleTimer variables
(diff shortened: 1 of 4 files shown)


packages/react/src/components/SuiteHeader/util/IdleTimer.test.js
```diff
@@ -57,11 +61,11 @@ describe('IdleTimer', () => {
     jest.runOnlyPendingTimers();
     // only onIdleTimeoutWarning should have been fired, and countdown should have decreased
     expect(timer.onIdleTimeoutWarning).toHaveBeenCalledWith(timer.COUNTDOWN_START);
-    expect(timer.countdown).toEqual(timer.COUNTDOWN_START - 1);
+    expect(timer.countdown).toEqual(timer.COUNTDOWN_START - 2);
```
kevinsperrine, line 64: Why minus 2 now?

kevinsperrine (approved):
Looks good and works as expected. Just the question about the changed test.

leetosc (approved):

packages/react/src/components/SuiteHeader/util/IdleTimer.test.js
kevinsperrine: Got it, thanks for the clarification. I don't think it needs a comment.

davidicus (approved):

PR #3294, opened 2022-02-15
feat(status-bar): add className prop

kevinsperrine:
Closes #

**Summary**

- Another small addition necessary for the StatusBar in graphite

- add className prop
- add tests
- add to docs

- tests pass

- tests here

### DIFF
--- a/packages/react/src/components/ProgressBar/ProgressBar.jsx
+++ b/packages/react/src/components/ProgressBar/ProgressBar.jsx
@@ -16,6 +16,7 @@ import {
 const { prefix, iotPrefix } = settings;
 
 const propTypes = {
+  className: PropTypes.string,
   /* the label show above the bar */
   label: PropTypes.string.isRequired,
   /* the text shown below the bar */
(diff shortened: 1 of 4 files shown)


packages/react/src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -76,17 +78,18 @@ const defaultProps = {
...
   light,
-  value,
-  valueUnit,
   max,
   renderIcon,
   renderIconByName,
   thresholds,
-  i18n,
+  value,
+  valueUnit,
```
kevinsperrine, line 92: I just re-ordered these by alphabetical automatically. sorry

bjornalm (approved):

packages/react/src/components/ProgressBar/ProgressBar.jsx
davidicus: Have I found a kindred spirit? I have thought about adding a prettier/eslint rule.

davidicus (approved):

packages/react/src/components/ProgressBar/ProgressBar.jsx
bjornalm: I also like the alphabetical order, if we could have that automated it would be great. Same goes for the proptypes

PR #3300, opened 2022-02-17
Repo: upgrade Carbon to 10.53

kevinsperrine:
Closes #

**Summary**

#### 10.50
- changed publicAPI.test to update warning message from MultiSelect.Filterable

#### 10.51
- no test failures -- updated snapshots

#### 10.52
- Had to add mocks for window.mediaMatches, but otherwise no new test failures

#### 10.53
- No new test failures

- upgraded to carbon 10.53
- added mock for window.matchMedia
- updated copy for warning message for deprecated MultiSelect.Filterable
- replaced temp ref div with forwardedRef in Table/Pagination
- replaced useDropDownTitleFixer hook with `renderSelectedItem` in new version of Carbon

- check titles on selected elements in ColorDropdown, IconDropdown, SimpleIconDropdown, TableViewDropdown
- check table pagination resizes correctly when below 500px (still have open issue to adjust this minimum)

- standard carbon upgrade work to check stories


bjornalm:
Only real difference I found comparing with next is this. Can't tell if it is an issue or a design update.
image
image

bjornalm (approved):
Compared all normal stories with next and didn't find any issues
Also tested all experimental stories

kevinsperrine:
> Only real difference I found comparing with next is this. Can't tell if it is an issue or a design update. image image

Looks like this was intentional. https://github.com/carbon-design-system/carbon/pull/10291

packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -34235,17 +34230,6 @@ Map {
   },
   "PasswordInput" => Object {
     "$$typeof": Symbol(react.forward_ref),
-    "defaultProps": Object {
```
davidicus, line 34238: do we need to provide any of these deleted default props? the className for instance.

kevinsperrine: The defaults are still there. Carbon has taken to inlining them instead of using `defaultProps`, so they're not showing up in our snapshots, but as for the className in particular. It was actually broken for a _very_ long time. Notice it's actually showing `${prefix}` instead of `bx` the default wasn't using template strings, but a normal string. https://github.com/carbon-design-system/carbon/blame/c2a62f9ce71c9570e9de21d266ccfe81f0ae3e18/packages/react/src/components/TextInput/PasswordInput.js#L339

davidicus (approved):

Issue #3304, opened 2022-02-18
[Table] Stories slow rerender


bjornalm:
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
Table stories noticeably rerenders when changing a knob 

See comment
https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3274#issuecomment-1040881902

### To Reproduce

1. Go to https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table--playground
2. Change any knob 
3. See the slow rerendering

> Please create a reduced test case in CodeSandbox:
> https://codesandbox.io/s/github/carbon-design-system/carbon-addons-iot-react/tree/next/packages/react/examples/codesandbox

### Environment/versions:

- OS: [e.g. MacOS, Windows]
- Browser: [e.g. chrome, safari]
- `carbon-addons-iot-react` version: [e.g. `v2.60.0`]

### Specific timeline issues / requests

Do you want this work within a specific time period? Is it related to an
upcoming release?

_NB: The core contributors will try to work with your timeline, but it's not
guaranteed. The earlier you make a request in advance of a desired delivery
date, the better!_

PR #3322, opened 2022-02-21
fix(table): advanced filter crash fix

### LINKED ISSUE #3319: [table] advanced filters throws error when columns are removed
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The advanced filters throws error when the columns prop is updated to an empty array after initial table loading.
This can for example happens when the columns are emptied while loading a new view with an in advanced unknown
number of columns.

### To Reproduce

Note! Will be reproducible in the table--playground
story after PR for issue https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3304 is merged into next.

2. Click on Tab "Sort & filter"
3. Enable knob advanced filters 
4. Click on Tab "States"
5. Enable knob "Demo empty columns in loading state (columns)"

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

bjornalm:
Closes #3319

**Summary**
Small fix to prevent the TableToolbarAdvancedFilterFlyout.jsx from crashing if the columns prop array is empty

- Unit test is passing

(The bug is not easily testable without modifying the table main story like is being done for #3304)

- Advanced filters work as before

### DIFF
--- a/packages/react/src/components/Table/TableToolbar/TableToolbarAdvancedFilterFlyout.test.jsx
+++ b/packages/react/src/components/Table/TableToolbar/TableToolbarAdvancedFilterFlyout.test.jsx
@@ -672,7 +672,7 @@ describe('TableToolbarAdvancedFilterFlyout', () => {
     );
   });
 
-  it('should not break on empty column arrry prop', () => {
+  it('should not break on empty column array prop', () => {
     const onApplyAdvancedFilter = jest.fn();
     const onCancelAdvancedFilter = jest.fn();
     render(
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Table/TableToolbar/TableToolbarAdvancedFilterFlyout.test.jsx
```diff
@@ -671,4 +671,50 @@ describe('TableToolbarAdvancedFilterFlyout', () => {
       'true'
     );
   });
+
+  it('should not break on empty column arrry prop', () => {
```
bjornalm, line 675:
```suggestion
  it('should not break on empty column array prop', () => {
```

kevinsperrine (approved):
Nice catch! Another bug bites the dust.

davidicus (approved):

PR #3324, opened 2022-02-22
docs(table): remove react wrapper for table stories

### LINKED ISSUE #3304: [Table] Stories slow rerender

bjornalm:
Closes #3304

**Summary**
Our stories using the StatefulTable had become slow due to the fact that the whole story was re-generated each time a knob was changed. This happened when the story was wrapped in a React element using the story `decorator`, which we did for each story that needed DragAndDrop or needed a react state.

- The Tables are no longer re-generated from scratch when a knob changes. Some knobs still explicitly re-renders the table for demo purposes, e.g. when view related knobs change for the StatefulTable we need to use `key` to get the change picked up.
- Removed DnD wrapper from the decorator and added the `<DragAndDrop>` element in the story itself.
- Created a non-react dependent version of React's `useState` called `useStoryState`. For simplicity of use I built it with the same API as react. In order to support our more advanced stories where different knobs also affects the state I added the option of providing a dependency array that when changed will "reset" the state to the current initial value.

Not re-generating the story when knob changes has exposed yet more bugs in the table. I have filed issues for the once I have found in my testing but there might be more.

1. Go to https://deploy-preview-3324--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table--playground
2. Click on the Title & Toolbar tab
3. Edit the knob `secondaryTitle` and make sure the table doesn't reload in a noticeable manner
4. Test with more knobs (remember that some knobs explicitly re-renders the table using the key)
5. Test the other "new" table stories to make sure there is no slow re-generation that makes the table flicker when knob changes

https://deploy-preview-3324--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-table--playground

All "new" table stories works as expected, no crashes when knobs are changed. Note, not all knobs in the stories are applied when changed when the type of table is StatefulTable (since the state of the knob has moved into the table after the initial render).

### DIFF
--- a/packages/react/src/internal/storyState.js
+++ b/packages/react/src/internal/storyState.js
@@ -1,0 +1,80 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import addons from '@storybook/addons';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { FORCE_RE_RENDER, STORY_RENDERED, STORY_CHANGED } from '@storybook/core-events';
+import { isEqual } from 'lodash-es';
+
+let callOrder = 0;
+let states = [];
+let dependencies = [];
+
+addons.getChannel().addListener(STORY_CHANGED, () => {
+  // Clear the global state and reset the capll order when
+  // the user selects another story
+  callOrder = 0;
+  states = [];
+  dependencies = [];
+});
+
+addons.getChannel().addListener(STORY_RENDERED, () => {
+  // Reset on every render so that the index of the useStoryState
+  // doesn't keep increasing.
+  callOrder = 0;
+});
+
+/**
+ * useStoryState can be used as a substitute for React's useState when you for
+ * performance reasons don't want to wrap the story in a react element.
+ * For more advanced stories it is also possible to add dependencies to the state
+ * that when changed will force the state to use the value passed in (i.e. a new initial value)
+ * and trigger a force rerender of the story to show the modified changes.
+ *
+ * Example 1
+ * The dependencies can be a one or more knob values affecting the initial value like the Table's
+ * 'hasLoadMore' affects the 'data' prop.
+ *
+ * Example 2
+ * The dependencies can be the value of a knob representing the same table prop as the state does
+ * so that that they can be used together, i.e. when the knob is modified the state
+ * is also updatet with the same value.
+ * @param {any} initialValue The inital value of the state
+ * @param {any} currentDep The dependencies that when changed will force a reset to the
+ * initial value and re-render
+ * @returns
+ */
+export default function useStoryState(initialValue, currentDep) {
+  // The call order is the "id" used to link a stored state with the actual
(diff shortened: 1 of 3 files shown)


kevinsperrine (approved):
Awesome work. How did you track this down?

bjornalm:
> Awesome work. How did you track this down?

Thanks, tedious trial and error... we did also have a couple of old StatefulTable stories where this didn't happen so I just had to compare with them to figure out what the difference was. The decorator was not my first guess, I suspected a knob or table setting initially.

packages/react/src/internal/storyState.js
```diff
@@ -0,0 +1,80 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { FORCE_RE_RENDER, STORY_RENDERED, STORY_CHANGED } from '@storybook/core-events';
+import { isEqual } from 'lodash-es';
+
+let callOrder = 0;
+let states = [];
+let dependencies = [];
+
+addons.getChannel().addListener(STORY_CHANGED, () => {
+  // Clear the global state and reset the capll order when
```
davidicus, line 12: typo:
```suggestion
  // Clear the global state and reset the call order when
```

packages/react/src/internal/storyState.js
```diff
@@ -0,0 +1,80 @@
...
+ * and trigger a force rerender of the story to show the modified changes.
+ *
+ * Example 1
+ * The dependencies can be a one or more knob values affecting the initial value like the Table's
+ * 'hasLoadMore' affects the 'data' prop.
+ *
+ * Example 2
+ * The dependencies can be the value of a knob representing the same table prop as the state does
+ * so that that they can be used together, i.e. when the knob is modified the state
+ * is also updatet with the same value.
```
davidicus, line 39: typo:
```suggestion
 * is also updated with the same value.
```

packages/react/src/internal/storyState.js
```diff
@@ -0,0 +1,80 @@
...
+
+  // Set the value to the initialValue if this is the first render, if not
+  // the value stays unchanged
+  const isInitialRender = states[index] === undefined;
+  const value = isInitialRender ? initialValue : states[index];
+  states[index] = value;
+
+  // If the dependencies has changed the state is "reset" to use the
+  // new initial value and force a rerender of the story just like
+  // a normal update would, thus the resetted value is picked up during
```
davidicus, line 59: typo: 
```suggestion
  // a normal update would, thus the reset value is picked up during
```

davidicus (approved):
awesome. great detective work.

PR #3325, opened 2022-02-22
refactor(table-card): remove styled components

### LINKED ISSUE #3315: [TableCard] Convert styled-components to scss

kevinsperrine:
Closes #3315 

**Summary**

- removes styled components from TableCard

- removes styled components from TableCard
- removes jest test related to styled components in expanded state and moves it to cypress
- updates snapshots

- tests pass 

- no visual changes to [TableCard stories](https://deploy-preview-3325--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-tablecard--with-multiple-actions)


packages/react/src/components/Dashboard/DashboardGrid.test.e2e.jsx
```diff
@@ -1162,6 +1162,7 @@ DashboardAllCardsAsResizable.propTypes = {
 describe('DashboardGrid', () => {
   beforeEach(() => {
     MockDate.set(1537538254000);
+    cy.viewport(1680, 900);
```
bjornalm, line 1165: Why are we setting the viewport to the same thing before and after all?

packages/react/src/components/TableCard/_table-card.scss
```diff
@@ -52,3 +52,105 @@
...
+    tr {
+      height: $spacing-07;
+    }
+  }
+
+  .#{$prefix}--data-table tbody tr {
+    height: $spacing-08;
+  }
+  .#{$prefix}--data-table-content + .#{$prefix}--pagination {
+    border: 1px solid #dfe3e6;
```
bjornalm, line 94: Special color or miss?

bjornalm (approved):
Looks good, just a question and possibly the hard coded color.

packages/react/src/components/TableCard/_table-card.scss
kevinsperrine: Just missed it. I'll find it's proper token. Thanks!

packages/react/src/components/Dashboard/DashboardGrid.test.e2e.jsx
kevinsperrine: I've realized cypress loses this value often. That's why in some of the automated tests items will get pushed off screen until you resize, so I've started adding it to tests as I come across it to keep the viewport rendering correctly. It's probably overkill to have it in both before/after here though.

blechner (approved):

packages/react/src/components/TableCard/_table-card.scss
```diff
@@ -52,3 +52,105 @@
...
+    height: 100%;
+    td {
+      white-space: nowrap;
+    }
+  }
+  .#{$prefix}--data-table thead tr:nth-child(2) {
+    height: $spacing-10;
+
+    th {
+      padding-top: 5px;
```
davidicus, line 110: did these not scream at you in our linter for using a pixel value?

davidicus (approved):
This looks good. We may want to use the rem function to convert our pixel values. I am  not sure why the linter did not bark for this.

packages/react/src/components/TableCard/_table-card.scss
kevinsperrine: Nope, I just tried running it and scanning that file directly, and it still passes. Not sure what's going on there. I wrapped all the px with rem them.

PR #3327, opened 2022-02-22
fix(SuiteHeader): Adding support for basic browser workflows in Suiteheader actions.

blechner:
**Summary**

- SuiteHeader navigational actions are now `<a>` elements and not `<button>` elements anymore, so that it is now possible to right click and select "open in a new tab/window" and also ctrl+click to force opening the page in a new tab.
- The rules and effects of the `onRouteChange` custom function are still maintained for left clicks and ctrl+left clicks. Also, the default behavior of existing SuiteHeader navigational actions (whether they open in the same tab or in another tab by default) is also maintained for left clicks (because ctrl+left click forces the navigation to open a new tab). However, for right click -> "open in a new tab/window" action, the `onClick` handlers of navigational buttons are bypassed and the `href` attribute contains the target URL that will open in a new tab/window. The only exception is the "Log Out" button, it is still a button html element that does not allow right click navigational actions. Even in the case of some SuiteHeader menu items that were `<a>` and showed the "open in a new tab", the new tab would open a `about:blank#blocked` because the `href` of these menu items had a `javascript:void(0)` string, and now they have the same URL that the `onClick` handler would use to do the redirect.
- The underlying PAL `Header` used by `SuiteHeader` did not support these browser workflows of `<a>` elements for top level actions that don't have `childContent`. So, a minor change was made in `HeaderAction.jsx` so that it forwards `href`, `rel` and `target` to the `HeaderGlobalAction` carbon component (which basically just renders a `Button`, and carbon `Button` renders an `<a>` if at least `href` is passed to it).

Before:
image

After:
image

Before:
image

After:
image

Before:
image

After:
image

Log Out button is still a button element:
image

- Open the SuiteHeader (and its underlying components) stories and check that you can right click and open the links in a new tab or window.
- Check that you can ctrl+click (cmd+click for Mac) navigational actions and they will open in a new tab.
- Go to the new story ("Navigational actions blocked") and check that clicks and ctrl+clicks don't trigger any navigational actions in the header actions and menu items.
- Check that the Log Out button does not support opening in a new tab from the context menu, nor opening in a new tab through ctrl+click.

- Some tests have been updated to adjust the screen selector code that were relying on `button` elements.
- A new test has been added for `HeaderAction` component to fulfill the minimum coverage requirement.


packages/react/src/components/Header/HeaderAction/HeaderAction.test.jsx
```diff
@@ -47,4 +47,23 @@ describe('HeaderAction', () => {
...
+      },
+      index: 0,
+      renderLabel: true,
+    };
+
+    render(<HeaderAction {...mockProps} />);
+    expect(true).toBeTruthy();
+    expect(screen.queryByText('myButton')).not.toBeInTheDocument();
+    expect(screen.getByTestId('menu-item-myhelpLabel-global')).toBeVisible();
+  });
```
kevinsperrine, line 68: @blechner Can we make this test a little more tangible? Maybe something like: 

```suggestion
    expect(screen.getByTestId('menu-item-myhelpLabel-global')).toBeVisible();
    expect(screen.getByTestId('menu-item-myhelpLabel-global').tagName).toEqual('A');
    expect(screen.getByTestId('menu-item-myhelpLabel-global')).toHaveAttribute('href', 'https://www.ibm.com');
  });
```

kevinsperrine (commented):
Looks good overall. Just the one note about making the test a little more robust. I still need to test it more thoroughly in storybook once the deploy preview is finished.

packages/react/src/components/Header/HeaderAction/HeaderAction.test.jsx
blechner: Sure, I hadn't added it before because I was not sure how to test if an html element was an `<a>`, I didn't know about this `.tagName` property of `getByTestId`. All the examples of checking if an element was `button` used the `getByRole`, and for some reason the `<a>` don't have the role attribute set, so I didn't know how to test :)
Thanks for the suggestions, I'll commit your suggestion.

kevinsperrine (approved):
LGTM

bjornalm:
Does it really makes sense to offer the possibility to open a menu/dropdown in a new window? I think that it only makes sense for navigational links/buttons

image

kevinsperrine:
> Does it really makes sense to offer the possibility to open a menu/dropdown in a new window? I think that it only makes sense for navigational links/buttons
> 
> image

He means the profile link here, not the menu itself.

bjornalm:
> > Does it really makes sense to offer the possibility to open a menu/dropdown in a new window? I think that it only makes sense for navigational links/buttons
> > 
> 
> He means the profile link here, not the menu itself.

I know, but I thought it was a new side effect now that the menu button itself had become a link and could be opened in a new window. Apparently it already works that way in next, so my comment can be ignored.

bjornalm (approved):
LGTM

kevinsperrine:
> 
Ah, understood.

blechner:
> Does it really makes sense to offer the possibility to open a menu/dropdown in a new window? I think that it only makes sense for navigational links/buttons
> 
> image

Sorry for the late response. I haven't changed the behavior of header actions that trigger dropdown menus. I only made sure that all navigational actions (header buttons, dropdown menu items and dropdown panel buttons) are now html links.

cgirani (approved):
LGTM

davidicus (approved):

PR #3335, opened 2022-02-23
refactor(wizard-footer): remove styled components

### LINKED ISSUE #3317: [WizardFooter] Convert styled-components to scss (and story)

kevinsperrine:
Closes #3317 

**Summary**

- removes styled components from WizardFooter

- replaces one styled component div with a div and scss class

- WizardInline story footers match those of next

- WizardInline story footers match those of next


bjornalm (approved):
Nice work, I'm glad to see the styled components getting removed

davidicus (approved):

PR #3338, opened 2022-02-23
feat(table): add overflow menu to batch actions

### LINKED ISSUE #3321: [Table] Add overflow menu to multiselect action bar
Depended on by https://jsw.ibm.com/browse/GRAPHITE-45682
Ability to pass a group of actions into a menu, "More", with a three-dot. These items appear in the menu intentionally, not based on whether they fit or not. 
Include common styles and functions as props:
• Danger
• Hidden
• Disabled
• Icon
Be sure that large text strings do not blow out of the menu (either wrap, or allow the menu to grow)

kevinsperrine:
Closes #3321 

**Summary**

- Adds an overflow menu to the table Batch actions bar. This follows the same dynamic as the overflow menu in the table toolbar. All items appearing in the overflow menu must be passed `isOverflow:true`. It also adds support for `hidden`, `disabled`, `isDelete`, `hasDivider`, etc. to match the other usages.

- update proptypes for table batch actions
- add `hidden`, `disabled` to batch actions (those not in overflow menu)
- add `isOverflow` and display those items that aren't hidden in the overflow menu
- add tests to confirm
- update snapshots
- update with selection and batch action story

- Go to the [with selection and batch actions story](https://deploy-preview-3338--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-selection-and-batch-actions)
- confirm the overflow menu exists
- confirm the actions are fired for clicking on those items (that aren't disabled)
- confirm there are only 3 items showing in the overflow menu (the 4th one is hidden)
- confirm the "Expunge" item is red when hovered
- confirm "Reassign" is disabled
- confirm there is a divider above both "Reassign" and "Expunge" (`hasDivider`)

- ensure batch actions still work as expected


kevinsperrine:
@JordanWSmith15 Does this meet your requirements?

bjornalm:
When I hover on a normal batch action the white divider is hidden:
image

But when I hover on the overflow button the divider is showing
image

kevinsperrine:
> When I hover on a normal batch action the white divider is hidden: image
> 
> But when I hover on the overflow button the divider is showing image

Good eye. Fixed on hover and focus.

packages/react/src/components/Table/Table.test.jsx
```diff
@@ -3050,4 +3050,167 @@ describe('Table', () => {
...
+          }}
+          actions={{
+            toolbar: {
+              onApplyBatchAction,
+            },
+          }}
+        />
+      );
+
+      userEvent.click(screen.getByRole('button', { name: 'open and close list of options' }));
```
bjornalm, line 3123: nit: we could put the string 'open and close list of options' in a const called `overflowButtonName` or something since it is a bit of a weird and long name

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -138,12 +139,27 @@ const propTypes = {
     customToolbarContent: PropTypes.node,
     /** available batch actions */
     batchActions: PropTypes.arrayOf(
-      PropTypes.shape({
-        id: PropTypes.string.isRequired,
-        labelText: PropTypes.string.isRequired,
-        icon: PropTypes.node,
-        iconDescription: PropTypes.string,
-      })
+      PropTypes.oneOfType([
```
bjornalm, line 142: Are these two shapes exact copies of any other props in the table? If so perhaps we can defined them in a single location

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -284,17 +310,49 @@ const TableToolbar = ({
...
-            disabled={!shouldShowBatchActions}
+            disabled={!shouldShowBatchActions || disabled}
             {...others}
           >
             {labelText}
           </TableBatchAction>
         ))}
+        {hasVisibleOverflowBatchActions ? (
+          <OverflowMenu
+            data-testid={`${testID || testId}-batch-actions-overflow-menu`}
```
bjornalm, line 326: do we need to add support for the deprecated testID?

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -284,17 +310,49 @@ const TableToolbar = ({
...
+            className={`${iotPrefix}--table-overflow-batch-actions`}
+            flipped={langDir === 'ltr'}
+            direction="bottom"
+            onClick={(e) => e.stopPropagation()}
+            renderIcon={OverflowMenuVertical20}
+            tabIndex={shouldShowBatchActions ? 0 : -1}
```
bjornalm, line 332: Why is this needed?

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -284,17 +310,49 @@ const TableToolbar = ({
...
+                <OverflowMenuItem
+                  data-testid={`${testID || testId}-batch-actions-overflow-menu-item-${id}`}
+                  itemText={renderTableOverflowItemText({
+                    action: { renderIcon, labelText },
+                    className: `${iotPrefix}--table-toolbar-aggregations__overflow-menu-content`,
+                  })}
+                  disabled={!shouldShowBatchActions || disabled}
+                  onClick={() => onApplyBatchAction(id)}
+                  key={`table-batch-actions-overflow-menu-${id}`}
+                  requireTitle={!renderIcon}
```
bjornalm, line 347: What does this actually do? The items with `renderIcon` still gets a hover title...

bjornalm (commented):
Nice work, just a couple of questions

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
kevinsperrine: To prevent the button from being tabbed to when the batch action bar is hidden.

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
kevinsperrine: We probably should until v3, right? Then remove it with all the rest?

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
kevinsperrine: It adds the `title` to items that _don't_ have an icon. It's for blocks of text that might be long and get truncated.

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
kevinsperrine: not exact, but I did make some re-usable objects to clean it up a bit.

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
bjornalm: I guess either way is fine, my thought was that we would not be breaking any functionality by only allowing testId for new features... if anything it would encourage developers to move away from the deprecated one.

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
bjornalm: Ok, so we have to "require" the title unless there is a renderIcon in which case there will always be a title added?

kevinsperrine: Yes, the renderIcon will always add a title. The underlying OverflowMenuItem will not unless `requireTitle` is true. Then, it'll use either `itemText || title` as the title. Instead of measuring the text for truncation we just set `requireTitle={!renderIcon}` to always force carbon to add the title for items that are text only.

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
kevinsperrine: It's not completely removed from the DOM. It's just hidden. That's why we had to add this fix a few months ago, because people were able to tab to these hidden elements.

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
kevinsperrine: We could, but if they're still using `testID` and we won't add it here it'll be `undefined-batch-action...`, right?

bjornalm: hehe, now it is getting complicated. It would actually use the default value for the `testId` which an empty string so we would get `-batch-action...` But the Table as only consumer of  the TableToolbar never sets a `testID` is only sets the `testId` so we could just skip the `testID` altogether ... **but** no wait, I see that we are still exporting the TableToolbar for some reason. Fun stuff :-)

kevinsperrine: Ha! That's ridiculous. I'll remove `testID` and see what we end up with in the snapshots. Re: exporting, I _think_ we removed that in v3, just in case someone was using it for some reason.

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
bjornalm: Ok, this is strange, the batch action bar is really completely removed from the DOM in this story https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table--with-selection-and-batch-actions (which is where I looked first) but in the playground story it is just hidden...

bjornalm (approved):
Nice work! Thanks explaining the small edge cases

packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
kevinsperrine: 🤔 Hmmm. It looks like that story is causing a full re-render, notice how the bar slides up smoothly from the top of the table header in the playground as items are selected or unselected, but just abruptly disappears in the selection and batch action story.

davidicus (approved):

PR #3343, opened 2022-02-24
chore(repo): try esbuild storybook plugin to speed up builds

### LINKED ISSUE #3344: [Repo] Speed up netlify builds
Netlify builds continue to timeout or run out of memory. We need to find a solution.

kevinsperrine:
Closes #3344 

**Summary**

- use storybook-addon-turbo-builds to optimize the netlify build process. This uses esbuild instead of terser for minification and turns off sourcemaps for production builds.

- I also modified the netlify configs slightly. I added `--ignore-scripts` to the `YARN_FLAGS` env variable. This prevents the post-install scripts that generate the `styles` package from running, since we don't need to generate that every time for React builds. I also changed the actual build command to only trigger the react build with `NODE_ENV=production yarn --cwd packages/react build:storybook`

- install storybook-addons-turbo-builds
- remove sourcesmaps for production builds in webpackFinal

- confirm the stories still function properly

- confirm the stories still function properly

### DIFF
--- a/packages/react/.storybook/main.js
+++ b/packages/react/.storybook/main.js
@@ -10,6 +10,7 @@ module.exports = {
     '@storybook/addon-a11y',
     'storybook-addon-rtl',
     '@storybook/addon-docs',
+    'storybook-addon-turbo-build',
   ],
   babel: async (options) => {
     // ensure all plugins are using loose: false (the default)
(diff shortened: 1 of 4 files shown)


packages/react/.storybook/main.js
```diff
@@ -29,7 +30,7 @@ module.exports = {
 
     // High quality 'original source' sourcemaps are slow to generate on initial builds and rebuilds.
     // Using cheap-module-eval-source-map speeds up builds and rebuilds in development while not sacrificing too much source map quality.
-    config.devtool = configType === 'DEVELOPMENT' ? 'cheap-module-eval-source-map' : 'source-map';
+    config.devtool = configType === 'DEVELOPMENT' ? 'cheap-module-eval-source-map' : '';
```
kevinsperrine, line 33: remove the sourcemap here for production builds, so we're no un-doing the work of the turbo-build addon.

davidicus (approved):

bjornalm (approved):
fantastic! Thank you!

PR #3347, opened 2022-02-25
feat(progress-bar): add inline variant

### LINKED ISSUE #3339: [ProgressBar] Inline variant
This is an extension to the recent [progress bar](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-progressbar--default) component that was built.

We're defining the two variations as stacked and inline, where the default (one that's already built) is the stacked, and then there is an inline variation where there is no label, and the percentage value is to the right of the bar. 

These are the [style specs](https://pages.github.ibm.com/ai-applications/design/components/progress-bar/style) defined on the PAL site. 
See the [usage guidance](https://pages.github.ibm.com/ai-applications/design/components/progress-bar/usage) tab as well for more details. 

progress-bar-specs

| Element            | Property             | Token                            |
| ------------------ | -------------------- | -------------------------------- |
| Label              | text style           | `$productive-heading-01`         |
|                    | text color           | `$text-01`                       |
| Percentage value   | text style           | `$body-short-01`, right-aligned (stacked), left-aligned (inline) |
|                    | text color           | `$text-01`                       |
| Remaining progress | color: default       | `$ui-02`                         |
|                    | color: light variant | `$ui-01`                         |
| Completed progress | color: default       | Blue 60                          |
|                    | color: multiple      | Blue 80                          |
|                    | color: caution       | `$support-03`, Yellow 60 outline |
|                    | color: warning       | `$support-01`                    |

kevinsperrine:
Closes #3339 

**Summary**

- adds `inline` prop to ProgressBar to hide label/icon/helperText and show the value and unit inlined with the bar

- add `inline` prop
- update docs
- add test
- update snapshots
- cleanup code

- go to the [ProgressBar story](https://deploy-preview-3347--carbon-addons-iot-react.netlify.app?path=/story/1-watson-iot-progressbar--default)
- turn on the inline prop
- confirm label and helperText disappear and value is inlined beside the bar with 8px of space in between
- repeat with icons/thresholds
- check RTL

- tests here

### DIFF
--- a/packages/react/src/components/ProgressBar/ProgressBar.jsx
+++ b/packages/react/src/components/ProgressBar/ProgressBar.jsx
@@ -23,6 +23,8 @@ const propTypes = {
   helperText: PropTypes.string,
   /* if true, hide the label  */
   hideLabel: PropTypes.bool,
+  /** When true, show the inline variant without an icon or label */
+  inline: PropTypes.bool,
   /* the number used to fill the bar */
   value: PropTypes.number.isRequired,
   /* the unit shown beside the value label text */
@@ -64,6 +66,7 @@ const defaultProps = {
   className: undefined,
   helperText: '',
   hideLabel: false,
+  inline: false,
   light: false,
   thresholds: [],
   max: 100,
@@ -82,6 +85,7 @@ const ProgressBar = ({
   helperText,
   hideLabel,
   i18n,
+  inline,
   label,
   light,
   max,
@@ -99,6 +103,63 @@ const ProgressBar = ({
   const fillColor = hasColorObject ? matchingThreshold.color.fill : matchingFillColor;
   const strokeColor = hasColorObject ? matchingThreshold.color.stroke : undefined;
 
+  /**
+   * When the progress bar is not using the inline variant, render the label and icon
+   * above the progress bar.
+   *
+   * @returns JSX Element
+   */
+  const renderProgressBarLabel = () => (
+    <div
+      className={classnames(`${iotPrefix}--progress-bar__label--right`, {
+        [`${prefix}--visually-hidden`]: hideLabel,
+      })}
+    >
+      {Icon ? (
+        <span className={`${iotPrefix}--progress-bar__icon`} data-testid="progress-bar-icon">
+          {renderIconByName && typeof Icon === 'string' ? (
+            renderIconByName(Icon, {
+              fill: fillColor,
+              stroke: strokeColor,
+              'aria-label': mergedI18n.iconLabel,
+            })
+          ) : (
+            <Icon fill={fillColor} stroke={strokeColor} aria-label={mergedI18n.iconLabel} />
+          )}
+        </span>
+      ) : null}
+      <span
+        className={classnames(`${iotPrefix}--progress-bar__value-label`, {
+          // allow styling the value label differently when above max
+          [`${iotPrefix}--progress-bar__value-label--over`]: value > max,
+        })}
+      >{`${value}${valueUnit}`}</span>
+    </div>
+  );
+
+  const renderProgressBarWithInlineWrapper = () => (
+    <div
+      className={classnames({
+        [`${iotPrefix}--progress-bar-wrapper--inline`]: inline,
+      })}
+    >
+      <CarbonProgressBar
+        label={label}
+        helperText={!inline ? helperText : undefined}
+        hideLabel={hideLabel || inline}
+        value={value}
+        max={max}
+      />
+      <span
+        className={classnames(`${iotPrefix}--progress-bar__value-label`, {
+          // allow styling the value label differently when above max
+          [`${iotPrefix}--progress-bar__value-label--over`]: value > max,
+          [`${iotPrefix}--progress-bar__value-label--inline`]: inline,
+        })}
+      >{`${value}${valueUnit}`}</span>
+    </div>
+  );
+
   return (
     <div
       className={classnames(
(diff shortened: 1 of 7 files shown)


bjornalm:
This looks good, just some edge case value scenarios we might wanna consider.

Not everything below is related to this PR  but I did some monkey testing with insane values and got this moving bar and a null value
image
when using large enough numbers

The bar can completely disappear if the value or unit is large enough. Also managed to get a line break if the the unit has a space
image

And the text can overflow
image

Overflowing for non inline version
image

packages/react/src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -99,6 +103,72 @@ const ProgressBar = ({
...
+   */
+  const renderProgressBarWithInlineWrapper = () => (
+    <div
+      className={classnames({
+        [`${iotPrefix}--progress-bar-wrapper--inline`]: inline,
+      })}
+    >
+      <CarbonProgressBar
+        label={label}
+        helperText={!inline ? helperText : undefined}
```
bjornalm, line 163: We know this is inline right?

packages/react/src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -99,6 +103,72 @@ const ProgressBar = ({
...
+        hideLabel={hideLabel || inline}
```
bjornalm, line 164: Same here, we can just always hide the label

packages/react/src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -114,38 +184,20 @@ const ProgressBar = ({
...
-        max={max}
-      />
+      {inline ? (
+        renderProgressBarWithInlineWrapper()
+      ) : (
+        <>
+          {renderProgressBarLabel()}
+          <CarbonProgressBar
+            label={label}
+            helperText={!inline ? helperText : undefined}
```
bjornalm, line 194: Here we know that it is not inline

packages/react/src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -114,38 +184,20 @@ const ProgressBar = ({
...
+            hideLabel={hideLabel || inline}
```
bjornalm, line 195: and here we can just use `hideLabel={hideLabel}`

packages/react/src/components/ProgressBar/ProgressBar.mdx
```diff
@@ -92,6 +93,8 @@ Thresholds can be given to change the color of the bar (and icon) at various val
 | i18n                    | shape                               |                     |                                                                                                                             |
 | i18n.iconLabel          | string                              | 'Progress bar icon' | aria-label used on the icon                                                                                                 |
 
+- Denotes required props
```
bjornalm, line 96:
```suggestion
\*) Denotes required props
```

packages/react/src/components/ProgressBar/ProgressBar.jsx
kevinsperrine: Fixed.

packages/react/src/components/ProgressBar/ProgressBar.jsx
kevinsperrine: Fixed.

packages/react/src/components/ProgressBar/ProgressBar.jsx
kevinsperrine: Fixed.

packages/react/src/components/ProgressBar/ProgressBar.mdx
kevinsperrine: Fixed.

kevinsperrine:
> This looks good, just some edge case value scenarios we might wanna consider.
> 
> Not everything below is related to this PR but I did some monkey testing with insane values and got this moving bar and a null value image when using large enough numbers
> 
> The bar can completely disappear if the value or unit is large enough. Also managed to get a line break if the the unit has a space image
> 
> And the text can overflow image
> 
> Overflowing for non inline version image

This should all be handled now. The moving bar is standard carbon behavior if no value is given. I hide the value/unit if it's not a number, and both the unit and the label should truncate when they are very long.

bjornalm:
Yes, it is working. But it raises the question, how long is very long in this context? It now truncates if I wanna show 450/500.
image

And if I would like to show value "1000" and unit "MiB" it also truncates, i.e. anything over 3 digit unit and 4 digit values
image

bjornalm (approved):
Looks good. Just the question regarding when to truncate... I guess that is a question for @dianatran18

kevinsperrine:
> Looks good. Just the question regarding when to truncate... I guess that is a question for @dianatran18

Yeah, I don't know. We're pretty limited, because we're wrapping carbon, and have to use absolute positioned elements. It would be better if we didn't have to set fixed limits instead of them shifting dynamically.

dianatran18:
I might have to look into this further, but generally the text on the right of the bar should only be a percentage value, which is a max of 3 numbers and the `%`. I noted that the progress bar width would be dynamic based on the use case, but are we limited to a certain width?

Also for the label on the stacked variant, I would have the text truncate when it hits 16px from the percentage value or status icon (if there is one). These are the specs for that. You can see the bounding box for the label and where it would stop in both use cases.

We also have usage guidance stating that "The label must be concise, not exceeding the length of the progress bar."


image

kevinsperrine:
Thanks, @dianatran18 and @bjornalm. I've adjusted the widths and padding so that both the default and inline versions can show `400/500` without truncating, but more than that will truncate. I've also adjusted the padding on the normal version to always have 16px between either the icon (or the text if no icon is given) and that's the width at which the truncation of the label will start. Seems the best middle ground.

packages/react/src/components/ProgressBar/ProgressBar.jsx
```diff
@@ -98,13 +102,74 @@ const ProgressBar = ({
...
   const fillColor = hasColorObject ? matchingThreshold.color.fill : matchingFillColor;
   const strokeColor = hasColorObject ? matchingThreshold.color.stroke : undefined;
+  const valueAndUnit = Number.isInteger(value) ? `${value}${valueUnit}` : '';
+
+  /**
+   * helper to render the value and unit
+   *
+   * @returns JSXElement
+   */
+  const renderValueAndUnit = () => (
```
davidicus, line 112: Why do we call these as functions rather than use the JSX?

kevinsperrine: Just habit to create a render function. No reason it can't be jsx in this case. Updated and pushed.

davidicus (approved):

PR #3349, opened 2022-02-26
fix(table): add align to tooltips

### LINKED ISSUE #3299: [table] column tooltip adds to table-content width
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
When the last column has a tooltip the tooltip will cause the table-content to expand to the right even when the
tooltip is not showing.

### To Reproduce

2. Click on tab "Column configuration"
3. Enable "Demo column tooltips"

### Expected behavior

The tooltip should not add to the with of the container. 

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context
image

image

upcoming release?

date, the better!_

jessieyan:
Closes #3299 

**Summary**

- set align on tooltips
- if last cell, set the align to the opposite 

- [TableCellRenderer.jsx] added new prop `align`
- [TableHead] added conditions to set `align` on  `TableCellRenderer` component

Left to right direction
image

Right to left direction
image

- tests here


bjornalm:
Not sure if this is just a case where the Story and the RTL knob aren't playing nice together but the tooltip of the "last" column in RTL still goes to the right

1. Go to https://deploy-preview-3349--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--playground
2. enable RTL
3. Enable "Demo column tooltips" 
4. The Tooltip of "String" still goes to the right

packages/react/src/components/Table/TableHead/TableHead.jsx
```diff
@@ -544,6 +553,7 @@ const TableHead = ({
                 truncateCellText={truncateCellText}
                 allowTooltip={false}
                 tooltip={matchingColumnMeta.tooltip}
+                align={isLast ? lastTooltipAlign : align}
```
bjornalm, line 556: I think we can use something like 
`tooltipDirection={flipTooltipDirection ? 'start' : undefined}`

bjornalm (changes requested):

packages/react/src/components/Table/TableHead/TableHead.jsx
jessieyan: > I think we can use something like `tooltipDirection={flipTooltipDirection ? 'start' : undefined}`

Hello Bjorn,

One question here. Do we allow customizing the tooltip alignment? Previously, it is always set as default which is `start` I  think (we did not pass in the align prop). If so, then I'll only focus on the `rightmostColumn`. Otherwise, we probably need flip the `leftmostColumn`  tooltip if `align` is set to `end`.

bjornalm: For the tooltip in the Table column headers there is no customisation, just the default.

jessieyan:
> 1. https://deploy-preview-3349--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--playground

Hello Bjorn,

I can see it goes to the left. Any other specifications?  I checked demo tooltips and enabled RTL. 

image

bjornalm (changes requested):
nice work, lets add a unit test for this and we should be good to go

jessieyan:
@bjornalm Added test for tooltips. I was not able to find a good way test directions `ltr` or `rtl`. I can set styles in the dom but not in the browser. `direction` is aways null since it's not in the browser. Do you have any suggestions on this?

bjornalm:
> @bjornalm Added test for tooltips. I was not able to find a good way test directions `ltr` or `rtl`. I can set styles in the dom but not in the browser. `direction` is aways null since it's not in the browser. Do you have any suggestions on this?

I think the `FlyoutMenu.test.e2e.jsx` has some rtl testing. It is browser testing but we can use that.

jessieyan:
> FlyoutMenu

Thanks! I finally found a way to mock up the window object. Now the unit tests work for both 'ltr' and 'rtl'.

bjornalm (approved):
Nice work

PR #3358, opened 2022-03-01
feat(ai-flyout-menu): track open/close state

### LINKED ISSUE #3334: [ai-flyout-menu] track open/close state of the filter-flyout menu
### What package is this for?

- [ ] React
- [x] Angular

### Summary

In the `ai-flyout-menu`, currently there's no way to react to the flyout's open/close state. Provide an @Output event to know the same?

**Additional context** Need this to close the pane when we use a custom header instead of the in-built header.

I'm working on utilising this in `orderlines-table` component for the CC team.

indieNik:
Closes #3334 

**Summary**

- Provide a way to track the filter menu's open/close state by emitting events from underlying elements.

- Added `flyoutOpenChange` event to track the filter's active state.
- You can also get the filter's current state by calling `getFlyoutState` public method on the component.

- Clicking on the filter trigger should output the filter's state in the console

- The changes shouldn't really break anything. Let me know if it does, though.

### DIFF
--- a/packages/angular/src/flyout-menu/flyout-menu.directive.ts
+++ b/packages/angular/src/flyout-menu/flyout-menu.directive.ts
@@ -5,6 +5,8 @@ import {
   ElementRef,
   ViewContainerRef,
   HostBinding,
+  Output,
+  EventEmitter,
 } from '@angular/core';
 import { EventService } from 'carbon-components-angular/utils';
 import { TooltipDirective, DialogService } from 'carbon-components-angular';
@@ -27,6 +29,10 @@ export class FlyoutMenuDirective extends TooltipDirective {
    * Controls wether the overflow menu is flipped
    */
   @Input() flip = false;
+  /**
+   * Emit the open/close state of the menu
+   */
+  @Output() flyoutStateChange = new EventEmitter;
   @HostBinding('class.iot--flyout-menu') menuClass = true;
   /**
    * bx--tooltip__trigger is inherited from TooltipDirective and it enables focus indication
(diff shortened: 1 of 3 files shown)


zvonimirfras (changes requested):
Thanks for doing this, Nik! Good first run.

Flyout menu directive is already a dialog directive that has `isOpenChange` event on it, which you noticed and used. 👍 
Since there's already an existing event, we don't need to define a new one, we can just reuse the one already there (yay for inheritance).

Using a `getFlyoutState` the way you planned and did in storybook isn't quite the pattern you'd wanna do in angular.
What would work much better is having an `isOpen` input (which will enable double binding with `isOpenChange` out of the box).
We can pass the value of component's `isOpen` directly to the directive's `isOpen`.

Feel free to ping me here or on slack if you need any additional clarification.

packages/angular/src/flyout-menu/flyout-menu.directive.ts
```diff
@@ -5,6 +5,8 @@ import {
   ElementRef,
   ViewContainerRef,
   HostBinding,
+  Output,
+  EventEmitter,
```
zvonimirfras, line 9:
```suggestion
```

zvonimirfras:
You should also run `yarn format` and `yarn lint --fix` before pushing.
You can check for any errors or warnings with `yarn format:check && yarn lint`.

zvonimirfras (approved):
Awesome!

indieNik:
Hey @davidicus,
Could you please let me know what the next steps would look like? Are we waiting for something to be added/changed to the PR? New here, so just curious about the process and if anything needs to be done from my side.

Thanks!

Donisius (approved):
Sorry for the delay, thanks for doing this! LGTM 👍

PR #3372, opened 2022-03-04
fix(resourcelist): convert style component to scss

### LINKED ISSUE #3312: [ResourceList] Convert styled-components to scss

Ritu-Thakur:
Closes #3312 

**Summary**

convert style component to scss fro Resourcelist

- added new scss file
- removed 3 style components from Resourcelist

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/ResourceList/ResourceList.jsx
+++ b/packages/react/src/components/ResourceList/ResourceList.jsx
@@ -76,7 +76,7 @@ class ResourceList extends Component {
         </svg>
       </div>
     );
-    const CustomActionContent = (rowId) => (
+    const CustomActionContent = ({ rowId }) => (
       <div
         className={`${prefix}--structured-list-td ${prefix}--structured-list-content--nowrap`}
         role="presentation"
--- a/packages/react/src/components/ResourceList/ResourceList.jsx
+++ b/packages/react/src/components/ResourceList/ResourceList.jsx
@@ -1,52 +1,12 @@
 import React, { Fragment, Component } from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
-import { spacing02 } from '@carbon/layout';
 
 import { settings } from '../../constants/Settings';
 import Button from '../Button';
-import { COLORS } from '../../styles/styles';
 import { handleSpecificKeyDown } from '../../utils/componentUtilityFunctions';
 import { keyboardKeys } from '../../constants/KeyCodeConstants';
 
-const { prefix } = settings;
-
-const ResourceListSection = styled.section`
-   {
-    &.${prefix}--structured-list--selection
-      .${prefix}--structured-list-row:hover:not(.${prefix}--structured-list-row--header-row)
-      > .${prefix}--structured-list-td {
-      border: none;
-    }
-
-    .${prefix}--btn {
-      display: flex;
-      align-items: flex-end;
-      margin: auto 0 auto auto;
-
-      svg {
-        fill: ${COLORS.blue60};
-        margin-left: ${spacing02};
-      }
-    }
-  }
-`;
-
-const InlineDiv = styled.div`
-   {
-    font-weight: normal;
-    padding-left: 0px !important;
-    padding-top: ${spacing02} !important;
-  }
-`;
-
-const CustomActionDiv = styled.div`
-   {
-    vertical-align: middle;
-    color: #3e70b2;
-    fill: #3e70b2;
-  }
-`;
+const { prefix, iotPrefix } = settings;
 
 class ResourceList extends Component {
   static propTypes = {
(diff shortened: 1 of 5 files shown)


packages/react/src/components/ResourceList/ResourceList.jsx
```diff
@@ -158,7 +118,9 @@ class ResourceList extends Component {
             <Fragment>
               <div className={`${prefix}--structured-list-td`}>
                 <strong>{title}</strong>
-                <InlineDiv className={`${prefix}--structured-list-td`}>{description}</InlineDiv>
+                <div className={`${iotPrefix}--inline-div ${prefix}--structured-list-td`}>
```
bjornalm, line 121: We might wanna name it
`${iotPrefix}--resource-list__inline-div`
where `resource-list` is the iot specific Block and `inline-div` is the element

packages/react/src/components/ResourceList/ResourceList.jsx
```diff
@@ -192,16 +154,17 @@ class ResourceList extends Component {
     });
 
     return (
-      <ResourceListSection
+      <section
         data-testid={testId}
         className={[
+          `${iotPrefix}--structured-list`,
```
bjornalm, line 160: Perhaps call this `${iotPrefix}--resource-list` ?

packages/react/src/components/ResourceList/__snapshots__/ResourceList.story.storyshot
```diff
@@ -198,14 +198,14 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/R
...
-          className="bx--structured-list-td bx--structured-list-content--nowrap ResourceList__CustomActionDiv-sc-27rnbq-2 gHaTLR"
+          className="bx--structured-list-td bx--structured-list-content--nowrap"
           role="presentation"
         >
           <button
             aria-describedby={null}
             aria-pressed={null}
             className="iot--btn bx--btn bx--btn--ghost"
-            data-testid="resource-list-custom-action-row-0"
+            data-testid="resource-list-custom-action-[object Object]"
```
bjornalm, line 208: the  [object Object] seems wrong

packages/react/src/styles.scss
```diff
@@ -149,7 +149,7 @@ $deprecations--message: 'Deprecated code was found, this code will be removed be
 @import 'components/PaginationNav/pagination-nav';
 @import 'components/UIShell/ui-shell';
 @import 'components/List/carbon-list';
-@import 'components/Table/TableBody/TableBodyLoadMoreRow/table-body-load-more-row';
+@import 'components/ResourceList/resourcelist';
```
bjornalm, line 152: Why is the `components/Table/TableBody/TableBodyLoadMoreRow/table-body-load-more-row` removed`?

bjornalm (changes requested):

bjornalm (commented):
Nice work, just a couple small changes needed

packages/react/src/styles.scss
Ritu-Thakur: Fixed it.

packages/react/src/components/ResourceList/__snapshots__/ResourceList.story.storyshot
Ritu-Thakur: It's done automatically. DO you know how to fix it?

packages/react/src/components/ResourceList/ResourceList.jsx
Ritu-Thakur: Fixed

packages/react/src/components/ResourceList/ResourceList.jsx
Ritu-Thakur: Fixed

Ritu-Thakur:
@bjornalm Needed your input on one of the above comments.

packages/react/src/components/ResourceList/__snapshots__/ResourceList.story.storyshot
bjornalm: The snapshot is updated automatically but the problem comes from the fact that at some point we are adding an object instead of the row-id to the string that makes up the data-testid. This might happen on a higher level where the data-testid is a prop called `testId` or `testID`...

Ritu-Thakur: Fixed

Ritu-Thakur:
@bjornalm Fixed the object issue too.

kevinsperrine (approved):
Thanks, @Ritu-Thakur!

davidicus (approved):

PR #3380, opened 2022-03-15
Carbon upgrade 10.56.0

### LINKED ISSUE #3384: [REPO] Update Carbon to get off Google  CDN
Google fonts have been found to be non-compliant with the GDPR. As a result the Carbon team have  changed the location of the CDN for the Plex fonts used in our package. We need to upgrade our Carbon deps to be on this new version. Release of this change should be merged in by March 24th. The changes are found in [this](https://github.com/carbon-design-system/carbon/pull/11002) PR. 

If need be we can just copy the changes from this PR to bring us into compliance without the upgrade.

kevinsperrine:
Closes #3384 

**Summary**

- No new test failures on any upgrade

- upgraded carbon from 10.53 to 10.56

- check stories. Carbon [Changelog here](https://github.com/carbon-design-system/carbon/releases) 

- general checks on storybook


packages/react/src/utils/__tests__/__snapshots__/publicAPI.test.js.snap
```diff
@@ -32832,19 +32834,6 @@ Map {
       "role": Object {
         "type": "string",
       },
-      "size": Object {
```
davidicus, line 32835: We have an API change?

davidicus (approved):
Seems ok. Just the one questions about the change in API snapshot

packages/react/src/components/OverflowMenu/OverflowMenu.test.jsx
```diff
@@ -13,4 +14,19 @@ describe('OverflowMenu', () => {
 
     expect(screen.getByTestId('overflow_menu')).toBeDefined();
   });
+
+  it('should call the default getMenuOffset from carbon when none given', () => {
```
bjornalm, line 18: Is that really what we are testing here?

bjornalm (commented):
I checked the stories but not compared them all with next. Seems ok, just the question about the new test

packages/react/src/components/OverflowMenu/OverflowMenu.test.jsx
kevinsperrine: Yeah, for some reason after this upgrade our coverage isn't being met on this file--specifically that the carbon getMenuOffset branch wasn't called. I don't know why it changed, but I added this test for coverage to make sure it doesn't blow up when menuOffset is null. 🤷

kevinsperrine:
@bjornalm and @davidicus This one is updated to 10.56 now, which includes the switch from Google to Akaimai for the fonts.

kevinsperrine:
FYI, This fixes it for carbon, but @carbon/charts is still using Google, so those show up in the network inspector, too.

packages/react/src/components/OverflowMenu/OverflowMenu.test.jsx
bjornalm: nit: I was just thinking that we aren't actually testing that the carbon getMenuOffset is being called, we are testing that our component handles menuOffset={null} and the onClick is still called

bjornalm (approved):

PR #3396, opened 2022-03-24
fix(card): accessibility violation in card header tooltip

### LINKED ISSUE #3395: [Card] Accessibility violation for no aria-label in card header tooltip.
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug
Card is used in Maximo Graphite and accessibility violation is thrown for the card header tooltip.
image

### To Reproduce

1. Go to storybook https://next.carbon-addons-iot-react.com//?path=/story/1-watson-iot-card-card--with-ellipsed-title-tooltip-external-tooltip
2. Check the accessibility tab

### Expected behavior

No accessibility violation for card header tooltip.

### Environment/versions:

- OS: [e.g. MacOS, Windows]
- `carbon-addons-iot-react` version: 2.149.0-next.10

kalyanixraut:
#3395

Closes #3395

**Summary**

- Card with a tooltip in the header throws serious accessibility violation for missing aria-label.
image

- Setting the `iconDescription` prop of `CardTitle`, sets the aria-label to the card header tooltip div and hence gets rid of accessibility violation.

- Go to storybook /?path=/story/1-watson-iot-card-card--with-ellipsed-title-tooltip-external-tooltip
- Check the Accessibility tab
- The serious violation 'Ensures every ARIA button, link and menuitem has an accessible name' is no more listed.

- As there is no functionality changes, so the Card storybook and all components referring Card should work as before.


bjornalm (commented):
code change looks good, but do you know where the snapshot deletions come from?

kalyanixraut:
> code change looks good, but do you know where the snapshot deletions come from?

Hi @bjornalm , thanks for reviewing. When I ran `yarn run test:base -u` to resolve snapshot failures those deletions happened automatically. Not sure whats the cause of it.

bjornalm (approved):
I also noticed the snapshots deletion in next. I think they will go away now if you pull in latest next and run a new test:update

kalyanixraut:
> I also noticed the snapshots deletion in next. I think they will go away now if you pull in latest next and run a new test:update

Done, thanks.

packages/react/src/components/Card/Card.jsx
```diff
@@ -516,6 +516,7 @@ const Card = (props) => {
                 triggerClassName={`${iotPrefix}--card--header--tooltip`}
                 id={`card-tooltip-${id}`} // https://github.com/carbon-design-system/carbon/pull/6744
                 triggerText=""
+                iconDescription={`card-header-tooltip-${id}`} // To fix accessibility violation.
```
davidicus, line 519: While we are doing this we may want to add the ability to provide a translatable string from the user. So we check if there is an `i18n.iconDescription`. If so we use that and if not we can use this as a fallback.

packages/react/src/components/Card/Card.jsx
```diff
@@ -516,7 +516,7 @@ const Card = (props) => {
                 triggerClassName={`${iotPrefix}--card--header--tooltip`}
                 id={`card-tooltip-${id}`} // https://github.com/carbon-design-system/carbon/pull/6744
                 triggerText=""
-                iconDescription={`card-header-tooltip-${id}`} // To fix accessibility violation.
+                iconDescription={strings.iconDescription || `card-header-tooltip-${id}`} // To fix accessibility violation.
```
bjornalm, line 519: We should add this prop to the packages/react/src/constants/CardPropTypes.js and perhaps call it titleTooltipIconDescription and also give that prop a default value like "Tooltip info icon".

bjornalm: And since we now merged https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3402 we need to move these changes to the new CardTitle component and also include this new iconDescription prop

davidicus (approved):

PR #3411, opened 2022-03-30
feat: adjust padding when using filters in table

### LINKED ISSUE #3231: [Table] Filterable column header padding adjustment
### What package is this for?

- [x] React
- [ ] Angular

### Summary

The padding around the filterable column headers has been adjusted to try and give the most space possible to the text field, while at the same time remaining visually separated from column-to-column. This has also been adjusted with better consideration to filterable right-aligned columns and selectable + filterable columns.

#### Proposed styling
Screen Shot 2022-01-26 at 10 44 41 AM

#### Red lines & interaction states
image

image

kevinsperrine:
Closes #3231 

**Summary**

This isn't a full 100% fix. The right-align items specifically were left out, because we don't currently change the alignment of the headers--only the column content when aligning right or center. Otherwise, the other items were addressed, I believe.

- it adds padding to the edges and bottom of the filters
- adjusts the height of the header based on table row size
- adjusts padding to align column title with edge of content/placeholder in filters
- added borders between filter sections when using column groupings
- updated snapshots

- confirm these changes match the styles in the original issue--minus the right-alignment specs

- check RTL/LTR with various knobs and demos in the playground to make sure edge cases weren't missed


bjornalm:
image

The sort button in RTL

bjornalm (changes requested):
This is a bit difficult to review since many details of our columns are still are different from the design specs attached to the issue. But looking at the change list of the PR I do find some new issues:

The last column's filter cell has a white border to right right
image

and in RTL it looks really strange:
image

For the XL size column group text is missing top padding/margin
image

The text of the input is now aligned with thet title of the column, which I think looks great, but it is not according to the specs I think

packages/react/src/components/Table/TableHead/FilterHeaderRow/_filter-header-row.scss
```diff
@@ -15,9 +16,56 @@
...
 }
 
+.#{$prefix}--data-table
+  th.#{$iot-prefix}--filter-header-row--header.#{$iot-prefix}--tableheader-filter {
+  padding-right: $spacing-02;
+  padding-left: $spacing-02;
+  padding-bottom: $spacing-03;
+
+  &.#{$iot-prefix}--tableheader-filter:not(:first-of-type) {
+    border-left-width: 2px;
```
davidicus, line 26: Should these be abstracted into a variable since we use it in a few places?

davidicus (approved):
LGTM 👍 Just one comment about possible variable but not a reason to hold up.

packages/react/src/components/Table/TableHead/FilterHeaderRow/_filter-header-row.scss
kevinsperrine: added a variable.

PR #3436, opened 2022-05-13
fix(side-panel): update icon imports for Angular side panel

yinglejia:
Closes #

**Summary**

We are using the Angular `SidePanel` in our Carbon Angular app. It works fine in the product (ESM), however, when we run the unit test code with jest and jest-preset-angular (CommonJS) on Travis CI, it always times out while loading the `SidePanel` component.

After some debugging, it seems the problem is related to loading icons from `@carbon/icons`. `SidePanel` imports the icons from the `@carbon/icons` package. For CommonJS build, `@carbon/icons/lib/index.js` contains hundreds of `require` calls to all icons. This is fast on a SSD drive, however, it is a lot slower on Travis CI.

This PR changes the imports to load individual icons instead. This is in line with the recommendation here https://github.com/carbon-design-system/carbon/tree/main/packages/icons, and also Carbon Angular implementation here https://github.com/IBM/carbon-components-angular/blob/master/src/icon/icon.module.ts#L9.

- items_here

- tests here

- tests here


yinglejia:
I've found a fix for unit test, so this proposed change is no longer required.

PR #3454, opened 2022-05-27
Feat on editdata item

EdmondCShaffer:
Closes #3911

**Summary**
Reworked the on click handler and add new prop onEditDataItem to support v2 api call to get downsample methods 

- card editor 

- on edit should look for downsample methods when V2

- tests here


davidicus (approved):

packages/react/src/components/HotspotEditorModal/HotspotEditorDataSourceTab/HotspotEditorDataSourceTab.test.jsx
```diff
@@ -146,19 +163,44 @@ describe('HotspotEditorDataSourceTab', () => {
...
+        hotspot={cardConfigWithPresets.content.hotspots[0]}
+        cardConfig={cardConfigWithPresets}
+        dataItems={dataItemsV2}
+        onChange={onChange}
+        translateWithId={() => {}}
+        onEditDataItem={onEditDataItem}
+      />
+    );
+    userEvent.click(screen.getAllByRole('button')[2]);
+    await waitFor(() => expect(screen.getByRole('dialog')).toBeInTheDocument());
```
davidicus, line 202: Isn't this the same assertion that we have on 204?

davidicus (commented):
just the one comment about the superfluous test.

davidicus (changes requested):
just the one comment about test and we should be able to merge

packages/react/src/components/HotspotEditorModal/HotspotEditorDataSourceTab/HotspotEditorDataSourceTab.test.jsx
EdmondCShaffer: Yes I removed 204 and 186

davidicus (approved):

PR #3481, opened 2022-06-21
fix(datetimepicker): added zIndex custom css attribute 

### LINKED ISSUE #3420: DateTime: Z-index solution for Graphite
For the date/time component, ensure that it will function properly with the z-index manager of Graphite, most likely by using a portal for the menu section
https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-date-time-range-selector-datetime-picker--selected-absolute

jessieyan:
Closes #3420

**Summary**

- added style prop and zIndex on the date menu so that it would respect parent's zIndex

- changes in both `<DateTimePicker>` and `<DateTimePickerv2>`
- included zindex knob in the Default Story

- tests here

- tests here


packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -682,6 +686,7 @@ const DateTimePicker = ({
           className={classnames(`${iotPrefix}--date-time-picker__menu`, {
             [`${iotPrefix}--date-time-picker__menu-expanded`]: isExpanded,
           })}
+          style={{ '--zIndex': zIndex }}
```
davidicus, line 689: we may want to spread the rest of the styles here in case people want to add additional styles via this prop.

davidicus (commented):
just the one comment about spreading the other styles and this should be good to go.

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
davidicus, line 689: @jessieyan, additionally if there is not z index we are rendering a style attribute and changing the zIndex custom property to undefined.

jessieyan: @davidicus  I've updated it. Please have another review.

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -682,6 +686,7 @@ const DateTimePicker = ({
           className={classnames(`${iotPrefix}--date-time-picker__menu`, {
             [`${iotPrefix}--date-time-picker__menu-expanded`]: isExpanded,
           })}
+          style={{ ...style, '--zIndex': zIndex ?? undefined }}
```
davidicus, line 689: The zIndex prop will already be undefined if not provided. We should either do a check that creates the style object with or without the zIndex or we should just have the alternative value be 0 which is the default you have set in the sass file. This same comment applies to the other DateTimePicker.

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -682,6 +686,7 @@ const DateTimePicker = ({
           className={classnames(`${iotPrefix}--date-time-picker__menu`, {
             [`${iotPrefix}--date-time-picker__menu-expanded`]: isExpanded,
           })}
+          style={{ ...updatedStyle }}
```
davidicus, line 689: for next time- this spread is unnecessary. Could just pass in the variable you created. 
```suggestion
          style={updatedStyle}
```

davidicus (approved):

PR #3483, opened 2022-06-23
fix(sidenav): support recentLinks

### LINKED ISSUE #3443: [Side-nav] Add support for "My recent applications" and pinned items at the top w/line
Maximo TPAE has long-standing logic where:
1. Apps recently clicked are stored and easily accessed again within the same session
2. Items such as "Start center" are docked at the top w/the search bar
In order to migrate customers from Maximo 7.6 to 8, we need to retain all existing features to reduce adoption inhibitors. Ensure that TPAE and Graphite are visually identical in terms of navigation is part of this effort.

jessieyan:
Closes #3443 

**Summary**

- added `recentLinks` prop to show recent links under pinned items 
- make sure search works in recentlinks
- added recentlinks story with state

- SideNav.jsx: added recentLinks prop
- SideNav.story.jsx: added state for recentLinks
- SideNav.test.jsx: added component tests

- Make sure recent links works well in the story


davidicus:
@jessieyan a few questions. Have these changes been brought into SuiteHeader? We will need to do so if not. The stateful example does not seem to be updating with new applications. Can we make sure this example updates our recent section? 

For the recent links prop I think the user should be passing in an array that has same signature as the links prop. In other words they should not have to pass in a parent item with icon, callback, and string. I guess this is a question for @JordanWSmith15  on this design. Do we want teams to have the flexibility to define their own icons, callbacks for this menu item? Another design question is when you click on the recent link item, what gets highlighted as selected, the recent one they click or the actual item in the menu that recent is a shortcut to?

jessieyan:
> @jessieyan a few questions. Have these changes been brought into SuiteHeader? We will need to do so if not. The stateful example does not seem to be updating with new applications. Can we make sure this example updates our recent section?
> 
> For the recent links prop I think the user should be passing in an array that has same signature as the links prop. In other words they should not have to pass in a parent item with icon, callback, and string. I guess this is a question for @JordanWSmith15 on this design. Do we want teams to have the flexibility to define their own icons, callbacks for this menu item? Another design question is when you click on the recent link item, what gets highlighted as selected, the recent one they click or the actual item in the menu that recent is a shortcut to?



> @jessieyan a few questions. Have these changes been brought into SuiteHeader? We will need to do so if not. The stateful example does not seem to be updating with new applications. Can we make sure this example updates our recent section?
> 
> For the recent links prop I think the user should be passing in an array that has same signature as the links prop. In other words they should not have to pass in a parent item with icon, callback, and string. I guess this is a question for @JordanWSmith15 on this design. Do we want teams to have the flexibility to define their own icons, callbacks for this menu item? Another design question is when you click on the recent link item, what gets highlighted as selected, the recent one they click or the actual item in the menu that recent is a shortcut to?



> @jessieyan a few questions. Have these changes been brought into SuiteHeader? We will need to do so if not. The stateful example does not seem to be updating with new applications. Can we make sure this example updates our recent section?
> 
> For the recent links prop I think the user should be passing in an array that has same signature as the links prop. In other words they should not have to pass in a parent item with icon, callback, and string. I guess this is a question for @JordanWSmith15 on this design. Do we want teams to have the flexibility to define their own icons, callbacks for this menu item? Another design question is when you click on the recent link item, what gets highlighted as selected, the recent one they click or the actual item in the menu that recent is a shortcut to?



> @jessieyan a few questions. Have these changes been brought into SuiteHeader? We will need to do so if not. The stateful example does not seem to be updating with new applications. Can we make sure this example updates our recent section?
> 
> For the recent links prop I think the user should be passing in an array that has same signature as the links prop. In other words they should not have to pass in a parent item with icon, callback, and string. I guess this is a question for @JordanWSmith15 on this design. Do we want teams to have the flexibility to define their own icons, callbacks for this menu item? Another design question is when you click on the recent link item, what gets highlighted as selected, the recent one they click or the actual item in the menu that recent is a shortcut to?

I don't think I've checked with SuiteHeader. Will do it now.

jessieyan:
@davidicus For the story with state, it only works with app links. E.g. we passed `App 1, App2, App3` in `recentLinks`.

Originally, it shows the order `App 1, App2, App3`. If clicked on `App3`, then it is `App3, App 1, App2`.  Are you saying we should have an `Application menu` (with App 1, App2, App3, App4...) in the story. When we click links under `Application`, it shows under `My recent links`?

image

davidicus:
> @davidicus For the story with state, it only works with app links. E.g. we passed `App 1, App2, App3` in `recentLinks`.
> 
> Originally, it shows the order `App 1, App2, App3`. If clicked on `App3`, then it is `App3, App 1, App2`. Are you saying we should have an `Application menu` (with App 1, App2, App3, App4...) in the story. When we click links under `Application`, it shows under `My recent links`?
> 
> image

@jessieyan I am asking for [clarity](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3443#issuecomment-1180460198) from Jordan in the issue. I at first thought these were recent items from the sideNav list. But if they are actually applications then maybe these do not belong in the sidenav but in the sliding drawer on the right that is part of the Header. If so, this change may be just for the `SuiteHeader` component.

JordanWSmith15:
@davidicus @jessieyan Only items in the left-hand side-nav would appear as "Recent applications". It has nothing to do with the Suite navigator, like "Manage" or "Monitor" (I understand those are also sometimes referred to as applications, but in this case, they are not part of "Recent apps").

jessieyan:
> @jessieyan I see you have added an "app" section the sidenav but I believe any link there should be added to recent links. Currently only clicking links in the apps section will change the recent link dropdown.

Updated the story. When click any of the links except for pinned link, it will show under` My recent links`. 

Copy the code from carbon that if child is active, then parent will be active as well. In this case, if I set child (links) active, then parent (`My recent links menu`) would be active as well. So I set all children as not active by default. 
image

davidicus (approved):

Issue #3489, opened 2022-06-30
[Table] empty tables generate conflicting ids


valZho:
### What package is this for?

- [x] React
- [] Angular

### Describe the bug

Empty table's `tbody` tags have the same (conflicting) id as their `table` tag

### To Reproduce

Create an empty table

### Expected behavior

Remove conflicting id

PR #3490, opened 2022-06-30
remove conflicting id from empty tables

valZho:
Closes #

**Summary**

- Added unique id` id={`${id}-empty-table`}`
- Updated snapshots

- items_here

- tests here

- tests here


davidicus:
@valZho Since there is no issue associated with this PR I have to guess at the actual issue. If the issue is that the empty table component id conflicts with the one provide to the Table element then this seems like a bug. I think the fix for this is not to remove the id but append an additional string to it in our use inside of the `Table` found [here](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/85d7b57d9129a9a405d1982d9136294dbf319fe9/packages/react/src/components/Table/Table.jsx#L1060).  

I have opened an [issue](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3492) for this that you can refer to in your PR if this is, in fact, your issue.

davidicus:
> I did, in fact, create an issue: #3489 — I named my commit with the correct issue number, and I thought I referenced it elsewhere, but I guess not. Sorry. 😞
> 
> There was a similar issue previously with non-empty tables which was solved by removing the id from the `<Table>` component. To be consistent with that, I thought it would be easy to just delete a couple of lines and throw up a PR. 🤷
> 
> On the aforementioned component, [line 968](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/85d7b57d9129a9a405d1982d9136294dbf319fe9/packages/react/src/components/Table/Table.jsx#L968), you can see that `id` is being supplied to the main `<CarbonTable>` wrapper... so any child items (including the Error Table that you pointed out) should not also use that same id value.

OK. Great. I will close the  issue I opened as a duplicate. We will want to pass a unique ID from the id  prop with an appended string like `${id}-error}`. That way if there are two tables with an empty table state we will not have similar id clashes as both would have the empty table default id.

valZho:
> OK. Great. I will close the issue I opened as a duplicate. We will want to pass a unique ID from the id prop with an appended string like `${id}-error}`. That way if there are two tables with an empty table state we will not have similar id clashes as both would have the empty table default id.

Just to make sure we are on the same page... the particular issue I was running into was with line 1125, but it looks like both lines 1125 _and_ 1060 will need to be appended to fix these particular ID conflicts.

davidicus:
@jessieyan if you have any time between big issues can you look into this?

davidicus (approved):

davidicus:
@valZho FYI

PR #3514, opened 2022-08-05
feat(suiteheader): Adding multi-workspace support

### LINKED ISSUE #3509: [SuiteHeader] Multi-workspace support
In the suite header/UI shell
1. Re-design the 9-dot nav menu as shown in the design spec
2. Add support for multi-workspace applications in the Navigator (nine dot) menu. This section only appears if the current user has access to applications in multiple work spaces. 

Design:
https://app.mural.co/invitation/mural/aiapplicationsdesign4698/1654283028012?sender=charlottewittels3890&key=86b4373c-f8b5-4e03-b67e-db819a5bd805

blechner:
Closes #3509 

**Summary**

- Adding `workspaces` and `globalApplications` properties to `SuiteHeader` for multi-workspace support while keeping support for legacy `applications` property for backwards compatibility.

- adding support for multi-workspace in SuiteHeader (when passing the new `workspaces` and `globalApplications` properties to the component).
- changing the app switcher background design to a dark themed one.
- including the current workspace name at the main header bar when more than one workspace is available, in workspace-based pages.

- Many tests and stories have been added to cover all possible scenarios: all combinations of pages (admin/workspace-based/non-workspace-based) and number of workspaces (one/many).

- Support for legacy version of `SuiteHeader` was kept intact when `workspaces` properties is not passed to the component. All existing tests for the `SuiteHeader` are passing.


JordanWSmith15:
@blechner @davidicus  Looks good. I can't seem to find the prop or example that has the workspace listed in the header while the suite nav pane is closed: 
Screen Shot 2022-08-16 at 2 40 32 PM
Is this completed? If so, how do I reach it? I am concerned about truncation for long strings.

JordanWSmith15:
Nevermind, I see it is defined here:
Screen Shot 2022-08-16 at 2 42 26 PM
This does not match the design, though I see why it would be easier to support (truncation again). Is the design team (Luke/Charlotte)\ OK with this change?

blechner:
> @blechner @davidicus Nevermind, I see it is defined here: Screen Shot 2022-08-16 at 2 42 26 PM This does not match the design, though I see why it would be easier to support (truncation again). Is the design team (Luke/Charlotte)\ OK with this change?

Yes, it does not match the design, but it wasn't possible to add any custom text in the right hand part of the header, that part is built from json configuration and it only renders action items. However, I've shown this approach of adding the workspace name on the left side to the designer and she was OK with the change.

blechner:
> @blechner @davidicus Looks good. I can't seem to find the prop or example that has the workspace listed in the header while the suite nav pane is closed: Screen Shot 2022-08-16 at 2 40 32 PM Is this completed? If so, how do I reach it? I am concerned about truncation for long strings.

The current workspace name only shows in the header for workspace-based pages (navigator and applications) and only when more than one workspace is available for the current user. Non-workspace based pages (admin page, for example) will never show a workspace name in the header

JordanWSmith15 (approved):

davidicus (approved):
Seems good. There are a lot changes here. It would be much easier if there was any way to break these changes up into smaller PRs. It would definitely allow me to get to it a lot quicker.

herleraja:
Is loading state suppose to be like this ? https://deploy-preview-3514--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-ui-shell-suiteheader-multi-workspace-multiworkspacesuiteheaderappswitcher--loading-state


image

PR #3523, opened 2022-08-23
fix(datetimepickerv2): add single select and replace range select with new timepickerdropdown

### LINKED ISSUE #3503: Update to Time input in Date Relative
• Update time input used in date relative component with David's new one
• Add property that hides options not shown in the below design (such as the relative vs absolute radio) 
• Add property for single date selection
• Add property for single time selection
• Create new side-by-side date+time input:
Screen Shot 2022-07-21 at 11 59 47 AM

jessieyan:
Closes #3503 

**Summary**

- add single select
- add new prop useNewTimeSpinner, if enabled, absolute will use `<TimePickerDropdown> `as the time picker. Single select only works when useNewTimeSpinner is enabled. 

` <DateTimePickerV2WithTimeSpinner>` is the component that use` <TimePickerDropdown>`
`<DateTimePickerV2WithoutTimeSpinner>` is the previous `<DateTimePickerV2> `
`<DateTimePickerV2>` is the wrapper. If `useNewTimeSpinner `enabled,  it renders `<DateTimePickerV2WithTimeSpinner>  `otherwise renders `<DateTimePickerV2WithoutTimeSpinner>`

Single select
image

Range select
image

- tests here


packages/react/src/components/DateTimePicker/DateTimePickerV2.test.e2e.jsx
```diff
@@ -107,24 +136,25 @@ describe('DateTimePickerV2', () => {
...
 
-    cy.findByLabelText(i18n.endTimeLabel).type(
-      '{backspace}{backspace}{backspace}{backspace}{backspace}11:61'
-    );
+    cy.get('#picker-test-end-time').type('11:61 AM');
     cy.findByText(i18n.applyBtnLabel).should('be.disabled');
-
-    // set time to 11:00
-    cy.findByLabelText(i18n.endTimeLabel).type('{backspace}{backspace}00');
-    cy.findByText(i18n.applyBtnLabel).should('not.be.disabled');
```
davidicus, line 127: is there a reason we took this out?

packages/react/src/components/TimePicker/TimePickerDropdown.jsx
```diff
@@ -297,7 +302,7 @@ const TimePickerDropdown = ({
             <TextInput
               onChange={handleOnChange}
               data-testid={`${testId}-input`}
-              id={id}
```
davidicus, line 300: changing id would be considered a breaking change.

packages/react/src/components/TimePicker/TimePickerDropdown.jsx
```diff
@@ -364,7 +369,7 @@ const TimePickerDropdown = ({
               onChange={handleOnChange}
               data-testid={`${testId}-input-1`}
               onFocus={() => setFocusedInput(0)}
-              id={`${id}-1`}
+              id={`${id}-start-time`}
```
davidicus, line 372: same comment about the id

packages/react/src/components/TimePicker/TimePickerDropdown.jsx
```diff
@@ -403,7 +408,7 @@ const TimePickerDropdown = ({
           >
             <TextInput
               data-testid={`${testId}-input-2`}
-              id={`${id}-2`}
+              id={`${id}-end-time`}
```
davidicus, line 411: breaking change

JordanWSmith15 (changes requested):
Tested on Mac Firefox
1. When I focus on the time picker and enter a number, focus is lost
2. I see "Invalid date" fairly often. 
a. It should give a standard invalid warning (red text below the input) if it's invalid, but still show what I selected
b. It appears seemingly for no reason (Hitting Apply)
Screen Shot 2022-08-23 at 3 47 34 PM

packages/react/src/components/TimePicker/_time-picker-dropdown.scss
```diff
@@ -66,14 +66,17 @@
 
     .#{$iot-prefix}--time-picker__wrapper-sm & {
       bottom: 1rem;
+      padding-right: 0rem;
```
davidicus, line 69: nit: zero values should have no units

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.e2e.jsx
jessieyan: > is there a reason we took this out?

This is the behavior in <TimePickerSpinner> that if put 00 in the input, it is not a invalid value. Now we use <TimePickerDropDown> it would cause invalid date error.

davidicus:
I try to open the time picker spinner dropdown and it does not open on [this](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--default) story.

davidicus (changes requested):
Some issues to address.

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
```diff
@@ -825,46 +932,85 @@ describe('DateTimePickerV2', () => {
...
     const timeRange = screen.getByTestId('date-time-picker__field');
 
     expect(timeRange).toHaveTextContent('2020-04-01 12:34 to 2020-04-06 11:49');
     expect(screen.getByText(i18n.applyBtnLabel)).toBeEnabled();
 
     // 2020-04-01 13:34 to 2020-04-06 11:49
-    userEvent.type(startTime, '{backspace}{backspace}{backspace}{backspace}{backspace}13:34');
-    expect(startTime).toHaveValue('13:34');
-    expect(endTime).toHaveValue('11:49');
-    expect(timeRange).toHaveTextContent('2020-04-01 13:34 to 2020-04-06 11:49');
```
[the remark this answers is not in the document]
jessieyan: > I see many of these old test deleted. Should we not keep these old test and add the new ones to ensure we are not breaking past functionality?

userEvent.type() does not work for some reason. I have to use fireEvent.chanage() to change a value. I use click on the spinners instead for the tests.

jessieyan:
> 2\. below



> Tested on Mac Firefox
> 
> 1. When I focus on the time picker and enter a number, focus is lost
> 2. I see "Invalid date" fairly often.
>    a. It should give a standard invalid warning (red text below the input) if it's invalid, but still show what I selected
>    b. It appears seemingly for no reason (Hitting Apply)
>    Screen Shot 2022-08-23 at 3 47 34 PM

For 2, I will update the story to expose the warning text. 
image

packages/react/src/components/TimePicker/TimePickerDropdown.jsx
jessieyan: I will change the ids back.

jessieyan:
> 



> Tested on Mac Firefox
> 
> 1. When I focus on the time picker and enter a number, focus is lost
> 2. I see "Invalid date" fairly often.
>    a. It should give a standard invalid warning (red text below the input) if it's invalid, but still show what I selected
>    b. It appears seemingly for no reason (Hitting Apply)
>    Screen Shot 2022-08-23 at 3 47 34 PM

For 1, it's some logic from the existing <DateTimePickerV2> . Will take another look.

JordanWSmith15:
> > 
> 
> > Tested on Mac Firefox
> > 
> > 1. When I focus on the time picker and enter a number, focus is lost
> > 2. I see "Invalid date" fairly often.
> >    a. It should give a standard invalid warning (red text below the input) if it's invalid, but still show what I selected
> >    b. It appears seemingly for no reason (Hitting Apply)
> >    Screen Shot 2022-08-23 at 3 47 34 PM
> 
> For 1, it's some logic from the existing . Will take another look.

This is what I'd expect for 2A:
Screen Shot 2022-08-23 at 4 09 58 PM

jessieyan:
> I try to open the time picker spinner dropdown and it does not open on [this](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--default) story.

Put zIndex setting is the story. Now it should work.

jessieyan:
> > > 
> > 
> > 
> > > Tested on Mac Firefox
> > > 
> > > 1. When I focus on the time picker and enter a number, focus is lost
> > > 2. I see "Invalid date" fairly often.
> > >    a. It should give a standard invalid warning (red text below the input) if it's invalid, but still show what I selected
> > >    b. It appears seemingly for no reason (Hitting Apply)
> > >    Screen Shot 2022-08-23 at 3 47 34 PM
> > 
> > 
> > For 1, it's some logic from the existing . Will take another look.
> 
> This is what I'd expect for 2A: Screen Shot 2022-08-23 at 4 09 58 PM

1. The focus should be fixed
2. The invalid text should now show under the start/end time input.

jessieyan:
The enter key problem should be fixed.

jessieyan:
I'm fixing the firefox browser issue that Jordan found. I need to format the hh:mm A date without usind dayjs.

jessieyan:
Updated some formats to make sure it works with Firefox and other browsers.

JordanWSmith15 (changes requested):
Two more things and then I think we're close:
1. I need to click the "Apply" primary button twice before the dialog closes
2. Null values don't seem to be demonstrated.
A. What does a null value look like, exactly? I would expect that placeholder grey text would show the browser's locale format:
Screen Shot 2022-08-29 at 8 15 38 AM
B. There is no way to clear the field back to null. I actually don't even see what the field looks like when null. 
Screen Shot 2022-08-29 at 8 03 53 AM

JordanWSmith15:
@jessieyan 
1. Maybe it's specific to Mac? Do you have a PC? 
2. I'm talking specifically about being able to clear the "Single select" version's value. Graphite will not be using the Relative version for this release, so I am very focused on this component: 
https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select

jessieyan:
Added clear button to reset date and time value in single select.

JordanWSmith15 (approved):
🥳 🎉

davidicus:
@jessieyan A few items:

- It looks like you took away the `TimePickerSpinner` option. The new input options were supposed to be put behind a prop. This component should work exactly as it did before unless this prop is provided. 
- Test for the old 24hour format should remain to make sure the old functionality still works. We should write additional test to test this new TimePickerDropdown functionality.
- When you are going from a [relative option with an existing value](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--selected-relative) to the absolute value the existing value is being turned to "invalid date to invalid date". This remains even when you go back to the relative option.  
- When you go to use the absolute value option on any [story](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--default) the input states "invalid date to invalid date" and the inputs are in the error state. This should not occur unless the user inputs an invalid date. 
- I am not seeing a story knob for switching between single and range or to use the old `TImePickerSpinner` vs. the new `TimePickerDropdown`

packages/react/src/components/DateTimePicker/dateTimePickerUtils.js
```diff
@@ -7,6 +7,30 @@ import dayjs from '../../utils/dayjs';
 
 const { iotPrefix } = settings;
 
+const customParseFormat = require('dayjs/plugin/customParseFormat');
```
davidicus, line 10: Should this not be imported

packages/react/src/components/TimePicker/TimePickerDropdown.jsx
```diff
@@ -215,13 +245,8 @@ const TimePickerDropdown = ({
       e.relatedTarget?.parentNode?.classList[0].includes('iot--list-spinner');
 
     if (!contained) {
-      // close dropdown and validate
+      // close dropdown
       setOpenState(false);
-      setInvalidState(!validate(inputRef.current.value));
```
davidicus, line 220: The removal of this code now allows for values other than numbers to be entered into the input.

davidicus: Go to [this](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-timepickerdropdown--default) story and try typing in letters. With your changes the [same story](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-timepickerdropdown--default) now allows for these inputs to go through.

packages/react/src/components/TimePicker/TimePickerDropdown.test.jsx
```diff
@@ -197,7 +197,6 @@ describe('TimePickerDropdown', () => {
     expect(qbt('time-picker-test-spinner')).toBeInTheDocument();
     userEvent.click(document.body);
     await waitFor(() => expect(qbt('time-picker-test-spinner')).not.toBeInTheDocument());
-    expect(screen.queryByText(/The time entered is invalid/)).toBeTruthy();
```
davidicus, line 200: This should not be removed. If all these steps in this test occur, this assertion should be truthy.

davidicus (changes requested):
There are a few items that still need to be addressed

jessieyan:
@EdmondCShaffer @herleraja  Hey guys, we want to the replace `<TimePickerSpinner> `with `< TimePickerDropDown>` in `<DateTimePickerV2>`. We need your feedback if you would like the keep` <TimePickerSpinner>`  or using the new ` < TimePickerDropDown>`.

The input and output of <DateTimePickerV2> will stay the same. Main differences are `<TimePickerDropDown> `uses a list spinner for hours, minutes and AM/PM. Also `<TimePickerDropDown>` uses a 12 hr format for displaying the time. 

Please let us know your preference and concerns.

herleraja:
> DateTimePickerV2

He @jessieyan I think its fine. Only to note that we use the 24-hour format in the dashboard card DateTimePickerV2. I think by default its 24 hours then we don't have much concern here.

Btw what is the difference between DateTimePickerV2 and DateTimePicker?

jessieyan:
@davidicus I renamed old `<DateTimePickerV2>` to `< DateTimePickerV2WithoutTimeSpinner>` and use `<DateTimePickerV2>` to wrap  `< DateTimePickerV2WithoutTimeSpinner>`  and  `< DateTimePickerV2WithTimeSpinner>`. I saw the PublicApi show some red lines. Should I define all props in the wrapper `<DateTimePickerV2> `as well?

jessieyan:
> @davidicus I renamed old `<DateTimePickerV2>` to `< DateTimePickerV2WithoutTimeSpinner>` and use `<DateTimePickerV2>` to wrap `< DateTimePickerV2WithoutTimeSpinner>` and `< DateTimePickerV2WithTimeSpinner>`. I saw the PublicApi show some red lines. Should I define all props in the wrapper `<DateTimePickerV2> `as well?

Fixed the the `propTypes` and `defaultProps` in `<DateTimePickerV2>`.

jessieyan:
> @jessieyan A few items:
> 
> * It looks like you took away the `TimePickerSpinner` option. The new input options were supposed to be put behind a prop. This component should work exactly as it did before unless this prop is provided.
> * Test for the old 24hour format should remain to make sure the old functionality still works. We should write additional test to test this new TimePickerDropdown functionality.
> * When you are going from a [relative option with an existing value](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--selected-relative) to the absolute value the existing value is being turned to "invalid date to invalid date". This remains even when you go back to the relative option.
> * When you go to use the absolute value option on any [story](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--default) the input states "invalid date to invalid date" and the inputs are in the error state. This should not occur unless the user inputs an invalid date.
> * I am not seeing a story knob for switching between single and range or to use the old `TImePickerSpinner` vs. the new `TimePickerDropdown`
> * When I click enter on the TimePickerDropdown component, inside DateTimePicker, I get an error thrown
>   image

@davidicus 
This should be addressed by adding a new prop `useNewTimeSpinner` and the wrapper component <DateTimePickerV2> which determines whether to render <DateTimePickerV2WithTimeSpinner> or <DateTimePickerV2WithoutTimeSpinner>. Added knob of `useNewTimeSpinner` in `Selected absolute` story and add a separate story Range select with new time spinner. 

image

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
```diff
@@ -528,47 +1047,133 @@ describe('DateTimePickerV2', () => {
...
       />
     );
+    jest.runAllTimers();
 
-    expect(screen.getByText('2021-08-01 12:34 to 2021-08-06 10:49')).toBeVisible();
+    userEvent.click(screen.getByTestId('date-time-picker__field'));
+    expect(screen.getByText('minutes')).toBeInTheDocument();
   });
 
-  it('should not render a time picker if hasTimeInput:false', () => {
```
davidicus, line 549: There still seems to be changed test here. These should stay to ensure the old functionality is being preserved, no?

packages/react/src/components/TimePicker/TimePickerDropdown.jsx
```diff
@@ -183,6 +186,12 @@ const TimePickerDropdown = ({
     }
   };
 
+  const handleOnKeyUp = (e) => {
+    if (e.key === 'Enter') {
+      setOpenState(true);
+    }
+  };
+
```
davidicus, line 194: This should not be applied to this component as it breaks the ability for [implicit submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission). The icon is what allows for the dropdown functionality on enter and is separate tab stop than the text input.

packages/react/src/components/TimePicker/_time-picker-dropdown.scss
```diff
@@ -66,14 +66,17 @@
 
     .#{$iot-prefix}--time-picker__wrapper-sm & {
       bottom: 1rem;
+      padding-right: 0;
```
davidicus, line 69: I believe that this element already has no padding applied. So these padding declarations are unnecessary.

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
jessieyan: > There still seems to be changed test here. These should stay to ensure the old functionality is being preserved, no?

The test was moved to line 1087.

packages/react/src/components/TimePicker/TimePickerDropdown.jsx
jessieyan: > This should not be applied to this component as it breaks the ability for [implicit submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission). The icon is what allows for the dropdown functionality on enter and is separate tab stop than the text input.

There is a bug in <DateTimePickerV2>. The keyup event can detect the enter key but the keydown event did not detect.

packages/react/src/components/TimePicker/_time-picker-dropdown.scss
jessieyan: Moved the padding-right out of <TimePickerDropdown>.

davidicus:
In [this](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select) story you can see the tooltip value is showing the mask value instead of the value of the input as it does in our current version. 
Next             |  This PR
  |

davidicus:
This may not something that we need to fix now but in the input when you click a on a date from an empty state you get a time as seen in this image. Should this value be added as the value for the TimePickerDropdown? image

davidicus:
In [this](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--selected-absolute) story when you go to a preset value then chose a relative value then go to the absolute value calendar, there is a discrepancy in what is in the input field and what is shown in the `TImePickerDropdown`
image

jessieyan:
> this



> In [this](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select) story you can see the tooltip value is showing the mask value instead of the value of the input as it does in our current version.
> 
> Next	This PR
> 	

This is the original design I think. The design before is if the whole value could not be displayed, then show it in tooltip. If the whole value could be shown, then show mask value in the tooltip.

davidicus:
@jessieyan there may be something in the code that does not allow for this check to be dynamic but when I make the input smaller I am not getting the value in the tool tip. Can we check if this is the case and what we can do about it. I think this would be the last thing and then we can get this merged in. Thanks for working through this. I know its a complicated component. 
image

jessieyan:
> In [this](https://deploy-preview-3523--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--selected-absolute) story when you go to a preset value then chose a relative value then go to the absolute value calendar, there is a discrepancy in what is in the input field and what is shown in the `TImePickerDropdown` image
> 
> @davidicus Yeah, there is a bug here. This bug is not new though. When change time values in relative, they are not picked up by absolute. I will see if i can fix it. Otherwise, I think we could open an issue to track it.

Hi @davidicus , I opened issue [3547](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3547) to track the issue that values are not updated when switching between Relative and Absolute.

davidicus (approved):

PR #3524, opened 2022-08-24
fix(angular): disable ivy compilation

siddharthcr:
Closes #

**Summary**

- setting `enableIvy` to false to disable ivy compilation.

- In angular 11, ivy enabled libraries were not meant to be published. The build script was removing the `prepublish` script that was added by `ng-packagr` to prevent Ivy enabled libs from being published.

- Create a new angular 11 app and add `@ai-apps/angular` as a dependency. When build or serve targets are run in the angular 11 app, `ngcc` compiler would compile `@ai-apps/angular` to become Ivy compatible.

- Same as above


Akshat55 (changes requested):
Should be approved after integration tests have been updated/added.

siddharthcr:
@Akshat55 copied over the integration tests. Could you please take a look? Thanks

Akshat55 (approved):

zvonimirfras (approved):

Issue #3526, opened 2022-08-26
[Card] add iconDescription to the card component


herleraja:
### What package is this for?

- [x] React
- [ ] Angular

### Summary

When we add custom card toolbars we can not add tooltips

![image](https://user-images.githubusercontent.com/18572477/186921518-eab44233-91a9-4836-803b-e120fbf16bb5.png)

PR #3527, opened 2022-08-26
feat(card): adding tooltip to the card extra actions

herleraja:
Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3526

**Summary**

- adding tooltip to the card extra actions

- feat(card): adding tooltip to the card extra actions

- tooltips can be passed from the props

- card should work with out any issues


davidicus:
It does not explicitly state it in the [spec](https://pages.github.ibm.com/ai-applications/design/components/card/usage) but the dropdown convention is for the 3 dot which is how it is shown through out the spec. @dianatran18 is it cool to allow for different type of dropdown icons for the card?

packages/react/src/components/Card/CardToolbar.jsx
```diff
@@ -192,9 +193,9 @@ const CardToolbar = ({
         </OverflowMenu>
       ) : extraActions.icon ? (
         <ToolbarSVGWrapper
-          title={mergedI18n.extraActionLabel}
+          title={extraActions.iconDescription || mergedI18n.extraActionLabel}
```
davidicus, line 196: Why can't we use the translation string that is already provided? This change will bring down the test coverage and is redundant, no?

packages/react/src/components/Card/CardToolbar.jsx
```diff
@@ -176,7 +176,8 @@ const CardToolbar = ({
         <OverflowMenu
           flipped={overflowMenuPosition}
           title={mergedI18n.overflowMenuDescription}
-          iconDescription={mergedI18n.overflowMenuDescription}
+          iconDescription={extraActions.iconDescription || mergedI18n.extraActionLabel}
```
davidicus, line 179: this causes a double tooltip. one for the button and one for the icon. Can we not use the the overflowMenuDescription translated string?

herleraja: overflowMenuDescription is present in different places and getting used in different place. So I added the extra prop.

packages/react/src/components/Card/CardToolbar.jsx
herleraja: I thought of making making it uniform by providing extraActions. but you are correct.  Let me revert it to extraActionLabel

packages/react/src/components/Card/CardToolbar.jsx
herleraja: Regarding double tooltip: It was there like that from long time. If its not a problem then I can remove that.

packages/react/src/components/Card/CardToolbar.jsx
herleraja: image

The reason behind that is if we have multiple card with multiple tooltip then its always taking the global ones. So to avoid that I have moved it into the `extraActions`

herleraja:
> @dianatran18 is it allowed for different type of dropdown icons for the card ?

packages/react/src/components/Card/CardToolbar.jsx
davidicus: In the old case it was not an issue because the string value was the same for both

packages/react/src/components/Card/CardToolbar.jsx
davidicus: The card is an isolated scope. You can pass in the extra action Icon description for the value of `i18n.extraActionLabel`, no?

herleraja: Unfortunately, the extraActionLabel is at the global scope. That's why we need to handle it in card-level scope.

packages/react/src/components/Card/CardToolbar.jsx
herleraja: I have removed it.  as you said it's kind of redundant.

herleraja:
merging issues, the branch seems to be not clean. So fixing the issue in https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3540

PR #3539, opened 2022-09-05
[Table]: loop in overflow menu state update

### LINKED ISSUE #3359: [Table] Last overflow menu can crash React
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When the last row of a non-paginated table is right at the bottom of the browser, I think the overflow menu gets into an infinite loop trying to reposition itself.

### To Reproduce

It's a tricky one to provide code to reproduce. It seems to be just a regular table with no pagination & an overflow menu on each row. If the last row is close to the bottom of the browser, react will crash with the below message:

`Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.`

After a look at the code & some debugging in the browser, it looks like it keeps calling `usePopoverPositioning.fixOverflow`.

I could fix this in our product by just giving it a bit of room at the bottom for the overflow menu to fit for the last row.

### Expected behaviour

I would expect the overflow menu to appear above the ellipses when there is no room below.

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: 2.149.0-next.17

erzhan-temir-mamyrov:
Closes #3359 

**Summary**

- Fix loop in React state update for OverflowMenu when at bottom of the browser.

- Update Storybook to reproduce issue
- Check for previous state before updating overflow direction.

- Go to [this story](https://deploy-preview-3539--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-inline-actions)
- Resize window vertically so that scrollbar appears.
- Scroll to the bottom of the table.
- Hover on the last row and click on "kebab" menu icon.
- Verify that overflow menu has appeared above the "kebab" menu icon.

- tests here

### DIFF
--- a/packages/react/src/components/Table/Table.main.story.jsx
+++ b/packages/react/src/components/Table/Table.main.story.jsx
@@ -934,6 +934,16 @@ export const WithInlineActions = () => {
       getHiddenRowAction(),
       getHiddenOverflowRowAction(),
     ]),
+    null,
+    null,
+    null,
+    null,
+    null,
+    objectWithSubstitution('Row actions for row-9 (data[9].rowActions)', [
+      getOverflowEditRowAction(),
+      getOverflowAddRowAction(),
+      getOverflowDeleteRowAction(),
+    ]),
   ];
 
   const rowActionsState = object(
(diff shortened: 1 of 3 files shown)


davidicus (approved):
LGTM 👍

PR #3540, opened 2022-09-06
feat(card): adding tooltip to the card extra actions

herleraja:
Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3526

**Summary**

- adding tooltip to the card extra actions

- feat(card): adding tooltip to the card extra actions

- tooltips can be passed from the props

- card should work with out any issues

### DIFF
--- a/packages/react/src/components/Card/CardToolbar.jsx
+++ b/packages/react/src/components/Card/CardToolbar.jsx
@@ -175,8 +175,8 @@ const CardToolbar = ({
       extraActions.children && extraActions.children.length ? (
         <OverflowMenu
           flipped={overflowMenuPosition}
-          title={extraActions.iconDescription || mergedI18n.extraActionLabel}
-          iconDescription={extraActions.iconDescription || mergedI18n.extraActionLabel}
+          title={extraActions.iconDescription || mergedI18n.overflowMenuDescription}
+          iconDescription={extraActions.iconDescription || mergedI18n.overflowMenuDescription}
           {...(extraActions.icon ? { renderIcon: extraActions.icon } : {})}
         >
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Card/CardToolbar.jsx
```diff
@@ -175,8 +175,9 @@ const CardToolbar = ({
       extraActions.children && extraActions.children.length ? (
         <OverflowMenu
           flipped={overflowMenuPosition}
-          title={mergedI18n.overflowMenuDescription}
-          iconDescription={mergedI18n.overflowMenuDescription}
+          title={extraActions.iconDescription || mergedI18n.extraActionLabel}
```
davidicus, line 178: Shouldn't this be 
```suggestion
          title={extraActions.iconDescription || mergedI18n. overflowMenuDescription}
          iconDescription={extraActions.iconDescription || mergedI18n.overflowMenuDescription}
```

davidicus: In order to preserve backwards compatibility.

herleraja: I think overflowMenuDescription was used in some other places. Also I feel its wrong, because its an extra action.  So I changed to make sure the value to be extraActionLabel so that its behave correctly.

davidicus: The overflow is not the extra action though. This title is used as the aria-label and title of the dropdown and the items in the dropdown are the extra actions. So it probably should not get the extra actions text at all.

davidicus (approved):

PR #3542, opened 2022-09-07
feat(readonlyvalue): Read only value updates

cgirani:
**Summary**

- Support to send a node in prop value of ReadOnlyValue  component

- Prop support to render a node

- Check that the value prop supports a node to be rendered and 

- Nothing changed in the already supported cases

### DIFF
--- a/packages/react/src/components/ReadOnlyValue/_read-only-value.scss
+++ b/packages/react/src/components/ReadOnlyValue/_read-only-value.scss
@@ -2,7 +2,8 @@
   display: block;
   margin-bottom: $spacing-06;
   label,
-  textarea {
+  textarea,
+  div {
     display: block;
     &:focus-visible {
       outline: unset;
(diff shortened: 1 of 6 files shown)


packages/react/src/components/ReadOnlyValue/_read-only-value.scss
```diff
@@ -2,7 +2,8 @@
   display: block;
   margin-bottom: $spacing-06;
   label,
-  textarea {
+  textarea,
+  div {
```
davidicus, line 6: this only works if they pass in a div. Should we wrap the node content in a div to ensure this style is applied?

davidicus (commented):
Looks good. Just the one comment on the styling of the div.

packages/react/src/components/ReadOnlyValue/_read-only-value.scss
cgirani: Actually, i'm wrapping the value prop into the div, at the code... check line 44 in `ReadOnlyValue.jsx`

davidicus: You are right. My mistake!

davidicus (approved):

PR #3543, opened 2022-09-08
feat(card): Card actions support to custom content

cgirani:
**Summary**

- Adding support to card actions to array with custom content, so we can send custom elements to be used

- extraActions support to array
- updates in MDX

- Check that custom content can be sent in the extraActions prop following the pattern of array with objects

- Check that none of the changes impacted the card header


davidicus:
@dianatran18 I think you have to weight in on this since this changes the existing style guidance for this component. You can see an example [here](https://deploy-preview-3543--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card-card--basic-with-custom-action)

dianatran18:
Hey @cgirani do you have an example of the design that is requesting the custom actions?

davidicus (commented):
I'm not quite sure why we do not have an extraActions in the prop type declarations or why eslint does not scream at you for it but we should probably add here to document. Otherwise, this looks good code wise. I have asked @dianatran18 to weigh in on the design as this changes the guidance. Also our translated string for extra actions label may not be adequate with multiples so may be something we need to address as well.

dianatran18:
I think custom actions are fine to have if that's a need for the team, but there should be some structure around spacing/padding, order of actions, and number of actions if so. We also have the footer in the card component that can also be used to place additional actions, so it doesn't overcrowd the card header where the title also lives.

cgirani:
@dianatran18 / @davidicus  this is what Luke had for us:
image

cgirani:
@dianatran18 thanks for following up with Luke on that!
just a small question @davidicus, we would still need this PR right, at least since at this moment the Switcher is not supported in actions or we will add support only for the switcher?

herleraja (approved):
LGTM

davidicus:
@cgirani yes. we will still need this PR. I am doing another review.

davidicus (changes requested):
Hey @cgirani. This looks good. We just need to address the two comments about the prop type and the translated string. Then if we can update the story to show the new design using the footer for the text button.

packages/react/src/constants/CardPropTypes.js
```diff
@@ -966,6 +966,7 @@ export const CardPropTypes = {
         hidden: PropTypes.bool,
       })
     ),
+    content: PropTypes.arrayOf(PropTypes.node),
```
davidicus, line 969: This should probably reflect the option of a single object which is the current prop signature?

cgirani: Yeah I changed it to be just the node... and also some other small updates

davidicus (approved):

PR #3550, opened 2022-09-15
fix(datetimepicker): disable apply button when absolute value is emtpy

### LINKED ISSUE #3474: [DateTimePicker] Absolute range - Clicking apply without selecting dates results in "Invalid date to Invalid date"
### What package is this for?

- [x ] React
- [ ] Angular

### Describe the bug

Selecting absolute range for DateTimePicker and then clicking apply before making any date selections results in component displaying "Invalid date to Invalid date"

### To Reproduce

1. Click on component
2. Click on custom range
3. Select absolute radio
4. Click apply without making any date selection

### Expected behavior

The apply button shouldn't be clickable without making date selections

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context

Can reproduce here https://carbon-addons-iot-react.com/?path=/story/1-watson-iot-date-time-range-selector-datetime-picker--default

upcoming release?

date, the better!_

jessieyan:
Closes #3474 

**Summary**

- When switching from preset to relative and then absolute, the Apply button is enabled when no date is selected.

- added a check for absolute value

- Go to Default story
- Click on `Custom Range`
- Click on `Absolute`
- check `Apply` button is disabled

- tests here

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -623,7 +623,9 @@ const DateTimePicker = ({
   const disableAbsoluteApply =
     isCustomRange &&
     customRangeKind === PICKER_KINDS.ABSOLUTE &&
-    (absoluteStartTimeInvalid || absoluteEndTimeInvalid);
+    (absoluteStartTimeInvalid ||
+      absoluteEndTimeInvalid ||
+      (absoluteValue.startDate === '' && absoluteValue.endDate === ''));
 
   const disableApply = disableRelativeApply || disableAbsoluteApply;
 
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.test.jsx
@@ -764,6 +764,16 @@ describe('DateTimePicker', () => {
     expect(applyBytton).toBeDisabled();
   });
 
+  it('should disable apply button when switching from presets to relative then to absolute with empty startDate and endDate', () => {
+    render(<DateTimePicker {...dateTimePickerProps} id="picker-test" />);
+    jest.runAllTimers();
+
+    userEvent.click(screen.getByTestId('date-time-picker__field'));
+    userEvent.click(screen.getByText('Custom Range'));
+    userEvent.click(screen.getByText('Absolute'));
+    expect(screen.getByRole('button', { name: 'Apply' })).toBeDisabled();
+  });
+
   it('should enable apply button when absolute DatePicker input has start and end date in different dates', () => {
     const { i18n } = DateTimePicker.defaultProps;
 


davidicus (changes requested):
We will want to make sure that the `DateTimePickerV2` has these changes as well.

davidicus:
When I am in [this](https://deploy-preview-3550--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select) story and I clear out the value I am not seeing the button getting disabled.

davidicus:
> When I am in [this](https://deploy-preview-3550--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select) story and I clear out the value I am not seeing the button getting disabled.

@jessieyan, There has been no response to this comment.

jessieyan:
> > When I am in [this](https://deploy-preview-3550--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select) story and I clear out the value I am not seeing the button getting disabled.
> 
> @jessieyan, There has been no response to this comment.

Sorry I missed the comment. It should be fixed by adding a check on the empty time value.

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
```diff
@@ -540,6 +540,38 @@ describe('DateTimePickerV2', () => {
...
+
+    // default value is 2020-04-01 12:34 AM
+    expect(screen.getByText('2020-04-01 12:34 AM')).toBeVisible();
+
+    // first open the menu
+    userEvent.click(screen.getAllByText(/2020-04-01 12:34 AM/i)[0]);
+
+    userEvent.clear(screen.getByTestId('date-time-picker-input'));
+
+    expect(screen.getByText(i18n.applyBtnLabel)).toBeDisabled();
```
sls-ca, line 572: Are we saying that if we enter a date/time then we can never clear it?
I suspect that this component needs to have a "required" attribute.  If the date is required, it should not allow the apply if the value is empty.  BUt if the component is not required, then I should be able to clear the date.  

@JordanWSmith15 @davidicus

sls-ca (changes requested):

sls-ca (approved):

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
jessieyan: > Are we saying that if we enter a date/time then we can never clear it? I suspect that this component needs to have a "required" attribute. If the date is required, it should not allow the apply if the value is empty. BUt if the component is not required, then I should be able to clear the date.
> 
> @JordanWSmith15 @davidicus

You can either clear the input manually or click on the `Clear` button as Jordan mentioned above. The test was saying if time input is cleared/empty we need to disable the `Apply` button

davidicus (approved):

PR #3557, opened 2022-09-20
fix(table): empty search result should show EmptyTable

### LINKED ISSUE #3368: [table] empty search result should show custom EmptyState
### What package is this for?

- [x] React
- [ ] Angular

### Summary

The empty search result in the StatefulTable should show a custom EmptyState adjusted for search the same way it does for empty result when filtering.

**Additional context** 

This is the EmptyState when filtering, we need something similar for the search
image

upcoming release?

date, the better!_

### Available extra resources

erzhan-temir-mamyrov:
Closes #3368 

**Summary**

- Display `EmptyTable` if search results are empty

- Adjust props in order to display `EmptyTable`
- Enable clearing search input by clicking on "Clear all filter" button in `EmptyTable`
- Update / Add tests
- Update docs

- Go to [this story](https://deploy-preview-3557--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--playground)
- In the Knobs section click on `Search` tab
- Enable `options.hasSearch` and `options.hasFastSearch` in order to display search input in the toolbar
- Type some irrelevant value in the search bar
- Verify that `EmptyTable` displayed
- Click on "Clear all filters" button in order to reset search input
- Verify that search input was cleared

- tests here

### DIFF
--- a/packages/react/src/components/Table/Table.test.jsx
+++ b/packages/react/src/components/Table/Table.test.jsx
@@ -603,6 +603,76 @@ describe('Table', () => {
     expect(screen.getByTestId(`${tableId}-table-toolbar-search`)).toHaveValue('ferrari');
   });
 
+  it('toolbar search should rerender with different search value', () => {
+    const tableId = 'table';
+
+    const { rerender } = render(
+      <Table
+        columns={tableColumns}
+        data={[tableData[0]]}
+        actions={mockActions}
+        options={{
+          hasSearch: true,
+        }}
+        view={{
+          toolbar: {
+            search: {
+              defaultValue: '',
+            },
+          },
+        }}
+        id={tableId}
+      />
+    );
+
+    expect(screen.getByTestId(`${tableId}-table-toolbar-search`)).toBeInTheDocument();
+    expect(screen.getByTestId(`${tableId}-table-toolbar-search`)).toHaveValue('');
+
+    rerender(
+      <Table
+        columns={tableColumns}
+        data={[tableData[0]]}
+        actions={mockActions}
+        options={{
+          hasSearch: true,
+        }}
+        view={{
+          toolbar: {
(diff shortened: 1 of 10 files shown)


packages/react/src/components/Table/Table.test.jsx
```diff
@@ -590,27 +592,40 @@ describe('Table', () => {
...
           },
         }}
+        id={tableId}
       />
     );
+    expect(screen.getByTestId(`${tableId}-table-toolbar-search`)).toBeInTheDocument();
+    expect(screen.getByTestId(`${tableId}-table-toolbar-search`)).toHaveValue('ferrari');
+  });
 
-    expect(wrapper.find(`.${prefix}--search-input`)).toHaveLength(1);
```
davidicus, line 600: Can we use the `rerender` method from testing library to see what happens when we update the defaultValue for search. I think this is something that we were asked to be able to support. I just want to be sure it's still functional.

erzhan-temir-mamyrov: @davidicus Thank you for bringing this up. It was a bug. If the `defaultValue` updated then `TableToolbar` wouldn't rerender as well as table data.
It is fixed in latest commits by modifying `tableRegister` action, which is rerendering whole table in case of new defaultValue. Relevant test is added for this feature.
Also, it can be tested:

- in [this story](https://deploy-preview-3557--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-searching)
- Modify the `view.toolbar.search.defaultValue` input in the knobs area
- Verify that search input in toolbar is being rerendered and table is filtered accordingly

davidicus: Great!

davidicus (approved):

PR #3564, opened 2022-09-22
Invalid and disabled props in DateTimePicker and DateTimePickerV2

### LINKED ISSUE #3558: [DateTimePicker(V2)] Cannot pass in a translated string for invalid Text
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

For both `DateTimePicker` and `DateTimePickerV2` there is no way to make the input, as a whole, be in an invalid state. We also have no invalidText string we pass into the `i18n` object. We need to be able to put the input into an invalid state and pass in a translated string.

jessieyan:
Closes #3558 #3569

**Summary**

- Support invalid state and an invalid helper text in `DateTimePicker` and `DateTimePickerV2` 
- Support disabled state in `DateTimePicker` and `DateTimePickerV2` 

- added `invalid`, `invalidText`, `invalidLabel` props in `DateTimePicker` 
- added `invalid`, `invalidText` props in `DateTimePickerV2` 
- updated `disabled` props to disable menu button 
- updated stories to reflect the new props

DateTimePicker
image

image

image

DateTimePickerV2
image

image

image

image

image

- tests here

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -653,6 +653,7 @@ const DateTimePicker = ({
         className={classnames(`${iotPrefix}--date-time-picker__box`, {
           [`${iotPrefix}--date-time-picker__box--light`]: light,
           [`${iotPrefix}--date-time-picker__box--invalid`]: invalidState,
+          [`${iotPrefix}--date-time-picker__box--disabled`]: disabled,
         })}
       >
         <div
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -317,7 +317,7 @@ const DateTimePicker = ({
   style,
   ...others
 }) => {
-  const strings = {
+  const mergedI18n = {
     ...defaultProps.i18n,
     ...i18n,
   };
@@ -429,7 +429,7 @@ const DateTimePicker = ({
       value.kind = PICKER_KINDS.PRESET;
     }
     setCurrentValue(value);
-    const parsedValue = parseValue(value, dateTimeMask, strings.toLabel);
+    const parsedValue = parseValue(value, dateTimeMask, mergedI18n.toLabel);
     setHumanValue(parsedValue.readableValue);
 
     return {
@@ -622,7 +622,7 @@ const DateTimePicker = ({
 
   const tooltipValue = renderPresetTooltipText
     ? renderPresetTooltipText(currentValue)
-    : getIntervalValue({ currentValue, strings, dateTimeMask, humanValue });
+    : getIntervalValue({ currentValue, mergedI18n, dateTimeMask, humanValue });
 
   const disableRelativeApply =
     isCustomRange &&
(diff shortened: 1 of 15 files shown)


jessieyan:
still need to update internal invalidValue state to reflect invalid.

jessieyan:
need to update invalid props base on internal state as well

packages/react/src/components/DateTimePicker/__snapshots__/DateTimePicker.story.storyshot
```diff
@@ -1740,12 +1740,12 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/D
                             className="bx--number__controls"
                           >
                             <button
-                              aria-label="Decrement number"
+                              aria-label="undefined undefined"
```
davidicus, line 1743: @jessieyan do we know what's going on here?

jessieyan: Looks like there is a bug. We should reference the `strings` object instead of the `i18n` object

jessieyan: > @jessieyan do we know what's going on here?

Looks like there is a bug. We should reference the strings object instead of the i18n object

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -793,9 +816,9 @@ const DateTimePicker = ({
                           onChange={onRelativeLastNumberChange}
                           translateWithId={(messageId) =>
                             messageId === 'increment.number'
-                              ? `${i18n.increment} ${i18n.number}`
+                              ? `${strings.increment} ${strings.number}`
```
herleraja, line 819: any specific reason `i18n` variable not used? `strings` seems to be too generic

herleraja, line 819: it would be nice to rename `strings` to `mergedI18n`. Just to make it more consistency with other components. Also `strings` is kind of data type, so we may get confused.

jessieyan: > it would be nice to rename `strings` to `mergedI18n`. Just to make it more consistency with other components. Also `strings` is kind of data type, so we may get confused.

Thanks for the feedback. I've update all `strings` to `mergedI18n`.

jessieyan:
Updated PR with disabled props related changes.

davidicus (changes requested):
Looks pretty good. Only thing is we need the disabled cursor like other inputs when in disabled state.

davidicus:
Is the same class not being supplied to all cases of being disabled? [This](https://deploy-preview-3564--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select) does not show the same disabled cursor or grayed out icon.

jessieyan:
> Is the same class not being supplied to all cases of being disabled? [This](https://deploy-preview-3564--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select) does not show the same disabled cursor or grayed out icon.

I missed one file. Just updated it.

davidicus (approved):

PR #3567, opened 2022-09-26
fix(emptystate): adjust proptypes schema

### LINKED ISSUE #3370: [Table] incorrect PropTypes for emptyState
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The Table has incorrect PropTypes for the prop `view.table.emptyState`. It uses `EmptyStatePropTypes` which defines a configuration object where all property values are of type `PropTypes.node`, but the actual underlying `EmptyState` component used expects strings for several of the props. This is true for `message`, `messageWithFilters`, `buttonLabel` and `buttonLabelWithFilters`. Only the actual `messageBody` can be a node.

### To Reproduce

na

### Expected behavior

PropTypes in `EmptyStatePropTypes` should match the types they are matched against in the underlying `EmptyState` component.

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

erzhan-temir-mamyrov:
Closes #3370 

**Summary**

- Change PropTypes schema for `EmptyState`

- PropTypes schema
- Public API snapshot update

- Go to [this commit](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3567/commits/0aeda1e8d8aba1b231e2904dc82f60a6d6ce261c)
- Verify that PropTypes fields were converted from single `node` to `oneOfType([PropTypes.node, PropTypes.string])`

- tests here


packages/react/src/components/Table/TablePropTypes.js
```diff
@@ -49,15 +49,15 @@ export const RowActionsStatePropTypes = PropTypes.arrayOf(
 
 export const EmptyStatePropTypes = PropTypes.oneOfType([
   PropTypes.shape({
-    message: PropTypes.node.isRequired,
-    messageBody: PropTypes.node,
+    message: PropTypes.oneOfType([PropTypes.node, PropTypes.string]).isRequired,
```
jessieyan, line 52: This should be only string.
message: PropTypes.string.isRequired,

packages/react/src/components/Table/TablePropTypes.js
```diff
@@ -49,15 +49,15 @@ export const RowActionsStatePropTypes = PropTypes.arrayOf(
...
+    messageBody: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
     /* Show a different message if no content is in the table matching the filters */
-    messageWithFilters: PropTypes.node,
-    messageWithFiltersBody: PropTypes.node,
+    messageWithFilters: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
```
jessieyan, line 55: only string here

packages/react/src/components/Table/TablePropTypes.js
```diff
@@ -49,15 +49,15 @@ export const RowActionsStatePropTypes = PropTypes.arrayOf(
...
+    messageWithFiltersBody: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
```
jessieyan, line 56: only string here

packages/react/src/components/Table/TablePropTypes.js
```diff
@@ -49,15 +49,15 @@ export const RowActionsStatePropTypes = PropTypes.arrayOf(
...
     /* If a label is not provided, no action button will be rendered */
-    buttonLabel: PropTypes.node,
+    buttonLabel: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
```
jessieyan, line 58: only string here

packages/react/src/components/Table/TablePropTypes.js
```diff
@@ -49,15 +49,15 @@ export const RowActionsStatePropTypes = PropTypes.arrayOf(
...
     /* Show a different button label if no content is in the table matching the filters */
-    buttonLabelWithFilters: PropTypes.node,
+    buttonLabelWithFilters: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
```
jessieyan, line 60: only string here

packages/react/src/components/Table/TablePropTypes.js
```diff
@@ -49,15 +49,15 @@ export const RowActionsStatePropTypes = PropTypes.arrayOf(
...
-    messageBody: PropTypes.node,
+    message: PropTypes.oneOfType([PropTypes.node, PropTypes.string]).isRequired,
+    messageBody: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
     /* Show a different message if no content is in the table matching the filters */
-    messageWithFilters: PropTypes.node,
-    messageWithFiltersBody: PropTypes.node,
+    messageWithFilters: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
+    messageWithFiltersBody: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
     /* If a label is not provided, no action button will be rendered */
-    buttonLabel: PropTypes.node,
```
jessieyan, line 58: only string here

jessieyan (changes requested):
Looked at `<EmptyTable>` and `<EmptyState> `component,  

There are string type:
messageWithFilters,
message,
buttonLabel,
 buttonLabelWithFilters,

These are either strings or node:
 messageBody,
 messageWithFiltersBody,

packages/react/src/components/Table/TablePropTypes.js
davidicus: Since this is a minor version bump we must keep  the node proptype for backwards compatibility.

davidicus:
Looking at the actual issue, I guess we need to see if we were getting proptype errors previously by using a node. If we were getting prop type errors no matter what then we might want to look at if EmptyState component to see if we refactor it to use a node. My original concern on first glance is that we were requiring the node prop type and are now changing it on our users.  @erzhan-temir-mamyrov If you want to discuss tomorrow  let me know and we can hop on a quick call to discuss. Thanks @jessieyan for pointing out the issues concerns.

packages/react/src/components/Table/TablePropTypes.js
erzhan-temir-mamyrov: @jessieyan Thank you for commenting on this. I changed `PropTypes.node` to `PropTypes.oneOfType([PropTypes.node, PropTypes.string])` for backward compatibility. In the last commits I modified `EmptyState` component itself, for prop types of `string` and `node`.

davidicus (approved):

PR #3568, opened 2022-09-26
chore(styles): replace divide symbol to math.div

### LINKED ISSUE #3439: [Repo] Find and replace all `/` used for division in sass files
Both our selves and our consumers will continue to get multiple warnings about the deprecation of this use until we are refactored. We should find  all instances of using the `/` symbol for division and replace it with `Math.div`.

[Relevant announcement](https://sass-lang.com/documentation/breaking-changes/slash-div) from Sass team about the deprecation 

### Update: 
The sass team has updated their docs since the [initial](https://web.archive.org/web/20210521230638/https://sass-lang.com/documentation/breaking-changes/slash-div) deprecation announcement. They now state that calc function's usage of the forward slash operator will not be deprecated. So only those uses of forward slash for division outside a calc function have to be updated.

For those wishing to silence warnings from dependencies they can either pass the [--quiet-deps flag](https://sass-lang.com/documentation/cli/dart-sass#quiet-deps) if they are using the command line or setting the [quietDeps option](https://sass-lang.com/documentation/js-api/interfaces/Options#quietDeps) to true when using the JS api.

erzhan-temir-mamyrov:
Closes #3439 

**Summary**

- Replace "/" symbol to `Math.div` in `scss` files

- Enable `sass:math` module (add `@use` directive to stylelint ignore)
- Affected component `ComboBox`
- Affected component `ImageCard`
- Affected component `MenuButton`
- Affected component `TileCatalog`

- Go to [this commit](https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3568/commits/54d1d17de2c4a6a443ee8937984c49613b6b8561)
- Verify that all "/" symbols have been replaced by `*` or `Math.div`

- tests here

### DIFF
--- a/packages/react/src/components/TileCatalog/_catalog-content.scss
+++ b/packages/react/src/components/TileCatalog/_catalog-content.scss
@@ -29,10 +29,10 @@
   padding-bottom: $spacing-03;
   max-width: calc(100vw - 20rem);
   @media screen and (min-width: $two-pane) {
-    max-width: calc(100vw / 2 - 15rem);
+    max-width: calc(math.div(100vw, 2) - 15rem);
   }
   @media screen and (min-width: $three-pane) {
-    max-width: calc(100vw / 3 - 15rem);
+    max-width: calc(math.div(100vw, 3) - 15rem);
   }
 }
 
--- a/packages/react/src/styles.scss
+++ b/packages/react/src/styles.scss
@@ -2,6 +2,12 @@
 // 🌍 Global
 //-------------------------
 
+// Enable math module to handle dividing operations due to deprecation of " / " operator
+// Reference https://sass-lang.com/documentation/breaking-changes/slash-div
+/* stylelint-disable */
+@use "sass:math";
+/* stylelint-enable */
+
 /// If true, includes font face mixins in `_css--font-face.scss` depending on the `css--plex` feature flag
 /// @access public
 /// @type Bool
--- a/packages/react/src/components/ComboBox/_combo-box.scss
+++ b/packages/react/src/components/ComboBox/_combo-box.scss
@@ -45,34 +45,34 @@
       right: $spacing-03;
       z-index: z('overlay') + 1;
       // half the height of the input/first list item
-      bottom: calc(#{$ai-apps-combobox-input-negative-height} / 2);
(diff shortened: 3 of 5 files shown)


jessieyan (approved):
Looks good to me

davidicus:
@erzhan-temir-mamyrov great job. This is just a note for anyone looking at this issue in the future. 

The sass team has updated their docs since the [initial](https://web.archive.org/web/20210521230638/https://sass-lang.com/documentation/breaking-changes/slash-div) deprecation announcement. They now state that calc function's usage of the forward slash operator will not be deprecated. So only those uses of forward slash for division outside a calc function have to be updated.

For those wishing to silence warnings from dependencies they can either pass the [--quiet-deps flag](https://sass-lang.com/documentation/cli/dart-sass#quiet-deps) if they are using the command line or setting the [quietDeps option](https://sass-lang.com/documentation/js-api/interfaces/Options#quietDeps) to true when using the JS api.

davidicus (approved):

packages/react/src/components/TileCatalog/_catalog-content.scss
```diff
@@ -29,10 +29,10 @@
   padding-bottom: $spacing-03;
   max-width: calc(100vw - 20rem);
   @media screen and (min-width: $two-pane) {
-    max-width: calc(100vw / 2 - 15rem);
+    max-width: calc(math.div(100vw, 2) - 15rem);
```
herleraja, line 32: why cant we do like above ? 100vw * 0.5 ? why do we need to use math.div ?

packages/react/src/styles.scss
```diff
@@ -2,6 +2,12 @@
 // 🌍 Global
 //-------------------------
 
+// Enable math module to handle dividing operations due to deprecation of " / " operator
+// Reference https://sass-lang.com/documentation/breaking-changes/slash-div
+/* stylelint-disable */
+@use "sass:math";
+/* stylelint-enable */
+
```
herleraja, line 10: In the downstream Maximo monitor component, this is causing the following error. 

image


I know we have to re-arrange the scss imports to over come the issue. but It may create CSS style related regressions.

PR #3576, opened 2022-09-28
Fix data filter for v2 device type

Ujjwal-UJ:
Closes Monitor - https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/4661

**Summary**

- Fixed DataFilter not available for V2 device types under dashboard editor

- fix(dataseriesformcontentmodal): added datafilter for v2 devicetypes
- fix(dataseriesformitemmodal): updated test cases

- Datafilter should be available for v2 device type on click of edit data item after adding it into a card in dashboard editor

- tests here

### DIFF
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditForm.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditForm.jsx
@@ -117,10 +117,10 @@ const defaultProps = {
   isSummaryDashboard: false,
   dataSeriesItemLinks: null,
   actions: {
-    onEditDataItem: null,
+    onEditDataItem: () => {},
     dataSeriesFormActions: {
-      hasAggregationsDropDown: null,
-      onAddAggregations: null,
+      hasAggregationsDropDown: () => {},
+      onAddAggregations: () => {},
     },
   },
 };
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditForm.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditForm.jsx
@@ -117,10 +117,11 @@ const defaultProps = {
   isSummaryDashboard: false,
   dataSeriesItemLinks: null,
   actions: {
-    onEditDataItem: () => {},
+    onEditDataItem: null,
     dataSeriesFormActions: {
-      hasAggregationsDropDown: () => {},
-      onAddAggregations: () => {},
+      hasAggregationsDropDown: null,
+      hasDataFilterDropdown: null,
+      onAddAggregations: null,
     },
   },
 };
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditForm.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditForm.jsx
@@ -119,7 +119,7 @@ const defaultProps = {
   actions: {
     onEditDataItem: null,
     dataSeriesFormActions: {
-      hideAggregationsDropDown: null,
+      hasAggregationsDropDown: null,
       onAddAggregations: null,
     },
   },
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
@@ -195,11 +195,11 @@ const defaultProps = {
   testId: 'aggregation-methods',
 
   actions: {
-    onEditDataItem: null,
+    onEditDataItem: () => {},
     dataSeriesFormActions: {
-      hasAggregationsDropDown: null,
-      hasDataFilterDropdown: null,
-      onAddAggregations: null,
+      hasAggregationsDropDown: () => {},
+      hasDataFilterDropdown: () => {},
+      onAddAggregations: () => {},
     },
   },
 };
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
@@ -197,8 +197,8 @@ const defaultProps = {
   actions: {
     onEditDataItem: null,
     dataSeriesFormActions: {
-      hideAggregationsDropDown: null,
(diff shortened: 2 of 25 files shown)


herleraja (approved):
LGTM

EdmondCShaffer (approved):
LGTM

packages/react/src/components/DashboardEditor/editorUtils.jsx
```diff
@@ -716,6 +716,7 @@ export const DashboardEditorActionsPropTypes = PropTypes.shape({
   onEditDataItem: PropTypes.func,
   dataSeriesFormActions: PropTypes.shape({
     hideAggregationsDropDown: PropTypes.func,
+    hideDataFilterDropdown: PropTypes.func,
```
davidicus, line 719: can we get an argument signature here? You may want to add some for the hideAggregationsDropdown as well. None of these functions are documented so it will be hard to write documentation for this component. I also do not see any test for this new functionality.

davidicus (changes requested):
Please document and test.

packages/react/src/components/DashboardEditor/editorUtils.jsx
EdmondCShaffer: @davidicus I see test for this Callback Function 
Screen Shot 2022-09-28 at 1 52 00 PM

davidicus: @EdmondCShaffer This test is not testing the functionality of the `hideDataFilterDropdown` function. It's just passing it in as a prop with a mocked function. 

This PR introduces a branch in the code depending on the return value of this function. We need to test that branched code. So you would mock a return value for this that would allow the branched code to render and then assert that it's there.

herleraja (changes requested):
I have asked you to add this in actions. I think its wrong. Lets move this into new options props than actions.

herleraja (approved):
LGTM

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -197,7 +197,8 @@ const defaultProps = {
   actions: {
     onEditDataItem: null,
     dataSeriesFormActions: {
-      hideAggregationsDropDown: null,
+      hasAggregationsDropDown: null,
```
davidicus, line 200: These defaults need to be a no-op (`() => {}`), otherwise your check on line 304 will error. Or am I missing a check somewhere?

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormContent.jsx
```diff
@@ -156,7 +156,7 @@ const defaultProps = {
   actions: {
     onEditDataItem: null,
     dataSeriesFormActions: {
-      hideAggregationsDropDown: null,
+      hasAggregationsDropDown: null,
```
davidicus, line 159: this function default should be a no-op

packages/react/src/components/CardEditor/CardEditor.jsx
```diff
@@ -178,7 +178,7 @@ const defaultProps = {
   actions: {
     onEditDataItem: null,
     dataSeriesFormActions: {
-      hideAggregationsDropDown: null,
+      hasAggregationsDropDown: null,
```
davidicus, line 181: need to be a no-op

packages/react/src/components/DashboardEditor/DashboardEditor.jsx
```diff
@@ -389,7 +389,7 @@ const defaultProps = {
   actions: {
     onEditDataItem: null,
     dataSeriesFormActions: {
-      hideAggregationsDropDown: null,
+      hasAggregationsDropDown: null,
```
davidicus, line 392: need to be a no-op

packages/react/src/components/HotspotEditorModal/HotspotEditorDataSourceTab/HotspotEditorDataSourceTab.jsx
```diff
@@ -98,7 +98,7 @@ const defaultProps = {
   actions: {
     onEditDataItem: null,
     dataSeriesFormActions: {
-      hideAggregationsDropDown: null,
+      hasAggregationsDropDown: null,
```
davidicus, line 101: need to be a no-op

davidicus (changes requested):
Just need to provide a no-op for defaults.

Ujjwal-UJ:
> Just need to provide a no-op for defaults.

Updated

davidicus (approved):

davidicus:
Looks like these changes has brought these files under the code threshold. You will want to add unit test to ensure that these thresholds are met. 
image

packages/react/src/components/CardEditor/CardEditForm/CardEditForm.jsx
```diff
@@ -117,10 +117,10 @@ const defaultProps = {
   isSummaryDashboard: false,
   dataSeriesItemLinks: null,
   actions: {
-    onEditDataItem: null,
+    onEditDataItem: () => {},
```
davidicus, line 120: add `/* istanbul ignore next */` above the default props declaration on line 83 to ignore these no-ops. Do the same for each file where no ops were added

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.test.jsx
```diff
@@ -187,6 +188,58 @@ describe('DataSeriesFormItemModal', () => {
     const legendColorLabel = screen.getByText('Line color');
     expect(label).toBeInTheDocument();
     expect(legendColorLabel).toBeInTheDocument();
+
+    const dataFilter = screen.getByRole('listbox', { name: 'Data filter' });
```
davidicus, line 192: We are testing that it renders without. We probably want to add test to ensure it renders with.

herleraja: @davidicus we already have a test case for that.

PR #3582, opened 2022-10-04
fix(sass): remove @use declaration

### LINKED ISSUE #3581: [Sass] Remove the @use declaration
Monitor has reported an issue with the `@use` sass:math import we recently added. I would like to do some test to see what the issue is. In the meantime, since its not strictly necessary, let's remove to unblock this team. 

This will also require us to change the Math function used in `packages/react/src/components/TileCatalog/_catalog-content.scss` on line 32

erzhan-temir-mamyrov:
Closes #3581 

**Summary**

- Remove `@use "sass:math"`

- Replace `math.div` to `*` or `/` operators
- Delete `@use "sass:math"` from root stylesheet

- Go to [this story](https://deploy-preview-3582--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-catalog-marketplace-tilecatalog--default)
- Change screen width and verify that `max-width` property in `iot--sample-tile-title` is changing accordingly

- tests here

### DIFF
--- a/packages/react/src/components/TileCatalog/_catalog-content.scss
+++ b/packages/react/src/components/TileCatalog/_catalog-content.scss
@@ -29,10 +29,10 @@
   padding-bottom: $spacing-03;
   max-width: calc(100vw - 20rem);
   @media screen and (min-width: $two-pane) {
-    max-width: calc(math.div(100vw, 2) - 15rem);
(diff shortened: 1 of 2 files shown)


herleraja (approved):
LGTM

davidicus (approved):
Thanks for getting this out so quick!

PR #3588, opened 2022-10-07
fix(table): create table action on search expand

### LINKED ISSUE #3341: [table] search callbacks not working
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

The Table search prop has two callbacks defined in the PropTypes definition and also mentioned in the documentation that do not work. It is the `view.toolbar.search.onChange` and the `view.toolbar.search.onExpand`. The `onChange` is overwritten in the TableToolbar.jsx and never works. The view.toolbar.search.onExpand works for the `Table` but not for the `StatefulTable` where the whole prop object `view.toolbar.search` is modified and the callbacks lost.

What strikes me here is that these callbacks should never have been placed in the `view` prop to begin with, they belong in the `actions` prop.

### To Reproduce

Modify a Table story with search and try to use these callbacks

### Expected behavior

1. Not sure if we need or should expose the `onChange` since we already have the `onApplySearch` with the `options.hasFastSearch` that currently handles the different search scenarios. The callback should probably just be removed in order to not cause confusion or future bugs. 
2. The `view.toolbar.search.onExpand` should be deprecated and a new prop `onSearchExpand` should be added to the `actions.toolbar` prop instead. This prop should work for both the Table and the StatefulTable
3. Docs and stories must be updated

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

erzhan-temir-mamyrov:
Closes #3341 

**Summary**

- Create new table action on search expand, deprecated previous one.

- Add table action that triggers on search open/close
- Add deprecation notice to previous search expand callback - `onExpand`
- Add deprecation notice to unused `onChange` callback in toolbar search
- Update test
- Update docs

- Go to [this story](https://deploy-preview-3588--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-searching)
- Open `Actions` tab
- Focus on search input
- Verify that `onSearchExpand` action has fired with event class and `true` parameters
- Clear search input and move focus away from search bar
- Verify that `onSearchExpand` action has fired with event class and `false` parameters
- In the Knobs tab change type of table to `Table`
- Verify that `onSearchExpand` triggers on the same action as in `StatefulTable`

- tests here


jessieyan (approved):
Looks good to me

davidicus (approved):

PR #3590, opened 2022-10-10
fix(SuiteHeader): Protecting the idle timer logic to avoid the creation of a cookie with an empty domain

blechner:
**Summary**

- When rendering the `IdleLogoutConfirmationModal` component, check if the domain is set. Otherwise the `IdleTimer` may be instantiated with an undefined domain, which will create a cookie in the wrong domain.

- Checking if domain is null or undefined before instantiating the `IdleLogoutConfirmationModal` and `IdleTimer`.

- No functional changes expected in this fix.

- No functional changes expected in this fix.

### DIFF
--- a/packages/react/src/components/SuiteHeader/IdleLogoutConfirmationModal/IdleLogoutConfirmationModal.jsx
+++ b/packages/react/src/components/SuiteHeader/IdleLogoutConfirmationModal/IdleLogoutConfirmationModal.jsx
@@ -113,7 +113,12 @@ const IdleLogoutConfirmationModal = ({
   const [stayLoggedInButtonDisabled, setStayLoggedInButtonDisabled] = useState(false);
   // eslint-disable-next-line consistent-return
   useEffect(() => {
-    if (idleTimeoutData && idleTimeoutData?.timeout > 0 && routes?.domain) {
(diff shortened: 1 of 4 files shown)


davidicus (commented):
Code looks good. Should we add some test?

packages/react/src/components/SuiteHeader/IdleLogoutConfirmationModal/IdleLogoutConfirmationModal.jsx
```diff
@@ -113,7 +113,12 @@ const IdleLogoutConfirmationModal = ({
   const [stayLoggedInButtonDisabled, setStayLoggedInButtonDisabled] = useState(false);
   // eslint-disable-next-line consistent-return
   useEffect(() => {
-    if (idleTimeoutData && idleTimeoutData?.timeout > 0 && routes) {
+    if (
```
sls-ca, line 116: Glad to see if fixed here as well.  I just got another defect where the cookie was being created using the wrong domain, and turns out someone was using this class directly vs SuiteHeader.

sls-ca (approved):

davidicus (approved):

PR #3596, opened 2022-10-12
feat(datetimepickerv2): support both 12 and 24 hours time format

### LINKED ISSUE #3595: [datetimepickerV2] support both 12 and 24 hours time format with new timepickerdropdown
### What package is this for?

- [ x] React
- [ ] Angular

### Summary

`DateTimePickerV2` should support the new is24hours prop on `TimePickerDropDown`
Add new prop is24hours in DateTimePickerV2 to specify time format

request here.

upcoming release?

date, the better!_

### Available extra resources

jessieyan:
Closes #3595 

**Summary**

- indicate 12 hours or 24 hours by the definition of prop `dateTimeMask`, specifically if the hour mask is `HH`, it means 24 hours, and `hh` means 12 hours , despite of whether  `dateTimeMask` has AM/PM for not. 
- make sure both 12 hours and 24 hours are supported for single select and range select
- updated stories and tests

For example,

If `dateTimeMask` is `YYYY-MM-DD HH:mm` this is 24 hours. The field shows value like `2020-04-01 13:34`. `TimePickerDropDown` is 24hours format like `13:34`.

If `dateTimeMask` is `YYYY-MM-DD HH:mm A` this is 24 hours with AM/PM. The field shows value like `2020-04-01 13:34 PM`. `TimePickerDropDown` is 24hours format like `13:34`.

If `dateTimeMask` is `YYYY-MM-DD hh:mm A` this is 12 hours with AM/PM.  The field shows value like `2020-04-01 01:34 PM `. `TimePickerDropDown` is 24hours format like `01:34 PM`.

For single select
- go to single select [story](https://deploy-preview-3596--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select)
- make sure is24hours is checked
- click on menu button to open flyout menu
- the time format on time picker should be 24 hours
- click on clock button to open time spinner
- the spinner should only show hours and minutes

For range select
- go to range select [story](https://deploy-preview-3596--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--selected-absolute-with-new-time-spinner)
- make sure is24hours is checked
- click on menu button to open flyout menu
- the time format on time picker should be 24 hours
- click on clock button to open time spinner
- the spinner should only show hours and minutes

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
@@ -29,7 +29,6 @@ const DateTimePicker = ({
   hasIconOnly,
   menuOffset,
   datePickerType,
-  is24hours,
   style,
   ...others
 }) => {
@@ -57,7 +56,6 @@ const DateTimePicker = ({
       hasIconOnly={hasIconOnly}
       menuOffset={menuOffset}
       datePickerType={datePickerType}
-      is24hours={is24hours}
       style={style}
       others={others}
     />
--- a/packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
@@ -29,6 +29,7 @@ const DateTimePicker = ({
   hasIconOnly,
   menuOffset,
   datePickerType,
+  is24hours,
   style,
   ...others
 }) => {
@@ -56,6 +57,7 @@ const DateTimePicker = ({
       hasIconOnly={hasIconOnly}
       menuOffset={menuOffset}
       datePickerType={datePickerType}
+      is24hours={is24hours}
       style={style}
       others={others}
     />
--- a/packages/react/src/components/DateTimePicker/DateTimePickerV2.story.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePickerV2.story.jsx
@@ -121,7 +121,6 @@ export const SelectedAbsolute = () => {
     <DateTimePicker
       id="datetimepicker"
       useNewTimeSpinner={boolean('useNewTimeSpinner', false)}
-      is24hours={boolean('is24hours', true)}
       defaultValue={defaultAbsoluteValue}
       dateTimeMask={text('dateTimeMask', 'YYYY-MM-DD HH:mm')}
       hasTimeInput={boolean('hasTimeInput', true)}
@@ -143,12 +142,12 @@ export const SelectedAbsolute = () => {
 SelectedAbsolute.storyName = 'Selected absolute';
 
 export const SingleSelect = () => {
-  const is24hours = boolean('is24hours', true);
+  const dateTimeMask = text('dateTimeMask', 'YYYY-MM-DD HH:mm');
   return (
     <DateTimePicker
       id="datetimepicker"
+      key={dateTimeMask}
       useNewTimeSpinner
-      is24hours={is24hours}
       defaultValue={{
         timeRangeKind: PICKER_KINDS.SINGLE,
         timeSingleValue: {
@@ -156,7 +155,7 @@ export const SingleSelect = () => {
           startTime: '12:34',
         },
       }}
-      dateTimeMask={text('dateTimeMask', 'YYYY-MM-DD HH:mm')}
+      dateTimeMask={dateTimeMask}
       hasTimeInput={boolean('hasTimeInput', true)}
       onApply={action('onApply')}
       onCancel={action('onCancel')}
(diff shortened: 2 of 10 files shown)


davidicus (changes requested):
Can we get a test when the mask is in 24 hour format and is24Hours prop is false? Just want to make sure our masking logic is still working with these changes.

jessieyan:
I added a check for `dateTimeMask`. We will respect `dateTimeMask` prior to `is24hours` prop

davidicus (dismissed):
This LGTM. @JordanWSmith15 Are we ok with the dateTimeMask overriding the is24Hour? Or do we want them to be independent?

JordanWSmith15:
> This LGTM. @JordanWSmith15 Are we ok with the dateTimeMask overriding the is24Hour? Or do we want them to be independent?

I deffer to @sls-ca

sls-ca:
> This LGTM. @JordanWSmith15 Are we ok with the dateTimeMask overriding the is24Hour? Or do we want them to be independent?

What is the point of the is24hrs flag?  If the mask controls this behaviour then I'm not sure why there is a separate property for this?

davidicus:
I would say that if the mask is going to dictate the behavior then we probably did not need an additional prop. We can deduce the format from the mask. If we want these two things to be independent then it makes sense to add the additional prop.

We can let @JordanWSmith15 and @stuckless inform us if Graphite needs these to be independent, otherwise we can merge this in.

davidicus:
> > This LGTM. @JordanWSmith15 Are we ok with the dateTimeMask overriding the is24Hour? Or do we want them to be independent?
> 
> What is the point of the is24hrs flag? If the mask controls this behaviour then I'm not sure why there is a separate property for this?

Right. These were originally independent of each other, is24Hour would dictate the dropdown spinner state and the mask would be the format that the value would be changed to. Now, the mask is overriding the is24hour prop and is redundant.

sls-ca:
From graphite's point of view we'd only set the mask, which might be 24hr or not... we don't care.  No one will ever set the is24hr flag since, it would require inspecting the mask, and then deduce if that flag should be enabled or not.  So that's the crux of my question... why does this prop exist?

So the mask controls how the value is formatted coming out of the component?  (we don't get an ISO date string? or Date object)?

davidicus:
> From graphite's point of view we'd only set the mask, which might be 24hr or not... we don't care. No one will ever set the is24hr flag since, it would require inspecting the mask, and then deduce if that flag should be enabled or not. So that's the crux of my question... why does this prop exist?
> 
> So the mask controls how the value is formatted coming out of the component? (we don't get an ISO date string? or Date object)?

Agreed,  The mask should be sufficient to dictate the state. The return value will be a date object. The mask dictates "the `dayjs.js` format for the human readable interval value" that is shown in the UI and in the Tooltip. This new prop was originally added in this PR to dictate the time format state for the `TImePickerDropdown`. My original thought was the same, that a new prop is not needed since we can deduce from the `DateTImeMask`. I did not object in case there was a use case that a user may want those to be different. Then these latest changes now have the mask overwriting the other prop, making it redundant. My question to Jordan was if there was an  actual need for them to be independent.

jessieyan:
> From graphite's point of view we'd only set the mask, which might be 24hr or not... we don't care. No one will ever set the is24hr flag since, it would require inspecting the mask, and then deduce if that flag should be enabled or not. So that's the crux of my question... why does this prop exist?
> 
> So the mask controls how the value is formatted coming out of the component? (we don't get an ISO date string? or Date object)?

I can remove the is24hours prop and only let dateTimeMask dictate whether 24 hours or 12 hours. @davidicus @sls-ca Is it the expected approach?

sls-ca:
Ok, let's remove the prop (I don't see a valid use case for it).

davidicus:
I don't think any additional props are necessary for the TimePickerDropdown. The only question is if the mask and the format need to be independent. If not, then we can use the Mask to tell us what format the TImePickerDropdown is supposed to be in and lose the `is24Hour` prop that was added to the DateTimePicker.

jessieyan:
Updated to indicate 12 hours or 24 hours by the definition of prop dateTimeMask, specifically if the hour mask is `HH`, it means 24 hours, and `hh` means 12 hours , despite of whether `dateTimeMask` has AM/PM for not.


For example, 

If dateTimeMask is YYYY-MM-DD HH:mm this is 24 hours. The field shows value like 2020-04-01 13:34. TimePickerDropDown is 24hours format like 13:34.

If dateTimeMask is YYYY-MM-DD HH:mm A this is 24 hours with AM/PM. The field shows value like 2020-04-01 13:34 PM. TimePickerDropDown is 24hours format like 13:34.

If dateTimeMask is YYYY-MM-DD hh:mm A this is 12 hours with AM/PM. The field shows value like 2020-04-01 01:34 PM . TimePickerDropDown is 24hours format like 01:34 PM.

davidicus (approved):

PR #3598, opened 2022-10-14
fix(barchartutils): condition check for categorydatasourceid

Ujjwal-UJ:
Closes Monitor - https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/4717

**Summary**

- summary_here

- [fix(barchartutils): condition check for categorydatasourceid rather than assigning undefined

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/BarChartCard/barChartUtils.js
+++ b/packages/react/src/components/BarChartCard/barChartUtils.js
@@ -242,8 +242,10 @@ export const formatChartData = (
           series.forEach((dataset) => {
             // if value is null, don't add it to the formatted chartData
             if (
-              !isNil(value[dataset.dataSourceId]) &&
-              (!isNil(value[categoryDataSourceId]) || timeDataSourceId)
+              (!isNil(value[categoryDataSourceId]) || !isNil(dataset.label)) &&
+              (timeDataSourceId && type !== BAR_CHART_TYPES.GROUPED
+                ? !isNil(value[timeDataSourceId])
+                : !isNil(value[categoryDataSourceId]))
             ) {
               data.push({
                 // if there's a dataset label, use it
--- a/packages/react/src/components/BarChartCard/barChartUtils.js
+++ b/packages/react/src/components/BarChartCard/barChartUtils.js
@@ -241,7 +241,10 @@ export const formatChartData = (
         group.forEach((value) => {
           series.forEach((dataset) => {
             // if value is null, don't add it to the formatted chartData
-            if (!isNil(value[dataset.dataSourceId])) {
+            if (
+              !isNil(value[dataset.dataSourceId]) &&
+              (!isNil(value[categoryDataSourceId]) || timeDataSourceId)
+            ) {
               data.push({
                 // if there's a dataset label, use it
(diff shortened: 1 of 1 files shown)


packages/react/src/components/BarChartCard/barChartUtils.js
```diff
@@ -241,7 +241,10 @@ export const formatChartData = (
         group.forEach((value) => {
           series.forEach((dataset) => {
             // if value is null, don't add it to the formatted chartData
-            if (!isNil(value[dataset.dataSourceId])) {
+            if (
+              !isNil(value[dataset.dataSourceId]) &&
+              (!isNil(value[categoryDataSourceId]) || timeDataSourceId)
```
herleraja, line 246: may be ? 

```
(!isNil(value[categoryDataSourceId]) || !isNil(value[timeDataSourceId]))
```

davidicus: What purpose does this second `!isNil(value[categoryDataSourceId])` in the ternary serve? This change is making this if statement look pretty complicated. Can we extract this out into a variable named in a descriptive way and use that in the if statement? ie. 
``` js
const isNilNonGroupedTimeDataSourceIdValue = timeDataSourceId && type !== BAR_CHART_TYPES.GROUPED ? !isNil(value[timeDataSourceId]) : true;
```

davidicus: Just as an FYI, this ternary is creating a new "branch", that if not checked by some test will lower the code coverage of this file. These small changes with no test are what can lead to coverage issues that are seemingly unrelated down the road.

davidicus:
Issues should outline how this bug can be reproduced in the PAL storybook if possible. The acceptance test part of this PR template should outline how we can check if this issues is resolved.

packages/react/src/components/BarChartCard/barChartUtils.js
Ujjwal-UJ: > What purpose does this second `!isNil(value[categoryDataSourceId])` in the ternary serve? This change is making this if statement look pretty complicated. Can we extract this out into a variable named in a descriptive way and use that in the if statement? ie.
> 
> ```js
> const nilCheckNonGroupedTimeDataSourceIdValue = timeDataSourceId && type !== BAR_CHART_TYPES.GROUPED ? !isNil(value[timeDataSourceId]) : true; // false if we want the check to fail without this
> ```

For Non-timeseries: if the data does not have value[categoryDataSourceId], it is assigning undefined and does not loads grouped bar chart at all. So `!isNil(value[categoryDataSourceId])` checks before assigning.
Also i will assign it to a variable as you suggested and add test cases too.

davidicus:
@Ujjwal-UJ How is this PR coming? Are we still waiting on test cases? It looks as if the condition were were discussing has changed but I do not see any other commits?

Ujjwal-UJ:
> @Ujjwal-UJ How is this PR coming? Are we still waiting on test cases? It looks as if the condition were were discussing has changed but I do not see any other commits?

@davidicus I was working on other issue, test cases are pending! will update that shortly

davidicus (approved):

PR #3608, opened 2022-10-18
fix(datetimepickerv2): flyout menu auto positioning

### LINKED ISSUE #3362: [DateTimePickerV2] Flyout scrolls with content when inside a sub-container
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

When the DateTimePickerV2 is rendered inside a scrollable area (inside of the body) it'll scroll along with the content instead of staying fixed in place. This is because DateTimePickerV2 is always rendered in a portal, and positioned absolutely relative to the body. However, when it's in a section that scrolls this positioning is no longer correct. We should expose the `renderInPortal` prop used on the `FlyoutMenu` within the DateTimePickerV2 to give consumers the option of rendering it in the portal or not. Additionally, this prop will need to be added to Cards so that they can pass it to the `CardToolbar` and eventually to the `DateTimePicker` for instances of cards having a built-in DateTimePicker.

### To Reproduce

This type of example will reproduce it. 
```jsx

    <div style={{ backgroundColor: 'blue' }}>
      <div
        style={{
          height: '20rem',
          paddingTop: '10rem',
          paddingBottom: '1000rem',
          width: 400,
          backgroundColor: 'red',
          overflow: 'scroll',
        }}
      >
        <DateTimePicker
          id="datetimepicker"
          dateTimeMask={text('dateTimeMask', 'YYYY-MM-DD HH:mm')}
          relatives={[
            {
              label: 'Yesterday',
              value: RELATIVE_VALUES.YESTERDAY,
            },
          ]}
          hasTimeInput={boolean('hasTimeInput', true)}
          onApply={action('onApply')}
          onCancel={action('onCancel')}
          hasIconOnly={boolean('hasIconOnly', false)}
        />
      </div>
    </div>
```

### Expected behavior

The expected behavior is for the picker to stay fixed in the correct position beside its trigger button and not scroll along with the content.

### Environment/versions:

- OS: [e.g. MacOS, Windows]

### Additional context

datetimepickerv2

upcoming release?

date, the better!_

### LINKED ISSUE #3618: [DateTimePicker(V2)] use prop renderInPortal and useAutoPositioning
### What package is this for?

- [x] React
- [ ] Angular

### Summary

`useAutoPositioning` offers when flyout menu offscreen we will adjust the direction of the menu automatically.
`renderInPortal` offers a way whether to render the flyout menu is a portal or not

request here.

upcoming release?

date, the better!_

### Available extra resources

jessieyan:
Closes #3618 

**Summary**

- expose `renderInPortal` prop so that user can set it to false for some scenarios
- expose `useAutoPositioning` prop to reposition flyout menu if offscreen

- added `renderInPortal` prop in DateTimePickerV2, Card, TableCard, CardToolBar
- added `useAutoPositioning` prop in DateTimePickerV2
- fixed the menuOffSet when useAutoPositioning is enabled and in RTL. The passed-in `menuOffset` prop only works for the passed-in `direction` prop. If repositioned, the `direction` will be changed, `menuOffset` no long apply to the new `direction`
before:
image
after:
image

- fixed the menuOffSet.left. set it to 288px in RTL
Before,
image

After:
image

- added tests for the scenarios when using `useAutoPositioning` and `renderInPortal`
- added a knob `useAutoPositioning` in single select and range select stories.

- go to single select [story](https://deploy-preview-3608--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select)
- deselect padding on story-container 
- add margin-left to -1rem or so to make it offscreen on the left
- click to open flyout menu, make sure flyout menu are with in screen
image

image

- go to range select or single select [story](https://deploy-preview-3608--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--selected-absolute-with-new-time-spinner)
- set renderInPortal to false, we should see `flyout-tooltip` is grandchild of `date-time-picker-datepicker-flyout-container`
image

- set renderInPortal to true,  `flyout-tooltip` is independent element outside of `date-time-picker-datepicker-flyout-container`
image


davidicus:
@jessieyan This usually needs to have it rendered in a portal, in order to break out of an overflow:hidden container, like our cards we need to be able to do so and still have the dropdown stay with the trigger. We can either close the portal on scroll or adjust the offset so it scrolls with the element. We have components in the codebase that manipulate the offset that you can try to look for.

davidicus (changes requested):
This is not a solution to the issue that is listed for this PR

jessieyan:
> @jessieyan This usually needs to have it rendered in a portal, in order to break out of an overflow:hidden container, like our cards we need to be able to do so and still have the dropdown stay with the trigger. We can either close the portal on scroll or adjust the offset so it scrolls with the element. We have components in the codebase that manipulate the offset that you can try to look for.

@davidicus  Do you mean the offered solution in [issue 3362](https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3362) is not valid? This is the change that requested in the description of 3362 that we need to expose renderInPortal prop. It does not solve the menu offset issue though.

davidicus:
We can surface a prop to allow people to render in portal or not but we need to have the dropdown not detach even when it is rendered in portal.

jessieyan:
@davidicus Change this PR to only resolve the issue #3618 which is to enable the auto positioning

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
```diff
@@ -1967,4 +1979,501 @@ describe('DateTimePickerV2', () => {
...
+  it('when flyout menu offscreen top left render direction to  RightStart', () => {
+    const originalGetBoundingClientRect = Element.prototype.getBoundingClientRect;
+    const mockGetBoundingClientRect = jest.fn(
+      generateBoundingClientRect({
+        x: -200,
+        y: -200,
+        height: 134,
+        width: 278,
+      })
+    );
```
davidicus, line 2042: I think the [spec](https://jestjs.io/docs/25.x/jest-object#jestfnimplementation) suggest the following for mock implementation. 
```suggestion
    const mockGetBoundingClientRect = jest.fn(generateBoundingClientRect);
```

jessieyan:
@davidicus I saw you left a review comment earlier but I couldn't find it in the PR.

davidicus:
> @davidicus I saw you left a review comment earlier but I couldn't find it in the PR.

I had deleted the comment after I saw you were returning a function. Was supposed to be part of the review.

davidicus (changes requested):

davidicus:
Issue with Card component and the DateTimePicker in [this](https://deploy-preview-3608--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card-card--with-date-time-picker-range-selector) story

This PR            |  Next
image | image

packages/react/src/components/Card/Card.jsx
```diff
@@ -215,6 +215,7 @@ export const defaultProps = {
     extra: false,
   },
   renderExpandIcon: undefined,
+  renderInPortal: false,
```
davidicus, line 218: Should this default be true? I think the DateTimePickerV2 was built specifically to render in a portal for the Cards. This is most likely why the story is broken

jessieyan:
> https://deploy-preview-3608--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select

I can see if set the padding-left to 80 rem. Then a horizontal bar showed up. The flyout menu will adjust its position if the horizontal bar moves.

packages/react/src/components/Card/CardToolbar.jsx
```diff
@@ -108,6 +110,7 @@ const defaultProps = {
   },
   testId: 'card-toolbar',
   dateTimeMask: 'YYYY-MM-DD HH:mm',
+  renderDateDropdownInPortal: false,
```
davidicus, line 113: I imagine this gets overwritten by Card default prop but should probably align

packages/react/src/constants/CardPropTypes.js
```diff
@@ -971,4 +971,6 @@ export const CardPropTypes = {
   overrides: PropTypes.shape({
     errorMessage: OverridePropTypes,
   }),
+  /** If set to true it will render outside of the current DOM in a portal, otherwise render as a child */
+  renderInPortal: PropTypes.bool,
```
davidicus, line 975: also need to rename the prop here.

davidicus (approved):
We can address some of the positioning issues in another PR.

jessieyan:
> I can see if set the padding-left to 80 rem. Then a horizontal bar showed up. The flyout menu will adjust its position if the horizontal bar moves. If set padding-left to 71 rem, The flyout menu will pop up on top of the input and will not move its position.

Opened #3620

PR #3611, opened 2022-10-19
fix(TableHead): add id prop on TableExpandHeader

### LINKED ISSUE #3485: [Table] [A11y] headers="expand" should be removed from first column <td>
### What package is this for?

- [X] React
- [?] Angular

### Describe the bug

The 'headers' attribute with value "expand" does not reference a valid 'id' in the document

### Expected behavior

The <td> tags for the first column (row expansion) in a table contain the property `header="expand"` which is supposed to point to the id of a <th> cell, but none of the <th> elements have ids (let alone something so generic). The header property should be removed.

### Environment/versions:

- `carbon-addons-iot-react` version: 2.150.0

### Additional context

Screen Shot 2022-06-23 at 3 16 44 PM

jessieyan:
Closes #3485 

**Summary**

- fix the a11y violation by adding id prop on `TableExpandHeader`

According to the rule [H43: Using id and headers attributes to associate data cells with header cells in data tables](https://www.w3.org/TR/WCAG20-TECHS/H43) 

Header should use `id` prop for the any table cell to reference with. The default header for the first cell in `TableExpandRow` is set to `headers='expand'`. But we did not set an `id` on `TableExpandHeader` for the cell to refer to. 

The solution is to set ` id={`${tableId}-expand`} `on `TableExpandHeader` and `expandHeader={`${tableId}-expand`}` on `TableExpandRow`:

-fix(TableBodyRow): add id prop on TableExpandHeader

- go to table story [with row expansion](https://deploy-preview-3611--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-row-expansion)
- scan with IBM Equal Access Accessibility
- check no errors shows like `'The 'headers' attribute value "expand" does not reference a valid 'id' in this document'`

before:
image

after:
image

- tests here

### DIFF
--- a/packages/react/src/components/PageTitleBar/__snapshots__/PageTitleBar.story.storyshot
+++ b/packages/react/src/components/PageTitleBar/__snapshots__/PageTitleBar.story.storyshot
@@ -398,7 +398,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/P
                           <th
                             className="bx--table-expand"
                             data-testid="null-table-head-row-expansion-column"
-                            id="expand"
+                            id="table-12-expand"
                             scope="col"
                           />
                           <th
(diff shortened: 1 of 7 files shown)


jessieyan:
Brought it to scrum today. We will use `${tableId}-expand` for the expandHeader on TableExpandRow and id on TableExpandHeader.

davidicus (approved):

PR #3617, opened 2022-10-25
fix(table): callback onRowSelected if radio buttons exist

### LINKED ISSUE #3616: [Table] callback issues when single select with Radio buttons props are set
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

There are two defects surrounding the single select behavior with radio buttons.

1. On line 252 of `TableBodyRow.jsx` we are calling `onRowSelected` inside an `onChange` event for the `TableCell`. We are also calling the `onRowSelected` when the `RadioButto` element is clicked on line 265. We should only be calling this once. 
You can see this in our storybook [here](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table--playground). Make sure `options.hasRowSelection` is set to 'single' and the `options.useRadioButtonSingleSelect` options is clicked in the 'Selections and actions' knob tab. Then open the actions panel and click on the radio button. You will see the callback fired twice. 
2. With the same story and knobs active try clicking on the row in an area outside of the radio button cell with the actions tab open. You will see that an `onRowSelected` and an `onRowClicked` callback are fired. Notice that the radio button does not show a selected state. This most likely stems to when we added the radio buttons for single selection. Previously, a user could click anywhere and it would select the row. @JordanWSmith15 @dianatran18 can you chime in with whether we should continue to allow the clicking on a row to fire the `onRowSelected` and update the radio button accordingly or  disallow row clicks to fire the `onRowSelected` when we are using radio buttons? 

@JordanWSmith15 This is related to the defect being worked by Tom for Graphite. 

### To Reproduce

1. Go to [this](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-table--playground) story
4. Got to the 'Selections and actions' knob tab in the knob pannel.
5. Make sure `options.hasRowSelection` is set to 'single' and the `options.useRadioButtonSingleSelect` options is clicked
6. Open the actions panel
7. Click on the radio button in the table
8. See the action firing twice.
9. Click anywhere in the row and see the `onRowSelected` callback being fired. 

We will want to make sure that our behavior for single select without the radio buttons and the multi-select functionality is preserved.

erzhan-temir-mamyrov:
Closes #3616 

**Summary**

- Resolve `onRowSelected` callback issues

- Remove duplication `onRowSelected` callback when radio buttons present
- Disable row selection on clicking the row outside of radio button (_note:_ this applies only with radio buttons)
- Add background for row selected state
- Update test

- Go to this story
- Select `hasRowSelection` to `single`
- Enable `useRadioButtonSingleSelect`
- Open `Actions` tab
- Click on some radio button
- Verify `background-color` change for selected row
- Verify that `onRowSelected` has been called only once
- _Note:_ Verify that `onRowClicked` has NOT been called

</br>

- In the same story
- Select `hasRowSelection` to `single`
- Disable `useRadioButtonSingleSelect`
- Open `Actions` tab
- Click on some row
- Verify `background-color` change for selected row
- Verify that `onRowSelected` has been called only once
- _Note:_ Verify that `onRowClicked` has been called once

- tests here

### DIFF
--- a/packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
+++ b/packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
@@ -452,6 +452,7 @@ const TableBodyRow = ({
     <TableRow
       className={classnames(`${iotPrefix}--table__row`, {
         [`${iotPrefix}--table__row--singly-selected`]: isSelected && !useRadioButtonSingleSelect,
+        [`${iotPrefix}--table__row--background`]: isSelected,
       })}
       key={id}
       onClick={() => {
--- a/packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
+++ b/packages/react/src/components/Table/TableBody/TableBodyRow/TableBodyRow.jsx
@@ -249,7 +249,6 @@ const TableBodyRow = ({
       <TableCell
         className={`${prefix}--radiobutton-table-cell`}
         key={`${id}-row-selection-cell`}
-        onChange={isSelectable !== false ? () => onRowSelected(id, !isSelected) : null}
         onClick={(e) => e.stopPropagation()}
       >
         <span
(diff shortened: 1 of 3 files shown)


davidicus (approved):
Great job. Thanks for the fast turnaround.

PR #3627, opened 2022-11-02
fix(FlyoutMenu): Flyout menu position

### LINKED ISSUE #3362: [DateTimePickerV2] Flyout scrolls with content when inside a sub-container

jessieyan:
Closes #3362 #3633 

**Summary**

FlyoutMenu does not move when parent is scrolling. 

- added css attribute `--top` and `--left` so that FlyoutMenu can move with the parent.
- temporarily updated the single story and absolute story to test. I will change it back later.
- fixed style PropTypes

- go to single select [story](https://deploy-preview-3627--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select)
- click on the Calendar button to open flyout menu
- scroll up and down in the red container
- scroll left and right in the red container
- notice the flyout menu is moving with the datetime input

https://user-images.githubusercontent.com/24279794/199554269-aee7c213-1872-44ea-8fa9-a335c309857d.mov

- tests here


jessieyan:
Fixing cypress test

davidicus (changes requested):
I am unable to test this. I think the story was reverted back because now there is no scrolling possible.

jessieyan:
Updated the single select story for testing.

jessieyan:
Netlify is not updating the deploy build for some reason.

davidicus:
@jessieyan I manually updated the build. There are still issues. I think if you simplify by taking out the internal red div and adding your height to the outer div you will see the issue. 


https://user-images.githubusercontent.com/1590966/204576600-19fbc60a-c4b2-400d-a094-d15fbc004e71.mov

davidicus:
I am also seeing some weirdness with the positioning in [our storybook](https://next.carbon-addons-iot-react.com/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select). Can we try to address this altogether?
image

jessieyan:
Hi @davidicus, I want to consult this with you.

When rendering `DateTimePickerV2`, the FlyoutMenu always render TimePickerDropDown with the Footer first and after a few more re-rendering, the `DatePicker` got rendered. Do you know what might cause the `DatePicker` rendered so late.

This is an issue causes the calculation of the height to be less than it should be. 

Screen shot of it first renders the time picker 
image

After a few re-renders, it renders the date picker
image

packages/react/src/components/DateTimePicker/DateTimePickerV2WithTimeSpinner.jsx
```diff
@@ -838,8 +850,27 @@ const DateTimePicker = ({
...
+    setTop(event.target.scrollTop);
+  };
+
+  // add scroll event listener to parent nodes
+  useEffect(() => {
+    let currentNode = containerRef.current?.parentNode;
+    while (currentNode) {
+      currentNode.addEventListener('scroll', getPosition);
+      currentNode = currentNode.parentNode;
+    }
```
davidicus, line 869: we need to return a function that removes these event listeners.

packages/react/src/components/DateTimePicker/DateTimePickerV2WithoutTimeSpinner.jsx
```diff
@@ -692,8 +702,21 @@ const DateTimePicker = ({
...
+  const getPosition = () => {
+    setLeft(containerRef.current?.parentNode.scrollLeft);
+    setTop(containerRef.current?.parentNode.scrollTop);
+  };
+
+  // Re-calculate X and Y when the parent is scroll by the user
+  useEffect(() => {
+    const currentNode = containerRef.current?.parentNode;
+    if (currentNode) {
+      currentNode.addEventListener('scroll', getPosition);
```
davidicus, line 714: same comment about cleaning up our listeners

davidicus (changes requested):
@jessieyan Looking good. Just the comments about cleaning up our event listeners.

davidicus (approved):
should be good to merge once checks pass

jessieyan:
I also tested with DateTimePicker without new time spinner. It should also work as expected.

PR #3635, opened 2022-11-07
fix(tabletoolbar): add aria-label to search input

### LINKED ISSUE #3400: [Table] Accessibility violation
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

Accessibility violation is found when `Table` has search enabled in the toolbar. The `input` field needs `aria-label` to be set.

### To Reproduce
 Though I am not sure why PAL's next storybook isn't showing this accessibility violation, but Maximo Graphite which uses the `Table` component does throw this in its storybook's accessibility tab.
image

Adding aria-label to the `input` seems to fix the accessibility violation. 
https://dequeuniversity.com/rules/axe/4.3/label-title-only
I think this needs to be done at carbon(carbon-components-react) end, which has the base component.

### Expected behavior

Aria-label is set in the `input` of search and Graphite storybook should show no _'Ensures every form element has a label'_ accessibility violations

### Environment/versions:

- OS: [e.g. MacOS, Windows]
- `carbon-addons-iot-react` version: 2.149.0-next.10

erzhan-temir-mamyrov:
Closes #3400 

**Summary**

- Add aria label to search input in table toolbar

- Add `aria-label="search"` to search input
- Update storyshots

- Go to [this story](https://deploy-preview-3635--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-searching)
- Inspect search input in the browser tool
- Verify that `aria-label` is set for the input field

- tests here


packages/react/src/components/Table/TableToolbar/TableToolbar.jsx
```diff
@@ -433,6 +433,7 @@ const TableToolbar = ({
           {hasSearch ? (
             <TableToolbarSearch
               {...search}
+              aria-label="search"
```
davidicus, line 436: Since this will be read by screen readers we should be using the translated string here.

davidicus (changes requested):
Small change with translated string.

davidicus:
Hey @erzhan-temir-mamyrov I have looked into this and I think this may be a false positive for Graphite's axe scans. There is a [known bug](https://github.com/dequelabs/axe-core/issues/1372) that was [fixed](https://github.com/dequelabs/axe-core/pull/1187). If you look at the scans in our Storybook you will  notice we do not have an a11y issue. Our version is ahead of graphite but quite a bit. `4.2.0` to Graphite's `3.5.6`

If you look at the [spec](https://www.w3.org/WAI/tutorials/forms/labels/#hiding-label-text), Carbon is following the recommendations of providing a visibly hidden label. 

So, I think this one can be closed out with no need of changes on our end.

PR #3677, opened 2022-12-15
feat(table): add option to pin column

### LINKED ISSUE #3636: [Table] Row actions/overflow menu should always be visible
### What package is this for?

- [x] React
- [ ] Angular

### Summary

When a table scrolls horizontally the last row actions column is not visible. There's no way for the user to act on the row without scrolling all the way right each time.

Row actions should probably be a sticky column that's always visible. Additionally, 

Additionally, there should be an option for the action column to take up as much table space as it needs. Currently, it always overlaps the col to the left. When there is just an overflow menu, that's OK, but if there are other inline actions, then those actions obscure the row content when visible.  Ideally the row actions would be "sticky", but if you scroll all the way to the end they would take up their own column so they don't overlap anymore. This feels a little like two requests, but they are both in service of working with many columns on a small display. 

See tags column obscured by row actions: 
image

request here.

This is partially achievable with extra CSS today: 

```
.bx--data-table.bx--data-table .iot--row-actions-cell--table-cell {
  position: sticky;
  right: 0;
}
```

That keeps it visible. It still overlaps content when scroll all the way to the right though. I tried adding width to the last column with CSS, but that broke column resizing.

erzhan-temir-mamyrov:
Closes #3636

**Summary**

- Add new prop to `options` to pin first or last column

- Add new css classes to freeze column position
- Setup with different properties, such as row selection and nesting
- Create new story in Storybook
- Update public API
- Add new tests

- _With row selection_
- Go to [Table / With pinned column](https://deploy-preview-3677--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-pinned-column)
- Resize window horizontally so that scrollbar appears
- Using horizontal scrollbar in table scroll to the right
- Verify that first column is pinned
- In the knobs area set `pinColumn: last`
- Verify that last column is pinned

</br>

- _With row nesting_
- Go to [Table / Playground](https://deploy-preview-3677--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--playground)
- In the `Nesting & expansion` set `hasRowExpansion: true` and `hasRowNesting: true`
- In the `Selection & actions` set `hasRowSelection: false`
- In the `Column configuration` set `pinColumn: first`
- Resize window horizontally so that scrollbar appears
- Using horizontal scrollbar in table scroll to the right
- Verify that first column is pinned
- _Latest update:_ If in `Data rendering` knob `wrapCellText: always` then overflow content will be hidden (For example, 5th row)

- tests here


bharathraomairpady:
Any estimate on when this item will be merged?

erzhan-temir-mamyrov:
@bharathraomairpady This item is still WIP. I will work on it during this sprint.

packages/react/src/components/Card/__snapshots__/Card.story.storyshot
```diff
@@ -1457,6 +1457,7 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/C
                 </thead>
                 <tbody
                   aria-live="polite"
+                  className=""
```
davidicus, line 1460: Not a big deal but it looks like we are applying a class attribute even if one is not provided.

davidicus: This is probably why we have so many snapshot changes

packages/react/src/components/Table/TableBody/TableBody.jsx
```diff
@@ -1,16 +1,18 @@
 import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { TableBody as CarbonTableBody } from 'carbon-components-react';
+import classNames from 'classnames';
```
davidicus, line 4: Since this gets imported so many different ways we have set the convention for this import to be `classnames`.

davidicus (changes requested):
Looks good, just a few comments.

packages/react/src/components/Card/__snapshots__/Card.story.storyshot
erzhan-temir-mamyrov: @davidicus Thanks for the note! Yes, that was the reason why snapshots were updated. I added conditional props to avoid it.

packages/react/src/components/Table/mdx/Table.mdx
```diff
@@ -533,6 +538,7 @@ import TableViewDropdownContent from '../TableViewDropdown/TableViewDropdownCont
...
+| pinColumn                   | 'none', 'first', 'last' .                                | 'none'   | Freezes selected column regardless of horizontal scrollbar position                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
sls-ca, line 541: Does that imply that I need to "select" the column in order for it to be pinned?   Or should this message be...
Can freeze the 'first' or 'last' column regardless of horizontal scroll position.  Default is 'none'.

And does this work in RTL mode as well?

sls-ca (approved):
Looks good just a comment on usage.

packages/react/src/components/Table/mdx/Table.mdx
erzhan-temir-mamyrov: Thanks for the note! I updated the doc comment.
Yes, it works in RTL mode as well.

jessieyan (approved):
LGTM

davidicus (approved):

PR #3681, opened 2022-12-21
fix(dataseriesformsettings): Set empty strings for missing values

### LINKED ISSUE #3680: [DataSeriesFormSettings] null and undefined values do not reset TextInput values
### What package is this for?

- [X] React
- [ ] Angular

### Describe the bug

TextInput values don't get cleared for parameters not present in new configuration.

### To Reproduce

1. Go to DataSeriesFormSettings
2. Set values to the following attributes/props
      **decimalPrecision**: PropTypes.number,
      **xLabel**: PropTypes.string,
      **yLabel**: PropTypes.string,
      **unit**: PropTypes.string
      **maximumDataPointsLabel**: PropTypes.string
3. Render component.
4. Re-render component with only **xLabel**
5. Inputs associated with missing attributes keep previous value.

### Expected behavior

It is expected to clear input value when attribute is absent from configuration.

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: v2.152.0-next.13

upcoming release?

date, the better!_

vaffonso:
Closes #3680

**Summary**

- Set empty string in case of null or undefined values

- DataSeriesFormSettings: use nullish coalescing operator to set empty strings

- Render component with all attributes from content filled
- Re-render component with some of them missing

- tests here

### DIFF
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormSettings.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormSettings.jsx
@@ -82,7 +82,7 @@ const DataSeriesFormSettings = ({ cardConfig, onChange, i18n }) => {
               content: { ...cardConfig.content, xLabel: evt.target.value },
             })
           }
-          value={content?.xLabel}
+          value={content?.xLabel ?? ''}
         />
       </div>
       <div className={`${baseClassName}--input`}>
@@ -96,7 +96,7 @@ const DataSeriesFormSettings = ({ cardConfig, onChange, i18n }) => {
               content: { ...cardConfig.content, yLabel: evt.target.value },
             })
           }
-          value={content?.yLabel}
+          value={content?.yLabel ?? ''}
         />
       </div>
       <div className={`${baseClassName}--input`}>
@@ -110,7 +110,7 @@ const DataSeriesFormSettings = ({ cardConfig, onChange, i18n }) => {
               content: { ...cardConfig.content, unit: evt.target.value },
             })
           }
-          value={content?.unit}
+          value={content?.unit ?? ''}
         />
       </div>
       <div className={`${baseClassName}--input`}>
@@ -127,7 +127,7 @@ const DataSeriesFormSettings = ({ cardConfig, onChange, i18n }) => {
               },
             })
           }
-          value={content?.decimalPrecision}
+          value={content?.decimalPrecision ?? ''}
         />
       </div>
       <div className={`${baseClassName}--input`}>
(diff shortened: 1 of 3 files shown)


vaffonso:
@herleraja @EdmondCShaffer could you take a look?

JordanWSmith15:
@davidicus FYI

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormSettings.jsx
```diff
@@ -144,7 +144,7 @@ const DataSeriesFormSettings = ({ cardConfig, onChange, i18n }) => {
               content: { ...cardConfig.content, maximumDataPoints },
             });
           }}
-          value={content?.maximumDataPoints}
+          value={content?.maximumDataPoints ?? ''}
```
herleraja, line 147: its numeric value. Is there no other way we can fix this issue ?

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItems/DataSeriesFormSettings.jsx
```diff
@@ -127,7 +127,7 @@ const DataSeriesFormSettings = ({ cardConfig, onChange, i18n }) => {
               },
             })
           }
-          value={content?.decimalPrecision}
+          value={content?.decimalPrecision ?? ''}
```
herleraja, line 130: I think this is also numeric value.  May be good to add type="number"

herleraja (commented):
It's fine to have an empty string literal for labels, but for numeric fields, its better be null or undefined.

davidicus:
Doing the steps outlined in this issue on this [story](https://next.carbon-addons-iot-react.com/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-dashboardeditor--with-initial-value), it seems like the fields get updated.

davidicus (changes requested):
Just the comment on the number inputs.

vaffonso:
Hi @davidicus @herleraja 
I was able to reproduce the issue in this [story](https://next.carbon-addons-iot-react.com/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-dashboardeditor--with-initial-value) following these steps:
- Select timeseries card
- Switch to Settings tab
- Note that Unit, Decimal Precision and Maximum Data Points are empty
- Add new card
- Select Time Series Line
- Give it a title Test
- Switch to Settings tab
- Fill Unit, Decimal Precision and Maximum Data Points with value
- Select timeseries card again
- Note that Unit, Decimal Precision and Maximum data points value are not reset

I have commented in the original defect:

> I see 3 possible approaches:
> 1. Provide empty strings for Data Series Form Settings attributes - Monitor UI
> 2. Check null and undefined values and replace with empty strings - PAL lib (solution tested)
> 3. Apply check for null/undefined check in TextInput - Carbon lib

Should we opt for the 1st option?

davidicus:
@vaffonso can we not provide a null value for these fields?

vaffonso:
I pushed a couple changes to this PR. Instead of empty strings, number inputs from DataSeriesFormSettings receive key prop to force a render.
In addition, field **decimal precision** was changed to number.

davidicus (approved):
LGTM 👍

Issue #3682, opened 2022-12-23
[Card] tooltip is offscreen for small cards


erzhan-temir-mamyrov:
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

If `Card` component is small and aligned to the left of the page then tooltip is not positioned correctly.

### To Reproduce

1. Go to [this story](https://next.carbon-addons-iot-react.com/?path=/story/1-watson-iot-card-card--with-ellipsed-title-tooltip-external-tooltip)
2. In the knobs area set size to `SMALL`
3. Disable padding using browser devtools 
4. Click on card title
5. See tooltip offscreen

### Expected behavior

Tooltip should auto adjust its alignment to `start` or `end`.

### Environment/versions:

- OS: MacOS
- Browser: chrome
- `carbon-addons-iot-react` version: `v2.152.0-next.13`

### Additional context

<img width="639" alt="Screenshot 2022-12-23 at 11 26 30" src="https://user-images.githubusercontent.com/112560065/209319744-78f2522a-7595-4815-a26f-516404b03380.png">

Issue #3278, opened 2022-02-10
[Table] update minimum TablePagination width to hide text


kevinsperrine:
[X] React

Summary: 

I did find an issue related to the pagination size. It seems that the breakpoint for when to activate `--pagination-text-display:none;` should be larger:

![image](https://user-images.githubusercontent.com/5167091/153480884-a06565ba-727c-453b-b640-adca8fd2136d.png)

Since this is not related to your change perhaps we create a new issue for this?

_Originally posted by @bjornalm in https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3277#pullrequestreview-879338111_

PR #3690, opened 2023-01-10
fix(table): pagination, toolbar toolips; (card): tooltip

erzhan-temir-mamyrov:
**Summary**

Changed approach to previous issues:
- #3682 Card tooltip - add throttle to scroll event listener (initial render of a tooltip caused micro-scrolling event, therefore tooltip closed itself, throttle for 10ms prevents this)
- #3278 Pagination in table - decrease padding slightly between `608px` and `500px` instead of hiding items per page block
- Carbon tooltips in table - change direction from `top` to `bottom` in table toolbar (due to `overflow: hidden`)

- Card tooltip: add `setTimeout` to add scroll event listener
- Table pagination: update css styles
- Tooltips in table toolbar: add props to change direction to `bottom`

**Card tooltip**
- Go to [this story](https://deploy-preview-3690--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card-card--with-ellipsed-title-tooltip-external-tooltip)
- Resize the window vertically in order to display vertical scrollbar
- Click on card or subtitle to show tooltip
- Scroll windows vertically
- Verify that tooltip was removed after the scrolling event

**Table pagination**
- Go to [this story](https://deploy-preview-3690--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-pagination)
- Resize the window horizontally till `500px`
- Verify that icon buttons to change pages do not overflow container

**Table toolbar tooltips**
- Go to [this story](https://deploy-preview-3690--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--playground)
- Set `hasFilter: true`, `hasSearch: true`, `hasAggregations: true` or any other button icons in tooltip
- Hover over icons in toolbar
- Verify that tooltips have direction `bottom`

- tests here

### DIFF
--- a/packages/react/src/components/Card/CardTitle.jsx
+++ b/packages/react/src/components/Card/CardTitle.jsx
@@ -2,6 +2,7 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import { Tooltip } from 'carbon-components-react';
+import throttle from 'lodash-es/throttle';
 
 import useHasTextOverflow from '../../hooks/useHasTextOverflow';
 import { usePopoverPositioning } from '../../hooks/usePopoverPositioning';
@@ -10,6 +11,11 @@ import { settings } from '../../constants/Settings';
 
 const { iotPrefix } = settings;
 
+const SCROLL_THROTTLE_TIMEOUT = 10;
+const SCROLL_THROTTLE_OPTIONS = {
+  leading: false,
+  trailing: false,
+};
 const isEngagingEventType = (type) => type === 'click' || type === 'keydown';
 
 const propTypes = {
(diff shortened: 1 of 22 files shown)


packages/react/src/components/Card/CardTitle.jsx
```diff
@@ -101,32 +107,81 @@ export const CardTitle = (
...
-    window.addEventListener('scroll', handleScroll);
-
-    return () => {
-      window.removeEventListener('scroll', handleScroll);
-    };
-  }, [tooltipsState]);
+    if (tooltipsState.title) {
+      window.addEventListener('scroll', handleTitleScroll);
+    } else {
+      window.removeEventListener('scroll', handleTitleScroll);
```
davidicus, line 161: We need to return a function that removes these listeners from the useEffect

packages/react/src/components/Card/CardTitle.jsx
```diff
@@ -101,32 +107,81 @@ export const CardTitle = (
...
+    } else {
+      window.removeEventListener('scroll', handleSubtitleScroll);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [tooltipsState.subtitle]);
+
+  /* istanbul ignore next */
+  useEffect(() => {
+    if (tooltipsState.info) {
+      window.addEventListener('scroll', handleInfoScroll);
```
davidicus, line 179: same comment about cleaning this up

davidicus (commented):
This code looks good and 2 of the 3 acceptance test worked as expected. The first one however does not close on scroll. The tooltip seems to stay with the content it is attached to so that seems acceptable but is this what the desired effect was supposed to be?

davidicus:
This is the acceptance test that does not work as stated. 

**Card tooltip**
- Go to [this story](https://deploy-preview-3690--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card-card--with-ellipsed-title-tooltip-external-tooltip)
- Resize the window vertically in order to display vertical scrollbar
- Click on card or subtitle to show tooltip
- Scroll windows vertically
- Verify that tooltip was removed after the scrolling event

erzhan-temir-mamyrov:
@davidicus Thanks for commenting on this! I replaced throttling with setTimeout, it seems to do the trick. Also, I added cleanup functions for timers and event listeners.

davidicus (approved):

PR #3697, opened 2023-01-17
fix(DateTimePickerV2): show invalid message when time fields are empty

### LINKED ISSUE #3691: [DateTimePickerv2] show error message when time is not selected
### What package is this for?

- [x] React
- [ ] Angular

### Describe the bug

We removed the default time value 00:00 for DateTimePIckerV2 but when time was empty or not picked. There was no clear error message to indicate that time is not selected.

### To Reproduce

image

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

jessieyan:
Closes #3691 

**Summary**

-when switching from relative to absolute, the time picker fields are empty without any warning message. Apply button is disabled. We should enable warning message to let user know that time fields are empty and they should fill in time value before they can click on apply button. 

- DateTimePickerV2WithTimeSpinner.jsx
- DateTimePickerV2WithoutTimeSpinner.jsx

Make sure warning message show up when time fields are empty
- go to [Default](https://deploy-preview-3697--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--default) story
- click on calendar icon to open drop down preset list
- click on custom range
- select Absolute radio button
- check the time fields are showing warning and apply button is disabled

https://user-images.githubusercontent.com/24279794/213009333-5475336e-3998-48c2-a993-efc1b2a598de.mov

- tests here


davidicus:
@JordanWSmith15 Didn't we just have a story in Graphite because this was the behavior?

JordanWSmith15:
@davidicus  I'm honestly not sure what the correct pattern is for this. I'll let @dianatran18 weigh in here. In this scenario, they just haven't gotten to the time input yet. Should we really he showing error messages? I am not a big fan of showing errors to the user unless they did something to cause the error. In Graphite, we have required indicators on the fields, and disable the save button. However, I know required indicators are not a pattern in Carbon.

jessieyan:
@JordanWSmith15 @dianatran18 @davidicus 

Would you please confirm here are the changes:

1. Apply button is aways enabled, we only validate date/time input after user click on the apply button ( like Modal)

2. After click on apply button and if time are not selected, we show invalid message on the time picker.  If Date not selected, we need to show some warning message somewhere as well?

3. Show date/time mask if date/time not selected. 
If no date and no time, we show text like YYYY-MM-DD HH:mm to YYYY-MM-DD HH:mm. 
If no date but have time, show text like YYYY-MM-DD 12:00 to YYYY-MM-DD 13:00.. 
If has date but not time, show text like 2023-01-19 HH:mm to 2023-01-19 HH:mm

4. We do not show invalid state on the top input field if time or date not selected.  The top input should be controlled by user.

dianatran18:
Hey @jessieyan @JordanWSmith15 @davidicus I've left comments based on what Jessie summarized, let me know if this makes sense. 

> Would you please confirm the changes:
> 
> 1. Apply button is aways enabled, we only validate date/time input after user click on the apply button ( like in Modal)

Correct. For context — we should avoid disabling the Apply button because that could lead the user to getting stuck in the flow, and not know how to make that disabled button enabled. If the user selects a date and hits Apply before setting a time, then display the error validation on the time inputs to let the user know that is a required step.


> 2. After click on apply button and if time not selected, we show invalid message on the time picker.  If Date not selected, we need to show some message somewhere as well?

Yes, I would agree with that. I don't think there is a current pattern for displaying an error on the calendar like that outside of the input itself, so that will have to be something custom.

> 3. Show date/time mask on the top input if date/time not selected
>    If no date and no time, we show text like YYYY-MM-DD HH:mm to YYYY-MM-DD HH:mm.
>    If no date but have time, show text like YYYY-MM-DD 12:00 to YYYY-MM-DD 13:00..
>    If has date but not time, show text like 2023-01-19 HH:mm to 2023-01-19 HH:mm

Let me confirm with Robin to see what the best text formats should be here. 

> 4. We do not show invalid state on the top input field if time or date not selected.  The top input should be controlled by user.

The invalid state should show up after the dropdown has closed so the user can see what the error message is. Before closing the dropdown the invalid states should be contained within the dropdown for the date or time selection. For example, if the user selected a date and time range, and then hit Apply, there could be a top input error that states that the date and time range is invalid. Though, playing around with the [Carbon date picker](https://react.carbondesignsystem.com/?path=/story/components-datepicker--playground&args=invalid:true;datePickerType:range), if the user closed the dropdown and sees the top input error message, if they open up the dropdown again, the error message should persist until the error has been fixed. 


Screenshot 2023-01-19 at 10 58 32 AM

jessieyan:
Will open a different PR.

PR #3709, opened 2023-02-02
Adding support for useCache in uiresources API helper and useUiResources hook

blechner:
**Summary**

In order to reduce the loading time of the `GET /uiresources` API, the `useUiResources` hook has been updated to include support for the `useCache` flag, which triggers `GET /uiresources` API twice, the first time with the `&cache=true` query parameter and the second time without it. The idea is to populate the result with data as fast as possible and only then refresh the result with the most up-to-date data. This is ok because the API response does not change very often and it is a slow API (when requesting the non-cached version of it).

- Adding support for `useCache` flag in `GET /uiresources` API helper
- Adding a `cacheduiresources.js` that runs 2 consecutive `GET /resources` API requests (one cached and the other non-cached)
- Adding support for `useCache` flag in the `useUiResources` hook, which triggers  the use of `getCachedUiResourcesData` from `cacheduiresources.js` to make the API requests.
- Added unit tests for the `useUiResources` hook.

- Run `jest src/components/SuiteHeader/hooks/useUiResources.test.js` and notice the that all tests pass.

- The `useUiResources` hook code has been copied as is (with the changes in this PR) to an application that uses `SuiteHeader` component and has access to the `GET /uiresources` API and it has been confirmed that the hook is working as expected.

### DIFF
--- a/packages/react/src/components/SuiteHeader/util/uiresources.js
+++ b/packages/react/src/components/SuiteHeader/util/uiresources.js
@@ -26,6 +26,7 @@ const getUiResourcesData = async ({
   surveyId = null,
   appId = null,
   workspaceId = null,
+  useCache = false,
   fetchApi = defaultFetchApi,
   isTest = false,
 }) => {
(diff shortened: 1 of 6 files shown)


packages/react/src/components/SuiteHeader/util/uiresources.js
```diff
@@ -42,9 +43,10 @@ const getUiResourcesData = async ({
   const surveyIdParam = surveyId ? `&surveyId=${surveyId}` : '';
   const appIdParam = appId ? `&appId=${appId}` : '';
   const workspaceIdParam = workspaceId ? `&workspaceId=${workspaceId}` : '';
+  const cacheParam = useCache ? `&cache=${true}` : '';
```
sls-ca, line 46: When build a URL dynamically you must use encodeURIComponent() so that you can't embed malicious content inside of surveyId, appId and workspaceId (cache is actually OK, since, it's not using a variable for the right hand side)

sls-ca (commented):
looks good, just a small thing around building the URL dynamically (that will eventually get picked up by a security audit)

davidicus (approved):
This looks good once we get the coverage up.

sls-ca (approved):

packages/react/src/components/SuiteHeader/util/uiresources.js
blechner: Done!

PR #3717, opened 2023-02-06
fix(DateTimePickerV2): dropdown menu should have custom height and internally scrollable

### LINKED ISSUE #3715: [DateTimePickerV2] internal scroll when dropdown height not fit in screen
### What package is this for?

- [x] React
- [ ] Angular

### Summary

When there is not enough space for dropdown to open in both top and bottom direction, we need to customize the height of the flyout menu and make internal content scrollable. 

request here.

upcoming release?

date, the better!_

### Available extra resources

jessieyan:
Closes #3715 

**Summary**

When not enough space on both top and bottom direction, dropdown (flyout menu) should have a customized height and the inside content of the dropdown should be able to scroll. 

- DateTimePicker/dateTimePickerUtils.jsx added a new hook useCustomHeight to adjust the height of the dropdown
- DateTimePickerV2WithTimeSpinner.jsx added custom heigh on the style of scroll menu 
- DateTimePickerV2WithoutTimeSpinner.jsx added custom heigh on the style of scroll menu 

- open single select [story](https://deploy-preview-3717--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select), make sure the knobs opens at the bottom
- move DateTimePickerV2 slightly up or down to adjust the distance between top of the input the top of the window (topGap) and the distance between the bottom of the input and the bottom of the window (bottomGap)
- If topGap > bottomGap, the dropdown opens up with the custom height of topGap. If topGap < bottomGap, the dropdown opens downwards with the custom height of bottomGap.

https://user-images.githubusercontent.com/24279794/217066192-285f9768-130a-4e27-8cb7-20e3e776e3c8.mov

- tests here


davidicus:
@jessieyan I got an error when trying to follow the reproduction steps
image

davidicus:
When I move the knob panel so there is room to open up fully I am seeing large gaps under the time picker spinner. 
image
I would think that this should have a max height equal to the original height from the spec.

davidicus:
Let's hide the overflow on the X axis so we do not get horizontal scrollbars. 
image

davidicus:
There is some weirdness that happens if the height changes. At first the height is still small and scrollable but then it jumps to full size when you are scrolling.

packages/react/src/components/DateTimePicker/dateTimePickerUtils.js
```diff
@@ -794,3 +794,28 @@ export const useCloseDropdown = ({
...
+ * A hook handling the height of the drop down menu
+ *
+ * @param {object} containerRef (Object): the ref to the container div of the drop down menu
+ * @returns Object An object containing:
+ *    offTop (boolean): if the menu is off top
+ *    offBottom (boolean): if the menu is off bottom
+ *    customHeight (string): the adjusted height of the drop down menu if both offTop and offBottom are true
+ */
+export const useCustomHeight = (containerRef) => {
+  const windowHeight = window.innerHeight || document.documentElement.clientHeight;
```
davidicus, line 808: You may want to recalculate and update on window resize.

davidicus (changes requested):
There are some minor issues that we will need to figure out before merging.

packages/react/src/components/DateTimePicker/dateTimePickerUtils.js
```diff
@@ -794,3 +794,102 @@ export const useCloseDropdown = ({
...
+
+  // add event listener on the scroll events of the parents
+  useEffect(() => {
+    let currentNode = containerRef?.current?.parentNode;
+    const parentNodes = [];
+    while (currentNode) {
+      parentNodes.push(currentNode);
+      currentNode.addEventListener('scroll', debouncedPosition);
+      currentNode = currentNode.parentNode;
+    }
```
davidicus, line 861: I think this could get expensive. If this component is deeply nested in the DOM this looks to put a scroll listener on every element in that parent chain. Even with the debounce. I worry it could have performance issues. @sls-ca thoughts?

davidicus (changes requested):
We need to rethink how to approach this.

davidicus:
I am also seeing some movement when I am trying to scroll the time picker spinner that causes some funkiness. Perhaps we remove the scroll when we are scrolling the children.

jessieyan:
David confirmed with Sean that we will close dropdown when scrolling. Update and remove the implementation previously for scrolling with the input.

davidicus (approved):
LGTM as long as checks pass.

Issue #3646, opened 2022-11-15
[List] HTML Title tag


RishabhSharma6:
### What package is this for?

- [ ] React
- [x] Angular

### Describe the bug
The title tag in HTML is missing for the list items

### Additional context

Here in the image you will see that the `title` tag is missing
The code below is seen in `ai-list` 

<img width="899" alt="b68e7a00-4ef8-11ed-8c34-0aedbfa42072" src="https://user-images.githubusercontent.com/111842352/201899570-86c0d7f6-093b-4dd5-a492-adff82d2fa22.png">

PR #3719, opened 2023-02-15
fix(listitem): title tag added

rkkp1023:
Closes : https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3646

**Summary**

Title tag is added in the list item

Just added tag in list-item component

### DIFF
--- a/packages/angular/src/list/list-item/ai-list-item.component.ts
+++ b/packages/angular/src/list/list-item/ai-list-item.component.ts
@@ -78,6 +78,7 @@ import { IconService } from 'carbon-components-angular';
             [ngClass]="{ 'iot--list-item--content--values--main__large': item.size === 'lg' }"
           >
             <div
+              title="Item {{ item.value }}"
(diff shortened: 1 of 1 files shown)


packages/angular/src/list/list-item/ai-list-item.component.ts
```diff
@@ -78,6 +78,7 @@ import { IconService } from 'carbon-components-angular';
             [ngClass]="{ 'iot--list-item--content--values--main__large': item.size === 'lg' }"
           >
             <div
+              title="Item {{ item.value }}"
```
davidicus, line 81: I would imagine you would want this to come from a translatable string.

rkkp1023: yes , It's fine  and I removed the hardcoded value `Item`  .

Akshat55 (approved):

zvonimirfras (approved):

Issue #3720, opened 2023-02-16
[TableCard] Provide a default timestamp format across all the dashboard cards


herleraja:
### What package is this for?

- [x] React
- [ ] Angular

### Summary
Provide a default timestamp format across all the dashboard cards. Table card using `L HH:mm` and other places we are using `L HH:mm:ss`
better to expose this as a prop and use this by default. 

### Available extra resources

![image](https://user-images.githubusercontent.com/18572477/219424878-3b5c0dcc-efbb-409c-bcb8-a31a55641386.png)

Time series card expand view
![image](https://user-images.githubusercontent.com/18572477/219426612-f36a2bba-8ff4-4242-beaa-7053f0b136ef.png)

PR #3721, opened 2023-02-17
feat: adding default date format pattern for table, timeseries, bar and combochart cards

herleraja:
Closes https://github.com/carbon-design-system/carbon-addons-iot-react/issues/3720
Related to https://github.ibm.com/wiotp/Maximo-Asset-Monitor/issues/5050

**Summary**

- adding default date format pattern for table, timeseries, bar and combochart cards. we exposed a optionally property `defaultDateFormatPattern` which has a default value `L HH:mm:ss` 

- feat: adding default date format pattern for table, timeseriesbar and combochart cards

- Before: The default date format is `L HH:mm`
image
Now: The default date format is `L HH:mm:ss` --> https://deploy-preview-3721--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card-tablecard--with-multiple-actions
image

- All the cards should work without any issues. 


packages/react/src/constants/CardPropTypes.js
```diff
@@ -160,7 +160,7 @@ export const TableCardPropTypes = {
           href: PropTypes.string,
           target: PropTypes.string,
         }),
-        /** optional type. If "TIMESTAMP", it will format to 'L HH:mm' */
+        /** optional type. If "TIMESTAMP", it will format to 'L HH:mm:ss' */
         type: PropTypes.string,
       })
```
jessieyan, line 165: The `defaultDateFormatPattern` prop us not added to the `CardPropTypes` for BarChartCard, ComboChartCard. TableCard and TimeSeriesCard?

jessieyan (changes requested):

vaffonso (approved):

packages/react/src/components/TableCard/TableCard.jsx
```diff
@@ -70,6 +70,7 @@ const defaultProps = {
   },
   renderDateDropdownInPortal: true,
   withToolbarTooltips: false,
+  defaultDateFormatPattern: 'L HH:mm:ss',
```
davidicus, line 73: same comment about defaults

packages/react/src/components/TableCard/tableCardUtils.jsx
```diff
@@ -23,7 +23,10 @@ export const determinePrecisionAndValue = (precision = 0, value, locale) => {
  * @param {Array<Object>} columns - Array of TableCard columns
  * @return {array} array of columns with formatted links and updated variable values
  */
-export const createColumnsWithFormattedLinks = (columns) => {
+export const createColumnsWithFormattedLinks = (
+  columns,
+  defaultDateFormatPattern = 'L HH:mm:ss'
```
davidicus, line 28: same comment about defaults

packages/react/src/components/ComboChartCard/comboChartHelpers.js
```diff
@@ -188,6 +188,7 @@ export const useChartOptions = (content) => {
       size,
       timeDataSourceId,
       tooltipDateFormatPattern,
+      defaultDateFormatPattern,
```
davidicus, line 191: we should set the default like we do in the tableCardUtils file but using the old format, 'L HH:mm'

packages/react/src/components/TimeSeriesCard/TimeSeriesCard.jsx
```diff
@@ -201,6 +201,7 @@ const defaultProps = {
   domainRange: null,
   tooltipDateFormatPattern: 'L HH:mm:ss',
   tooltipShowTotals: true,
+  defaultDateFormatPattern: 'L HH:mm:ss',
```
davidicus, line 204: Same comment about defaults

davidicus (changes requested):
@herleraja By adding a default where none existed before we are causing a breaking change to our down stream consumers. This looks good to go but we just need to add the default props to be the old value. This will probably allow us to remove the snapshot changes and the changes to the unit test. I would also like to see a unit test where we provide this prop and see that the value then changes.

packages/react/src/components/TableCard/TableCard.jsx
herleraja: Done

packages/react/src/components/ComboChartCard/ComboChartCard.jsx
```diff
@@ -43,6 +43,7 @@ const defaultProps = {
   showTimeInGMT: false,
   domainRange: null,
   tooltipDateFormatPattern: 'L HH:mm:ss',
+  defaultDateFormatPattern: 'L HH:mm:ss',
```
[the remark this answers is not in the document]
herleraja: Done

packages/react/src/components/BarChartCard/BarChartCard.jsx
```diff
@@ -61,6 +61,7 @@ const defaultProps = {
   showTimeInGMT: false,
   tooltipDateFormatPattern: 'L HH:mm:ss',
   values: null,
+  defaultDateFormatPattern: 'L HH:mm:ss',
```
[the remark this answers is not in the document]
herleraja: Done

packages/react/src/components/ComboChartCard/comboChartHelpers.js
herleraja: Done

packages/react/src/components/TableCard/tableCardUtils.jsx
herleraja: Done

vaffonso (approved):

davidicus (approved):

PR #3729, opened 2023-02-22
Fix test build

davidicus:
This PR addresses a breakdown in our testing pipeline. The test build was blowing up because it started using node v18. These changes enforce v16.


jessieyan:
The update-v3.md did not pass prettier format check. Previously it's yml. Prettier does not check yml file.

jessieyan:
The update-v3.md did not pass format check. Before it was yml, prettier does not check yml file.

davidicus:
@jessieyan I changed to md file to stop that test from being ran. We need to fix merge conflicts before that will pass again. I fixed formatting issues with it and it is now passing.

jessieyan (approved):

PR #3749, opened 2023-03-09
fix(walkme): Defining the required Walkme global variable and function in a useEffect hook because they were not being set inside the script tag for some reason.

blechner:
The `_walkmeConfig` object and `walkme_get_language()` function, which are required for Walkme to work properly, were being left undefined for some reason when the Walkme component rendered more than once. This happens more frequently when the `lang` property availability in the component is delayed, which triggers more re-renders. This has always worked in the past but, for some reason, it is not working anymore. The solution was to move the assignments of the variable and function outside of the script tag and put them inside `useEffect` hooks. 
Here is an article that warns about the issues of using `<script>` tags in React: https://codingshower.com/adding-and-executing-script-tags-in-react-render/ 

- Moved the definition of `_walkmeConfig` and `walkme_get_language()` functions to `useEffect` hooks to make sure they are defined and can be used by Walkme script when needed.

- I've used the component in an web app that integrates a Walkme script that supports multiple languages and I was able to see translated content `when` property lang is set to a language different than `en`. 

- There was no noticeable side effect after this change.

### DIFF
--- a/packages/react/src/components/Walkme/Walkme.jsx
+++ b/packages/react/src/components/Walkme/Walkme.jsx
@@ -25,7 +25,7 @@ const Walkme = ({ path, lang }) => {
   useEffect(() => {
     window.walkme_get_language = function () {
       return lang === 'en' ? '' : lang;
-    }
+    };
   }, [lang]);
 
   return (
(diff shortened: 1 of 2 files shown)


davidicus (approved):
LGTM 👍

PR #3750, opened 2023-03-09
fix(idlelogoutconfirmationmodal): Fixing an issue in the idle timer logic when query parameters are appended to the page

blechner:
Closes #

**Summary**

- Fixing an issue in the idle timeout confirmation dialog where changes in the current url (like appending a query parameter) without reloading the page caused the idle timer to restart and enter an invalid state where the countdown dialog would show up frozen. 

- Assembling the `originHref` parameter to be added to the logout URLs in such a way that the current page's query parameters are not included. An easier solution would be to just remove `originHref` for good, since MAS is no longer using it, but we cannot assume that other consumers of the component aren't using it, so it is better to keep it in order to not introduce breaking changes.
- Adding a null check in `surveyData` to avoid runtime errors if this property becomes null in subsequent component re-renders. 

- When using the IdleLogoutConfirmationModal in a MAS application, reduce the idle timeout (to avoid having to wait 30 minutes), them run this in the development console to include an extra query parameter to the page url: `window.history.replaceState(undefined, 'XXXX', window.location.href + "?tab=x")`. Wait for the idle timeout dialog to pop up and notice that the countdown happens normally, i.e. should keep decreasing by 1s. No frozen countdown number is expected.

- Some unit tests were adjusted and all of them are passing.

### DIFF
--- a/packages/react/src/components/SuiteHeader/IdleLogoutConfirmationModal/IdleLogoutConfirmationModal.jsx
+++ b/packages/react/src/components/SuiteHeader/IdleLogoutConfirmationModal/IdleLogoutConfirmationModal.jsx
@@ -98,8 +98,11 @@ const IdleLogoutConfirmationModal = ({
   let logoutRoute = routes?.logout;
   try {
     const url = new URL(routes.logout);
-    if (window.location.href) {
-      url.searchParams.append('originHref', window.location.href);
+    if (window.location) {
+      url.searchParams.append(
+        'originHref',
+        [window.location.protocol, '//', window.location.host, window.location.pathname].join('')
+      );
     }
     if (appId) {
       url.searchParams.append('originAppId', appId);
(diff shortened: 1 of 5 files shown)


herleraja (approved):
LGTM

cgirani (approved):

packages/react/src/components/SuiteHeader/IdleLogoutConfirmationModal/IdleLogoutConfirmationModal.jsx
```diff
@@ -99,7 +99,10 @@ const IdleLogoutConfirmationModal = ({
   try {
     const url = new URL(routes.logout);
     if (window.location.href) {
-      url.searchParams.append('originHref', window.location.href);
+      url.searchParams.append(
+        'originHref',
+        [window.location.protocol, '//', window.location.host, window.location.pathname].join('')
```
davidicus, line 104: In the future I would prefer a template literal here. 
```suggestion
        `${window.location.protocol}//${window.location.host}${window.location.pathname}`
```

davidicus (approved):

PR #3767, opened 2023-04-06
feat(table): add custom input to header columns

EKong-IBM:
[GRAPHITE-58370](https://jsw.ibm.com/browse/GRAPHITE-58370)

**Summary**

- Adding property to allow custom column inputs for Table. This will also allow for extended filters to be added.

- FilterHeaderRow and testing file. Added property to TableHead.

- When creating a table, set customInput to an element and it should display that element instead of the current multiselect, combobox, or textinput.

### DIFF
--- a/packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
+++ b/packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
@@ -60,6 +60,7 @@ class FilterHeaderRow extends Component {
         ),
         /** if isMultiselect and isFilterable are true, the table is filtered based on a multiselect */
         isMultiselect: PropTypes.bool,
+        customInput: PropTypes.element,
       })
     ).isRequired,
     /** internationalized string */
@@ -361,6 +362,8 @@ class FilterHeaderRow extends Component {
           const headerContent =
             column.isFilterable !== undefined && !column.isFilterable ? (
               <div />
+            ) : column.customInput !== undefined ? (
+              <>{column.customInput}</>
             ) : column.options ? (
               column.isMultiselect ? (
                 <FilterableMultiSelect
--- a/packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.test.jsx
+++ b/packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.test.jsx
@@ -1034,4 +1034,16 @@ describe('FilterHeaderRow', () => {
       expect(screen.getByTestId('filter-row-icon')).toBeVisible();
     });
   });
+
+  it('should display custom input when not undefined', () => {
+    render(
+      <FilterHeaderRow
+        {...commonFilterProps}
+        ordering={[{ columnId: 'col1' }, { columnId: 'col2' }]}
+        columns={[{ id: 'col1', customInput: 'customInput' }, { id: 'col2' }]}
+      />
+    );
+
+    expect(screen.getAllByText('customInput')[0]).toBeInTheDocument();
+  });
 });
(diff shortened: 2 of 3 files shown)


packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.test.jsx
```diff
@@ -1034,4 +1034,16 @@ describe('FilterHeaderRow', () => {
       expect(screen.getByTestId('filter-row-icon')).toBeVisible();
     });
   });
+
+  it('should display custom input when not undefined', () => {
+    render(
+      <FilterHeaderRow
+        {...commonFilterProps}
+        ordering={[{ columnId: 'col1' }, { columnId: 'col2' }]}
+        columns={[{ id: 'col1', customInput: 'customInput' }, { id: 'col2' }]}
```
davidicus, line 1043: You are adding a string here which would set off the prop type warning since it is expecting an element.  You can either pass in rendered output (`<div>customInput</div>`) or the output of a React component (`<MyElement/>`)

davidicus (changes requested):
You are almost there. Just a few more tweaks and then the addition to the Table story to showcase this new feature as well as adding it to the docs

packages/react/src/components/Table/TableHead/TableHead.jsx
```diff
@@ -669,6 +669,7 @@ const TableHead = ({
             isFilterable: !isNil(column.filter),
             isMultiselect: column.filter?.isMultiselect,
             width: column.width,
+            customInput: column?.customInput,
```
davidicus, line 672: I would imagine we would want this to be part of the filter object. Like above on line 670,

packages/react/src/components/Table/Table.story.helpers.jsx
```diff
@@ -286,6 +286,11 @@ export const getTableColumns = () => [
       },
     },
   },
+  {
+    id: 'customInput',
+    name: 'Custom Input',
+    customInput: <div>customInput</div>,
```
davidicus, line 292: This will not render without the filter property present.

EKong-IBM:
I wasn't quite able to resolve the unit test issues, but I had followed David's review and the display seemed to work properly. 

packages/react/src/components/Table/TableHead/TableHead.jsx
`customInput: column?.customInput,` -> `customInput: column.filter?.customInput,`

packages/react/src/components/Table/Table.story.helpers.jsx
`customInput: <div>customInput</div>,` -> `filter: { customInput: <div>customInput</div> },`

packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
```diff
@@ -361,6 +362,8 @@ class FilterHeaderRow extends Component {
           const headerContent =
             column.isFilterable !== undefined && !column.isFilterable ? (
               <div />
+            ) : column.customInput !== undefined ? (
+              column.customInput
```
davidicus, line 366: If we do change this to be an elementType you can render like so:
image
Where you define a local variable like `const CustomInput = column.customInput`

packages/react/src/components/Table/TableHead/TableHead.jsx
davidicus, line 672:
```suggestion
            customInput: column.filter?.customInput,
```

davidicus (changes requested):
Some more comments on clarity and how you are able to fix that test issue. I did test that the failure comes from not having a ref on the filter. So changing to elementType will resolve it and render the custom input that we want. We just need to find out if that works for the needed use case in Graphite. Otherwise, you can try to clone the element and see if that works.

sls-ca (commented):
@erzhan-temir-mamyrov @JordanWSmith15 Is this PR still valid?

erzhan-temir-mamyrov:
@sls-ca I am not sure. I can follow up with Graphite ticket and resolve merge conflicts.

JordanWSmith15:
@sls-ca it is needed still, it just isn't high priority any more. Let's let it sit for a while, and some other dev will pick it up and finish it.

PR #3770, opened 2023-04-11
fix(filterheaderrow): display filter icon without input

### LINKED ISSUE #3768: [FilterHeaderRow] throws null when getting filter icon button position
### What package is this for?

- [ x] React
- [ ] Angular

### To Reproduce

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

erzhan-temir-mamyrov:
Closes #3768 

**Summary**

- Replace `input` as an anchor for Filter Icon position to `th`

- Remove query selector for input because when it may not exist when there is no filter for the last column.

- Go to [this story](https://deploy-preview-3770--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-filtering)
- In the knobs area activate "Display icon button in filter row"
- Verify that button has appeared in the correct position

- tests here


jessieyan (approved):
Looks good. Thanks for fixing it.

PR #3778, opened 2023-05-02
fix(DateTimePickerv2): update placeholder text to show XM instead of 'a' or 'A'

### LINKED ISSUE #3762: [DateTimePickerV2] inconsistency in placeholder text
### What package is this for?

- [x ] React
- [ ] Angular

### Describe the bug

In `<TimePickerDropDown>`, we use placeholder text 'hh:mm XM' for 12 hours format and 'HH:mm' for 24 hours format
In `<DateTimePickerV2>`, we use dateTimeMask as the format. There are 2 places indicate incosistency.

1. The meridian, XM vs A
2. When hours are in single h or H format, time picker is not able to show it in placeholder text.
image

### To Reproduce

4. Click on '....'
5. Scroll down to '....'
6. See error

### Environment/versions:

- OS: [e.g. MacOS, Windows]

upcoming release?

date, the better!_

jessieyan:
Closes #3762 

**Summary**

- `TimePickerDropDown` follows the carbon guideline to show  placeholder text like `hh:mm XM` if it is 12hours format. To be consistent, we update DateTimePickerV2 single select placeholder text to show `XM` instead of `a` or `A`. 
- It should look like `YYYY-DD-MM hh:mm XM`

- Update in parseValue() in dateTimePickerUtils.js, specifically to single select. 

- go to single select [story](https://deploy-preview-3778--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-datetimepickerv2--single-select)
- update dateTimeMask knob to use 12 hour format like `YYYY-DD-MM hh:mm A`
- open flyout menu
- click on clear button
- check placeholder should use `XM` instead of `A`

- tests here

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
@@ -607,8 +607,8 @@ describe('DateTimePickerV2', () => {
         }}
       />
     );
-    // default value is YYYY-MM-DD hh:mm A
-    expect(screen.getByText('YYYY-MM-DD hh:mm A')).toBeVisible();
+    // default value is YYYY-MM-DD hh:mm XM
+    expect(screen.getByText('YYYY-MM-DD hh:mm XM')).toBeVisible();
   });
 
   it('should clear date and time fields when click clear button in single select', () => {
@@ -641,7 +641,7 @@ describe('DateTimePickerV2', () => {
 
     userEvent.click(screen.getByText(/clear/i));
 
-    expect(screen.getByText('YYYY-MM-DD hh:mm A')).toBeVisible();
+    expect(screen.getByText('YYYY-MM-DD hh:mm XM')).toBeVisible();
     expect(screen.queryByRole('dialog')).toBeNull();
     expect(mockOnClear).toHaveBeenCalledWith({
       timeRangeKind: 'SINGLE',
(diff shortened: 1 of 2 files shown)


erzhan-temir-mamyrov (approved):
LGTM 👍

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
```diff
@@ -607,8 +607,8 @@ describe('DateTimePickerV2', () => {
         }}
       />
     );
-    // default value is YYYY-MM-DD hh:mm A
-    expect(screen.getByText('YYYY-MM-DD hh:mm A')).toBeVisible();
+    // default value is YYYY-MM-DD hh:mm XM
+    expect(screen.getByText('YYYY-MM-DD hh:mm XM')).toBeVisible();
```
sls-ca, line 611: I don't understand.  Why would expect the datetime mask to be XM ?  That's not even a valid dayjs mask (at least not as used)

sls-ca (changes requested):
I don't understand how replacing the A with XM in the dayjs mask is going to fix a problem here.  Can you explain this change, since I'm probably missing something.

packages/react/src/components/DateTimePicker/DateTimePickerV2.test.jsx
jessieyan: @sls-ca  Hello Sean, this is an issue reported in Graphite that placeholder texts were inconsistent. Like in the screenshot, time picker follows carbon guideline to use 'XM' as the meridian and date picker uses the time format. User reported it was confusing to see the differences. And 'A' is not as intuitive as 'XM' described in the guideline. To make it consistent, we made sure dateTimePicker use 'XM' in the placeholder text.

image

sls-ca (approved):

PR #3830, opened 2023-10-16
feat(codeeditor): New Code editor component

cgirani:
**Summary**

New component CodeEditor

- Update version of `@monaco-editor/react` package
- New component CodeEditor

- Test component and it's knobs

- Make sure CardCodeEditor keep working after package upgrade
- 


sls-ca (approved):

herleraja:
can you check if this effect dashboard code editor 
image
example : https://deploy-preview-3830--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-dashboardeditor--with-editable-title

* add a card
* click on open json editor
image
* then click on save. Ideally, it should save.

herleraja (changes requested):
Thanks for upgrading monaco-editor. However, the dashboard editor is not working as expected.

packages/react/src/components/CardCodeEditor/CardCodeEditor.test.e2e.jsx
```diff
@@ -64,7 +64,7 @@ describe('CardCodeEditor loaded editor test', () => {
 
     // This component throws a network error with too many calls to the cdn script it loads so adding snapshot to existing instance
     onlyOn('headless', () => {
-      cy.findByTestId('ComposedModal').compareSnapshot('CardCodeEditor');
+      cy.findByTestId('ComposedModal').compareSnapshot('CardCodeEditor', 0.2);
```
herleraja, line 67: any specific reason for this change?

packages/react/src/components/CodeEditor/CodeEditor.jsx
```diff
@@ -0,0 +1,216 @@
...
+      button.removeAttribute('disabled');
+    }
+  }, [disabled]);
+
+  /**
+   *
+   * @param {func} _editorValue - a method that returns the current value of the editor
+   * @param {object} val - instance of the monaco editor
+   */
+  const handleEditorDidMount = (_editorValue, val) => {
```
herleraja: In CardCodeEditor you have the following code. 
  ```
  /**
   *
   * @param {object} editor - instance of the editor
   * @param {object} _monaco - instance of monaco
   */
  // eslint-disable-next-line no-unused-vars
  const handleEditorDidMount = (editor, _monaco) => {
    editorValue.current = editor;
  };
  
  ```
  
  where you are setting the first parameter as the current value, whereas you are setting 2nd parameter in here. 
  
  Is this correct? I have just done a visual code review I'm simply observing this. I may be completely wrong here!

packages/react/src/components/CardCodeEditor/CardCodeEditor.test.e2e.jsx
cgirani: Just adding a threshold, instead of 0, locally the tests were passing but not here, maybe some  compare not 100% equal, so just added a bit of threshold.

packages/react/src/components/CodeEditor/CodeEditor.jsx
cgirani: Good catch, updated the `CodeEditor`, to be like `CardCodeEditor `... the function return has changed to updating it to both be the same ;)

herleraja (approved):
LGTM

PR #3831, opened 2023-10-17
feat(table): enable percentage-based column width for the table header

abpaul1993:
Closes # https://jsw.ibm.com/browse/GRAPHITE-66562

**Summary**
[PAL][Table] When the enablePercentageColumnWidth property is set to true and column widths are not provided, the system will automatically calculate the column widths based on the total number of columns, distributing them as a percentage of the total width.
And After that if user start Resizing the Column it will get reset to default column width 150px.

- summary_here

- items_here

- tests here

- tests here

- [x] The code should pass a11y scans (The storybook a11y knob should show no violations). New components should have a11y test written.
- [x] Unit test should be written and should have a coverage of 90% or higher in all areas.
- [x] All components should be passing visual regression test. For new styles or components either a visual regression test should be written for all permutations or the base image updated.
- [x] PR should link and close out an existing issue

### DIFF
--- a/packages/react/src/components/Table/TableHead/TableHead.jsx
+++ b/packages/react/src/components/Table/TableHead/TableHead.jsx
@@ -473,12 +473,49 @@ const TableHead = ({
   );
 
   useEffect(() => {
-    if (enablePercentageColumnWidth && columns.length && percentageMode) {
-      const width = `${parseInt(100 / columns.length, 10)}`;
+    let tableWidth;
+    let columnWidth;
+    const filteredColumns = columns.filter((colOrder) => {
+      const hiddenCol = ordering.find((col) => col.columnId === colOrder.id);
+      return !hiddenCol || !hiddenCol.isHidden;
+    });
 
-      const updatedColumns = columns.map((col) => {
+    const hasOverflowMenu = filteredColumns.some(
+      (column) => column.overflowMenuItems && column.overflowMenuItems.length > 0
+    );
+    const totalColumns = filteredColumns.length;
+    if (
+      enablePercentageColumnWidth &&
+      filteredColumns.length &&
+      percentageMode &&
+      !hasOverflowMenu
+    ) {
+      if (hasRowSelection === 'multi' && !(hasRowExpansion || hasRowNesting)) {
+        tableWidth = `calc(100% - 54px)`;
+        columnWidth = `calc(${tableWidth} / ${totalColumns})`;
+      } else if (
+        (hasRowExpansion || hasRowNesting) &&
+        !(
+          hasRowSelection === 'multi' ||
+          (useRadioButtonSingleSelect && hasRowSelection === 'single')
+        )
+      ) {
+        tableWidth = `calc(100% - 40px)`;
+        columnWidth = `calc(${tableWidth} / ${totalColumns})`;
+      } else if (
+        (hasRowExpansion || hasRowNesting) &&
+        (hasRowSelection === 'multi' ||
+          (useRadioButtonSingleSelect && hasRowSelection === 'single'))
+      ) {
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Table/TableHead/columnWidthUtilityFunctions.js
```diff
@@ -27,7 +27,12 @@ function createWidthsMap(ordering, columnWidths, adjustedCols) {
       ? columnWidths.find((col) => col.id === orderedColumn.columnId)
       : columnWidths[orderedColumn.columnId];
     newColumnWidths[orderedColumn.columnId] = {
-      width: current && current.width !== undefined ? parseInt(current.width, 10) : undefined,
+      width:
+        current && current.width !== undefined
```
hectordavis: We can just have `current?.width` here instead of `current && current.width`

packages/react/src/components/Table/TableHead/TableHead.jsx
```diff
@@ -428,9 +469,31 @@ const TableHead = ({
     // since it would be directly overridden by the column props. This effect can be removed
     // with issue https://github.com/IBM/carbon-addons-iot-react/issues/1224
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [hasResize, columns, ordering, previousColumns]
+    [hasResize, columns, ordering, previousColumns, percentageMode]
   );
 
+  useEffect(() => {
+    if (enablePercentageColumnWidth && columns.length && percentageMode) {
+      const width = `${parseInt(100 / columns.length, 10)}`;
```
sls-ca: Does this take into account hidden columns?
Does this take into account the columns for selection, overflow menu, etc, (ie, I would think those would not be a part of the % calculation)

packages/react/src/components/Table/TableHead/TableHead.jsx
```diff
@@ -428,9 +469,31 @@ const TableHead = ({
...
+
+      const updatedColumns = columns.map((col) => {
+        const updatedCol = { ...col };
+        updatedCol.width = `${width}%`;
```
sls-ca: So what the puprose of this... it ensures that ALL columns are the same % width?  Can I have some fixed width columns and some % width columns?  Can I have columns with different % widths?

sls-ca (changes requested):

packages/react/src/components/Table/TableHead/TableHead.jsx
abpaul1993: The PAL table does not do a good job with default width(pixel width) even we have enough space left in the table. It should, by default, provide a % column width based on how many columns there are based on a props. And when we will pass that props there will not be any fixed pixel column width, all columns will have same % width.
1 column = 100%
2 columns = 50/50
3 columns = 33/33/33

packages/react/src/components/Table/TableHead/columnWidthUtilityFunctions.js
abpaul1993: I have upgraded my code and used current?.width

packages/react/src/components/Table/TableHead/TableHead.jsx
abpaul1993: The user will not be able to specify a fixed percentage width for any column; instead, it will be calculated automatically based on the number of columns present in the table.

packages/react/src/components/Table/TableHead/TableHead.jsx
abpaul1993: I have considered hidden columns, selection(multi and single with radio button), overflow menu, row expansion)while calculating % of width

sls-ca (approved):

PR #3833, opened 2023-11-24
feat(table): table column filter custom input render

abpaul1993:
Closes #
https://jsw.ibm.com/browse/GRAPHITE-64123
**Summary**
Added another check in FilterHeaderRow.jsx to render customInput component passed from storybook or Graphite xml.
- summary_here

Featurw: Allowing Table column filters: to render Custom (Smart) inputs in PAL
- items_here

- tests here

- tests here

- [x] Unit test should be written and should have a coverage of 90% or higher in all areas.

### DIFF
--- a/packages/react/src/components/Table/Table.main.story.jsx
+++ b/packages/react/src/components/Table/Table.main.story.jsx
@@ -1137,8 +1137,6 @@ WithFiltering.parameters = {
 export const WithCustomInputFiltering = () => {
   const {
     selectedTableType,
-    hasFilter,
-    hasAdvancedFilter,
     hideClearAllFiltersButton,
     hasEmptyFilterOption,
     hasMultiSelectFilter,
@@ -1146,18 +1144,12 @@ export const WithCustomInputFiltering = () => {
   } = getTableKnobs({
     knobsToCreate: [
       'selectedTableType',
-      'hasFilter',
-      'hasAdvancedFilter',
       'hideClearAllFiltersButton',
       'hasEmptyFilterOption',
       'hasMultiSelectFilter',
       'hasFilterRowIcon',
     ],
     getDefaultValue: (knobName) => {
-      if (knobName === 'hasAdvancedFilter') {
-        return false;
-      }
-
       if (knobName === 'hideClearAllFiltersButton') {
         return false;
       }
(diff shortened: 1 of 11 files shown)


packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.test.jsx
```diff
@@ -790,6 +791,51 @@ describe('FilterHeaderRow', () => {
...
+      );
+    };
+    render(
+      <FilterHeaderRow
+        {...commonFilterProps}
+        ordering={[{ columnId: 'col1' }, { columnId: 'col2' }]}
+        columns={[
+          {
+            id: 'col1',
+            customInput: ({ onChange }) => <Text onChange={onChange} />,
```
cgirani: why send the `onChange` function if this is not being used in the `Text` component you created at line 807?

packages/react/src/components/Table/Table.story.helpers.jsx
```diff
@@ -98,6 +99,8 @@ export const getSelectDataOptions = () => [
   },
 ];
 
+const itemToString = (item) => (item ? item.text : '');
```
cgirani, line 102: just a short hand 
```suggestion
const itemToString = (item) => (item.text ?? '');
```

packages/react/src/components/Table/Table.main.story.jsx
```diff
@@ -1133,6 +1134,106 @@ WithFiltering.parameters = {
...
+      value: 'whiteboard',
+    },
+  ]);
+  const activeBar = select(
+    'Show filter toolbar (view.toolbar.activeBar)',
+    ['filter', undefined],
+    'filter'
+  );
+
+  const filteredData = data;
```
cgirani: better if you  remove this line and use the `data` variable in line 1212 directly

packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
```diff
@@ -352,7 +353,11 @@ class FilterHeaderRow extends Component {
           '--filter-header-dropdown-max-height': dropdownMaxHeight,
         }}
       >
-        {hasDragAndDrop && <TableHeader className={`${iotPrefix}--filter-header-row--header`} />}
+        {hasDragAndDrop && (
+          /* istanbul ignore next */ <TableHeader
+            className={`${iotPrefix}--filter-header-row--header`}
+          />
+        )}
```
hectordavis: Please avoid `&&` when rendering elements. Use `hasDragAndDrop ? (<> ... </>) : null` instead. For more info, read [this article](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx)

packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
```diff
@@ -373,11 +378,37 @@ class FilterHeaderRow extends Component {
...
+                  if (event.persist) {
+                    event.persist();
+                  }
+                  this.setState(
+                    (state) => ({
+                      filterValues: {
+                        ...state.filterValues,
+                        [column.id]: event.selectedItem
+                          ? getFilterValue(event.selectedItem)
+                          : event.selectedItems
```
hectordavis: Shouldn't you check if `event.selectedItems?.length > 0` here?

packages/react/src/components/Table/TableHead/TableHead.jsx
```diff
@@ -803,6 +803,7 @@ const TableHead = ({
             isFilterable: !isNil(column.filter),
             isMultiselect: column.filter?.isMultiselect,
             width: column.width,
+            customInput: column.filter?.customInput,
```
hectordavis, line 806: You're adding a new prop to the column that is currently not expected. Please update the `components/Table/TablePropTypes.js` file with this new "customInput" filter prop

hectordavis (changes requested):

hectordavis (approved):

packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.test.jsx
```diff
@@ -790,6 +791,51 @@ describe('FilterHeaderRow', () => {
...
+        ]}
+      />
+    );
+
+    // Get the custom input field by its placeholder text
+    const filterInput = screen.getByPlaceholderText('Filter col1');
+    expect(filterInput).toHaveValue('initial value');
+    // Simulate typing in the filter input
+    fireEvent.change(filterInput, { target: { value: 'Value A' } });
+    expect(filterInput).toHaveValue('Value A');
```
sls-ca: While this test is good, it would be better if the Test() component was setup like Test({value, onChange}) and then if you could validate if a default filter value existed then it was passed to Test and when changing the filter value that the FilterheaderRow also got the changes from the Input.  This test, as is, is only really testing that the filter is rendering and not that the filter is actually communicating with the other component.

packages/react/src/components/Table/TablePropTypes.js
```diff
@@ -125,6 +125,7 @@ export const TableColumnsPropTypes = PropTypes.arrayOf(
       placeholderText: PropTypes.string,
       /** if isMultiselect is true, the table is filtered based on a multiselect */
       isMultiselect: PropTypes.bool,
+      customInput: PropTypes.elementType,
```
sls-ca, line 132: Please add a large comment describing what the expectation of this customInput will be.. eg,

```
/**
 * customInput should be a React component that will accept the following props.  value, onChange, (list all props we are sending)
 * The expectation is that the input component will use value as the default value and onChange will be called when the value changes.  The FilterTableRow will debounce the values so so the component does not need to do this
 */
```

packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
```diff
@@ -60,6 +60,7 @@ class FilterHeaderRow extends Component {
         ),
         /** if isMultiselect and isFilterable are true, the table is filtered based on a multiselect */
         isMultiselect: PropTypes.bool,
+        customInput: PropTypes.elementType,
```
sls-ca, line 67: add a fully description of this prop so that when we read the docs we know how to use it, and what to expect for input and outputs, etc.

packages/react/src/components/Table/TableHead/FilterHeaderRow/FilterHeaderRow.jsx
```diff
@@ -373,11 +378,37 @@ class FilterHeaderRow extends Component {
...
+                          ? getFilterValue(event.selectedItem)
+                          : event.selectedItems?.length > 0
+                          ? event.selectedItems.map(getMultiselectFilterValue)
+                          : event.target.value,
+                      },
+                    }),
+                    debounce(this.handleApplyFilter, 1000)
+                  );
+                }}
+                value={filterValues[column.id]}
```
sls-ca, line 417: We should pass as much information to the CustomInput so that someing that is rendering the input can have as much information about the data.  We should send columns and column so that the input component can make other decisions around what to render (ie, it can use the information in the column as well)

sls-ca (changes requested):
Just some minor stuff but this looks good.

sls-ca (approved):

PR #3846, opened 2024-02-29
feat: Empty Table code restructured as before to compatible with Graphite 

anishkumar-bhut:
[GRAPHITE-72230](https://jsw.ibm.com/browse/GRAPHITE-72230)
**Summary**
This is code correction for above story because its creating unit test case fail in graphite.

- summary_here
**isFilterEmptyTable** and **defaultEmptyTable** component should not compile before emptyState validation, previously it does so code reverted as before. 

### DIFF
--- a/packages/react/src/components/Table/EmptyTable/EmptyTable.jsx
+++ b/packages/react/src/components/Table/EmptyTable/EmptyTable.jsx
@@ -87,7 +87,7 @@ const EmptyTable = ({
               ) : (
                 <EmptyState
                   icon={emptyStateIcon ?? 'empty'}
-                  title={message || ''}
+                  title={message}
                   body={messageBody || ''}
                   action={
                     onEmptyStateAction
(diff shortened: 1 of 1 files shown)


cgirani (approved):

packages/react/src/components/Table/EmptyTable/EmptyTable.jsx
```diff
@@ -102,7 +69,36 @@ const EmptyTable = ({
...
+                          label: buttonLabelWithFilters,
+                          onClick: onEmptyStateAction,
+                          kind: 'secondary',
+                        }
+                      : null
+                  }
+                />
+              ) : (
+                <EmptyState
+                  icon={emptyStateIcon ?? 'empty'}
```
herleraja, line 89: This is breaking monitor code. The change should not break downstream application.
emptyStateIcon is by default empty or '' so '' ?? 'empty' always evaluated to ''

anishkumar-bhut: If User will pass props for empty icon then it will take that else it will take 'empty' icon.
we have below possible option for emptyStateIcon.

const icons = {
  error: ErrorImage,
  error404: Error404Image,
  empty: EmptyImage,
  'not-authorized': NotAuthImage,
  'no-result': NoResultImage,
  success: SuccessImage,
};

PR #3850, opened 2024-03-19
revert: table empty state icon changes

herleraja:
**Summary**

- In the commit https://github.com/carbon-design-system/carbon-addons-iot-react/commit/ad2c04288fd28689ca120d04def04a060c697460 we are not setting the empty value by default causing the downstream application to fail

- revert: table empty state icon changes

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/Table/EmptyTable/EmptyTable.jsx
+++ b/packages/react/src/components/Table/EmptyTable/EmptyTable.jsx
@@ -86,7 +86,7 @@ const EmptyTable = ({
                 />
               ) : (
                 <EmptyState
-                  icon={emptyStateIcon ?? 'empty'}
+                  icon={emptyStateIcon !== '' ? emptyStateIcon : 'empty'}
                   title={message}
                   body={messageBody || ''}
                   action={
--- a/packages/react/src/components/Dashboard/__snapshots__/Dashboard.story.storyshot
+++ b/packages/react/src/components/Dashboard/__snapshots__/Dashboard.story.storyshot
@@ -2090,7 +2090,251 @@ exports[`Storybook Snapshot tests and console checks Storyshots 1 - Watson IoT/D
                                   <div
                                     className="iot--empty-state--content"
                                   >
-                                    
+                                    <svg
+                                      alt=""
+                                      className="iot--empty-state--icon"
+                                      data-testid="EmptyState-icon"
+                                      height={80}
+                                      width={80}
+                                    >
+                                      <defs>
+                                        <linearGradient
+                                          id="emptystate-Default-icon_svg__a-defaultIcon"
+                                          x1="37.502%"
+                                          x2="55.959%"
+                                          y1="34.928%"
+                                          y2="57.808%"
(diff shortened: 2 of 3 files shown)


packages/react/src/components/Table/EmptyTable/EmptyTable.jsx
```diff
@@ -86,7 +86,7 @@ const EmptyTable = ({
                 />
               ) : (
                 <EmptyState
-                  icon={emptyStateIcon ?? 'empty'}
+                  icon={emptyStateIcon !== '' ? emptyStateIcon : 'empty'}
```
stuckless, line 89: I see this is reverting, but what was the original code before you did any of this enumeration stuff.

The default value for emptyStateIcon is '', so in the previous change if emptyStateIcon is undefined, then it would bet set to '' and therefor no icon.  In the new change if emptyStateIcon is undefined it will be set to '' and based on the condition it will use 'empty'.  Is that the correct behaviour?

was the original behaviour such that if the prop was no specified then it would use the 'empy' icon?  if so, then I think we are good.

cgirani: The empty fallback would work only if `emptyStateIcon` is **undefined or null** in the `defaultProps` since this is an empty string, that is a "valid value" for the Nullish coalescing operator... that's why "empty" is not being used.

So if we want it to work with `??` the default props need to be `undefined/null` otherwise makes sense to revert to this code

cgirani (approved):

packages/react/src/components/Table/EmptyTable/EmptyTable.jsx
stuckless: @cgirani The challenge (and what I'm trying to get to, is the default behaviour before the changes were in place).

This "revert" comes as a request from a downstream consumer where this change caused changes in the default behaviour of the component.  (I suspect that previously they were not setting the emptyStateIcon and was seeing empty, and after the bullish coalescing operator was used they got no icon (because the default value is an empty string it can never be undefined).

as a rule of thumb when we update components adding new features, etc, we never want to change the default behaviour of the component and I think this change did that.

vaffonso (approved):

packages/react/src/components/Table/EmptyTable/EmptyTable.jsx
herleraja: @stuckless yup. you are correct. Default behavior was having an empty icon. I could have updated the monitor code but the issue is we can't update our old codebase easily (8.11.x and 8.10.x). So just to keep the exact same behaviour I have set this to default. 

@cgirani  One more thing to note is changing the prop name also can break the downstream system. example in the Previous commit `noDataIcon` changed to `emptyStateIcon`.
we should avoid this kind of change in patch releases so that we dont break downstream systems.

 Since the monitor does not use `noDataIcon` and depends on default behavior we are not facing any issues.

stuckless (approved):

PR #3865, opened 2024-05-20
fix: ReadOnlyValue component changes

simson1:
Closes #

**Summary**

- Made some changes to ReadOnlyValue Component.

- Allow attributes to be passed in textarea.
- Removed label loading in the ReadOnlyValue Component as label should be visible.

- There will be no Regression of this.

### DIFF
--- a/packages/react/src/components/ReadOnlyValue/ReadOnlyValue.jsx
+++ b/packages/react/src/components/ReadOnlyValue/ReadOnlyValue.jsx
@@ -17,6 +17,7 @@ const ReadOnlyValue = ({
   isLoading,
   skeletonLoadingLabel,
   skeletonLoadingValue,
+  textAreaAttributes = {},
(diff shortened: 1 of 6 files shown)


herleraja (approved):
LGTM

packages/react/src/components/ReadOnlyValue/ReadOnlyValue.jsx
```diff
@@ -44,6 +43,7 @@ const ReadOnlyValue = ({
               name={value}
               value={value}
               readOnly
+              {...textAreaAttributes}
```
herleraja: Move this at the beginning. else if someone pass value in textAreaAttributes it will overwrite it.  Or you can specific the fields that you allow in textAreaAttributes instead of all props.

sls-ca (approved):

PR #3871, opened 2024-06-24
feat: add tooltip to the tablecard column and value card label

herleraja:
Closes [MASMON-1704](https://jsw.ibm.com/browse/MASMON-1704)

**Summary**

- add tooltip to the tablecard column and value card label
  - Table card :  https://deploy-preview-3871--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card-tablecard--with-multiple-actions
  image
  - Valuecard: https://deploy-preview-3871--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-card-valuecard--with-link-and-measurement-unit
  image

Dashboard Editor: https://deploy-preview-3871--carbon-addons-iot-react.netlify.app/?path=/story/2-watson-iot-experimental-%E2%98%A2%EF%B8%8F-dashboardeditor--with-initial-value 
Click the edit data item button to view the modal
image

- feat: add tooltip to the tablecard column and value card label
- chore: adding amrith as a code owner to get PR reviews

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
+++ b/packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
@@ -104,6 +104,7 @@ const propTypes = {
         value: PropTypes.oneOfType([PropTypes.bool, PropTypes.number, PropTypes.string]),
       })
     ),
+    tooltip: PropTypes.string,
   }),
   setEditDataItem: PropTypes.func,
   /** an object where the keys are available dimensions and the values are the values available for those dimensions
@@ -147,6 +148,14 @@ const propTypes = {
    * Used to override the default behaviour of handleDataItemEdit. if we dont pass any function then it uses handleDefaultDataItemEdit function by default
    */
   handleDataItemEdit: PropTypes.func,
+
+  options: PropTypes.shape({
+    hasColorDropdown: PropTypes.bool,
+    hasUnit: PropTypes.bool,
+    hasDecimalPlacesDropdown: PropTypes.bool,
+    hasThresholds: PropTypes.bool,
+    hasTooltip: PropTypes.bool,
+  }),
 };
 
 const defaultProps = {
@@ -170,6 +179,7 @@ const defaultProps = {
     dataItemEditorDataItemFilter: 'Data filter',
     dataItemEditorDataItemThresholds: 'Thresholds',
     dataItemEditorDataItemAddThreshold: 'Add threshold',
+    dataItemEditorDataItemTooltip: 'Tooltip',
     dataItemEditorBarColor: 'Bar color',
     dataItemEditorLineColor: 'Line color',
     dataItemEditorAddAggregationMethodLabel: 'Add aggregation method',
@@ -208,6 +218,13 @@ const defaultProps = {
     },
   },
   handleDataItemEdit: handleDefaultDataItemEdit,
+  options: {
+    hasColorDropdown: false,
+    hasUnit: false,
+    hasDecimalPlacesDropdown: false,
+    hasThresholds: false,
+    hasTooltip: false,
+  },
 };
 
 const DATAITEM_COLORS_OPTIONS = [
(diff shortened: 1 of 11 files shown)


simson1 (approved):

sls-ca (approved):

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -449,47 +474,68 @@ const DataSeriesFormItemModal = ({
               />
             </div>
           )}
-        </div>
 
-        {(type === CARD_TYPES.VALUE || type === CARD_TYPES.IMAGE) && (
-          <div className={`${baseClassName}--input-group`}>
+          {hasTooltip && (
```
cgirani, line 478: I would suggest we keep the previous validation, by just deleting that, we can impact others that use this component, and maybe add `deprecation` if we are removing the `type` prop

cgirani (changes requested):

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -449,47 +474,68 @@ const DataSeriesFormItemModal = ({
...
             <div className={`${baseClassName}--input-group--item`}>
               <TextInput
-                id={`${id}_attribute-unit`}
-                labelText={mergedI18n.dataItemEditorDataItemUnit}
+                id={`${id}_attribute-tooltip`}
+                labelText={mergedI18n.dataItemEditorDataItemTooltip}
```
blechner, line 482: I think that changing expected property names could cause backward compatibility issues. If you need to change it, maybe use a fallback logic to also accept the previous one.

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -449,47 +474,68 @@ const DataSeriesFormItemModal = ({
...
-                placeholder={`${mergedI18n.example}: %`}
                 onChange={(evt) =>
                   setEditDataItem({
                     ...editDataItem,
-                    unit: evt.target.value,
+                    tooltip: evt.target.value,
                   })
                 }
-                value={editDataItem.unit}
+                value={editDataItem.tooltip}
```
blechner, line 490: I think that changing expected property names could cause backward compatibility issues. If you need to change it, maybe use a fallback logic to also accept the previous one.

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
```diff
@@ -560,7 +606,7 @@ const DataSeriesFormItemModal = ({
             </div>
           )}
 
-        {(type === CARD_TYPES.VALUE || type === CARD_TYPES.IMAGE || type === CARD_TYPES.TABLE) && (
+        {hasThresholds && (
```
blechner, line 609: Here perhaps the old logic should be kept in case `hasThreshold` hasn't been provided, also for backward compatibility.

blechner (changes requested):
The changes to add support for tooltip seem ok, however, I think that we should always keep backward compatibility since we are not generating a major version.

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
herleraja: I have deleted here. But in the parent component ,I kept it same. So functionality wise there is no change. 
image

packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/DataSeriesFormItemModal.jsx
herleraja: same comment as https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3871#discussion_r1650967130

blechner (approved):
LGTM

PR #3903, opened 2024-12-09
fix: locale format for cards

simson1:
Closes #

**Summary**

- Fix locale format

- For all cards default date and tooltip format changed to 'L HH:mm:ss'.
- <DatePicker of the carbon library need locale in 2 digits for that we don't need to set locale of 2 digit in carbon-addons code.
- Ex. For locale 'en-gb' and 'en' date format is diffrent. so we should not set locale in 2 digits.

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.story.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.story.jsx
@@ -62,6 +62,7 @@ export const Default = () => {
             value: RELATIVE_VALUES.YESTERDAY,
           },
         ]}
+        dateTimeMask={text('dateTimeMask', 'YYYY-MM-DD HH:mm')}
         hasTimeInput={boolean('hasTimeInput', true)}
         invalid={boolean('invalid', false)}
         disabled={boolean('disabled', false)}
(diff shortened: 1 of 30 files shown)


packages/react/src/components/DateTimePicker/DateTimePicker.story.jsx
```diff
@@ -56,7 +56,6 @@ export const Default = () => {
     >
       <DateTimePicker
         id="datetimepicker"
-        dateTimeMask={text('dateTimeMask', 'YYYY-MM-DD HH:mm')}
```
herleraja, line 59: any reason it was removed ?

simson1: we already have default value in DateTimePicker.

herleraja: This is for story book. In storybook they can change the value and test it out.

herleraja: image

simson1: reverted.

herleraja (approved):
LGTM

PR #3913, opened 2025-01-07
fix: header and suite header fixes to support rtl

hectordavis:
**Summary**
Changes in header and suite header scss files to support rtl (Right to left)

- paddings, margins and borders updated to use inline and block instead of right/left/top/bottom
- Suite header survey toast rtl fix

Check header and suite header in RTL

Check that header and suite header continue to be displayed correctly in LTR after the change


herleraja (approved):
LGTM

PR #3919, opened 2025-01-20
fix: date time picker absolute range

simson1:
Closes #

**Summary**

-fix: date time picker absolute range

- As per the carbon-11 upgrade, there is design change for range selector. (https://react.carbondesignsystem.com/?path=/story/components-datepicker--range-with-calendar&args=short:!true;closeOnSelect:!false;allowInput:!true)
- Also, In the current design we were facing issue while using absolute in custom range radio buttons. 
- I have attached a video showing the issues.

https://github.com/user-attachments/assets/4c60f08a-2df9-4fce-bde3-47edddf9b8f9

There are the images that contain new change.
image
image

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePicker.jsx
@@ -29,10 +29,8 @@ import {
   invalidStartDate,
   parseValue,
   useAbsoluteDateTimeValue,
-  useDateTimePickerFocus,
   useDateTimePickerKeyboardInteraction,
   useDateTimePickerRangeKind,
-  useDateTimePickerRef,
   useRelativeDateTimeValue,
   useCloseDropdown,
   useDateTimePickerClickOutside,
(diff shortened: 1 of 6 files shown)


packages/react/src/components/DateTimePicker/DateTimePicker.jsx
```diff
@@ -448,51 +448,13 @@ const DateTimePicker = ({
...
-      if (prevFocusOnFirstField) {
-        flatpickr.jumpToDate(newAbsolute.start);
-      } else {
-        flatpickr.jumpToDate(newAbsolute.end);
-      }
-    } else {
-      setFocusOnFirstField(false);
-      flatpickr.jumpToDate(newAbsolute.start);
-    }
+    newAbsolute.startDate = start ? dayjs(newAbsolute.start).format('MM/DD/YYYY') : '';
```
sls-ca, line 454: Interesting that the datemask is hard coded.  Doesn't seem right, but, the previous code also hard coded the date mask.

sls-ca (approved):

packages/react/src/components/DateTimePicker/DateTimePicker.jsx
simson1: `datetimemask` will be used while rendering the value. there is one state `humanValue`.

herleraja (approved):
LGTM

PR #3926, opened 2025-02-06
expose close on select date props to use outside carbon iot

anishkumar-bhut:
Closes #

**Summary**

In DateTimePicker previously date picker is not closed after select date, default is false, so I enable with props and expose that props so we can set true/false as required.

- Created a default props for closeOnSelect and default value is False

- Go to Storybook -> 2. Watson IOT experimental -> Single select with new time spinner. added knobs at last **Close On Select**

### DIFF
--- a/packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
+++ b/packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
@@ -33,6 +33,7 @@ const DateTimePicker = ({
   renderInPortal,
   useAutoPositioning,
   style,
+  closeOnSelect,
   ...others
 }) => {
   return useNewTimeSpinner ? (
@@ -64,6 +65,7 @@ const DateTimePicker = ({
       renderInPortal={renderInPortal}
       useAutoPositioning={useAutoPositioning}
       style={style}
+      closeOnSelect={closeOnSelect}
       others={others}
     />
   ) : (
@@ -93,6 +95,7 @@ const DateTimePicker = ({
       renderInPortal={renderInPortal}
       useAutoPositioning={useAutoPositioning}
       style={style}
+      closeOnSelect={closeOnSelect}
       others={others}
     />
   );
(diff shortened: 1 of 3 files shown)


packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
```diff
@@ -93,6 +95,7 @@ const DateTimePicker = ({
       renderInPortal={renderInPortal}
       useAutoPositioning={useAutoPositioning}
       style={style}
+      closeOnSelect={closeOnSelect}
```
herleraja: I dont think we expose this or use this in DateTimePickerOld

packages/react/src/components/DateTimePicker/DateTimePickerV2WithTimeSpinner.jsx
```diff
@@ -1220,7 +1224,7 @@ const DateTimePicker = ({
                             : null
                         }
                         locale={locale?.split('-')[0]}
-                        closeOnSelect={false}
+                        closeOnSelect={closeOnSelect}
```
herleraja, line 1227: There is nothing wrong with what you have done. But frankly speaking its overkill, only the Monitor team and Graphite team are using this carbon 11 branch. The monitor team is not using this component so it's safe to assume that we dont have to pass these props down the lane. I think we should make the component simpler with fewer parameters.

sls-ca: @anishkumar-bhut I read your summary, but I assume this is being done to fix a problem in graphite, can you explain more about why this change is needed?

anishkumar-bhut: @sls-ca We get query [**MAXUIF-1668**] from App team like Date picker is not close after select date so I checked, In PAL it self its false all the time so I required to expose from PAL so after that App team problem solve.

packages/react/src/components/DateTimePicker/DateTimePickerV2.jsx
anishkumar-bhut: Ok so I can remove from DateTimePickerOld, will do and push again

herleraja (approved):
LGTM

PR #3935, opened 2025-03-07
fix(ValueCard) Remove XSMALL from LEGACY_CARD_SIZES because we are using and displaying unwanted warnings

anishkumar-bhut:
**Summary**
in ValueCard component have a CARD_SIZES and in that XSMALL is under LEGACY_CARD_SIZES but in Operation dashboard and other places XSMALL is using so in console unwanted warning displaying so added  XSMALL in CARD_SIZES

- tests here

- tests here


packages/react/src/constants/LayoutConstants.js
```diff
@@ -9,6 +9,7 @@ export const COLORS = {
 };
 
 export const CARD_SIZES = {
+  XSMALL: 'XSMALL',
```
herleraja, line 12: is this supported ? As far as I'm aware support for XSMALL was removed. and when ever someone use XSMALL we used to change it to 'SMALL' 

This may have impact on monitor dashboards.

JordanWSmith15: @herleraja Hello.  Support for xsmall was removed, but we have a scenario in Maximo where the use of it is required again. It will only be allowed for certain simple Value cards, and there is likely some CSS cleanup and other work that will need to go in to that card type.  What would work for Monitor here? Is there a condition or check we could add to isolate you from this change? @anishkumar-bhut @hectordavis @stuckless

hectordavis: Where exactly is the warning coming from? From what I can see [here](https://github.com/carbon-design-system/carbon-addons-iot-react/blob/next/packages/react/src/constants/CardPropTypes.js#L851), ValueCard still uses the LEGACY_CARD_SIZES as expected sizes

herleraja: @JordanWSmith15 Just checked. I think the monitor should not be affected by this change. 

We override the props accordingly. 
 
image
 
 
 Also, this may need to be well tested, because its used directly or indirectly
example if the size is small we won't display the error message icons. so we need to add the XSMALL size here too. 
image

packages/react/src/utils/cardUtilityFunctions.js
```diff
@@ -92,9 +92,7 @@ export const compareGrains = (grain1, grain2) => {
 
 export const getUpdatedCardSize = (oldSize) => {
```
herleraja, line 93: If its needed for value card may be you can do it specifically for value card. looks like getUpdatedCardSize function getting used in almost all the cards
image

packages/react/src/constants/LayoutConstants.js
herleraja: @hectordavis the warning thrown in getUpdatedCardSize but that's thrown only in development mode. It should not effect in production. 
image

PR #3943, opened 2025-03-25
feat(table) AILabel with Table Header column

anishkumar-bhut:
Closes #

**Summary**
I upgraded Carbon 1.72.0 to 1.78.0 because I need to use AILabel component introduced in Carbon lib. after that I create a props for table column and name is decorator. we can pass decorator with column information and check root element is AILabel then only it will render in column header.

- summary_here

- items_here

- tests here

- tests here

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -179,7 +179,7 @@
     "@carbon/icons-react": "^11.53.0",
     "@carbon/layout": "^11.28.0",
     "@carbon/pictograms-react": "^11.69.0",
-    "@carbon/react": "^1.72.0",
+    "@carbon/react": "^1.78.0",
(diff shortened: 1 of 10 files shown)


anishkumar-bhut:
@sls-ca @hectordavis @herleraja Please review this PR

herleraja:
Table header got moved for all columns. I think some CSS got messedup

New: 
https://deploy-preview-3943--carbon-addons-iot-react.netlify.app/?path=/story/1-watson-iot-table--with-sorting
image

Old: 
https://carbon-design-system.github.io/carbon-addons-iot-react/?path=/story/1-watson-iot-table--with-sorting
image

herleraja (approved):
LGTM

PR #3948, opened 2025-04-30
fix: aligned nested Load More Button in Table and List

vivekdomadiya:
Closes #

**Summary**

- Aligned Load More button with nested children row in Table
- Aligned View More button in List

- fix: aligned nested Load More Button in Table

image

image

- tests here

- [x] Make sure all references to `iot` or `bx` class prefix is using the prefix variable

### DIFF
--- a/packages/react/src/components/Table/TableBody/TableBodyRowRenderer.jsx
+++ b/packages/react/src/components/Table/TableBody/TableBodyRowRenderer.jsx
@@ -333,6 +333,7 @@ const TableBodyRowRenderer = (props) => {
       hasRowNesting={hasRowNesting}
       hasRowSelection={hasRowSelection}
       showExpanderColumn={showExpanderColumn}
+      nestingLevel={nestingLevel}
       columns={columns}
       totalColumns={totalColumns}
       isLoadingMore={loadingMoreIds.includes(row.id)}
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Table/TableBody/TableBodyRowRenderer.jsx
```diff
@@ -365,7 +366,7 @@ const TableBodyRowRenderer = (props) => {
               {...props}
               key={`load-more-row-${row.id}`}
               row={{ id: row.id, isLoadMoreRow: true }}
-              nestingLevel={nestingLevel}
+              nestingLevel={nestingLevel + 1}
```
herleraja, line 369: we are doing concat of the item. I think increasing the level by one is not correct!

vivekdomadiya: LoadMore should align with children of Item, so need to increase nestingLevel, same as done for children in line 351,
https://github.com/carbon-design-system/carbon-addons-iot-react/pull/3948/files#diff-dcd9f29a8236dd448882cf5e4f363881ed05927f73a67c343a8c403841677e0bR351

herleraja (approved):
LGTM

sls-ca (approved):

abpaul1993 (approved):
LGTM

PR #3991, opened 2025-07-08
fix(tablecellrenderer): fix text truncate and tooltip with WrapCellText

vivekdomadiya:
Closes # [MASMON-4317](https://jsw.ibm.com/browse/MASMON-4317)

**Summary**

- fix(tablecellrenderer): fix text truncate and tooltip with WrapCellText

- fix(tablecellrenderer): fix text truncate and tooltip with WrapCellText

- 
image

- tests here

### DIFF
--- a/packages/react/src/components/Table/TableCellRenderer/TableCellRenderer.jsx
+++ b/packages/react/src/components/Table/TableCellRenderer/TableCellRenderer.jsx
@@ -1,10 +1,9 @@
 import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { DefinitionTooltip } from '@carbon/react';
+import { DefinitionTooltip, Tooltip } from '@carbon/react';
 import warning from 'warning';
 
-import { Tooltip } from '../../Tooltip';
 import { settings } from '../../../constants/Settings';
 import { WrapCellTextPropTypes } from '../../../constants/SharedPropTypes';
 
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Table/TableCellRenderer/TableCellRenderer.jsx
```diff
@@ -98,9 +97,10 @@ const TableCellRenderer = ({
     ) : (
       <Tooltip
         showIcon={false}
-        triggerText={element}
+        label={element}
```
simson1, line 100: i think we need to pass chidlren.toString() here instead of whole element. by using that we can avoid css white-space: 'unset'.

packages/react/src/components/Table/TableCellRenderer/_table-cell-renderer.scss
```diff
@@ -6,19 +6,32 @@
...
+
+    &:focus {
+      outline: none;
+    }
+  }
+
+  .#{$iot-prefix}--tooltip,
+  .#{$prefix}--tooltip {
+    display: block;
+  }
```
simson1, line 29: also check if someone passes tooltip for table cell renderer, definition tooltip render correclty,

vivekdomadiya: working fine

packages/react/src/components/Table/TableCellRenderer/TableCellRenderer.jsx
vivekdomadiya: as discuss, no need

abpaul1993 (approved):

simson1 (approved):

PR #4004, opened 2025-08-19
For PAL make sure no external css fonts are exported. (no font-face's created with urls) 

Frank-Yournet:
Closes #
MAXUIF-2087
https://jsw.ibm.com/browse/MAXUIF-2087

**Summary**
_______________________________________________________________
MAXUIF-2087 ITEMS:

1)For PAL make sure not external css fonts are exported
	-Checked for explicit @font-face templates in the sass files that referenced external URLs
	-Realized the @font-face was implicitly defined and removed the line “@use '@carbon/styles/scss/fonts' as *”

5)Upgrade carbon charts in PAL                 
6)Make Carbon in PAL a peer dependency
	-Moved @carbon/react, @carbon/charts, @carbon/charts-react to peer dependencies and gave a range from the current charts version of 1.23.10 - 2.0.0. Applications still on 1.23.10 would work the same but allows for Graphite to move to the latest 1.23.17 with no complaints. Also upgraded to 1.23.17 in devDependencies.

MAXUIF-2380:
	-Upgraded @carbon/charts and @carbon/charts-react from 1.23.10 —> 1.23.17 and made them peer deps
_______________________________________________________________

- items_here

- tests here

- tests here


packages/react/package.json
```diff
@@ -173,13 +173,10 @@
   "dependencies": {
     "@babel/plugin-transform-optional-chaining": "^7.24.7",
```
sls-ca: not sure that should be a dependency, nor @babel/runtime.  Can you verify if those should ever be dependencies, vs dev dependencies?

packages/react/package.json
```diff
@@ -173,13 +173,10 @@
...
     "@babel/runtime": "^7.26.9",
-    "@carbon/charts": "1.23.10",
-    "@carbon/charts-react": "1.23.10",
     "@carbon/colors": "11.34.0",
```
sls-ca: I think all of the carbon libraries should be dev/peer dependency

packages/react/rollup.config.mjs
```diff
@@ -121,6 +121,7 @@ export default [
           '@carbon/icons-react': 'CarbonIconsReact',
           '@carbon/react': 'CarbonReact',
           '@carbon/charts': 'CarbonCharts',
+          '@carbon/charts-react': 'CarbonChartsReact',
```
sls-ca, line 124: What does putting this in globals do?

sls-ca (changes requested):
Looks good, but can you provide some answers to the comments I left.

packages/react/rollup.config.mjs
Frank-Yournet: From my understanding, it avoids multiple copies of @carbon/charts-react and aligns with our peer dep declared in Graphite for example. This is only useful in UMD builds.

packages/react/package.json
Frank-Yournet: @babel/plugin-transform-optional-chaining can be moved to dev, as it's only used at build-time. @babel/runtime however, is used in imports all over the build files and should likely remain a dep.

hectordavis (commented):
I agree with @sls-ca's comments on this. LGTM otherwise

Frank-Yournet:
Going to merge and open a new PR with the new branch.

PR #4013, opened 2025-08-28
Datepicker not able to open in model issue fixed 

anishkumar-bhut:
Closes #

**Summary**
Fix DateTimePicker inside Modal

The DateTimePicker component had an issue where the TimePicker dropdown wouldn’t open correctly when rendered inside a modal. This happened because the dropdown was always being portaled to document.body, which caused positioning conflicts with the modal layer.

To fix this, the portal logic has been updated:

If a modal is open, the TimePicker now renders its portal inside the modal body instead of document.body.

Positioning is recalculated based on the modal container to ensure the dropdown opens in the correct place.

This change ensures the DateTimePicker works seamlessly both inside and outside of modals.

- summary_here

- items_here

- tests here

- tests here

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -17,27 +17,27 @@
 # These users own any files in the following directory at the root of
 # the repository and any of its subdirectories.
 
-* @davidicus @jessieyan @sls-ca @herleraja @abpaul1993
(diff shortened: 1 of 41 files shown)


anishkumar-bhut:
have a issue so closing

PR #4019, opened 2025-09-09
feat(tearsheet):  add shouldPreventClose prop to control tearSheet closing via promise

chaudarohit:
Closes #

**Summary**

- Introduced a new prop `shouldPreventClose` for `TearSheet`.  
This prop is a promise that resolves to `false` to allow closing or `true` to block closing.  
By default, it resolves to `false`.

- https://jsw.ibm.com/browse/MAXUIF-2911

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/TearSheet/TearSheet.jsx
+++ b/packages/react/src/components/TearSheet/TearSheet.jsx
@@ -1,4 +1,4 @@
-import React, { cloneElement, useMemo } from 'react';
+import React, { cloneElement, useCallback, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import { Close } from '@carbon/react/icons';
 import classnames from 'classnames';
@@ -33,6 +33,7 @@ const propTypes = {
   }),
   children: PropTypes.node,
   testId: PropTypes.string,
+  shouldPreventClose: PropTypes.func,
 };
 
 const defaultProps = {
@@ -51,6 +52,7 @@ const defaultProps = {
   children: undefined,
   // TODO: set default testId in v3.
   testId: '',
+  shouldPreventClose: () => new Promise((res) => res(false)),
 };
 
 const TearSheet = ({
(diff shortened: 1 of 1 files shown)


packages/react/src/components/TearSheet/TearSheet.jsx
```diff
@@ -66,15 +68,18 @@ const TearSheet = ({
...
 }) => {
-  const onCloseButton = () => {
+  const onCloseButton = useCallback(async () => {
+    const preventClose = await shouldPreventClose();
     if (onClose) {
       onClose();
-      goToPreviousSheet();
-    } else {
+    }
+    if (!preventClose) {
```
hectordavis, line 75: This if condition should wrap both onClose and goToPrevious sheet, no just goToPrevious sheet.
The main thing we want to prevent if preventClose is true is onClose...

hectordavis (changes requested):

packages/react/src/components/TearSheet/TearSheet.jsx
chaudarohit: @hectordavis 
If I also wrap onClose inside the condition, the function I passed in MAS common UI on 'onClose' does not get triggered.

hectordavis: What does your function do? Your logic should be in shouldPreventClose instead. That's the first thing that runs when the tear sheet close button is clicked

chaudarohit: This is what I put in MAS common UI, Calling onCancel function since the behaviour should be same onClose.

Screenshot 2025-09-10 at 7 35 07 PM

hectordavis: Your shouldPreventClose is always resolving as true. That's not the correct approach. The value it returns should be dependent on what you clicked in the cancel modal. Ideally the promise should only resolve as false (allowing the close) after the user clicked the button in the modal that is supposed to close the tear sheet (whatever that button is). Do you have a branch in mas-common-ui with your changes so I could take a look?

chaudarohit: Done, Please review again.

hectordavis (approved):

PR #4029, opened 2025-10-09
chore: upgraded latest PAL version to fix numberinput issue

abpaul1993:
Closes #
Fixing Graphite NumberInput issue. DT number: https://jsw.ibm.com/browse/MASISMIG-67871
Summary
https://github.com/carbon-design-system/carbon/pull/20580

When entering meter readings with a wrong decimal separator there is no message indicating that the decimal separator is wrong and the meter reading is saved by ignoring the decimal separator.
RESULT: 
A wrong meter reading is saved, e.g., 19 instead of 1.9. 
EXPECTED RESULT: 
The user is notified that the value entered is not correct.
**Summary**

- summary_here

- items_here

- tests here

- tests here


packages/react/package.json
```diff
@@ -210,7 +210,7 @@
     "@carbon/icons-react": "^11.61.0",
     "@carbon/layout": "^11.35.0",
     "@carbon/pictograms-react": "^11.69.0",
-    "@carbon/react": "^1.88.0",
+    "@carbon/react": "1.93.0",
```
Frank-Yournet: Should always begin with a leading caret to continue giving freedom to the consuming application. "^1.93.0"

hectordavis (changes requested):
@abpaul1993 please provide a clear PR description. Why is this change being made? Is there a defect related to this? Is it to fix a graphite component?

Frank-Yournet (changes requested):
These changes also require a sibling PR in Graphite that upgrades PAL once merged. (as a reminder) This PR here should be linked in that description when the time comes. 

Also, when I took the build and moved it into Graphite... I received 9 failing tests in graphite/packages/React-Components, so unsure if that is related to these changes.

Frank-Yournet (approved):

hectordavis (approved):

PR #4049, opened 2025-12-03
fix: updated the fix for access related issues in dashboard component

Maha14-ui:
closes : https://jsw.ibm.com/browse/MAXUIF-3195?filter=-1 

**Summary**

- here I have updated the fix for access related issues for dashboard value component.
- issues : 
- Content is not within a landmark element
- Accessible name does not match or contain the visible label text
components from carbon-addon-iot-react mentioned below.
 1) valueCard component
2) gaugeCard

- items_here

- tests here
Before : 
Screenshot 2025-12-03 at 4 24 37 AM
Screenshot 2025-12-03 at 4 25 30 AM
Screenshot 2025-12-03 at 4 25 40 AM
After : 
Screenshot 2025-12-03 at 12 33 44 AM
Screenshot 2025-12-03 at 6 19 33 PM
Screenshot 2025-12-03 at 6 20 09 PM

- tests here

- [x] (React) All major areas have a `data-testid` attribute. New test ids should have test written to ensure they are not changed or removed.
- [x] PR should link and close out an existing issue


packages/react/src/components/ValueCard/ValueContent.jsx
```diff
@@ -57,47 +57,51 @@ const ValueContent = ({
...
-            locale={locale}
-            isEditable={isEditable}
-            title={title}
-            renderIconByName={others.renderIconByName}
-            value={determineValue(attribute.dataSourceId, values, attribute.dataFilter)}
-            secondaryValue={
-              attribute.secondaryValue && {
-                ...attribute.secondaryValue,
-                value: determineValue(attribute.secondaryValue.dataSourceId, values),
+    <section id="main-content" aria-label="value content">
```
aqibibm: Can't we use same div to give aria-label instead of having <section>

packages/react/src/components/GaugeCard/GaugeCard.jsx
```diff
@@ -93,93 +93,96 @@ const GaugeCard = ({
...
-                  />
-                  <circle
-                    className={`${iotPrefix}--gauge-fg`}
-                    cx={gaugeSize / 2}
-                    cy={gaugeSize / 2}
-                    r={radius}
-                  />
-                  <text
-                    id="gauge-label"
+    <section id="main-content" aria-label="region">
```
aqibibm: Why do we need this?

packages/react/src/components/ValueCard/ValueContent.jsx
Maha14-ui: hi @aqibibm , I have tried to repeat the issues from reference-app. issue (Content is not within a landmark element) was not repeating and main tag and role is already available from main page so these changes are not required. I have reverted these changes. and updated the PR. please review. Thank you!.

packages/react/src/components/GaugeCard/GaugeCard.jsx
Maha14-ui: hi @aqibibm ,as I said above. I have reverted these changes. and updated the PR. please review. Thank you!.

Maha14-ui:
hi @aqibibm and Team , I was able to reproduce the issue “Accessible name does not match or contain the visible label text.”
I have fixed it in the index.js file and verified the fix in the reference-app through Graphite using my local PAL changes.
updated test ss below. please review the latest changes. Thank you!
Before : 
--------
Screenshot 2025-12-09 at 5 17 12 PM
After : 
------
Screenshot 2025-12-10 at 6 28 36 PM
Screenshot 2025-12-10 at 6 28 43 PM

JordanWSmith15 (approved):

packages/react/src/components/GaugeCard/GaugeCard.jsx
aqibibm: I can still see its wrapped in `section` tag. The question is do we really need this `section` tag and why? @Maha14-ui

Maha14-ui: hi @aqibibm , yes section is not required now I have updated the fix without section. its working fine. could you please review it once.
changes in the PR : 1) section removed in value card and gauge card
2) and index.jsx changes for fix

aqibibm (approved):

PR #4054, opened 2025-12-31
fix: add tooltip to drag handler #MAXUIF-3526

darshan1290:
Closes # [MAXUIF-3526](https://jsw.ibm.com/browse/MAXUIF-3526)

**Summary**

- Add tooltip to drag handler to data-table

image


packages/react/src/components/Table/TableBody/TableBodyRow/TableDragHandle.jsx
```diff
@@ -17,12 +18,30 @@ const propTypes = {
...
+
+  /**
+   * Tooltip text to display on hover.
+   */
+  tooltipText: PropTypes.string,
+};
+
+const defaultProps = {
+  isDragging: false,
+  tooltipText: 'Click and drag to new location',
```
hectordavis: defaultProps is going to be removed in React 19 (see https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-deprecated-react-apis). This causes error in the console and we should be moving away from it. Please remove it and use [object destructuring default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring#default_value) instead - for the component props destructuring.

packages/react/src/components/Table/TableBody/TableBodyRow/TableDragHandle.jsx
```diff
@@ -34,11 +53,28 @@ const TableDragHandle = forwardRef(function TableDragHandle({ onStartDrag, rowId
...
+        >
+          <Draggable />
+        </IconButton>
+      )}
     </div>
   );
 });
 
 TableDragHandle.propTypes = propTypes;
+TableDragHandle.defaultProps = defaultProps;
```
hectordavis: This should be removed once the above is addressed.

packages/react/src/components/Table/TableBody/TableBodyRow/TableDragHandle.jsx
```diff
@@ -34,11 +53,28 @@ const TableDragHandle = forwardRef(function TableDragHandle({ onStartDrag, rowId
...
+          <Draggable />
+        </div>
+      ) : (
+        // When not dragging, render with tooltip
+        <IconButton
+          className="draggble-icon-wrapper"
+          kind="ghost"
+          size="sm"
+          label={tooltipText}
+          align="right"
```
hectordavis: Why use IconButton instead of Carbon's Tooltip component directly? Also, we should rely on autoAlign to prevent future issues by always having it on the right

hectordavis (changes requested):

packages/react/src/components/Table/TableBody/TableBodyRow/TableDragHandle.jsx
darshan1290: removed the defaultProps

packages/react/src/components/Table/TableBody/TableBodyRow/TableDragHandle.jsx
darshan1290: removed the defaultProps

packages/react/src/components/Table/TableBody/TableBodyRow/TableDragHandle.jsx
darshan1290: I was thought of using the definition tooltip but its not fit in requirement but yeah the carbon tooltip fullfil the requirement. Thank you.

hectordavis (approved):

PR #4070, opened 2026-01-23
chore: updated @carbon/react version to bring in a fix

patel1643:
Closes # MASISMIG-68987

**Summary**

- Updated the `@carbon/react` version to accommodate new changes/bug fix for `FileUploader` component

- package.json

- It can be tested by adding `maxFileSize` prop (this prop accepts number in bytes so for example: maxFileSize={100} means maximum acceptable size should be 100 bytes) to the FileUploader component and when a file is uploaded that's higher than 100bytes it would result in an error popup.

- Ensure the other components are still working as expected


hectordavis (approved):
LGTM. We can merge this once the workflows pass

PR #4075, opened 2026-02-02
fix: thresold value breaks if value is equal to zero

kajalbijlani:
Closes #

**Summary**

- Issue: The threshold value input field was breaking when users entered a value of 0 or values starting with 0. The  component was treating 0 as a falsy value, causing the input to display incorrectly or lose the value.
- fix: Zero values are properly parsed and stored. Values starting with 0 (like 0.5, 01) are correctly handled.

- fix: thresold value breaks if value is equal to zero

- tests here

- tests here

closes: https://jsw.ibm.com/browse/MASMON-5626 


packages/react/src/components/CardEditor/CardEditForm/CardEditFormItems/ThresholdsFormItem.jsx
```diff
@@ -275,13 +275,11 @@ const ThresholdsFormItem = ({
                     invalid={false} // don't allow invalid state
                     value={threshold.value?.toString() || 0}
                     onChange={(event, { value }) => {
+                      const parsedValue = Number(value ?? event.imaginaryTarget.value);
```
simson1: i would say just use value. `event.imaginaryTarget.value` is not supported

simson1: also, parent component will return number value only. no need to convert  Number(value).

simson1 (changes requested):

simson1 (approved):

simson1 (approved):

PR #4082, opened 2026-03-20
Update TearSheet component

chaudarohit:
Closes #

**Summary**

- Added support for custom role attribute to the TearSheet component to improve accessibility

- https://jsw.ibm.com/browse/MAXUIF-3224

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/TearSheet/TearSheet.jsx
+++ b/packages/react/src/components/TearSheet/TearSheet.jsx
@@ -34,6 +34,7 @@ const propTypes = {
   children: PropTypes.node,
   testId: PropTypes.string,
   shouldPreventClose: PropTypes.func,
+  // eslint-disable-next-line react/require-default-props
   role: PropTypes.string,
 };
 
@@ -54,7 +55,6 @@ const defaultProps = {
   // TODO: set default testId in v3.
   testId: '',
   shouldPreventClose: () => new Promise((res) => res(false)),
-  role: '',
 };
 
 const TearSheet = ({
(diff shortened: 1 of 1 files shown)


packages/react/src/components/TearSheet/TearSheet.jsx
```diff
@@ -85,6 +88,7 @@ const TearSheet = ({
       // TODO: use only testId in v3.
       data-testid={testId || `${iotPrefix}--tear-sheet-${idx}`}
       className={classnames(`${iotPrefix}--tear-sheet`, className)}
+      role={role || undefined}
```
hectordavis: No need for undefined here as long as you don't pass the default value in defaultProps

packages/react/src/components/TearSheet/TearSheet.jsx
```diff
@@ -53,6 +54,7 @@ const defaultProps = {
   // TODO: set default testId in v3.
   testId: '',
   shouldPreventClose: () => new Promise((res) => res(false)),
+  role: '',
```
hectordavis: This can be removed.

hectordavis (changes requested):

packages/react/src/components/TearSheet/TearSheet.jsx
chaudarohit: I added role to defaultProps to address the ESLint warning (react/require-default-props). But now I’ve removed this from defaultProps and added "eslint-disable-next-line react/require-default-props" in propTypes

packages/react/src/components/TearSheet/TearSheet.jsx
chaudarohit: Done

packages/react/src/components/TearSheet/TearSheet.jsx
hectordavis: Actually, instead of disabling the eslint, you can just set the role's default prop to undefined. It's the correct approach

hectordavis (changes requested):

packages/react/src/components/TearSheet/TearSheet.jsx
chaudarohit: Yes, made the changes

hectordavis (approved):

PR #4111, opened 2026-06-08
feat: gradient background added in card

mahiman321:
Closes #
https://jsw.ibm.com/browse/MAXUIF-4011

**Summary**
Added backgroundGradient prop in Card, user can add any gradient ad background
Added dark prop in Card, setting dark = true will switch card toolbar icon color to white.

### with backgroundGradient = linear-gradient(14deg, #7C3DD6 0%, #02274A 97.44%)
image

### with backgroundGradient = linear-gradient(14deg, #7C3DD6 0%, #02274A 97.44%) and dark = true
image

### When icon is clicked icon changes back to black as background color is white
image

- items_here

- tests here

- tests here

### DIFF
--- a/packages/react/src/components/Card/Card.jsx
+++ b/packages/react/src/components/Card/Card.jsx
@@ -529,9 +529,8 @@ const Card = (props) => {
               ...(backgroundGradient && { background: backgroundGradient }),
             }
       }
-      className={classnames(`${iotPrefix}--card`, className, {
+      className={classnames(`${iotPrefix}--card`, `${iotPrefix}--card--${theme}`, className, {
         [`${iotPrefix}--card--resizing`]: isResizing,
-        [`${iotPrefix}--card--dark`]: theme === 'dark',
       })}
     >
       {!hideHeader && !hideTitle && (
--- a/packages/react/src/components/Card/Card.jsx
+++ b/packages/react/src/components/Card/Card.jsx
@@ -210,6 +210,7 @@ export const defaultProps = {
   isExpanded: false,
   /** performance option: only render the content of the card to the ReactDOM if the card is visible on screen */
   isLazyLoading: false,
+  dark: false,
   /** For now we will hide the per card actions when we're editing */
   availableActions: {
     edit: false,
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Card/Card.jsx
```diff
@@ -279,6 +280,7 @@ export const defaultProps = {
   data: null,
   content: null,
   shouldUseTranslatedLabels: false,
+  backgroundGradient: undefined,
```
sachin-chauhan2: replace undefined with null

packages/react/src/components/Card/Card.story.jsx
```diff
@@ -261,6 +261,8 @@ export const BasicWithRenderProp = () => {
         tabIndex={0}
         hideHeader={boolean('hideHeader', true)}
         hideTitle={boolean('hideTitle', false)}
+        backgroundGradient={text('backgroundGradient', undefined)}
```
sachin-chauhan2: use null instead undefined

sachin-chauhan2:
Change font color to white as well for `dark`

packages/react/src/constants/CardPropTypes.js
```diff
@@ -993,4 +993,8 @@ export const CardPropTypes = {
   /** If set to true it will render outside of the current DOM in a portal, otherwise render as a child */
   renderDateDropdownInPortal: PropTypes.bool,
   extraHeaderContent: PropTypes.element,
+  /** CSS gradient value to apply as the card background. */
+  backgroundGradient: PropTypes.string,
+  /** Enable dark mode styling for the card */
+  dark: PropTypes.bool,
```
sachin-chauhan2: I think we should rename this prop, `dark` to `theme` and have a value as `dark`, so that it can be expandable in future to other themes as well

packages/react/src/components/Card/Card.jsx
```diff
@@ -531,7 +531,7 @@ const Card = (props) => {
       }
       className={classnames(`${iotPrefix}--card`, className, {
         [`${iotPrefix}--card--resizing`]: isResizing,
-        [`${iotPrefix}--card--dark`]: dark,
+        [`${iotPrefix}--card--dark`]: theme === 'dark',
```
sachin-chauhan2: i think we can write this class directly as ``${iotPrefix}--card--${theme}``

sachin-chauhan2 (approved):

sls-ca (approved):

